PR #6, opened 2016-04-04
Initial import of specs

vbatts:
Per the proposal for this image-spec group, this is bringing in the base for the distributable image format from https://github.com/docker/distribution/blob/bf9f80eaffb0eabc768762bc4ff03ded277ea594/docs/spec/manifest-v2-2.md and https://github.com/docker/docker/blob/99a396902f0ea9d81ef87a683489b2435408f415/image/spec/v1.md

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -1,0 +1,284 @@
+<!--[metadata]>
++++
+draft = true
++++
+<![end-metadata]-->
+
+# Image Manifest Version 2, Schema 2
+
+This document outlines the format of of the V2 image manifest, schema version 2.
+The original (and provisional) image manifest for V2 (schema 1), was introduced
+in the Docker daemon in the [v1.3.0
+release](https://github.com/docker/docker/commit/9f482a66ab37ec396ac61ed0c00d59122ac07453)
+and is specified in the [schema 1 manifest definition](./manifest-v2-1.md)
+
+This second schema version has two primary goals. The first is to allow
+multi-architecture images, through a "fat manifest" which references image
+manifests for platform-specific versions of an image. The second is to
+move the Docker engine towards content-addressable images, by supporting
+an image model where the image's configuration can be hashed to generate
+an ID for the image.
+
+# Media Types
+
+The following media types are used by the manifest formats described here, and
+the resources they reference:
+
+- `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
+- `application/vnd.docker.distribution.manifest.v2+json`: New image manifest format (schemaVersion = 2)
+- `application/vnd.docker.distribution.manifest.list.v2+json`: Manifest list, aka "fat manifest"
+- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
+- `application/vnd.docker.container.image.v1+json`: Container config JSON
+
+## Manifest List
+
+The manifest list is the "fat manifest" which points to specific image manifests
+for one or more platforms. Its use is optional, and relatively few images will
+use one of these manifests. A client will distinguish a manifest list from an
+image manifest based on the Content-Type returned in the HTTP response.
+
+## *Manifest List* Field Descriptions
+
+- **`schemaVersion`** *int*
+	
+  This field specifies the image manifest schema version as an integer. This
+  schema uses the version `2`.
+
+- **`mediaType`** *string*
+
+    The MIME type of the manifest list. This should be set to
+    `application/vnd.docker.distribution.manifest.list.v2+json`.
+
+- **`manifests`** *array*
+
+    The manifests field contains a list of manifests for specific platforms.
+
+    Fields of a object in the manifests list are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object. This will generally be
+        `application/vnd.docker.image.manifest.v2+json`, but it could also
+        be `application/vnd.docker.image.manifest.v1+json` if the manifest
+        list references a legacy schema-1 manifest.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+    - **`platform`** *object*
+
+        The platform object describes the platform which the image in the
+        manifest runs on. A full list of valid operating system and architecture
+        values are listed in the [Go language documentation for `$GOOS` and
+        `$GOARCH`](https://golang.org/doc/install/source#environment)
+
+        - **`architecture`** *string*
+
+            The architecture field specifies the CPU architecture, for example
+            `amd64` or `ppc64le`.
+
+        - **`os`** *string*
+
+            The os field specifies the operating system, for example
+            `linux` or `windows`.
+
+        - **`os.version`** *string*
+
+            The optional os.version field specifies the operating system version,
+            for example `10.0.10586`.
+
+        - **`os.features`** *array*
+
+            The optional os.features field specifies an array of strings,
+            each listing a required OS feature (for example on Windows
+            `win32k`).
+
+        - **`variant`** *string*
+
+            The optional variant field specifies a variant of the CPU, for
+            example `armv6l` to specify a particular CPU variant of the ARM CPU.
+
+        - **`features`** *array*
+
+            The optional features field specifies an array of strings, each
+            listing a required CPU feature (for example `sse4` or `aes`).
+
+## Example Manifest List
+
+*Example showing a simple manifest list pointing to image manifests for two platforms:*
+```json
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.docker.image.manifest.v2+json",
+      "size": 7143,
+      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
+      "platform": {
+        "architecture": "ppc64le",
+        "os": "linux",
+      }
+    },
+    {
+      "mediaType": "application/vnd.docker.image.manifest.v2+json",
+      "size": 7682,
+      "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270",
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux",
+        "features": [
+          "sse4"
+        ]
+      }
+    }
+  ]
+}
+```
+
+# Image Manifest
+
+The image manifest provides a configuration and a set of layers for a container
+image. It's the direct replacement for the schema-1 manifest.
+
+## *Image Manifest* Field Descriptions
+
+- **`schemaVersion`** *int*
+	
+  This field specifies the image manifest schema version as an integer. This
+  schema uses version `2`.
+
+- **`mediaType`** *string*
+
+    The MIME type of the manifest. This should be set to
+    `application/vnd.docker.distribution.manifest.v2+json`.
+
+- **`config`** *object*
+
+    The config field references a configuration object for a container, by
+    digest. This configuration item is a JSON blob that the runtime uses
+    to set up the container. This new schema uses a tweaked version
+    of this configuration to allow image content-addressability on the
+    daemon side.
+
+    Fields of a config object are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object. This should generally be
+        `application/vnd.docker.container.image.v1+json`.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+- **`layers`** *array*
+
+    The layer list is ordered starting from the base image (opposite order of schema1).
+
+    Fields of an item in the layers list are:
+    
+    - **`mediaType`** *string*
+    
+        The MIME type of the referenced object. This should
+        generally be `application/vnd.docker.image.rootfs.diff.tar.gzip`.
+    
+    - **`size`** *int*
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+## Example Image Manifest
+
+*Example showing an image manifest:*
+```json
+{
+    "schemaVersion": 2,
+    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
+    "config": {
+        "mediaType": "application/vnd.docker.container.image.v1+json",
+        "size": 7023,
+        "digest": "sha256:b5b2b2c507a0944348e0303114d8d93aaaa081732b86451d9bce1f432a537bc7"
+    },
+    "layers": [
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 32654,
+            "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f"
+        },
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 16724,
+            "digest": "sha256:3c3a4604a545cdc127456d94e421cd355bca5b528f4a9c1905b15da2eb4a4c6b"
+        },
+        {
+            "mediaType": "application/vnd.docker.image.rootfs.diff.tar.gzip",
+            "size": 73109,
+            "digest": "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736"
+        }
+    ],
+}
+```
+
+# Backward compatibility
+
+The registry will continue to accept uploads of manifests in both the old and
+new formats.
+
+When pushing images, clients which support the new manifest format should first
+construct a manifest in the new format. If uploading this manifest fails,
+presumably because the registry only supports the old format, the client may
+fall back to uploading a manifest in the old format.
+
+When pulling images, clients indicate support for this new version of the
+manifest format by sending the
+`application/vnd.docker.distribution.manifest.v2+json` and
+`application/vnd.docker.distribution.manifest.list.v2+json` media types in an
+`Accept` header when making a request to the `manifests` endpoint. Updated
+clients should check the `Content-Type` header to see whether the manifest
+returned from the endpoint is in the old format, or is an image manifest or
+manifest list in the new format.
+
+If the manifest being requested uses the new format, and the appropriate media
+type is not present in an `Accept` header, the registry will assume that the
+client cannot handle the manifest as-is, and rewrite it on the fly into the old
+format. If the object that would otherwise be returned is a manifest list, the
+registry will look up the appropriate manifest for the amd64 platform and
+linux OS, rewrite that manifest into the old format if necessary, and return
+the result to the client. If no suitable manifest is found in the manifest
+list, the registry will return a 404 error.
+
+One of the challenges in rewriting manifests to the old format is that the old
+format involves an image configuration for each layer in the manifest, but the
+new format only provides one image configuration. To work around this, the
+registry will create synthetic image configurations for all layers except the
+top layer. These image configurations will not result in runnable images on
+their own, but only serve to fill in the parent chain in a compatible way.
+The IDs in these synthetic configurations will be derived from hashes of their
+respective blobs. The registry will create these configurations and their IDs
+using the same scheme as Docker 1.10 when it creates a legacy manifest to push
+to a registry which doesn't support the new format.
--- a/serialization.md
+++ b/serialization.md
@@ -1,0 +1,573 @@
+# Docker Image Specification v1.0.0
+
+An *Image* is an ordered collection of root filesystem changes and the
+corresponding execution parameters for use within a container runtime. This
+specification outlines the format of these filesystem changes and corresponding
+parameters and describes how to create and use them for use with a container
+runtime and execution tool.
+
+## Terminology
+
+This specification uses the following terms:
+
+<dl>
+    <dt>
+        Layer
+    </dt>
+    <dd>
+        Images are composed of <i>layers</i>. <i>Image layer</i> is a general
+        term which may be used to refer to one or both of the following:
+
+        <ol>
+            <li>The metadata for the layer, described in the JSON format.</li>
+            <li>The filesystem changes described by a layer.</li>
+        </ol>
+
+        To refer to the former you may use the term <i>Layer JSON</i> or
+        <i>Layer Metadata</i>. To refer to the latter you may use the term
+        <i>Image Filesystem Changeset</i> or <i>Image Diff</i>.
+    </dd>
+    <dt>
+        Image JSON
+    </dt>
+    <dd>
+        Each layer has an associated JSON structure which describes some
+        basic information about the image such as date created, author, and the
+        ID of its parent image as well as execution/runtime configuration like
+        its entry point, default arguments, CPU/memory shares, networking, and
+        volumes.
+    </dd>
+    <dt>
+        Image Filesystem Changeset
+    </dt>
+    <dd>
+        Each layer has an archive of the files which have been added, changed,
+        or deleted relative to its parent layer. Using a layer-based or union
+        filesystem such as AUFS, or by computing the diff from filesystem
+        snapshots, the filesystem changeset can be used to present a series of
+        image layers as if they were one cohesive filesystem.
+    </dd>
+    <dt>
+        Image ID <a name="id_desc"></a>
+    </dt>
+    <dd>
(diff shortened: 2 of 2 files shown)


serialization.md
```diff
@@ -0,0 +1,573 @@
+# Docker Image Specification v1.0.0
```
jonboulle: update

serialization.md
```diff
@@ -0,0 +1,573 @@
...
+2. For every image layer, in order from root ancestor and descending down,
+extract the contents of that layer's filesystem changeset archive into a
+directory which will be used as the root of a container filesystem.
+
+    - Extract all contents of each archive.
+    - Walk the directory tree once more, removing any files with the prefix
+    `.wh.` and the corresponding file or directory named without this prefix.
+
+
+## Implementations
```
jonboulle: remove

serialization.md
```diff
@@ -0,0 +1,573 @@
...
+    }
+}
+```
+
+Every key in this object is the name of a repository, and maps to a collection
+of tag suffixes. Each tag maps to the ID of the image represented by that tag.
+
+## Loading an Image Filesystem Changeset
+
+Unpacking a bundle of image layer JSON files and their corresponding filesystem
```
jonboulle: These files don't follow our [style guide](https://github.com/opencontainers/image-spec#markdown-style) fwiw

manifest.md
```diff
@@ -0,0 +1,272 @@
...
+    
+        The size in bytes of the object. This field exists so that a client
+        will have an expected size for the content before validating. If the
+        length of the retrieved content does not match the specified length,
+        the content should not be trusted.
+    
+    - **`digest`** *string*
+
+        The digest of the content, as defined by the
+        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
```
jonboulle: Let's rather define this within the scope of the spec

jonboulle:
actually before going any further, some big open questions: 
- are we considering this a complete fork of the Docker V2 format, and removing all possible references therein? What are the compatibility requirements (if any?)
- What need is there to preserve language referencing other Docker specifications/technologies, as it does today?

serialization.md
vbatts: ... working on it ...

vbatts:
@jonboulle good question. That topic did not come up during the TOB discussions. In some ways this would be considered a fork, though by defining itself a mediatype, then IIRC it could be sourced _by_ a docker v2.s2 manifest and still have predictable processing of the document.

manifest.md
philips: agreed, we can file an issue and leave this as-is for now, I think.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+        {
+            "mediaType": "application/vnd.oci.image.rootfs.tar.gzip",
+            "size": 73109,
+            "digest": "sha256:ec4b8955958665577945c89419d1af06b5f7636b4ac3da7f12184802ad867736"
+        }
+    ],
+}
+```
+
+# Backward compatibility
```
philips: probably makes sense to drop this section?

serialization.md
```diff
@@ -0,0 +1,564 @@
...
+## Terminology
+
+This specification uses the following terms:
+
+<dl>
+    <dt>
+        Layer
+    </dt>
+    <dd>
+        Images are composed of <i>layers</i>. <i>Image layer</i> is a general
```
philips: needs some markdown love, I am OK merging and cleaning up later.

philips:
@vbatts I closed and reopened this to trigger Travis CI. Seems to work.

vbatts:
cool. I initially did not sign the WIP commit for this reason! :-)

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+- `application/vnd.oci.image.serialization.v1+json`: Container config JSON
+- `application/vnd.oci.image.rootfs.tar.gzip`: "Layer", as a gzipped tar
+
+## Manifest List
+
+The manifest list is the "fat manifest" which points to specific image manifests for one or more platforms.
+Its use is optional, and relatively few images will use one of these manifests.
+A client will distinguish a manifest list from an image manifest based on the Content-Type returned in the HTTP response.
+
+## *Manifest List* Field Descriptions
```
brendandburns: Why italics for Manifest List?  (I might expect code font)

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+- **`schemaVersion`** *int*
```
brendandburns: personally I find inline YAML + comments is more readable than markdown bullets.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+  - **`size`** *int*
+
+    The size in bytes of the object.
+    This field exists so that a client will have an expected size for the content before validating.
+    If the length of the retrieved content does not match the specified length, the content should not be trusted.
+
+  - **`digest`** *string*
+
+    The digest of the content, as defined by the [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
```
brendandburns: "by the Docker Registry V2 ..."

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+  - **`platform`** *object*
+
+    The platform object describes the platform which the image in the manifest runs on.
+    A full list of valid operating system and architecture values are listed in the [Go language documentation for `$GOOS` and `$GOARCH`](https://golang.org/doc/install/source#environment)
```
brendandburns: Why do we use golang's space here?  (indeed why do we restrict OS and ARCH at all?)

manifest.md
brendandburns: ah, nm, I see the "let's define this in the scope of this system" below I'd prefer that too.  Apply that comment here as well.

manifest.md
philips: We can just borrow the language from appc: https://github.com/appc/spec/blame/master/spec/aci.md#L224

We do need to specify a list of well-known OS and Arch combinations, and starting from the Go list is reasonable.

philips: @vbatts I am OK merging this without this tweak. Maybe @brendandburns can come back through and add more specific language from appc.

manifest.md
philips: @brendandburns example?

Personally I am OK as-is and we can reformat later.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+The third goal is to be translatable to the [OpenContainers/runtime-spec](https://github.com/opencontainers/runtime-spec)
+
+# Media Types
+
+The following media types are used by the manifest formats described here, and the resources they reference:
+
+- `application/vnd.oci.image.manifest.v1+json`: Image manifest format
+- `application/vnd.oci.image.manifest.list.v1+json`: Manifest list, aka "fat manifest"
+- `application/vnd.oci.image.serialization.v1+json`: Container config JSON
+- `application/vnd.oci.image.rootfs.tar.gzip`: "Layer", as a gzipped tar
```
duglin: I think we need one for the Image itself,no?  Meaning the tar file for the complete image (all configs+layers, etc...)

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+## Manifest List
+
+The manifest list is the "fat manifest" which points to specific image manifests for one or more platforms.
+Its use is optional, and relatively few images will use one of these manifests.
```
duglin: Let's remove the commentary about "few images will use one of these".  Not really relevant for a spec, plus I'm not sure its true since I know of some folks who are working hard to port the popular images to other platforms.

manifest.md
duglin: We should consider making this a string with proper versioning semantics since I suspect we may have minor releases that are fully backwards compatible while major releases don't necessarily have to be backwards compatible.   Just a single int doesn't allow us to convey that info.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+A client will distinguish a manifest list from an image manifest based on the Content-Type returned in the HTTP response.
+
+## *Manifest List* Field Descriptions
+
+- **`schemaVersion`** *int*
+
+  This field specifies the image manifest schema version as an integer. This
+  schema uses the version `2`.
```
duglin: '2' ?  Even though the MIME types use "v1" ?  Seems kind of confusing.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+- **`mediaType`** *string*
+
+  The MIME type of the manifest list. This should be set to
```
duglin: Let's be a bit more precise w.r.t. RFC2119 stuff.  I'd suggest:
```
This REQUIRED property contains the MIME type of the manifest list.
It MUST be set to `....`
```
Did you mean for it to be that specific string or did you really mean "SHOULD" ?

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+  `application/vnd.oci.image.manifest.list.v1+json`.
+
+- **`manifests`** *array*
+
+  The manifests field contains a list of manifests for specific platforms.
```
duglin:
```
This REQUIRED property contains a list of manifests for specific platforms. While the property MUST be present, the size of the array MAY be zero.
```

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+  Fields of a object in the manifests list are:
+
+  - **`mediaType`** *string*
+
+    The MIME type of the referenced object. This will generally be
```
duglin:
```
This REQUIRED property contains the MIME type of the referenced object.
```

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+    `application/vnd.oci.image.manifest.v1+json`, but it could also
+    be `application/vnd.oci.image.manifest.v1+json` if the manifest
```
duglin: I think there's a typo - you show the same MIME type twice in the comparison.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+    list references a legacy schema-1 manifest.
+
+  - **`size`** *int*
```
duglin: I'd prefer to not have this. Updating a referenced image means we then need to update all possible fat manifests too - which will quickly become a real pain.  Let's just use normal transport specific mechanisms (e.g. http's content-length) if people want to know in advance how big something is.  To use this to verify trust isn't really good/sufficient so implying is it shouldn't be done.  We have other mechanisms for that.

manifest.md
duglin: I agree with @brendandburns let's not use any language specific values - worst case, just copy the info needed into this spec.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+    This field exists so that a client will have an expected size for the content before validating.
+    If the length of the retrieved content does not match the specified length, the content should not be trusted.
+
+  - **`digest`** *string*
+
+    The digest of the content, as defined by the [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+
+  - **`platform`** *object*
+
+    The platform object describes the platform which the image in the manifest runs on.
```
duglin:
```
This REQUIRED property describes the platform which the image in the manifest runs on.
```

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+    A full list of valid operating system and architecture values are listed in the [Go language documentation for `$GOOS` and `$GOARCH`](https://golang.org/doc/install/source#environment)
+
+    - **`architecture`** *string*
+
+        The architecture field specifies the CPU architecture, for example `amd64` or `ppc64le`.
```
duglin: `This REQUIRED property specified the CPU architecture...`

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+    - **`os`** *string*
```
duglin: `This REQUIRED property specifies the operating systems, for example...`

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+        The os field specifies the operating system, for example `linux` or `windows`.
+
+    - **`os.version`** *string*
+
+        The optional os.version field specifies the operating system version, for example `10.0.10586`.
```
duglin: `This OPTIONAL field specifies the operating system version, for example 10.0.10586`

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+    - **`os.features`** *array*
+
+        The optional os.features field specifies an array of strings, each listing a required OS feature (for example on Windows `win32k`).
```
duglin: `This OPTIONAL property specifies an array of strings, each specifying a mandatory OS feature (for example, on Windows win32k`)`

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+    - **`variant`** *string*
+
+        The optional variant field specifies a variant of the CPU, for example `armv6l` to specify a particular CPU variant of the ARM CPU.
```
duglin: `This OPTIONAL property specifies the variant of the CPU, for example 'arm61' to specify a particular CPU variant of the ARM CPU.`

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+    - **`features`** *array*
+
+        The optional features field specifies an array of strings, each listing a required CPU feature (for example `sse4` or `aes`).
```
duglin: `This OPTIONAL property specifies....`

manifest.md
duglin: Why is this property needed?  Don't we already know this JSON is a fat manifest?  Kind of like a word doc saying its a word doc inside of it - feels odd.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux",
+        "features": [
+          "sse4"
+        ]
+      }
+    }
+  ]
+}
```
duglin: Thanks for the complete example, it helps see everything all in one shot.
Couple of questions:
- should there be some kind of URI pointing to where the referenced image can be downloaded from? (optional?)
- should we allow for referenced images to be "in-lined"?  It would be really nice if we could "export" a fat-manifest and share it by just passing around a single file instead of a set of them.  People will probably end up tar'ing them up anyway, so we might as well formalize it.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+```
+
+# Image Manifest
+
+The image manifest provides a configuration and a set of layers for a container image.
+It's the direct replacement for the schema-1 manifest.
+
+## *Image Manifest* Field Descriptions
+
+- **`schemaVersion`** *int*
```
duglin: Like above, let's make this a string (x.y).

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+  This field specifies the image manifest schema version as an integer.
```
duglin: `This REQUIRED property specifies the image manifest schema version.`

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+The manifest list is the "fat manifest" which points to specific image manifests for one or more platforms.
+Its use is optional, and relatively few images will use one of these manifests.
+A client will distinguish a manifest list from an image manifest based on the Content-Type returned in the HTTP response.
+
+## *Manifest List* Field Descriptions
+
+- **`schemaVersion`** *int*
+
+  This field specifies the image manifest schema version as an integer. This
```
jonboulle: Can we go the appc or ociruntime route (`acVersion`, `ociVersion`) - this allows it to double as a magic marker.

Also +1 on using a SemVer string

jonboulle: Actually this raises the question that now `ociVersion` is unambiguous in the `runtime-specs` project.. perhaps one of you maintainers @philips @vbatts can fix that? ;-)

manifest.md
jonboulle: How do we already know it?

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+- **`mediaType`** *string*
+
+  The MIME type of the manifest list. This should be set to
+  `application/vnd.oci.image.manifest.list.v1+json`.
+
+- **`manifests`** *array*
+
+  The manifests field contains a list of manifests for specific platforms.
+
+  Fields of a object in the manifests list are:
```
jonboulle: each object

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+The image manifest provides a configuration and a set of layers for a container image.
+It's the direct replacement for the schema-1 manifest.
+
+## *Image Manifest* Field Descriptions
+
+- **`schemaVersion`** *int*
+
+  This field specifies the image manifest schema version as an integer.
+  This schema uses version `2`.
```
duglin: I really don't understand why we're starting with version 2 instead of 1.  We have an OCI specific mime type so there shouldn't be any overlap with someone else's v1.

jonboulle: Because this is a fork of the Docker V2 manifest. That explains a bunch of the typos etc you're seeing.

manifest.md
jonboulle: +1

manifest.md
duglin: Yea but this is "v1" for us - and we have an OCI defined mime type which should help scope it.

manifest.md
duglin: From the mime type when we downloaded it. Or are we assuming people will be given a random piece of json and from that we should know what it is?  Seems like it'll always be within the context of something else - like an HTTP response (which will have the content-type) or part of some other blob/data.

An aside... this kind of sounds like the outer-most XML element telling you what the doc (json in this case) is about  :-)

manifest.md
philips: A docker v2.2 image is this JSON manifest and then a collection of tarballs that make up the image layers. We might define a format that puts all of these things into a single file but I don't think there is a ton of utility since you sign this json manifest, and this would create a second signable object in the system.

manifest.md
philips: fine by me

manifest.md
philips: agreed, I am fine starting at v2 in the mime types

manifest.md
philips: This is a security mechanism for hashing.

manifest.md
philips: Agreed we need to discuss how we find the referenced images using a discovery mechanism. I think this will be a longer discussion. Please file an issue @duglin.

Currently distribution is out of scope for the image spec. But, this is something the TOB needs to look at fixing soon. I think we should skip this discussion on the initial import pass.

manifest.md
philips: I think starting from v2 is fine

philips:
@vbatts I would personally prefer you do a best effort pass on this then just merge it. Then we can follow back up with smaller PRs to fix all of the smaller issues.

stevvooe:
@philips @jonboulle I understand there may be problems with this document, but if we are importing, shouldn't we have an "untouched" commit? It is easy for small changes to work their way in when formatting is being applied.

philips:
@stevvooe Probably a good idea for vbatts to rebase on top of a clean doc. Sorry vbatts.

manifest.md
stevvooe: The original intent for breakdown here was to avoid the propagation of micro-formats. We also wanted to avoid fragmentation in os/arch representation. Go's values were used here as they provide a reasonably recent topology of architectures that can be mapped into other formats (ie. GNU triplets).

I agree that "can together be considered to describe the syscall ABI this image requires" is a very good start, but some curation is required to ensure compatibility across implementations.

The correct approach, as suggested by @philips, to start a registry of guaranteed compatible platform specifications, along with any aliases. Other implementations may use different values for the fields, but no compatibility or support guarantee is provided unless they become registered.

manifest.md
```diff
@@ -0,0 +1,261 @@
...
+
+The first goal is content-addressable images, by supporting an image model where the image's configuration can be hashed to generate an ID for the image and its components.
+The second goal is to allow multi-architecture images, through a "fat manifest" which references image manifests for platform-specific versions of an image.
+The third goal is to be translatable to the [OpenContainers/runtime-spec](https://github.com/opencontainers/runtime-spec)
+
+# Media Types
+
+The following media types are used by the manifest formats described here, and the resources they reference:
+
+- `application/vnd.oci.image.manifest.v1+json`: Image manifest format
```
stevvooe: If we are starting a fork, should there not be a base commit before changing mediatypes? Do we intend to do a full compatibility break with Docker V2.2?

It might be appropriate to declare a mapping of equivalent mediatypes from OCI. For example V2.2 becomes, V1 of OCi and it is compatible with `application/vnd.docker...`.

manifest.md
stevvooe: There might be the role of a "bundle pack" (similar to `docker save`). Effectively, this is an tar with everything required, such that the contents, identified by hash, would be enough to run the image without network IO or content resolution.

In practice, this tends to be inefficient.

philips: @stevvooe Right, can you file a bug to define something like this? I see it as similar to git packfiles or something.

manifest.md
stevvooe: @duglin Mostly, this functionality would be used by software distributors, which don't represent the bulk of users of the software. It might better to declare that this is intended to distribute cross-platform containers. I'm not sure if this is really in-scope, a _manifest list_ is targeted at distributing several images.

manifest.md
stevvooe: @duglin This already represents the minor version of the spec. Major version is represented within the mediatype. Typically, unless explicitly identified, differing mediatypes are considered incompatible, while a revision here indicate the inclusion of new fields.

Generally, you want to avoid having a manifest know too much about its version. This allows the mediatype to act like a "lense" into the interpretation of a manifest.

serialization.md
stevvooe: @philips Not sure about now, but at the time, definition lists weren't supported by markdown.

manifest.md
stevvooe: Does this mean we intend to drop compatibility?

manifest.md
stevvooe: > should there be some kind of URI pointing to where the referenced image can be downloaded from? (optional?)

This defeats the purpose of providing content addresses. Pushing down the fetch content also over-couples the description of the image from the distribution of the image.

> should we allow for referenced images to be "in-lined"? It would be really nice if we could "export" a fat-manifest and share it by just passing around a single file instead of a set of them. People will probably end up tar'ing them up anyway, so we might as well formalize it.

This was considered. However, it burdens verification of the "in-lined" manifest. The identity becomes dependent on the ability to re-serialize the embedded format. Re-serialization should be avoided at all costs in a robust content-addressable system.

manifest.md
stevvooe: @duglin `Content-Length` is typically insufficient for this purpose. In addition, this specification cannot and should not assume that the underlying distribution mechanism knows about the complete size.

It is much more efficient to mark this during assembly. This is mostly just the size of the manifest json, which is usually small and at hand anyways.

manifest.md
stevvooe: @jonboulle Would it be possible to maintain both? This can be fixed for from the perspective of docker interpretation.

manifest.md
stevvooe: @duglin I agree. This shouldn't have been here.

Let's leave it as a deprecated field.

stevvooe:
@philips @vbatts @duglin @jonboulle I agree with much of the sentiment in the review of this specification, but the V2.2 manifest schema is already quite flexible for the purposes of describing a configuration target and set of image layers. It was designed to be quite generic and went through almost a year of revisions before being merged. Arguably, the flexibility of the format has not yet been fully realized.

While I understand that compatibility may not be the primary goal, let's make sure breaks are based on technical merit. Let's consider defining a baseline that is at least structurally compatible. This would mean we maintain all the fields, as they are. Additions, such as `ociVersion` can be added without breaking the format. We can maintain a registry of backwards compatible media types to the docker equivalent.

Does this seem reasonable?

manifest.md
vbatts: I did commit the two as separate commits first before starting the WIP commit. 70c1c837ec3a4c64adac0ed01a4957964d7f0739 and fd4580cedbaebdc13fedb5b500780d40781cedf5

I'm not sure I follow the compatible thought. Could explain that further?

manifest.md
vbatts: much of this is just carry over styling that has yet to be visited.

manifest.md
vbatts: So, I'm still churning on the the `schemaVersion` carried over here, but it definitely has me reconsidering the `ociVersion` in runtime-spec, and how that file is distinguished (should it now have a mediatype?). `ociVersion` is not really clear in this regard.

manifest.md
vbatts: oh. I hadn't gotten to all these references yet, and figured starting the mime type at 2 would be confusing for a v1.0 release :-)

vbatts:
@stevvooe understandable that fields can be added without breaking, but regardless I think this would be a new mediatype and binding ourselves to not breaking the docker v2.s2 format could limit conversation. By having a departure into its own mediatype, it allows to build on the year of iteration and not limit ourselves.

Also, having the "untouched" commits, they are here c22ca79 and c1ff36f

Some points, even like the ".wh." diff type rootfs is something that will be revisited.

philips:
@vbatts +1 on a new media type.

If at all possible I would like to keep the image layers bit-for-bit compatible for the initial v1 release to make it possible for people to use existing s3 objects for layers.

manifest.md
vbatts: sorry. I'm missing the reference on the preferred wording here

vbatts:
@philips bit-for-bit compatible [with the docker v2.s2] for the initial v1 release [of the oci image spec] ?

philips:
@vbatts It would be ideal if the serialization format for the layer was bit-for-bit compatible for the sanity of the container registries like gcr.io, quay, hub, etc so they don't need to duplicate 98% of the data to change some JSON manifest. If necessary we can make changes to the manifest.

philips:
LGTM for now. This discussion has gotten too splintered to be useful.

vbatts:
hold on. I'm trying to get the simple line cleanups done. Then I'll take
the WIP off.

vbatts:
All. I've rebased this PR to be _only_ the initial import of the the specs cited, and opened https://github.com/opencontainers/image-spec/pull/12 for the adjustments.

philips:
LGTM!

stevvooe:
@vbatts Reading back, I wasn't very clear above. I agree that we should have new mediatypes. The spec would have a table declaring which mediatypes are compatible with the docker mediatypes.

I don't want to limit the conversation but let's make sure that if a break is done, there is a _very good_ technical reason for doing so.

LGTM

philips:
I am going to click the button tomorrow morning unless anyone has objections.

PR #12, opened 2016-04-07
Begin adjustments on the imported specs

vbatts:
As #6 was tangled up with WIP modifications, this splits out into a new PR all the modifications.

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -4,113 +4,90 @@ draft = true
 +++
 <![end-metadata]-->
 
-# Image Manifest Version 2, Schema 2
+# Image Manifest
 
-This document outlines the format of of the V2 image manifest, schema version 2.
-The original (and provisional) image manifest for V2 (schema 1), was introduced
-in the Docker daemon in the [v1.3.0
-release](https://github.com/docker/docker/commit/9f482a66ab37ec396ac61ed0c00d59122ac07453)
-and is specified in the [schema 1 manifest definition](./manifest-v2-1.md)
-
-This second schema version has two primary goals. The first is to allow
-multi-architecture images, through a "fat manifest" which references image
-manifests for platform-specific versions of an image. The second is to
-move the Docker engine towards content-addressable images, by supporting
-an image model where the image's configuration can be hashed to generate
-an ID for the image.
+There are three main goals of the Image Manifest Specification.
+The first goal is content-addressable images, by supporting an image model where the image's configuration can be hashed to generate a unique ID for the image and its components.
+The second goal is to allow multi-architecture images, through a "fat manifest" which references image manifests for platform-specific versions of an image.
+The third goal is to be translatable to the [OpenContainers/runtime-spec](https://github.com/opencontainers/runtime-spec)
 
 # Media Types
 
-The following media types are used by the manifest formats described here, and
-the resources they reference:
+The following media types are used by the manifest formats described here, and the resources they reference:
 
-- `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
-- `application/vnd.docker.distribution.manifest.v2+json`: New image manifest format (schemaVersion = 2)
-- `application/vnd.docker.distribution.manifest.list.v2+json`: Manifest list, aka "fat manifest"
-- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
-- `application/vnd.docker.container.image.v1+json`: Container config JSON
+- `application/vnd.oci.image.manifest.list.v1+json`: Manifest list, aka "fat manifest"
+- `application/vnd.oci.image.manifest.v1+json`: Image manifest format
+- `application/vnd.oci.image.rootfs.tar.gzip`: "Layer", as a gzipped tar archive
+- `application/vnd.oci.image.serialization.v1+json`: Container config JSON
 
 ## Manifest List
 
-The manifest list is the "fat manifest" which points to specific image manifests
-for one or more platforms. Its use is optional, and relatively few images will
-use one of these manifests. A client will distinguish a manifest list from an
-image manifest based on the Content-Type returned in the HTTP response.
+The manifest list is the "fat manifest" which points to specific image manifests for one or more platforms.
+While the use of a manifest list is OPTIONAL for image providers, image consumers SHOULD be prepared to process them.
+A client will distinguish a manifest list from an image manifest based on the Content-Type returned in the HTTP response.
 
 ## *Manifest List* Field Descriptions
 
 - **`schemaVersion`** *int*
-	
-  This field specifies the image manifest schema version as an integer. This
-  schema uses the version `2`.
+
+  This REQUIRED property specifies the image manifest schema version.
+  This schema uses the version `2`.
 
 - **`mediaType`** *string*
 
-    The MIME type of the manifest list. This should be set to
-    `application/vnd.docker.distribution.manifest.list.v2+json`.
+  This REQUIRED property contains the MIME type of the manifest list.
+  For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.list.v1+json`.
 
 - **`manifests`** *array*
 
-    The manifests field contains a list of manifests for specific platforms.
+  This REQUIRED property contains a list of manifests for specific platforms.
+  While the property MUST be present, the size of the array MAY be zero.
 
-    Fields of a object in the manifests list are:
-    
-    - **`mediaType`** *string*
-    
-        The MIME type of the referenced object. This will generally be
-        `application/vnd.docker.image.manifest.v2+json`, but it could also
-        be `application/vnd.docker.image.manifest.v1+json` if the manifest
-        list references a legacy schema-1 manifest.
-    
-    - **`size`** *int*
-    
-        The size in bytes of the object. This field exists so that a client
-        will have an expected size for the content before validating. If the
-        length of the retrieved content does not match the specified length,
-        the content should not be trusted.
-    
-    - **`digest`** *string*
+  Fields of each object in the manifests list are:
 
-        The digest of the content, as defined by the
-        [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
+  - **`mediaType`** *string*
 
-    - **`platform`** *object*
+    This REQUIRED property contains the MIME type of the referenced object.
+    (i.e. `application/vnd.oci.image.manifest.v1+json`)
 
-        The platform object describes the platform which the image in the
-        manifest runs on. A full list of valid operating system and architecture
-        values are listed in the [Go language documentation for `$GOOS` and
-        `$GOARCH`](https://golang.org/doc/install/source#environment)
+  - **`size`** *int*
 
-        - **`architecture`** *string*
+    This REQUIRED property specifies the size in bytes of the object.
+    This field exists so that a client will have an expected size for the content before validating.
+    If the length of the retrieved content does not match the specified length, the content should not be trusted.
 
-            The architecture field specifies the CPU architecture, for example
-            `amd64` or `ppc64le`.
+  - **`digest`** *string*
 
-        - **`os`** *string*
+    The digest of the content, as defined by the [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter).
 
-            The os field specifies the operating system, for example
-            `linux` or `windows`.
+  - **`platform`** *object*
 
-        - **`os.version`** *string*
+    This REQUIRED property describes the platform which the image in the manifest runs on.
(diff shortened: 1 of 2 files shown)


philips:
@vbatts can you rebase on top of master?

vbatts:
done

manifest.md
```diff
@@ -4,124 +4,100 @@ draft = true
...
+A client will distinguish a manifest list from an image manifest based on the Content-Type returned in the HTTP response.
 
 ## *Manifest List* Field Descriptions
 
 - **`schemaVersion`** *int*
-	
-  This field specifies the image manifest schema version as an integer. This
-  schema uses the version `2`.
+
+  This REQUIRED property specifies the image manifest schema version.
```
stevvooe: This reset will make the format incompatible with Docker's manifest format. I'd rather we maintain this at 2 and deprecate the field.

serialization.md
```diff
@@ -422,14 +362,13 @@ bin/my-app-binary
 bin/my-app-tools
 ```
 
-The TarSum checksum for the archive file is then computed and placed in the
-JSON metadata along with the execution parameters.
+The TarSum checksum for the archive file is then computed and placed in the JSON metadata along with the execution parameters.
```
stevvooe: I'm not sure if this is in scope for this PR, but a TarSum isn't used here. This should probably just say `digest`.

serialization.md
```diff
@@ -1,10 +1,7 @@
-# Docker Image Specification v1.0.0
```
stevvooe: Unfortunately, this specification is fairly out of date after the Docker 1.10 changes. Most of these changes are documented [here](https://gist.github.com/aaronlehmann/b42a2eaf633fc949f93b). We are going to update this to correspond with the current implementation described in the referenced document, which can be used as a more up to date base.

Given that, it seems prudent to keep this PR focused on the manifest specification.

serialization.md
stevvooe: Please see https://github.com/opencontainers/image-spec/pull/12/files#r59071393. Unfortunately, this specification is out of date.

manifest.md
vbatts: k. changed.

philips: We will need a note as to why it is at 2. Leaving it at 2 is fine by me.

serialization.md
vbatts: ah nice. That should probably be removed from the distribution spec too. :-)

manifest.md
vbatts: perhaps as a part of the compatibility matrix/table

serialization.md
stevvooe: @vbatts That work item is https://github.com/docker/docker/issues/21938, for reference.

vbatts: ah. nice.

manifest.md
```diff
@@ -246,39 +215,22 @@ image. It's the direct replacement for the schema-1 manifest.
...
-format involves an image configuration for each layer in the manifest, but the
-new format only provides one image configuration. To work around this, the
-registry will create synthetic image configurations for all layers except the
-top layer. These image configurations will not result in runnable images on
-their own, but only serve to fill in the parent chain in a compatible way.
-The IDs in these synthetic configurations will be derived from hashes of their
-respective blobs. The registry will create these configurations and their IDs
-using the same scheme as Docker 1.10 when it creates a legacy manifest to push
-to a registry which doesn't support the new format.
+The registry will continue to accept uploads of manifests in both the old and new formats.
```
stevvooe: After some thought, we may need to word this in a way that doesn't assume the existence of a registry. We can use the `Accept`/`Allow` headers to frame a fallback story, but we should talking about expressing compatibility in transport agnostic manner.

The behavior here is going to be very important to ensure we have a specification that is beneficial to the community of users. Part of the problem here is that content-addressability (content security, really), needs to be balanced against backwards compatibility and usability.

philips: @stevvooe I think we can just refer to the non-existent mime-type table: https://github.com/opencontainers/image-spec/issues/14

Then add concrete language around handling those mime-types in the optional distribution layer: https://github.com/opencontainers/image-spec/issues/15

stevvooe: We really need to add distribution-agnostic fallback. A mimetype table and optional distribution layer aren't really sufficient for describing fallback that will work across transports. Otherwise, we risk losing compatibility across transports.

Perhaps, this section should describe the structure of the table.

philips: @stevvooe It is a bit hard to design with a strawman of "distribution-agnostic". Do you have a concrete example?

serialization.md
philips: @vbatts are you going to remove this then?

philips:
overall LGTM.

serialization.md
vbatts: Oh sure I can. This `serialization.md` needs loads more. It seemed like we may punt, to make headway on `manifest.md` first.
I'll do a once-over of this doc right now.

manifest.md
stevvooe: I'm not sure if "distribution-agnostic" or "transport-agnostic" is a "strawman". It simply means to describe mediatype fallback outside of the context of http, which may refer to `Accept` headers. A concrete example would be if `application/vnd.oci.image.manifest.v1+json` is encountered, and only `application/vnd.docker.image.manifest.v2+json` is supported, what should be done? 

Another issue is expressing a fallback precedence. For example, one might declare they can handle `application/vnd.oci.image.manifest.v1+json` and `application/vnd.docker.image.manifest.v2+json` but only supports the latter. Both are provided. Should the client convert or fallback to the format they support natively?

philips: Do you have an example transport in mind? I am confused what you are trying to ask for here. Perhaps we can move it to a separate issue?

Do you just want language like what is in the HTTP spec about quality factors?

```
Quality factors allow the user or user agent to indicate the relative degree of preference for that media-range, using the qvalue scale from 0 to 1 (section 3.9). The default value is q=1.

The example

       Accept: audio/*; q=0.2, audio/basic
SHOULD be interpreted as "I prefer audio/basic, but send me any audio type if it is the best available after an 80% mark-down in quality."
```

stevvooe: This section discusses fallback in the context of http. Let's remove the transport-specific items and describe what was here generically. 

I'm really confused as to why this is an unreasonable request for this stage of the specification process. If you believe this is outside the scope of this PR, we can file a separate issue, but let's make sure it is followed up.

manifest.md
```diff
@@ -4,124 +4,100 @@ draft = true
...
-and is specified in the [schema 1 manifest definition](./manifest-v2-1.md)
+# Image Manifest
 
-This second schema version has two primary goals. The first is to allow
-multi-architecture images, through a "fat manifest" which references image
-manifests for platform-specific versions of an image. The second is to
-move the Docker engine towards content-addressable images, by supporting
-an image model where the image's configuration can be hashed to generate
-an ID for the image.
+The first goal is content-addressable images, by supporting an image model where the image's configuration can be hashed to generate an ID for the image and its components.
```
duglin: Let's add some kind of intro sentence to this section - even something as simple as:
```
There are three main goals of the Image Manifest.
```

duglin: s/generate an ID/generate a unique ID/

manifest.md
```diff
@@ -4,124 +4,100 @@ draft = true
...
-The following media types are used by the manifest formats described here, and
-the resources they reference:
+The following media types are used by the manifest formats described here, and the resources they reference:
 
-- `application/vnd.docker.distribution.manifest.v1+json`: schema1 (existing manifest format)
-- `application/vnd.docker.distribution.manifest.v2+json`: New image manifest format (schemaVersion = 2)
-- `application/vnd.docker.distribution.manifest.list.v2+json`: Manifest list, aka "fat manifest"
-- `application/vnd.docker.image.rootfs.diff.tar.gzip`: "Layer", as a gzipped tar
-- `application/vnd.docker.container.image.v1+json`: Container config JSON
+- `application/vnd.oci.image.manifest.v1+json`: Image manifest format
```
duglin: Minor thing, but can order these so that we start at the outer most wrapper and work our way in... so:
- fat manifest
- single image manifest
- layer
- container config

It just makes it easier to read/follow.

manifest.md
```diff
@@ -4,124 +4,100 @@ draft = true
...
+- `application/vnd.oci.image.rootfs.tar.gzip`: "Layer", as a gzipped tar
 
 ## Manifest List
 
-The manifest list is the "fat manifest" which points to specific image manifests
-for one or more platforms. Its use is optional, and relatively few images will
-use one of these manifests. A client will distinguish a manifest list from an
-image manifest based on the Content-Type returned in the HTTP response.
+The manifest list is the "fat manifest" which points to specific image manifests for one or more platforms.
+Its use is optional, and relatively few images will use one of these manifests.
```
duglin: I'd prefer to remove this editorial comment since its not really that true for some people who work in multi-arch envs a lot.  Perhaps just:
```
While the use of a manifest list is OPTIONAL for image providers, image consumers SHOULD be prepared to process them.
```

manifest.md
```diff
@@ -4,124 +4,100 @@ draft = true
...
-  schema uses the version `2`.
+
+  This REQUIRED property specifies the image manifest schema version.
+  This schema uses the version `2`.
 
 - **`mediaType`** *string*
 
-    The MIME type of the manifest list. This should be set to
-    `application/vnd.docker.distribution.manifest.list.v2+json`.
+  This REQUIRED property contains the MIME type of the referenced object.
```
duglin: s/referenced object/manifest list/

manifest.md
```diff
@@ -4,124 +4,100 @@ draft = true
...
+  This should be set to `application/vnd.oci.image.manifest.list.v1+json`.
```
duglin: `For this version of the specification, this MUST be set to ...`

manifest.md
```diff
@@ -4,124 +4,100 @@ draft = true
...
-    
-    - **`size`** *int*
-    
-        The size in bytes of the object. This field exists so that a client
-        will have an expected size for the content before validating. If the
-        length of the retrieved content does not match the specified length,
-        the content should not be trusted.
-    
-    - **`digest`** *string*
+  Fields of each object in the manifests list are:
```
duglin: This seems to assume that there is no in-lining of the referenced images - meaning, we're not defining a way for someone to download a *single* entity that contains the fat manifest plus all referenced images. Was this intentional? I think it would be good to allow for the sharing of single entities instead of forcing people to always break things up.

manifest.md
vbatts: alrighty.

manifest.md
vbatts: ah yeah.

manifest.md
vbatts: alrighty.

manifest.md
vbatts: alrighty. And I'll carry this over to similar sentences.

manifest.md
vbatts: alrighty.

manifest.md
vbatts: Are you saying that the `manifests` array is only used when referencing multiple objects, otherwise the other mediatypes be sub-structures of in this manifest-list document?

duglin: This wasn't really a statement about the manifests array itself, its just that this is where I was in the doc when it dawned on me that downloading this manifest json by itself feels like the main usecase being promoted - so I started to wonder about the "big bundle" case.  So, should we consider creating a blob (tar) containing this manifest and all referenced images as files in there?

vbatts: This is the confusion around having 0 mapping to even an OPTIONAL api, like
https://github.com/opencontainers/image-spec/issues/15

manifest.md
```diff
@@ -4,124 +4,101 @@ draft = true
...
+- `application/vnd.oci.image.rootfs.tar.gzip`: "Layer", as a gzipped tar archive
+- `application/vnd.oci.image.serialization.v1+json`: Container config JSON
 
 ## Manifest List
 
-The manifest list is the "fat manifest" which points to specific image manifests
-for one or more platforms. Its use is optional, and relatively few images will
-use one of these manifests. A client will distinguish a manifest list from an
-image manifest based on the Content-Type returned in the HTTP response.
+The manifest list is the "fat manifest" which points to specific image manifests for one or more platforms.
```
stevvooe: We can deprecate the term "fat manifest". I'm not sure if it helps the understanding here.

manifest.md
stevvooe: If we do embedding, it should probably be with a `data` attribute, with base64 encoded data. There are problems with pulling the full tar up into something like that, but intermediates, like manifests and config, can see gains from this.

philips:
@vbatts Can we please just merge this, @stevvooe has some additional comments but we can fix those elsewhere. I have some other fixes I want to make but this is going to conflict with this huge change.

jonboulle:
Needs a rebase before merging.

vbatts:
rebased. merging now.

Issue #7, opened 2016-04-04
JSON schema all the things


vbatts:
Carrying over the approach from https://github.com/opencontainers/runtime-spec/tree/v0.4.0/schema
We might ought to have JSON schema for the structures defined in this project and easy validation of them.

vbatts:
do other @opencontainers/image-spec-maintainers have an opinion on this?

vbatts:
I do like this from the swagger/openAPI https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#info-object

vbatts:
Now that #18 is merged, next steps might be:
* [ ] Schema for the serialization document
* [ ] Decide best way to have common definitions across OCI projects (i.e. `defs.json`, etc)
* [ ] Validation tooling for an OC image archive (e.g. https://github.com/opencontainers/ocitools )

philips:
@vbatts can you give an update on where this is now? I think serialization doc is now covered, right?

philips:
@vbatts perhaps open a new issue for sharing and then we can close this out?

vbatts:
There is nothing done for the serialization.md file yet, but I can track that separately. It's a more intricate document than the manifests.

PR #18, opened 2016-04-12
Json schema for image manifest and manifest list

vbatts:
related to #7 

testing against https://gist.github.com/vbatts/b3e4fc231fbc92a9cbc28b9184f0b2d2

### DIFF
--- a/schema/defs-image.json
+++ b/schema/defs-image.json
@@ -6,6 +6,11 @@
       "type": "string",
       "pattern": "^[a-z]+/[0-9a-zA-Z.+]+$"
     },
+    "digest": {
+      "description": "the cryptographic checksum digest of the object, in the pattern '<hash>:<hexadecimal digest>'",
+      "type": "string",
+      "pattern": "^[a-z0-9]+:[a-fA-F0-9]+$"
+    },
     "blobObject": {
       "id": "https://opencontainers.org/schema/image/blobObject",
       "type": "object",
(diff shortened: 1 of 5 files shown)


wking:
On Tue, Apr 12, 2016 at 11:45:14AM -0700, Vincent Batts wrote:
>   * schema: copy common from runtime
> …
>     A schema/defs.json (160)

While it's nice to have local copies of this sort of thing for offline
use, it's more DRY and probably easier to maintain to just $ref the
runtime-spec defs.json.  Stuff like:

  "$ref": "https://raw.githubusercontent.com/opencontainers/runtime-spec/v0.5.0/schema/defs.json#/definitions/mapStringString"

should work fine, although it's probably nicer if we just host these
under opencontainers.org/schema/… or some such.

vbatts:
I mulled on that too. This is a first step.

philips:
LGTM, can you file issues for steps after this?

schema/defs-image.json
```diff
@@ -0,0 +1,99 @@
...
+      "id": "https://opencontainers.org/schema/image/mediaType",
+      "type": "string",
+      "pattern": "^[a-z]+/[0-9a-zA-Z.+]+$"
+    },
+    "digest": {
+      "description": "the cryptographic checksum digest of the object, in the pattern '<hash>:<hexadecimal digest>'",
+      "type": "string",
+      "pattern": "^[a-z0-9]+:[a-fA-F0-9]+$"
+    },
+    "blobObject": {
```
stevvooe: This is actually called a "Descriptor".

vbatts: where is that defined?

stevvooe: This is what we have called it from the start. The godoc is [here](https://godoc.org/github.com/docker/distribution#Descriptor). I am not sure how it never made into this specification.

Other types with further fields, such as the manifest descriptor, are extensions to the descriptor type.

vbatts: ah fine. A name of a struct. Feel free to open a PR to include this in the
spec. I can update it here.

vbatts:
haha. I see I added the definitions for the manifest list, but didn't `git add` the actual schema for the manifest list. smh

schema/defs-image.json
stevvooe: Yea, I'll do that. Just reviewing old notifications.

PR #19, opened 2016-04-13
Fix up a pointer to the docker v1 spec

duglin:
I'm not sure this is correct, but the ref to "manifest-v2-1.md" was returning a 404 and I think it should point to the docker image v1 spec - at the same time I also renamed "serialization.md" to "docker-image-v1.0.0" since that's more descriptive - assuming that's the right name for it.

Signed-off-by: Doug Davis <dug@us.ibm.com>


manifest.md
```diff
@@ -10,7 +10,7 @@ This document outlines the format of of the V2 image manifest, schema version 2.
 The original (and provisional) image manifest for V2 (schema 1), was introduced
 in the Docker daemon in the [v1.3.0
 release](https://github.com/docker/docker/commit/9f482a66ab37ec396ac61ed0c00d59122ac07453)
-and is specified in the [schema 1 manifest definition](./manifest-v2-1.md)
+and is specified in the [schema 1 manifest definition](./docker-image-v1.0.0.md)
```
vbatts: ah, this should just be `manifest.md`

duglin: So it should point to itself?

vbatts: ah. this whole sentence could be axed.

philips: +1 on axing it

philips:
@duglin Ping. Otherwise I will fix this myself today.

duglin:
I think it's already been done, no?

philips:
@duglin wait, what? You renamed serialization to docker v1? I don't think that was the goal. I thought you were just going to remove that sentence.  https://github.com/opencontainers/image-spec/pull/19#discussion_r59558935

duglin:
yea I was focusing on that sentence too but it appears to have vanished from master - unless I'm looking at the wrong spot

philips:
ah, in that case I will close this out.

duglin:
ok

PR #21, opened 2016-04-14
README: update to point to GitHub milestones

philips:
Open question is whether we should just delete this section and replace it with milestones all together.


jonboulle:
> Open question is whether we should just delete this section and replace it with milestones all together.

Yes, let's do that once we create milestones with appropriate issues, but this LGTM for now.

PR #26, opened 2016-04-14
Specify which artifact.

timthelion:


philips:
Hrm, @vbatts @stevvooe at some conversation on this repo, which I can't find, I thought tarsum was droppable. Was this the field?

serialization.md
```diff
@@ -59,7 +59,7 @@ This specification uses the following terms:
         Image Checksum
     </dt>
     <dd>
-        The checksum is a cryptographic digest of the artifact.
+        The checksum is a sha256 cryptographic digest of the layer.tar file. This is a unicode string of the format "tarsum.v1+sha256:<checksum>".
```
vbatts: no more tarsum!

vbatts:
@timthelion super glad to see you pile in. Truly. The current state of these docs is a carry-over from the docker/docker and docker/distribution repos. Much updating and amending is needed. So, like you are finding and opening issues/PRs for, the presence of some of the specifics does not mean final release state. :smile_cat:

stevvooe:
@philips We should not mention tarsum, if at all possible. I think `checksum` was the storage location for digests before some of the v2 overhaul.

That whole spec needs an update, which we are in the process of getting together for OCI. I am unsure if any updates are worthwhile until then (but thanks for reading!).

timthelion:
@stevvooe so `checksum` should become a no-op perhaps?

philips:
@stevvooe Huh? Can you share more information on what needs a full overhaul?

stevvooe:
@philips With 1.10, we overhauled the approach to image and layer hashes to ensure the merkle DAG provided a strong guarantee, end to end. The changes are described [here](https://gist.github.com/aaronlehmann/b42a2eaf633fc949f93b#new-image-config) and they need to be merged with the current image specification.

We've filed an issue to get this resolved here: https://github.com/docker/docker/issues/21938 and are working on it now. Effectively, that gist needs to be merged with the spec. We should work from the result of that merge.

I looked into the use of `checksum`. It is only used for v1 pulls and is set [here](https://github.com/docker/docker/blob/master/distribution/push_v1.go#L441) and calculated [here](https://github.com/docker/docker/blob/master/registry/session.go#L549).  It contains the concatenation of layer checksum and the image json, calculated after v1 push and is actually a tarsum in that case.

philips:
@stevvooe What is the timeline for this 1.10 work? If it is longer than a week or so we can just proceed with the previous edition in the OCI.

stevvooe:
@philips We're going to have a PR out this week.

philips:
I believe this is fixed in https://github.com/opencontainers/image-spec/pull/45. Closing for that.

PR #29, opened 2016-04-14
Specify process for naming future architectures

timthelion:


philips:
Not opposed but I am guessing you meant for this to be on the arch line, not the OS line?

timthelion:
Oops, fixed...

serialization.md
```diff
@@ -167,7 +167,8 @@ Here is an example image JSON file:
             <li>amd64</li>
             <li>arm</li>
         </ul>
-        More values may be supported in the future and any of these may or may not be supported by a given container runtime implementation.
+        More values may be supported in the future and any of these may or may not be supported by a given container runtime implementation. Future values will be consistent with those already canonicalized by the <a href="https://www.debian.org/ports/">Debian ports list</a>.
```
jonboulle: this is already kinda broken by "arm", right? :/

timthelion: Well Debian doesn't use `386` but `i386`. They do seem to have an architecture named `arm`, there are lost of Debian arm architectures based on the build flags. I don't demand that we follow Debian's lead, but it was the closest thing that I could find to a standard for specifying actual BUILD TIME architectures! I think that it is good to copy someone, even if there are some differences in the naming. I think that it is better than to just come up with random names for things. And actually, I think that in the case of arm, following Debian's standard brings along a lot of valulable technical know-how (AKA, the taxonomy of arm build setups ;) ).

timthelion: Indeed, for the sake of Arm, I think that it is necesary to either copy either Debian or Fedora, otherwise, we won't take into acount the specifics and there will be incompatibility problems at the kernel level. AKA, not all "ARM" images will run on all "ARM" kernels.

timthelion:
I just realized that the manifest spec already specifies the use of the official list for $GOARCH. https://golang.org/doc/install/source#environment

jonboulle:
You're correct. The problem is, the `GOARCH` values are not specific enough by themselves (e.g. for ARM, necessitating the [additional `GOARM` variable](https://github.com/golang/go/wiki/GoArm)), which is part of the reason we ended up avoiding using them directly in appc. I don't think we got it entirely right there, but I think there are some learnings we can take from it: e.g. [armv7l vs armv7b](https://github.com/appc/spec/pull/304), [arm64 vs aarch64](https://github.com/appc/spec/pull/297#commitcomment-11287834))

The manifest spec (which as you noted already uses the `GOARCH` values) [attempts to take this into consideration with the additional `variant` property](https://github.com/opencontainers/image-spec/blame/master/manifest.md#L84), but this just raises the point that there's a mismatch between here and there.

timthelion:
@jonboulle is the debian ports list specific enough? Pardon me, that I have no actual knowledge of the Arms race or whatever we're going to call this conundrum.

jonboulle:
@timthelion well, from the brief exchange we had with ARM engineers (see below), I had thought that endianness was a requirement, but I also am not extremely well versed in this area

> On ARM64 Linux, "uname -m" reports "aarch64". When you run a 32-bit on arm64, it still reports "aarch64" unless you run it with a 32-bit personality. So "linux32 uname -m" reports "armv8l" for backwards compatibility (independently of whether uname is a 32 or 64-bit application). On big-endian, Linux reports "aarch64_be" and "armv8b" respectively.

> For the older architectures, using the output of ‘uname –m’ would also make sense and negate the armel/armhf/armv7l confusion. Doing this would mean ‘armv7l' would be the correct “tag” to use for little-endian, and ‘armv7b’ for big endian.

timthelion:
Perhaps this would be another possible source for standard names: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/arch

wking:
On Fri, Apr 15, 2016 at 05:56:00AM -0700, Timothy Hobbs wrote:
> Perhaps this would be another possible source for standard names:
> https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/arch

Not unless you want to dig down in there to find obscure architectures
like x86_64 [1] ;).

[1]: https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/arch/x86/Kconfig?h=v4.5#n13

wking:
Maybe just reference [1] here, and move this conversation over to that
repository?  I don't think we want separate versions of this
conversation in each repository, and that's the lower-level spec.

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#platform-specific-configuration

stevvooe:
@jonboulle We studied this problem early on in OCI and during multi-arch considerations for the V2.2 manifest. Indeed, `GOARCH` is insufficient, but it is a good starting point.

The intent with these field sets was to define a class of machines on which the image would run. The real work here is defining the classes and aliasing in such way to make these fields useful in multi-arch scenarios. In some case, this may actually be better left to implementations, leaving OCIs role to define the common architectures which _must_ be compatible (amd64, etc.), along with the data model and process for expressing the os/architecture (platform) for an image.

That said, we can say that these are seeded by `GOARCH`, but not limited by.

Ultimately, this is going to have to be a real community effort and that will be as successful as the process we adopt.

serialization.md
```diff
@@ -167,7 +167,9 @@ Here is an example image JSON file:
             <li>amd64</li>
             <li>arm</li>
         </ul>
-        More values may be supported in the future and any of these may or may not be supported by a given container runtime implementation.
+        A full list of possible values can be found in the <a href=https://golang.org/doc/install/source#environment>Go language documentation for $GOOS and $GOARCH</a>.
+        Any of these may or may not be supported by a given container runtime
```
vbatts: perhaps pull this sentence up to be all on one line.

serialization.md
```diff
@@ -180,8 +182,9 @@ Here is an example image JSON file:
             <li>freebsd</li>
             <li>linux</li>
         </ul>
-        More values may be supported in the future and any of these may or may
-        not be supported by a given container runtime implementation.
+        A full list of values can be found in the <a href=https://golang.org/doc/install/source#environment>Go language documentation for $GOOS and $GOARCH</a>.
+        Any of these may or may not be supported by a given container runtime
```
vbatts: ditto

vbatts:
This seems like something that could utilize `annotations` of a prescribed namespace.
This PR seems LGTM, and we can iron out particular hardware details as needed.

philips:
@vbatts I would rather this spec be authoritative on archs and OS strings than not. This will be an important namespace for the spec to own.

If we reference the GOARCH / GOOS spec I think it should be more along the lines of "New architectures and OSes will be added over time and through discussion on the OCI Image Project. A source for inspiration for new entries is the Go docs etc etcd"

stevvooe:
@philips While agree we need to call these out here, it is unlikely that anyone will ever "own" the platform namespace ;).

Deferring to GOOS/GOARCH as the inspiration, but not the canonical source, seems to be the right balance.

philips:
@stevvooe So, are you saying you are OK merging as-is? Or you want it to be more along the lines of "future entries should be inspired by". I am not seeing what you want.

stevvooe:
@philips I agree with "future entries should be inspired by" and https://github.com/opencontainers/image-spec/pull/29#issuecomment-215577479. We're going to have to curate these and I don't see a way around it.

I'm not sure about merging this PR. We can merge this but then it will be rolled back later.

philips:
@stevvooe Ack.

@timthelion Sorry, but would you mind making the language be more like what @stevvooe suggests? If our indecisiveness is getting too annoying LMK and I can fixup the language and merge on your behalf.

philips:
Closing in favor of: https://github.com/opencontainers/image-spec/pull/60

PR #33, opened 2016-04-15
schema: switch Object to Descriptor

vbatts:

### DIFF
--- a/schema/defs-image.json
+++ b/schema/defs-image.json
@@ -11,8 +11,8 @@
       "type": "string",
       "pattern": "^[a-z0-9]+:[a-fA-F0-9]+$"
     },
-    "blobObject": {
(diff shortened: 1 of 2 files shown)


stevvooe:
Bah, you beat me to it, since I had to read about json-schema... 😺 

Here is my commit message for anyone who is curious about reasoning here:

```
    schema: change name of blobObject to descriptor
    
    Historically, the structure that describes a content target with a
    mediaType, size and digest has been known as a "descriptor". This is to
    differentiate it from actual blobObject when used in context. By
    differentiating the pointer from the target, discussing the role of a
    "descriptor" becomes much clearer.
    
    Other types, such as "manifestObject", actually inherit fields from the
    descriptor, but there doesn't seem to be a way to represent this in
    JSONSchema at this time.
```

LGTM

philips:
LGTM

PR #40, opened 2016-04-22
schema: use correct reference for manifestDescriptor

stevvooe:
cc @vbatts 

Signed-off-by: Stephen J Day <stephen.day@docker.com>

PR #41, opened 2016-04-22
examples: validate markdown examples

### LINKED ISSUE #34: Create golden standards to test against json schema
We need a few "golden standards" to test against the json schema specs to catch unintentional changes. It will also help in having them as part of PRs that show the schema change and how it affects the "golden standard".

@vbatts Thoughts?

stevvooe:
To ensure that examples in the specification match up with the json
schemas, we propose a validation approach that will extract the fenced
code blocks from the markdown document and validate them. Examples are
marked with specification mediatypes which are then validated against
the respective json shema doc.

From here, we can propose that all examples be part of the
specification. Each act of clarification would be represented within the
specification and validated against the schema.

The tool can be run from the root repository directory with the
following command:

```console
$ go run examples.go < manifest.md
ok	 application/vnd.oci.image.manifest.list.v1+json
ok	 application/vnd.oci.image.manifest.v1+json
```

This is a work in progress to validate the concept. The following needs
to be addressed before merging:

1. Decide on output format.
2. Provide more context for each example.
3. Integrate with top-level Makefile.
4. Move tools (validate.go, examples.go) to `cmd/` directory.
5. Run validation as part of travis CI testsuite.

Closes #34, requires #40


examples.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+var (
+	verbose bool
+	skipped bool
+)
+
+var specsByMediaType = map[string]string{
+	"application/vnd.oci.image.manifest.v1+json":      "schema/image-manifest-schema.json",
+	"application/vnd.oci.image.manifest.list.v1+json": "schema/manifest-list-schema.json",
```
wking: Probably better to have a pattern like “remove the `application/vnd.oci.` prefix and replace `+` with `.`”  to get filenames from media types.

manifest.md
```diff
@@ -92,7 +92,7 @@ A client will distinguish a manifest list from an image manifest based on the Co
 ## Example Manifest List
 
 *Example showing a simple manifest list pointing to image manifests for two platforms:*
-```json
+```application/vnd.oci.image.manifest.list.v1+json
```
wking: Won't this break GitHub's syntax highlighting?  How about leaving it as:

    ```json

and having the checker do:

1. Is it valid JSON?
2. Is there a root `mediaType` with a recognized value?
3. Does it validate with that media type's JSON Schema?

wking:
On Thu, Apr 21, 2016 at 05:39:16PM -0700, Stephen Day wrote:
> To ensure that examples in the specification match up with the json
> schemas, we propose a validation approach that will extract the fenced
> code blocks from the markdown document and validate them.

The drawback to this approach (vs. stand-alone files will all embedded
examples being subsets of the stand-alone files [1]), is that all your
embedded examples need to completely fill a given schema.  You'll need
a different approach if you want to validate subsets.  I don't see any
subsets like that in this repository at the moment, but the
runtime-spec repo uses lots of them (e.g. [2]).

[1]: https://github.com/opencontainers/image-spec/issues/34#issuecomment-210901748
[2]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#specification-version

manifest.md
vbatts: It technically does, but this is pretty logical.

philips: I would rather break syntax highlighting. +1 on the solution.

philips:
@wking if it gets annoying we can reconsider.

examples.go
vbatts: eh? I don't follow.

@stevvooe it does make me think the schema files could be named more aligned with the mime type

philips: @vbatts Why not name the files exactly the same as the mime type?

vbatts: gosh. I must be dyslexic. 
This is what I was thinking

wking: On Fri, Apr 22, 2016 at 07:26:53AM -0700, Brandon Philips:
“Why not name the files exactly the same as the mime type?”

Reasons why I like:

  $ echo application/vnd.oci.image.manifest.v1+json |
  >   sed 's|application/vnd.oci.||;s|[+]|.|'
  image.manifest.v1.json

more are:

1. application/stuff is a filename on Windows and a path into a
   subdirectory on POSIX.
2. vnd.oci. is noise in this repo (although that noise isn't a
   terrible problem if you want to keep it).
3. +json is an odd file extension (although again, you can keep it if
   you like).

philips:
@stevvooe Shall we just merge this as is?

manifest.md
stevvooe: It could be argued that github's syntax highlighting is broken... this just seemed very natural. Let's go with this until it breaks down.

examples.go
stevvooe: @vbatts @philips @wking There is already quite a but of magic going on here. I thought it would sensible to have this file mapping explicitly called out. I suspect this list will remain small, so maintaining it should be of minimal burden.

stevvooe:
@philips I've been out with a surgery for a few days. I'll take a clean up pass over this and we can go forward with it.

examples.go
philips: WFM for now

philips: @stevvooe It would be good to stay consistent. For example image-manifest-schema and image-manifest-list-schema. manifest-list is missing image-.

#bikeshed

philips:
@stevvooe Oh! Well, hope you have a speedy recovery. If you want one of us can take it over and merge it in. Let us know!

examples.go
stevvooe: @philips I noticed that. I'm also not sure about the `image-` prefix.

Let's fix it up in another PR.

(sidenote: international emoji for bikeshed is :bike: :house:)

philips: ack

philips:
@stevvooe Shall I just merge this as-is then do follow-up PRs?

manifest.md
stevvooe: Looks like there is a way to preserve it:

```json,mediatype=asdf&foo=with%20space
{
  "id": "asdf"
}
```

This has ````json,mediatype=asdf&foo=with%20space` as the opening stanza. This will let us tag with other info, such as digests and titles to improve output.

stevvooe:
@philips I'm putting the changes in right now.

stevvooe:
Changes:

- Cleaned up code and output
- Changed to future-proof format for `lang` in fenced blocks
- Moved validation tools to cmd/ hierarchy
- Moved Makefile to top-level

Follow up items:

- Maintainers should play with tool
- Project build/layout considerations
- Add checks to travis
- Possibly merge validation tools

Caveats:

- There is a problem with recursive schema validation. Try editing a manifest descriptor in the specification; the errors don't show up. Please confirm and I'll file a bug with gojsonschema.

Recommendation:

- Merge this as is and take on follow up tasks

philips:
LGTM

@s-urbaniak is looking at merging validation tools via #43

vbatts:
cool cool. @stevvooe is there value in keeping these as two commits?

stevvooe:
@vbatts I separated the implementation and the Makefile move.

vbatts:
LGTM

Issue #38, opened 2016-04-20
Consider a packed archive format for content addressable content


stevvooe:
With `docker save`, we typically have a complete image that can be transported independent of a registry. It includes configuration, layers and pointers to the actual content.

For OCI, we should define a similar tree-structured format focused on distributing OCI manifests and constituent content addressable content. The resulting archive can be used to transport complete OCI images based on manifests.

philips:
I am certain that this is the same spirit of @vbatts issue on https://github.com/opencontainers/image-spec/issues/23

stevvooe:
@philips No, it is quite different. I understood his issue is identifying the actual artifacts involved in an image, whereas this is a format for packing them into a single file.

Using multimedia encoding as an example, imagine #23 as being the video encoding format and this issue as defining the container format, such as mkv or mp4.

vbatts:
I'm confused. This sounds like what I was (poorly) attempting to arrive at. 
With a user story like "As an opensource project, I want to produce an OCI image for the world to consume" what do I do to get this packed archive?

stevvooe:
@vbatts Perhaps, with the original title. The new title is now "Outline the transportable objects". I saw that as defining the inner components (manifests, layers, image config, etc.).

I was imagining it like this:

```
[runtime bundle] -> [image spec components] --+----> [pack] --> [transports]
                                              |
                                              +----> [transports]
```

Are separate these concerns?

crimeminister:
I am curious as to whether or not OCI is committed to tar as the only export format for images. Is there any value in looking at other options, e.g. 7z, SquashFS, etc.?

wking:
On Wed, Apr 20, 2016 at 02:28:56PM -0700, Robert Medeiros wrote:
> I am curious as to whether or not OCI is committed to tar as the
> only export format for images. Is there any value in looking at
> other options, e.g. 7z, SquashFS, etc.?

I think the OCI specifying a particular family of serializations and
distribution APIs is independent of whether a discovery registry
*requires* those serializations and distribution APIs [1].  In that
case, OCI work like this issue doesn't exclude alternative
serializations from the discovery-registry ecosystem (although a
particular fetcher may not end up supporting all protocols).

[1]: https://github.com/opencontainers/image-spec/issues/15#issuecomment-209090608
     See the “registered protocol” stuff.

kamalmarhubi:
[...]

Issue #23, opened 2016-04-14
Outline the transportable objects


vbatts:
From a single tar archive of tar archives and JSON documents, or a manifest JSON document that points to other objects needing to be fetched.

UserStory:
* As an opensource project, I want to produce an OCI image for the world to consume.

See also https://groups.google.com/a/opencontainers.org/d/msg/dev/VKpZNs-qYoI/RzPhj71ODgAJ

jonboulle:
Is this essentially the enhancement/replacement of serialization.md?

vbatts:
@jonboulle sorry, this was more of placeholder thought. The serialization doc does have tar archive info. Though I was thinking more of the manifest. This may be a non-topic, if the culmination of referenced layers in a manifest are handled independently, but the docs need to be cleaned up to show the transportable states of the image.

vbatts:
Updated the title and comment. Hopefully that helps a bit more.

vbatts:
Once you've gotten all the objects on disk specified in a manifest, there will be a validation and translation to the runtime-spec bundle, or similar. This needs to be outlined as well.

philips:
I think there are two things here.

Spec out a directory layout for manifests and blobs that mirrors how the image manifests work. My suggestion is that the manifest is the primary object in the system so the other "things" are adjacent. 

```
$ cd my-oci-image/
$ find .
.
./blobs
./blobs/sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
./blobs/sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
./manifests
./manifests/v1.0
./manifests/v1.0-debug
```

Build a tool that consumes this directory and creates a runnable OCI bundle.

```
ocitool runtime-bundle file:///path/to/my-oci-image/manifests/v1.0
```

wking:
On Wed, Apr 20, 2016 at 03:20:44PM -0700, Brandon Philips wrote:
> $ find .
> .
> ./blobs
> ./blobs/sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
> ./blobs/sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
> ./manifests
> ./manifests/v1.0
> ./manifests/v1.0-debug

I'd expect the format should also include signatures and the name ↔
[...]

PR #46, opened 2016-04-28
Media types page

vbatts:
layout the media/mime types into their own page. Also add a type for the serialized combined layout (related to #23 and #38).

Next steps would be to make a compatibility matrix of types.

### DIFF
--- a/media-types.md
+++ b/media-types.md
@@ -1,0 +1,8 @@
+# Media Types
+
+The following `mediaType` MIME types are used by the formats described here, and the resources they reference:
+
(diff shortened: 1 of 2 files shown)


media-types.md
```diff
@@ -0,0 +1,11 @@
+
```
philips: newline?

philips:
LGTM! Like the idea of having a central place and having compat there.

media-types.md
vbatts: removed

PR #47, opened 2016-04-28
[ Image Format Spec ] Remove Code of Conduct file

RobDolinMS:
Since the ReadMe now points to the shared Code of Conduct in the OCI TOB repo, we can remove this file.  

Signed-off-by: Rob Dolin <RobDolin@microsoft.com>


wking:
On Thu, Apr 28, 2016 at 02:23:29PM -0700, Rob Dolin (MSFT) wrote:
> Since the ReadMe now points to the shared Code of Conduct in the OCI
> TOB repo, we can remove this file.

The runtime-spec decision was to keep the local copy for use in the
HTML/PDF docs [1], although if that's the only use case fetching over
the network at build time should be possible.

[1]: https://github.com/opencontainers/runtime-spec/pull/375#discussion_r60603475

philips:
LGTM. @vbatts can make the call on the PDF thingie.

RobDolinMS:
@wking I'm thinking that @vbatts can either:
a) Have the HTML and PDF generators pull the code of conduct from the TOB repo
b) Just have the text hyperlink to the Code of Conduct in the TOB repo (there are a few IETF RFCs referenced in the spec and we don't need to include their full text.)

NOTE: I'm not in any way trying to minimize the importance of having a Code of Conduct, just trying to ensure any updates to shared files are O(1) instead of O(n) as our number of repos expands.

vbatts:
LGTM

PR #49, opened 2016-04-29
docs: make target for PDF and HTML

### LINKED ISSUE #16: Produce PDF and HTML for releases

vbatts:
This adds a `docs` make target for producing a PDF and HTML.

The output of `tree` in `./serialization.md` does not jive with pandoc.
Not sure what better visual layout of a directory tree we can use.

This also pins the version of code-of-conduct, instead of just master.

Fixes #16

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,14 @@
 
+DOCKER ?= $(shell which docker)
+# These docs are in an order that determines how they show up in the PDF/HTML docs.
+DOC_FILES := \
+	README.md \
+	code-of-conduct.md \
+	project.md \
+	media-types.md \
+	manifest.md \
+	serialization.md
+
 default: help
 
 help:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -10,6 +20,34 @@ help:
...
+	-it \
+	--rm \
+	-v $(shell pwd)/:/input/:ro \
+	-v $(shell pwd)/output/:/output/ \
+	-u $(shell id -u) \
+	vbatts/pandoc -f markdown_github -t html5 -o /output/docs.html $(patsubst %,/input/%,$(DOC_FILES)) && \
+	ls -sh $(shell readlink -f $@)
+
+code-of-conduct.md:
+	curl -o $@ https://raw.githubusercontent.com/opencontainers/tob/d2f9d68c1332870e40693fe077d311e0742bc73d/code-of-conduct.md
```
wking: Maybe [float with master][1]?  While I like reproducible builds, I think “we've pulled an inappropriately modern code of conduct when rebuilding the old v0.1.0 docs” is not a big enough issue to justify bumping this commit hash.  And a more modern code of conduct will probably be an uncontroversial improvement to the old docs anyway.

[1]: https://raw.githubusercontent.com/opencontainers/tob/master/code-of-conduct.md

README.md
```diff
@@ -65,7 +65,7 @@ The specification and code is licensed under the Apache 2.0 license found in the
 
 ## Code of Conduct
 
-Participation in the OpenContainers community is governed by [OpenContainer's Code of Conduct](https://github.com/opencontainers/tob/blob/master/code-of-conduct.md).
+Participation in the OpenContainers community is governed by [OpenContainer's Code of Conduct](https://raw.githubusercontent.com/opencontainers/tob/d2f9d68c1332870e40693fe077d311e0742bc73d/code-of-conduct.md).
```
wking: Why this change?  Folks clicking on the link will probably appreciate GitHub's Markdown → HTML rendering.

RobDolinMS:
Thanks for doing this @vbatts.  As a consumer of the HTML and PDF versions.  I'm excited for the OCI Image Format spec to be available in these formats as well

philips:
LGTM overall

README.md
vbatts: ah. the raw carry over was from the makefile for the pdf. I can use the fancy github rendering

PR #51, opened 2016-04-29
manifest: Remove trailing comma from annotations.key2

wking:
Should fix [this][1].

[1]: https://travis-ci.org/opencontainers/image-spec/builds/126777631#L176


vbatts:
LGTM

Issue #54, opened 2016-05-02
ambiguity of manifest.json in Combined Image JSON + Filesystem Changeset Format


s-urbaniak:
https://github.com/s-urbaniak/image-spec/blob/master/serialization.md#combined-image-json--filesystem-changeset-format specifies a couple of files, none of which have any media types defined:

- `VERSION` (plain text file) schema version of the following `json` file
- `json` (json file), pointing to a "version 1" metadata format of the, which show 404.
- `repositories` (json file)
- `layer.tar` (layer tar file)
- `manifest.json` (json file) (is this the "combined Image JSON"?)

It is not clear to me how this relates to the Image Manifest file (https://github.com/s-urbaniak/image-spec/blob/master/manifest.md#image-manifest) and how we can reference a combined image from the "Image Manifest".

Also it is unclear to which of the above files the specified `application/vnd.oci.image.serialization.combined.v1+json` media type points to, the `+json` indicates a JSON format, is it `manifest.json`?

s-urbaniak:
xref https://github.com/opencontainers/image-spec/issues/56

vbatts:
This is unclear to me too.

philips:
@stevvooe Can you help clear this one up?

stevvooe:
@philips @vbatts This is the specification for the "docker save" format. 

Here is what they are, in relation to docker:

- The `json` file is just the image config corresponding to the specified layer. Used pre 1.10
- `VERSION` is the version of the `json` file, which is just as useful as `json` itself. Used pre 1.10.
- `repositories` links names to tags. Used pre 1.10.
- `manifest.json` identifies configs by tag at the top-level in 1.10+. The file `manifest.json` has nothing to do with the https://github.com/s-urbaniak/image-spec/blob/master/manifest.md#image-manifest.
- `<id>.json` are the 1.10 updated image configurations

In relation to OCI, there are few options:

1. Try to adopt everything here. It is not wildly impossible, but there are several layers. Basically, you have a pre-1.10 layer (`json`, `repositories`) and a post-1.10 layer (`manifest.json`, `<id>.json`) that need to be handled.
2. Specify 1.10 layer, removing `json`, et al.
3. Diverge from `docker save` approach and use something similar to that discussed in #38.

I hope that helps.

I am not sure where I lean here, but likely more towards the 2-3 spectrum.

philips:
@s-urbaniak can you take @stevvooe and offer a cleanup PR?

[...]

PR #55, opened 2016-05-02
media-types: show relations between media types

s-urbaniak:
Please ignore this PR if you think it is not useful, it is mere manifestation of my notes and the relations between all the JSON files in this spec.

I am still unsure of the combined output format, see https://github.com/opencontainers/image-spec/issues/54.


wking:
I don't think specs for
application/vnd.oci.image.serialization.config.v1+json have landed yet
(it seems to be based on discussion in #23), but I'd expect it will
include everything up through
application/vnd.oci.image.manifest.v1+json and possibly
application/vnd.oci.image.manifest.list.v1+json.

It will likely go beyond that, and include several images [1], and
possibly signatures asserting names for those images, but I don't
think there is a consensus yet on how that information will be
structured.

[1]: https://github.com/opencontainers/image-spec/issues/23#issuecomment-212751623

Makefile
```diff
@@ -58,3 +57,7 @@ oci-validate-json: validate.go
 oci-validate-examples: cmd/oci-validate-examples/main.go
 	go build ./cmd/oci-validate-examples
 
+%.dot.png: %.dot
```
vbatts: I like make patterns, but nothing is depending on the `media-types.dot.png` target

s-urbaniak: good catch, thanks, also it doesn't appear in the generated doc, will fix.

philips:
@s-urbaniak serialization.rootfs and serialization.config are siblings of image.manifest. Otherwise I think this is helpful for folks and we should probably add language around it.

philips:
@s-urbaniak you have two copies checked in now. Can you remove the outdated one?

s-urbaniak:
@philips fixed, thanks!

philips:
LGTM

@vbatts thoughts?

vbatts:
@s-urbaniak fixed my concerns. LGTM

PR #60, opened 2016-05-03
Specify process for naming future architectures

philips:
Signed-off-by: Timothy Hobbs <timothyhobbs@seznam.cz>

### DIFF
--- a/serialization.md
+++ b/serialization.md
@@ -173,6 +173,7 @@ It has been added to this example for clarity.
             <li>arm</li>
         </ul>
         More values may be supported in the future and any of these may or may not be supported by a given container runtime implementation.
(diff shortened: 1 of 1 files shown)


wking:
I think it makes sense to also have something like:

  New entries SHOULD be submitted to [1] for standardization.

which is along the lines of what you had in [2].  I agree with
@stevvooe that the OCI will not own the namespace [3], but we can
encourage consistency between implementations without requiring it.

[1]: https://github.com/opencontainers/image-spec/issues
[2]: https://github.com/opencontainers/image-spec/pull/29#issuecomment-215577479
[3]: https://github.com/opencontainers/image-spec/pull/29#issuecomment-215578968

philips:
@wking improved the language, thanks.

vbatts:
LGTM

PR #61, opened 2016-05-03
serialization: remove references to docker

philips:
This spec started from the Docker v1.10 serialization notes but we need
to give it its own identity. So, remove mentions of Docker that aren't
material to the spec.


jonboulle:
lgtm

vbatts:
LGTM

PR #64, opened 2016-05-04
serialization: add headers to config JSON example

### LINKED ISSUE #63: The "image JSON"/"Container config JSON" does have a media-type but no schema, no headers
The "image JSON" also described as "Container config JSON" https://github.com/opencontainers/image-spec/blob/master/serialization.md#image-json-description does have a media-type `application/vnd.oci.image.serialization.config.v1+json` but no schema defined.

Also the example image/config JSON does not have the `schemaVersion` and `mediaType` headers defined. I am aware this was taken over, but it seems not to be consistent with the manifest/manifest list schemas.

Finally I suggest to have a unique wording for this JSON file, either "Image JSON" or "Container config JSON" and use this wording consistently across the spec.

s-urbaniak:
This adds the schemaVersion, and mediaType headers to the serialization
config example.

Partially fixes #63

### DIFF
--- a/serialization.md
+++ b/serialization.md
@@ -91,6 +91,8 @@ Here is an example image JSON file:
 
 ```
 {  
+    "schemaVersion": 2,
+    "mediaType": "application/vnd.oci.image.serialization.config.v1+json",
     "created": "2015-10-31T22:22:56.015925234Z",
     "author": "Alyssa P. Hacker &ltalyspdev@example.com&gt",
     "architecture": "amd64",


serialization.md
```diff
@@ -91,6 +91,8 @@ Here is an example image JSON file:
 
 ```
 {  
+    "schemaVersion": 2,
```
vbatts: Honestly I'm unclear that it applies here, but there ought to be something like this. And more well defined in the `serialization.md`

philips:
I think we should add these fields eventually but requiring them intially will break compat with the configs that registries have today in Docker format. I am going to put it as post-v1.0.0.

vbatts:
or OPTIONAL? Question, if they're present, will docker just ignore them?

philips:
@vbatts OPTIONAL is fine for pre-v1.0.0  if Docker will ignore the additional fields. My hunch would be yes because the JSON parsing in Go drops fields it doesn't know. But, maybe @stevvooe has a solid idea.

stevvooe:
These fields really aren't necessary. We _can_ add the mediatype, but the `schemaVersion` is mostly for backwards compatibility in the registry for manifests.

Adding these is just adopting cruft from other portions of the specification, so I am not sure this PR is necessary. Either way, they will be ignored.

serialization.md
stevvooe: `schemaVersion` isn't required here.

philips:
Closing for now.

PR #65, opened 2016-05-04
oci-image-tool: initial cut

s-urbaniak:
This is work in progress for the oci-image-tool, current TODOs:

- [x] Initial https://github.com/spf13/cobra support
- [x] Autodetect type
- [x] https://github.com/xeipuuv/gojsonschema only reads single documents or files from the physical filesystem. Hence https://github.com/jteeuwen/go-bindata cannot be used since this uses `http.FileSystem`. I am implemented support for `http.FileSystem` in gojsonschema now.
- [x] Validate manifest
- [x] Validate manifest list
- [ ] Validate config
- [ ] Validate unpacked image layout on disk
- [ ] Validate compressed image
- [ ] Unpack image
- [ ] Introduce vendored dependencies
- [ ] rethink build system

All commits in this PR will be squashed into one once the above points are implemented.

__DO NOT MERGE__


s-urbaniak:
gojsonschema: upstream PR https://github.com/xeipuuv/gojsonschema/pull/101

Makefile
```diff
@@ -62,9 +62,18 @@ oci-validate-json: validate.go
 oci-validate-examples: cmd/oci-validate-examples/main.go
 	go build ./cmd/oci-validate-examples
```
philips: we can replace this once you get validation done.

s-urbaniak: ack

s-urbaniak:
@stevvooe can it be you mentioned to have manifest examples I can validate against? It'd be great to get those.

Makefile
stevvooe: Not really. This tool extracts the examples from the spec for validation. Do we intend to replace the markdown extraction?

stevvooe:
@s-urbaniak They are extracted directly from the specification by `oci-examples-validate`. We can have that tool extract them and pipe them to another tool that can validate the examples, but that tool links the specification against the schema definitions.

cmd/oci-image-tool/validate.go
```diff
@@ -0,0 +1,207 @@
...
+
+		return nil
+	}
+
+	return fmt.Errorf("type %q unimplemented", typ)
+}
+
+// withFile opens the named file for reading and executes the given function fn if no error occured.
+// If the given function returns an error it will be returned.
+func withFile(name string, fn func(f io.Reader) error) error {
```
stevvooe: This is an odd pattern for Go. Just open the file in `validatePath`. `withFile` isn't used anywhere else.

oci/validation.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/opencontainers/image-spec/schema"
+	"github.com/pkg/errors"
+	"github.com/xeipuuv/gojsonschema"
+)
+
+var (
+	fs = schema.FS(false)
+
+	specs = map[string]string{
```
stevvooe: Do you mind updating `oci-examples-validate` to use this table?

Makefile
```diff
@@ -62,9 +62,18 @@ oci-validate-json: validate.go
...
+	go build ./cmd/oci-image-tool
+
 media-types.png: media-types.dot
 
 %.png: %.dot
 	dot -Tpng $^ > $@
 
-.PHONY: validate-examples
+fs:
+	esc -o schema/fs.go -pkg="schema" -ignore=".*go" schema
```
stevvooe: Why not use [`go generate`](https://github.com/mjibson/esc#go-generate) here?

stevvooe: Also, reference that this is https://github.com/mjibson/esc.

Makefile
```diff
@@ -62,9 +62,18 @@ oci-validate-json: validate.go
...
+
+.PHONY: \
+	validate-examples \
+	statik \
```
stevvooe: Where is this target?

oci/validation.go
```diff
@@ -0,0 +1,67 @@
...
+type ValidationError struct {
+	Errs []error
+}
+
+func (e ValidationError) Error() string {
+	return fmt.Sprintf("%v", e.Errs)
+}
+
+// ValidateManifest validates the contents from the given reader as a manifest.
+func ValidateManifest(src io.Reader) error {
```
stevvooe: Might be better to declare `Validate` method as part of a var/const instance. This way, validation can be something like this:

```go
if err := MediaTypeManifest.Validate(rd); err != nil { /*...*/ }
```

This still uses the table lookup, but avoids having a method for every single type.

Makefile
s-urbaniak: good catch, leftover

cmd/oci-image-tool/validate.go
```diff
@@ -0,0 +1,207 @@
...
+	f, err := os.Open(name)
+	if err != nil {
+		return errors.Wrap(err, "unable to open file")
+	}
+	defer f.Close()
+
+	return fn(f)
+}
+
+// autodetect returns one of the type.. constants of the content of the given path
```
philips: I don't understand this comment.

cmd/oci-image-tool/validate.go
```diff
@@ -0,0 +1,207 @@
...
+		return typeImageLayout, nil
+	}
+
+	f, err := os.Open(path)
+	if err != nil {
+		return "", errors.Wrap(err, "unable to open file") // os.Open includes the filename
+	}
+	defer f.Close()
+
+	buf, err := ioutil.ReadAll(io.LimitReader(f, 512)) // read up to 512 bytes for content detection
```
philips: I would just say "read some initial bytes to detect content" instead of putting the amount in the comment

oci/validation.go
s-urbaniak: ok, why not, good idea. I'll move the whole validation logic into the `schema` package then too.

cmd/oci-image-tool/validate.go
s-urbaniak: ok, removed it, we're validating files right now only anyways.

Makefile
s-urbaniak: fixed

Makefile
s-urbaniak: fixed

oci/validation.go
s-urbaniak: done

cmd/oci-image-tool/validate.go
s-urbaniak: corrected, PTAL

cmd/oci-image-tool/validate.go
s-urbaniak: corrected, PTAL

Makefile
s-urbaniak: I am thinking of moving the markdown extraction to a markdown_test.go invocable via `go test` and not have a separate binary.

s-urbaniak:
@stevvooe OK, I was assuming you have more samples, nevermind. As mentioned above I'll try to migrate the functionality of `oci-examples-validate` to a Go test function. The manifest markdown links against the media type so this should be doable, right?

stevvooe:
> OK, I was assuming you have more samples, nevermind. 

Sorry about that. I think the idea is that we would add these variations to specification to introduce them in context.

> The manifest markdown links against the media type so this should be doable, right?

Yep. I think you can port this almost directly. Most of `oci-examples-validate` should be pulled into the presented package.

Makefile
stevvooe: That sounds great. I am not sure why I didn't do that initially.

s-urbaniak: @stevvooe moved `oci-validate-examples` as a unit test, PTAL https://github.com/opencontainers/image-spec/pull/65/commits/71e594293d741708c37a13331edafb8aec6a6ff0. I did not do big refactorings, just a minimalistic 1:1 transition.

s-urbaniak: Also removed `oci-validate-json` since it is being replaced by `oci-image-tool`

schema/spec_test.go
```diff
@@ -1,120 +1,73 @@
...
-func init() {
-	flag.BoolVar(&verbose, "verbose", false, "display examples no matter what")
-	flag.BoolVar(&skipped, "skipped", false, "show skipped examples")
-	flag.StringVar(&schemaPath, "schema", "./schema", "specify location of schema directory")
-}
-
-func main() {
-	flag.Parse()
+func TestSpecExamples(t *testing.T) {
+	m, err := os.Open("../manifest.md")
```
stevvooe: Let's throw a TODO here to have this encompass all specification files.

schema/schema.go
```diff
@@ -0,0 +1,50 @@
...
+
+	specs = map[jsonManifest]string{
+		MediaTypeManifest:     "image-manifest-schema.json",
+		MediaTypeManifestList: "manifest-list-schema.json",
+	}
+)
+
+// FindValidator returns the validator for the given media type
+// or an error if none could be found.
+func FindValidator(mediaType string) (Validator, error) {
```
s-urbaniak: I actually don't like this, will change to `func Validate(mediaType string, r io.Reqder) error`

stevvooe: I agree this function is awkward, but don't we already have this through casting? Something like `Validator(mediaType).Validate(r)` is workable.

For reference, this pattern is very close to [crypto.Hash](https://golang.org/pkg/crypto/#Hash). It may help to look there for inspiration.

s-urbaniak: Indeed, many thanks for the pointer! I'll reiterate on this, my approach is a bit different than `crypto.Hash` but looking over it I like it much better.

stevvooe: We also did this for the [`distribution/digest.Algorithm`](https://godoc.org/github.com/docker/distribution/digest#Algorithm) type, if you're curious.

s-urbaniak: Ah great, I am implementing this very scheme, thanks again! :-)

liangchenye:
CooI!  I think it might better be a part of  https://github.com/opencontainers/ocitools,
especially for the remain tasks, 
``` 
 Validate unpacked image layout on disk
 Validate compressed image
 Introduce vendored dependencies
 rethink build system
```
What do you think? @s-urbaniak @mrunalp @philips

schema/media_types.go
```diff
@@ -0,0 +1,75 @@
...
+	// specs maps OCI schema media types to schema files.
+	specs = map[ManifestValidator]string{
+		MediaTypeManifest:     "image-manifest-schema.json",
+		MediaTypeManifestList: "manifest-list-schema.json",
+	}
+)
+
+// ManifestValidator wraps a media type string identifier
+// and implements validation against a JSON schema.
+type ManifestValidator string
```
stevvooe: Small nit: these won't always be "manifests".

philips:
@liangchenye right, @s-urbaniak please don't worry about validating the ondisk format. We can rely on the oci-runtime-tool for that.

philips:
@s-urbaniak can we merge the tool as-is and do a follow-up PR? Essentially just add the dependencies and then we can do followup PRs for each of the remaining featrues?

s-urbaniak:
@philips sure, we can mere as is! Let me squash all of this in a sec. I working on an unpacker now. What do you mean with "Essentially just add the dependencies"? Do you mean vendoring the deps?

s-urbaniak:
@philips squashed, ready to merge once green

philips:
LGTM, can you file individual issues for the remaining things?

s-urbaniak:
@philips filed #74, #75, and #76 to catch remaining TODOs and updated the description.

stevvooe:
LGTM

I have no problems with merging this PR, but let's make sure to get a few more LGTMs in the future.

vbatts:
This was a bigger kick off of work, but so far we've been merging with one,
sometimes two LGTMs.

PR #71, opened 2016-05-07
image manifest: add source annotation

gregkh:
For a number of images, source code needs to be provided due to the
license of the files in the image, so provide an annotation to make it
easy to determine where that source can be found.

Signed-off-by: Greg Kroah-Hartman <gregkh@linuxfoundation.org>

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -201,6 +201,7 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
     * **authors** contact details of the people or organization responsible for the image (freeform string)
     * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
     * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
+    * **source** URL to get the source code for the binary files in the image (string, must be a URL with scheme HTTP or HTTPS)
 
 
 ## Example Image Manifest


vbatts:
So much. I debated on this field, and/or LICENSE. And what the value would be.
Otherwise LGTM

gregkh:
Thanks, LICENSE wouldn't work given the normal mix of tons of different licenses in a single image.  But it really would be nice if we could add it on a per-file basis if possible, let me look into that...

cgwalters:
For things like this I'd expect a little bit more referencing of prior art (deb/rpm/etc).

For example - should this URL attempt to reference an exact git commit for something you happened to `git clone` in your `Dockerfile`, or is it more like just a URL to the git repo?

And what about multiple sources?

vbatts:
I was thinking to a tarball or repo. Perhaps folks could comma delimit a
list

gregkh:
@cgwalters as it is, this will support what Debian does for their docker containers, see: https://hub.docker.com/_/debian/ 

I would have loved to build off of the LICENSE stuff from RPM and DEB, but due to an image just being a big mass of files, all with potentially different licenses, that doesn't work, hence this patch.

It's up to the creator of the image to properly provide source (hint, almost no one does, this is not good), so this gives people a way to start to do that correctly.

If you can think of a way to do this on a per-file basis, that would be wonderful, but I don't see how to do that right now.  Unless I'm missing something?

philips:
@gregkh I am not seeing what debian does. Can you add a bit onto this PR that says "(e.g. a Debian base image might use the URL https://debian.org/foo/bar)".

Thanks!

philips:
Oh, and I am +1 for this change.

gregkh:
@philips this PR has already been merged, I can't change it now.  Do you mean for me to add the example into the spec itself?

philips:
Submitting the example as a new PR is great.

Issue #70, opened 2016-05-07
pandoc: issue with displaying images


vbatts:
with https://github.com/opencontainers/image-spec/releases/tag/v0.1.0
The media-types.png caused issues in the `make docs` target. See https://github.com/opencontainers/image-spec/issues/66#issuecomment-217563622

As a make-shift effort, I switched it to inline html. While that caused pandoc to not fail, it did not render the image in the resulting document.
Here's the diff that let pandoc run successfully.
```diff
diff --git a/media-types.md b/media-types.md
index 3bf7221..de6cc3b 100644
--- a/media-types.md
+++ b/media-types.md
@@ -46,6 +46,6 @@ This section shows where the OCI Image Specification is compatible with formats
 
 The following figure shows how the above media types reference each other:
 
-![Media Types](media-types.png)
+<img src="media-types.png" alt="Media Types"/>
 
 A reference is defined as the target content digest, as defined by the [Registry V2 HTTP API Specificiation](https://docs.docker.com/registry/spec/api/#digest-parameter). The manifest list being a "fat manifest" references one or more image manifests per target platform. An image manifest references exactly one target configuration and possibly many layers.
```

Other option would be a full base64 inline of the image, in an html tag. 😛

philips:
sounds like base64 it is! :)

philips:
This is fixed now. Closing.

PR #73, opened 2016-05-19
inline images: media-types.png is inline

vbatts:
Due to pandoc pdf and html wanting a uri to reference an image, we'll
need images to be inlined fo rthe document. So with this, the document
will get an base64 blob of images that match the target 'inline-png-%.md'

See also https://github.com/opencontainers/image-spec/issues/70#issuecomment-217656058

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ DOC_FILES := \
 	code-of-conduct.md \
 	project.md \
 	media-types.md \
+	inline-png-media-types.md \
 	manifest.md \
 	serialization.md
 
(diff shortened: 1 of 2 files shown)


philips:
elegant :( LGTM

vbatts:
it's html ... and base64 ...

image

s-urbaniak:
bummer ... LGTM then

vbatts:
Perhaps we can find a nicer way to handle it in the future? Suggestions welcome :-\

PR #81, opened 2016-05-19
schema: add config/image JSON schema and validation

### LINKED ISSUE #74: oci-image-tool: Implement config validation
The "Image JSON", media type `oci-image-tool: Implement config validation (media type application/vnd.oci.image.serialization.config.v1+json` validation needs to be implemented in the oci-image-tool

s-urbaniak:
Fixes #74

### DIFF
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -1,0 +1,33 @@
+{
+  "description": "OpenContainer Config Specification",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://opencontainers.org/schema/image/config",
+  "type": "object",
+  "properties": {
+    "created": {
+      "type": "string",
+      "format": "date-time"
+    },
+    "author": {
+      "type": "string"
+    },
+    "architecture": {
+      "type": "string"
+    },
+    "os": {
+      "type": "string"
+    },
+    "config": {
+      "$ref": "defs-config.json#/definitions/config"
+    },
+    "rootfs": {
+      "$ref": "defs-config.json#/definitions/rootfs"
+    },
+    "history": {
+	  "type": "array",
+      "items": {
+        "$ref": "defs-config.json#/definitions/history"
+      }
+    }
+  }
+}
--- a/schema/spec_test.go
+++ b/schema/spec_test.go
@@ -33,9 +33,17 @@ var (
 	errFormatInvalid = errors.New("format: invalid")
 )
 
+func TestValidateManifest(t *testing.T) {
+	validate(t, "../manifest.md")
+}
+
+func TestValidateSerialization(t *testing.T) {
+	validate(t, "../serialization.md")
+}
+
 // TODO(sur): include examples from all specification files
-func TestSpecExamples(t *testing.T) {
-	m, err := os.Open("../manifest.md")
+func validate(t *testing.T, name string) {
+	m, err := os.Open(name)
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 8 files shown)


s-urbaniak:

schema/config-schema.json
```diff
@@ -0,0 +1,33 @@
...
+      "type": "string"
+    },
+    "architecture": {
+      "type": "string"
+    },
+    "os": {
+      "type": "string"
+    },
+    "config": {
+      "$ref": "defs-config.json#/definitions/config"
```
stevvooe: This is probably outside the scope of this PR: let's make sure we have some negative examples of the config to ensure that the validation is working with these recursively definitions. I know I had some issues with this when I first started experimenting with the validators.

s-urbaniak: 👍 I have this on my radar, as well as the support for providing the erroneous line number in case of validation failure.

schema/spec_test.go
```diff
@@ -46,6 +54,10 @@ func TestSpecExamples(t *testing.T) {
 	}
 
 	for _, example := range examples {
+		if example.Err == errFormatInvalid { // ignore
```
jonboulle: hmm why?

s-urbaniak: There are many examples/snippets in the spec not meant for validation.
Hmm, but you're right, we should only ignore if there is no MediaType set.

s-urbaniak:
@jonboulle PTAL

jonboulle:
LGTM

schema/config-schema.json
stevvooe: @s-urbaniak I've filed #83. Feel free to expand on anything I've missed.

vbatts: what is recursive here?

stevvooe: @vbatts `config-schema.json` pulls in `defs-config.json`. During my experimentation, I was not convinced this was working correctly. Either way, we should have negative examples to confirm this and ensure that bad schemas are actually being rejected.

schema/config-schema.json
```diff
@@ -0,0 +1,33 @@
+{
+  "description": "OpenContainer Config Specification",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://opencontainers.org/schema/image/config",
+  "type": "object",
+  "properties": {
+    "created": {
+      "type": "string",
+      "format": "date-time"
```
vbatts: is this format too strict for `"2015-10-31T22:22:56.015925234Z"`?

schema/config-schema.json
```diff
@@ -0,0 +1,33 @@
...
+    },
+    "rootfs": {
+      "$ref": "defs-config.json#/definitions/rootfs"
+    },
+    "history": {
+	  "type": "array",
+      "items": {
+        "$ref": "defs-config.json#/definitions/history"
+      }
+    }
```
vbatts: I was about to comment on the lack of an array of requires, but then the `./serialization.md` doesn't show much for required fields ...

vbatts:
overall LGTM, but a couple of questions

schema/config-schema.json
wking: On Mon, May 23, 2016 at 01:51:33PM -0700, Vincent Batts wrote:
> > +    "created": {
> > +      "type": "string",
> > +      "format": "date-time"
> 
> is this format too strict for `"2015-10-31T22:22:56.015925234Z"`?

The v4 JSON Schema RFC doesn't mention date-time [1], but it's listed
here [2] referencing RFC 3339, section 5.6.  And your string is
compatible with that [3].

[1]: https://tools.ietf.org/html/draft-zyp-json-schema-04
[2]: http://json-schema.org/latest/json-schema-validation.html#anchor108
[3]: http://tools.ietf.org/html/rfc3339#section-5.6

schema/config-schema.json
vbatts: ah, so. Not working is an understandable concern. I just did not see recursion. :-)

stevvooe:
LGTM

vbatts:
LGTM

PR #82, opened 2016-05-22
layout: implement unpacking and validation

### LINKED ISSUE #75: oci-image-tool: validate and unpack image
The oci-image-tool should be able to validate and unpack an oci-image from 1. a directory and 2. a tar file.

Proposed spec: https://github.com/opencontainers/image-spec/issues/23#issuecomment-212636147

s-urbaniak:
Fixes #75

### DIFF
--- a/cmd/oci-image-tool/autodetect.go
+++ b/cmd/oci-image-tool/autodetect.go
@@ -1,0 +1,103 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"encoding/json"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"os"
+
+	"github.com/opencontainers/image-spec/schema"
+	"github.com/pkg/errors"
+)
+
+// supported autodetection types
+const (
+	typeImageLayout  = "imageLayout"
+	typeImage        = "image"
+	typeManifest     = "manifest"
+	typeManifestList = "manifestList"
+	typeConfig       = "config"
+)
+
+// autodetect detects the validation type for the given path
+// or an error if the validation type could not be resolved.
+func autodetect(path string) (string, error) {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return "", errors.Wrapf(err, "unable to access path") // err from os.Stat includes path name
+	}
+
+	if fi.IsDir() {
+		return typeImageLayout, nil
+	}
+
+	f, err := os.Open(path)
+	if err != nil {
+		return "", errors.Wrap(err, "unable to open file") // os.Open includes the filename
+	}
+	defer f.Close()
+
+	buf, err := ioutil.ReadAll(io.LimitReader(f, 512)) // read some initial bytes to detect content
+	if err != nil {
+		return "", errors.Wrap(err, "unable to read")
+	}
+
+	mimeType := http.DetectContentType(buf)
+
+	switch mimeType {
+	case "application/x-gzip":
+		return typeImage, nil
+
+	case "application/octet-stream":
+		return typeImage, nil
+
+	case "text/plain; charset=utf-8":
+		// might be a JSON file, will be handled below
+
+	default:
+		return "", errors.New("unknown file type")
+	}
+
+	if _, err := f.Seek(0, os.SEEK_SET); err != nil {
+		return "", errors.Wrap(err, "unable to seek")
+	}
+
+	header := struct {
+		SchemaVersion int         `json:"schemaVersion"`
+		MediaType     string      `json:"mediaType"`
+		Config        interface{} `json:"config"`
+	}{}
+
+	if err := json.NewDecoder(f).Decode(&header); err != nil {
+		return "", errors.Wrap(err, "unable to parse JSON")
+	}
+
+	switch {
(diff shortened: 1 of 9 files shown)


s-urbaniak:
Blocked by https://github.com/opencontainers/image-spec/pull/81: I need the config validator to thoroughly validate the image/image layout.

philips:
@s-urbaniak #81 is merged, what is the next step?

s-urbaniak:
@philips implementing image layout validation now, should be ready by EOD

s-urbaniak:
@philips @vbatts @stevvooe PTAL; I implemented this using self-made image available here https://github.com/s-urbaniak/oci-image-test.

philips:
@s-urbaniak Are you generating the OCI Runtime config.json? https://github.com/opencontainers/runtime-spec/blob/master/config.md We need to generate that from the serialization.config thingie.

s-urbaniak:
@philips nope, as discussed OOB, this is just "pure" unpacking. The runtime config.json would have to be solved in a separate PR.

stevvooe:
@s-urbaniak I'm looking at this today. Sorry for the delay.

image/unpack.go
```diff
@@ -0,0 +1,223 @@
...
+			// success, continue below
+		default:
+			return errors.Wrapf(err, "error advancing tar stream")
+		}
+
+		path := filepath.Join(dest, filepath.Clean(hdr.Name))
+		info := hdr.FileInfo()
+
+		if strings.HasPrefix(info.Name(), `.wh.`) {
+			path = strings.Replace(path, `.wh.`, "", 1)
```
philips: any reason to use backtick instead of regular quotes?

stevvooe:
@s-urbaniak This is looking very solid. Is there a corresponding specification for this format?

The only large change I'd suggest is that rather than `manifests`, this should be `refs`. These files should simply contain a descriptor pointing to the manifest blob. This will allow us to avoid the half-deserialized manifests.

image/unpack.go
```diff
@@ -0,0 +1,223 @@
...
+			if info.IsDir() {
+				return nil
+			}
+
+			dd, err := filepath.Rel("blobs", filepath.Clean(path))
+			if err != nil || d != dd {
+				return nil // ignore
+			}
+
+			if err := extractLayer(dest, r); err != nil {
```
stevvooe: If we unpack directly to disk, these layers really need to be verified against the manifest's digest.

philips: +1 on that.

philips:
@stevvooe what do you mean by "avoid the half-deserialized manifests"?

philips:
@stevvooe are you thinking something like this for the refs schema:

```
$ cat refs/v1.0
{
  "media-types": {
    "application/vnd.oci.image.manifest.v1+json": {"size": "4096", "digest": "sha256:afff3928"
  }
}
```

s-urbaniak:
@stevvooe @philips thanks a lot for the review! just fyi: I'm sitting between flights though so won't make to implement changes before tomorrow.

cmd/oci-image-tool/autodetect.go
```diff
@@ -0,0 +1,103 @@
...
+	"net/http"
+	"os"
+
+	"github.com/opencontainers/image-spec/schema"
+	"github.com/pkg/errors"
+)
+
+// supported autodetection types
+const (
+	typeImageLayout  = "imageLayout"
```
stevvooe: Why not use the mediatypes directly?

stevvooe:
@philips I really referring to https://github.com/opencontainers/image-spec/pull/82/files#diff-1d4f0c730d731e154e77caf8c8ce36a4R80.

But, I was thinking about something like this:

```
$ cat refs/v1.0
{"size": 4096, "digest": "sha256:afff3928", "mediatype": "application/vnd.oci.image.manifest.v1+json"}
```

philips:
@stevvooe I think we need an array of media types though as we need a way of emulating Accept with a filesystem. What if my ref is serving a v1 and v1.1 manifest? Let me file a new issue on this.

wking:
On Thu, May 26, 2016 at 02:14:36PM -0700, Brandon Philips wrote:
> What if my ref is serving a v1 and v1.1 manifest?

Make the root reference point to a manifest.list [1]?

[1]: https://github.com/opencontainers/image-spec/blob/master/manifest.md#manifest-list

philips:
@wking I think that is the approach we are going to take. Let me work up a proposal.

philips:
@stevvooe here is a concrete format specification: https://github.com/opencontainers/image-spec/pull/94

image/unpack.go
```diff
@@ -0,0 +1,223 @@
...
+	digests, err := diffIDs(w, manifest)
+	if err != nil {
+		return errors.Wrap(err, "error reading diff IDs")
+	}
+
+	return extractLayers(w, digests, dest)
+}
+
+func diffIDs(w walker, manifest string) ([]string, error) {
+	manifest = filepath.Join("manifests", manifest)
```
philips: can you change this code to follow the spec here: https://github.com/opencontainers/image-spec/pull/94

image/unpack.go
s-urbaniak: no reason, just too fast typing ;-) fixed

cmd/oci-image-tool/autodetect.go
s-urbaniak: this gives me finer grained control, i.e. `typeImageLayout`, and `typeImage` don't have media types at all and must be treated differently.

image/unpack.go
s-urbaniak: fixed

image/unpack.go
s-urbaniak: fixed

s-urbaniak:
@stevvooe @philips refactored this towards the proposed layout spec in https://github.com/opencontainers/image-spec/pull/94.

philips:
Overall LGTM for a starting point. 

Anyone else give this an LGTM @opencontainers/image-spec-maintainers

cmd/oci-image-tool/autodetect.go
stevvooe: Of course!

jonboulle:
LGTM

PR #89, opened 2016-05-25
Makefile: remove oci-validate-examples target

stevvooe:
This binary was removed in a previous commit but the make targets were
not. The make targets have now been removed and the `validate-examples`
targets run the schema tests on the spec files.


philips:
LGTM! Merge at will.

PR #91, opened 2016-05-26
serialization: add language improvement to serialization

philips:
A small fix from the v1.1 merge that happened inside of the docker repo
after we pulled this in.

See https://github.com/docker/docker/pull/22264


philips:
Build failure is because of gopkg.in being down. Nothing to do with the patch. 

Can I get a review @opencontainers/image-spec-maintainers?

stevvooe:
LGTM

Confirmed this was in docker/docker#22264

PR #93, opened 2016-05-27
img: move images into the img dir

philips:
Trying to cleanup the objects in the root directory so there is less
random stuff to sort through.


jonboulle:
lgtm

Issue #92, opened 2016-05-26
media-type: add a media-type for descriptor


philips:
For the "image layout" that can be tar'd, zip'd, etc we need to create the spec and media-type for the "descriptor"

https://github.com/opencontainers/image-spec/blob/master/schema/defs-image.json#L23

Relevant IRC convo:

```
4:08 PM <philips> stevvooe: this response? https://github.com/opencontainers/image-spec/pull/82#issuecomment-221980346
4:09 PM <stevvooe> philips: yes
4:09 PM <stevvooe> philips: it just means that we deserialize a descriptor, rather than a few fields of the manifest
4:09 PM <philips> stevvooe: I don't know what you mean by half-deserialized still
4:09 PM <philips> stevvooe: what is a descriptor?
4:09 PM <stevvooe> philips: "partially-serialized" would probably be more accurate
4:10 PM <stevvooe> stevvooe: this structure: {"size": 4096, "digest": "sha256:afff3928", "mediatype": "application/vnd.oci.image.manifest.v1+json"}
4:10 PM <philips> stevvooe: ah, ok. well, did you see my concern about mime types?
4:10 PM <stevvooe> philips: yea, and now i think i understand your example
4:11 PM <philips> stevvooe: like my structure is identical but allows for N media types: https://github.com/opencontainers/image-spec/pull/82#issuecomment-221760143
4:11 PM <stevvooe> philips: i think if we push that dispatch to a manifest list, we can avoid a large amount of unverified content
4:12 PM <philips> stevvooe: so you suggest always force a manifest list?
4:12 PM <philips> stevvooe: like the manifest list is always the thingie we point to in this "container"
4:12 PM <stevvooe> philips: so, the options are <descriptor> -> <manifest list> or <new format: [descriptor]> -> <arbitrary>
4:13 PM <stevvooe> philips: that might be the right approach
4:13 PM <philips> stevvooe: yea, not opposed to the descriptor thing for "file transport containers"
4:13 PM <philips> stevvooe: and always pointing to a manifest list makes sense, I think.
4:13 PM <stevvooe> philips: btw, here is the descriptor def : https://github.com/opencontainers/image-spec/blob/master/schema/defs-image.json#L14
4:15 PM <philips> stevvooe: got it, we only ever refer to it that way in code instead of in the spec so I wasn't familiar with the term 
4:15 PM <philips> stevvooe: thanks, makes sense
[...]

PR #94, opened 2016-05-27
image-layout: add an initial image layout spec

philips:
In order to build tooling that converts the set of OCI Image objects
into an OCI Runtime object we need an Image Layout that test tools can
be built around.

Longer term this Image Layout could be used to define a "single object
image" that is a tar/zip/etc that could be posted over https/ftp/etc.

The layout comes from several different discussions:

https://github.com/opencontainers/image-spec/issues/23
https://github.com/opencontainers/image-spec/issues/92

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -1,0 +1,62 @@
+## Open Container Initiative Image Layout Specification
+
+The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
+This layout MAY be used in a variety of different transport mechanisms: archive formats (e.g. tar, zip), shared filesystem environments (e.g. nfs), or networked file fetching (e.g. http, ftp, rsync).
+Given an image layout a tool can convert a given ref into a runnable OCI Image Format by finding an appopriate manifest from the manifest list, unpacking the filesystem serializations in the correct order, and then converting the image configuration into an OCI Runtime config.json.
+
+The image layout has two top level directories:
+
+- "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque.
+- "refs" contains descriptors pointing to an image manifest list
+
+It also contains a file that is used to identify the layout version:
+
+- "oci-layout" MUST contain a JSON object with a version field `{"imageLayoutVersion": "1.0.0"}` and MAY include additional fields.
+
+This is an example image layout:
+
+```
+$ cd example.com/app/
+$ find .
+.
+./blobs
+./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
+./blobs/sha256-5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
+./blobs/sha256-e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
+./oci-layout
+./refs
+./refs/v1.0
+./refs/v1.1
+```
+
+Blobs are named by their contents:
+
+```
+$ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
+afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
+```
+
+Object names in the refs and blobs MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
+Hash algorithm identifiers containing the colon `:` will be converted to the hyphen `-`.
+For example `sha256:5b` will map to the layout `blobs/sha256-5b`.
+The blobs directory MAY contain blobs which are not referenced by any of the refs.
+The blobs directory MAY be missing referenced blobs, in which case the missing blobs SHOULD be fulfilled by an external blob store.
+
+Each object in the refs subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
+In general the `mediatype` of this descriptor object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
+
+This illustrates the expected contents of a given ref and the manifest list it points to.
+
+```
+$ cat ./refs/v1.0
+{"size": 4096, "digest": "sha256:afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51", "mediatype": "application/vnd.oci.image.manifest.list.v1+json"}
+```
+```
+$ cat ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
+  "manifests": [
+    {
+...
+```


image-layout.md
```diff
@@ -0,0 +1,45 @@
+## Open Container Initiative Image Layout Specification
+
+The OCI Image Layout is a slash separated layout of OCI content-addressable
```
wking: "content-adressable ..."?  I think you're missing"objects." or similar.

image-layout.md
```diff
@@ -0,0 +1,45 @@
...
+- "refs" contains descriptors pointing to an image manifest list
+
+This is an example image layout:
+
+```
+$ cd example.com/app/
+$ find .
+.
+./blobs
+./blobs/sha256:afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
```
stevvooe: Suggest we go with a directory per algorithm. `:` denotes alternate data streams on windows.

image-layout.md
```diff
@@ -0,0 +1,45 @@
...
+./blobs/sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
+./blobs/sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
+./refs
+./refs/v1.0
+./refs/v1.1
+```
+
+The blobs directory MAY contain objects which are not referenced by any of the refs.
+Each object in the refs subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
+And the `mediatype` of this descriptor object SHOULD be `application/vnd.oci.image.manifest.list.v1+json` although future versions of the spec may use a different mediatype.
```
stevvooe: Do we have a weaker word than SHOULD?

image-layout.md
```diff
@@ -0,0 +1,45 @@
+## Open Container Initiative Image Layout Specification
+
+The OCI Image Layout is a slash separated layout of OCI content-addressable 
+This layout MAY be used in a variety of different transport mechanisms from archive formats (e.g. tar, zip) or used in shared filesystem environments (e.g. nfs) or in networked file fetching systems (e.g. http, ftp, rsync).
+Given an image layout a tool can convert a given ref into a runnable OCI Image Format by finding an appopriate manifest from the manifest list, unpacking the filesystem serializations in the correct order, and then converting the image configuration into an OCI Runtime config.json.
+
```
stevvooe: A few requirements we may want to lay out:

1. No symbolic linking. All linking is through file contents.
2. File object names should restricted to a simple character set, such that are supported across a wide variety of protocols.

image-layout.md
```diff
@@ -0,0 +1,45 @@
...
+./blobs
+./blobs/sha256:afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
+./blobs/sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
+./blobs/sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
+./refs
+./refs/v1.0
+./refs/v1.1
+```
+
+The blobs directory MAY contain objects which are not referenced by any of the refs.
```
wking: I think it's also worth saying explicitly that `blobs` MAY (SHOULD?) contain all referenced objects.  That leaves the door open to folks who want to delegate a few branches/leaves to other CAS engines/locations.  git-bundle(1) is an example of this approach.

For example, say I want to put my Python script into a Debian rootfs.  I reference layers for my Python script as one layer and the Debian rootfs as another.  But then only the Python script (tarball) is in `blobs`, and I expect folks will be able to find the Debian rootfs (via it's hash) through another channel.  We haven't defined mechanisms for that “find the Debian rootfs through another channel” yet, but I don't think we need to specify them before partial blob-stores are useful.

image-layout.md
wking: On Thu, May 26, 2016 at 08:28:37PM -0700, Stephen Day wrote:
> > +And the `mediatype` of this descriptor object SHOULD be `application/vnd.oci.image.manifest.list.v1+json` although future versions of the spec may use a different mediatype.
> 
> Do we have a weaker word than SHOULD?

MAY [1].

[1]: https://tools.ietf.org/html/rfc2119#section-5

stevvooe: Clarifying: it would be acceptable to point directly to a manifest.

image-layout.md
wking: And this page should link to a list somewhere that spells out the
algorithm behind each short form (e.g. ‘sha256’ → “SHA-256 in [1]”).

[1]: http://dx.doi.org/10.6028/NIST.FIPS.180-4

image-layout.md
wking: On Thu, May 26, 2016 at 08:40:00PM -0700, Stephen Day wrote:
> > +And the `mediatype` of this descriptor object SHOULD be
> > `application/vnd.oci.image.manifest.list.v1+json` although future
> > versions of the spec may use a different mediatype.
> 
> Clarifying: it would be acceptable to point directly to a manifest.

And as far as the image-layout is concerned, I don't see a problem
with supporting:

  {"size": …, "digest": "sha256:…", "mediatype": "image/png"}

That's unlikely to successfully unpack to an OCI image, but there
should be a distinction between what image-layout supports (where I
think it's useful to be generic) and what image-spec clients can
successfully unpack to an OCI config/bundle (where I think it's useful
to SHOULD specific choices for compatibility, but still leave the door
open for unspecified innovation).

image-layout.md
philips: Why is symlinking bad?

And I think the set {[a-z], [0-9], ':', '/'} is OK for most all things.

image-layout.md
philips: Surely : can be escaped on Windows. Paging @jstarks.

image-layout.md
philips: Hrm, I agree in theory but perhaps we can file an issue for how to deal with this in a follow-up PR? The initial motivation for this thing is to simply help show the vision of the image format turning into a "Runtime Bundle" and adding external CAS resolution language will confuse this early on.

image-layout.md
philips: I will weaken the language to say "In general blah blah" instead of SHOULD

image-layout.md
wking: On Thu, May 26, 2016 at 09:45:56PM -0700, Brandon Philips wrote:
> … adding external CAS resolution language will confuse this early on.

I agree that *that* should be punted.  All I'm asking for is a line
like:

  The blobs directory MAY omit objects referenced by the refs or other blobs.

image-layout.md
philips: added character limit

image-layout.md
philips: fixed!

image-layout.md
philips: Language weakened!

philips:
Any other @opencontainers/image-spec-maintainers want to chime in?

image-layout.md
```diff
@@ -0,0 +1,47 @@
...
+./blobs/sha256:afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
+./blobs/sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
+./blobs/sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
+./refs
+./refs/v1.0
+./refs/v1.1
+```
+
+Object names in the refs and blobs MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen (-), the colon (:), the dot (.), and the underscore (_).
+The blobs directory MAY contain objects which are not referenced by any of the refs.
```
wking: Looks like GitHub wants you to quote your single-characters, otherwise the underscore leads to a lot of italics (at least in the /pull/94/files view).

And while I'm generally against explicit charset restrictions, in this case I'm fine with them, because the “real” name (e.g. whatever is used for crypo-verification) is going to have to be stored in the contents of a blob.  The refs/{nick} entries are just convenient nicknames.

image-layout.md
```diff
@@ -0,0 +1,47 @@
...
+
+Each object in the refs subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
+In general the `mediatype` of this descriptor object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
```
wking: I'd drop the “MAY” → “may”, since the spec is about imposing constraints on implementations and consumers/producers.  We don't want a chicken-and-egg constraint on the spec itself.

image-layout.md
```diff
@@ -0,0 +1,47 @@
+## Open Container Initiative Image Layout Specification
+
+The OCI Image Layout is a slash separated layout of OCI content-addressable objects and human significant references (refs)
```
wking: “v1.0” is (without more context) not really significant to me.  I'd recommend “[location-addressable][1] references.” (with the trailing period too ;).

[1]: https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed

image-layout.md
```diff
@@ -0,0 +1,47 @@
+## Open Container Initiative Image Layout Specification
+
+The OCI Image Layout is a slash separated layout of OCI content-addressable objects and human significant references (refs) 
+This layout MAY be used in a variety of different transport mechanisms from archive formats (e.g. tar, zip) or used in shared filesystem environments (e.g. nfs) or in networked file fetching systems (e.g. http, ftp, rsync).
```
wking: nit: the “from … or … or” phrasing is a bit awkward.  How about “transport mechanisms, including archive formats (…), shared filesystems (…), and networked transport protocols (…).  Although at this level, the distinction between NFS and HTTP is pretty irrelevant, so it's probably not worth looking to hard for good bucket names.

image-layout.md
philips: done, thanks.

image-layout.md
stevvooe: I'd like to hear @jstarks response but checkout this [superuser question](http://superuser.com/questions/461320/get-a-colon-in-a-windows-filename):

> Your options are to strip the colons on the source, or __replace them with a unicode character that looks like a colon but isn't.__

Not sure if that is a viable option. 😄 

> And this page should link to a list somewhere that spells out the
algorithm behind each short form (e.g. ‘sha256’ → “SHA-256 in [1]”).

These probably belong in https://github.com/opencontainers/image-spec/blob/master/media-types.md or we can pull in the digest portion of the manifest specification.

wking: On Thu, May 26, 2016 at 11:54:07PM -0700, Stephen Day wrote:
> > And this page should link to a list somewhere that spells out the
> > algorithm behind each short form (e.g. ‘sha256’ → “SHA-256 in
> > [1]”).
>
> These probably belong in
> https://github.com/opencontainers/image-spec/blob/master/media-types.md
> or we can pull in the digest portion of the manifest specification.

I don't have a strong opinion about where the docs live (or even that
they live in this repository).  For example, I'm fine linking [1] with
a note that those subdirectory names mean the standard mentioned in
the comment there (e.g. ‘sha1’ means “the SHA1 hash algorithm as
defined in RFC 3174”, etc.).  That binds the supported set of hash
names to the Go folks, but we can always in-house the list if/when we
need a standard name for a hash that Go doesn't support (unlikely).

[1]: https://golang.org/pkg/crypto/#pk-subdirectories

philips: eek, ok, so I guess I will just define sha256: to be transformed into sha256-.

Ping @jstarks again. :)

image-layout.md
```diff
@@ -0,0 +1,47 @@
...
+Given an image layout a tool can convert a given ref into a runnable OCI Image Format by finding an appopriate manifest from the manifest list, unpacking the filesystem serializations in the correct order, and then converting the image configuration into an OCI Runtime config.json.
+
+The image layout has two top level directories: 
+
+- "blobs" contains content-addressable blobs
```
stevvooe: I'm not sure I have an opinion either way, I'll regret not bringing this up: we use the term "blobs" because these are free-form binary data. They are not "objects" because they have no required structure, other than that imposed by their access path (ie, through a descriptor).

Using something like "objects" would imply structure or a header, similar to git.

I'm not sure if we include this distinction in the spec.

wking: On Fri, May 27, 2016 at 12:28:05PM -0700, Stephen Day wrote:
> > +- "blobs" contains content-addressable blobs
>
> I'm not sure I have an opinion either way, I'll regret not bringing
> this up: we use the term "blobs" because these are free-form binary
> data. They are not "objects" because they have no required
> structure, other than that imposed by their access path (ie, through
> a descriptor).

I'm not sure what your comment is in response to, but I'm ok with
“objects” → “blobs” throughout this PR.

stevvooe: I am defending the use of the term "blobs" and wondering if we should enshrine said defense. Up until, we've just used it because that is what we've used, whereas there is a deliberate difference between an "object" and a "blob".

wking:
Thinking over the backwards-compat discussion in #95, I think we want
to say something here about how we see that working with this layout.
Basically, we want to make it easy and extremely unambiguous to
identify content following this layout from content following some
other layout.

My guess is that we want to give this format a specific name
(application/vnd.oci.image.layout.v1?) in this PR, but punt exposing
that name to the transport mechanisms (e.g. file-based transports can
put it in a magic header, HTTP-based transport can have have a
/version path, etc.).  Or maybe we want to require a ‘version’ file
holding the name, and folks who want to *also* add a magic header are
free to do so?

image-layout.md
wking: On Fri, May 27, 2016 at 12:35:56PM -0700, Stephen Day wrote:
> > +- "blobs" contains content-addressable blobs
>
> I am defending the use of the term "blobs" and wondering if we
> should enshrine said defense…

How about:

  The image layout does not specify an internal structure for blobs;
  any semantics must be inferred from the blob content or the context
  from which the blob was referenced.

philips:
I am fine adding a root /oci-layout file with contents like:

```
{"ociLayoutVersion": "1.0.0"}
```

image-layout.md
philips: I will add some blobs are opaque language.

wking:
On Fri, May 27, 2016 at 02:16:17PM -0700, Brandon Philips wrote:
> I am fine adding a root /oci-layout file with contents like:
> 
> ```
> {"ociLayoutVersion": "1.0.0"}
> ```

Sounds fine to me, although I don't see a point to putting ‘oci’ in
both the filename and JSON content.

philips:
@wking @stevvooe all feedback addressed.

@vbatts can you give this a review?

image-layout.md
```diff
@@ -0,0 +1,55 @@
+## Open Container Initiative Image Layout Specification
+
+The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and human significant [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
```
wking: I [still][1] don't see much point in the hard-to-define “human significant”, especially since location-addressable is sufficient to distinguish refs from blobs.

[1]: https://github.com/opencontainers/image-spec/pull/94#discussion_r64858029

image-layout.md
```diff
@@ -0,0 +1,55 @@
...
+./refs
+./refs/v1.0
+./refs/v1.1
+```
+
+Object names in the refs and blobs MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
+Hash algorithm identifiers containing the colon `:` will be converted to the hyphen `-`.
+For example `sha256:5b` will map to the layout `blobs/sha256-5b`.
+The blobs directory MAY contain blobs which are not referenced by any of the refs.
+The blobs directory MAY be missing referenced blobs the missing blobs SHOULD be fulfilled by an external blob store.
```
wking: “… missing referenced blobs*, in which case* the missing blobs… ” (emphasizing the words I think you're missing).

image-layout.md
```diff
@@ -0,0 +1,55 @@
...
+Given an image layout a tool can convert a given ref into a runnable OCI Image Format by finding an appopriate manifest from the manifest list, unpacking the filesystem serializations in the correct order, and then converting the image configuration into an OCI Runtime config.json.
+
+The image layout has two top level directories:
+
+- "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque. It is addressed by its hashed contents from byte 0 to the end of the file.
+- "refs" contains descriptors pointing to an image manifest list
+
+It also contains a file that is used to identify the layout version:
+
+- "oci-layout" MUST contain a JSON object with a version field `{"version": "1.0.0"}` and MAY include additional fields.
```
wking: Do we want this filename to contain `image` somewhere (for namespacing vs. other OCI projects)?

image-layout.md
```diff
@@ -0,0 +1,55 @@
+## Open Container Initiative Image Layout Specification
+
+The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and human significant [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
+This layout MAY be used in a variety of different transport mechanisms: archive formats (e.g. tar, zip), shared filesystem environments (e.g. nfs), or networked file fetching (e.g. http, ftp, rsync).
+Given an image layout a tool can convert a given ref into a runnable OCI Image Format by finding an appopriate manifest from the manifest list, unpacking the filesystem serializations in the correct order, and then converting the image configuration into an OCI Runtime config.json.
+
+The image layout has two top level directories:
+
+- "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque. It is addressed by its hashed contents from byte 0 to the end of the file.
```
wking: Can the last sentence be “The full blob is hashed as it is stored.”?  I understand you're trying to rule out headers, compression, etc., but “from byte 0 to the end of the file” is a bunch of words in a somewhat tangential direction.  Maybe post an example (`sha256sum blobs/sha256-aff…` → `aff…`)?

vbatts:
Would this be an alternative to the combined layout?

philips:
@vbatts yes, combined has been removed because it is based on a bunch of organically grown schemas that aren't super documented. See #95

wking:
On Tue, May 31, 2016 at 02:18:46PM -0700, Vincent Batts wrote:
> Would this be an alternative to the combined layout? 

A replacement (see #95).

vbatts:
oh wow. I'll move on from protecting the combined format then, once this is incorporated.

So definitely a shim or `docker load --format=oci.v1 ..` will be needed.

philips:
@vbatts yes.

image-layout.md
```diff
@@ -0,0 +1,55 @@
...
+This is an example image layout:
+
+```
+$ cd example.com/app/
+$ find .
+.
+./blobs
+./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
+./blobs/sha256-5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
+./blobs/sha256-e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
```
vbatts: I would break this down further. In the even of a _lot_ of objects, this would be a mess. 
`./blobs/sha256/af/ff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51`
or even just 
`./blobs/sha256/afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51`

wking:
On Tue, May 31, 2016 at 02:23:07PM -0700, Vincent Batts wrote:
> So definitely a shim or `docker load --format=oci.v1 ..` will be
> needed.

The version file (discussion starting around [1], current filename is
‘oci-layout’ [2]) is designed to make the format easy to auto-detect,
so I doubt we need a --format field.  Being able to set that
explicitly may be nice, but I don't think users should have to bother.

[1]: https://github.com/opencontainers/image-spec/pull/94#issuecomment-222242438
[2]: https://github.com/opencontainers/image-spec/pull/94/files#r65251233

image-layout.md
philips: adding an example instead

image-layout.md
```diff
@@ -0,0 +1,55 @@
...
+The blobs directory MAY be missing referenced blobs the missing blobs SHOULD be fulfilled by an external blob store.
+
+Each object in the refs subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
+In general the `mediatype` of this descriptor object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
+
+This illustrates the expected contents of a given ref and the manifest list it points to.
+
+```
+$ cat ./refs/v1.0
+{"size": 4096, "digest": "sha256:afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51", "mediatype": "application/vnd.oci.image.manifest.list.v1+json"}
```
vbatts: will this doc have a struc and schema somewhere too?

image-layout.md
```diff
@@ -0,0 +1,55 @@
...
+For example `sha256:5b` will map to the layout `blobs/sha256-5b`.
+The blobs directory MAY contain blobs which are not referenced by any of the refs.
+The blobs directory MAY be missing referenced blobs the missing blobs SHOULD be fulfilled by an external blob store.
+
+Each object in the refs subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
+In general the `mediatype` of this descriptor object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
+
+This illustrates the expected contents of a given ref and the manifest list it points to.
+
+```
```
vbatts: json

vbatts: well.. shell ... hrm

image-layout.md
wking: On Tue, May 31, 2016 at 02:26:40PM -0700, Vincent Batts wrote:
> I would break this down further. In the even of a _lot_ of objects, this would be a mess. 
> `./blobs/sha256/af/ff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51`
> or even just 
> `./blobs/sha256/afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51`

How many layers do you expect to put in a single lump?  I think this
is premature optimization that should be punted to later versions ;).
And a performant file format containing lots of blobs is probably
going to have to abandon tar at some point for a binary format like
[1], although then you lose the transport-agnosticism of the current
format.

[1]: http://git.kernel.org/cgit/git/git.git/tree/Documentation/technical/pack-format.txt?id=v2.8.3

philips: For others: the reason things like [git](http://programmers.stackexchange.com/questions/301400/why-is-the-git-git-objects-folder-subdivided-in-many-sha-prefix-folders) and [diskv](https://github.com/peterbourgon/diskv#basic-idea) (used in rkt) do this is because POSIX filesystems like ext4 just suck at having 32,000+ dentries. And some CLI tools like `ls` can get grumpy too. This generally isn't applicable to things like HTTP, s3, ceph, etc. 

To fix this we can transform `sha256-ff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51` to `sha256/af/ff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d5`

I am OK with this change. @stevvooe @wking ?

image-layout.md
philips: see #92

image-layout.md
philips: see #92

image-layout.md
wking: On Tue, May 31, 2016 at 02:43:32PM -0700, Brandon Philips wrote:
> To fix this we can transform
> `sha256-ff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51`
> to
> `sha256/af/ff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d5`
> 
> I am OK with this change. @stevvooe @wking ?

I'm not intimately familiar with tar, but tar(5) does not talk about
indexes or other super-record structures.  So I don't think this
change provides a useful speedup for tarred layouts.  This means out
of our available transport mechanisms, we're optimizing for only one
(local filesystem access), and then for a subset of filesystems.
Until someone says “oh man, image $NAME really drags without
filename-based blob sharding”, I think we should leave the blob
directory as it stands [1].

[1]: https://github.com/opencontainers/image-spec/pull/94/files/c6144032bb3679f1b0c317c69924fa7e4c53a63a#r65266627

philips: @vbatts It looks like ext4 is mostly OK: http://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory

philips: Looking at the performance on ext4 I am OK punting on this for image layout version 2. cc @vbatts ?

philips:
@opencontainers/image-spec-maintainers Can people please review and give an LGTM?

We have one outstanding question from @vbatts which I am OK fixing up later based on current understanding of Linux filesystems: https://github.com/opencontainers/image-spec/pull/94#discussion_r65265938

image-layout.md
vbatts: So `TODO plug the descriptor stuffs here`.

image-layout.md
vbatts: I don't think it's that clean, but we can punt.

vbatts:
All-in-all LGTM

philips:
@stevvooe LGTM from you as you have been following this closely?

philips:
LGTM

philips:
FWIW, this is missing two things:

1. The merging of #111 for the descriptor type that appears in here (I would say this is optional as the descriptor appears in other parts of the spec)
2. An LGTM from another @opencontainers/image-spec-maintainers

image-layout.md
stevvooe: Sorry, I missed this comment. ext4 is okay but many file systems do still struggle without the prefixed hash directories. Partitioning can help with remote file systems, as well.

There are also cases where you want to partition based on algorithm, which easier if you can just pick up the whole directory, rather than match a file prefix.

stevvooe:
@philips I really think we should go with partition blob storage, over the `-` prefixed files.

We may also want to have `sha256/aaa/data`, over `sha256/aaa`. This allows you to store companion data.

wking:
On Tue, Jun 07, 2016 at 09:23:45PM -0700, Stephen Day wrote:
> We may also want to have `sha256/aaa/data`, over `sha256/aaa`. This
> allows you to store companion data.

“Content-addressable violation” warning lights started flashing all
over at my house once I read “companion data” ;).

image-layout.md
wking: On Tue, Jun 07, 2016 at 09:21:52PM -0700, Stephen Day wrote:
> ext4 is okay but many file systems do still struggle without the
> prefixed hash directories…

Under blob counts that you'd still consider stuffing into a single
tarball?  I'd still like to see someone post a link to a real-life
tarball that uses this layout to motivate any optimizations like this,
since the tar-layer approach makes me think it unlikely that you have
more than a few dozen blobs in a given layout.

> There are also cases where you want to partition based on algorithm…

Can you point to docs on those cases?

philips:
@stevvooe @wking Yea, I don't understand what "companion data" means and why "companion data" can't live in `companion-data`. Can you please file a separate issue with some use cases.

image-layout.md
philips: @stevvooe What operations on a remote filesystem? Do you have benchmarks?

100% agree this feels like a premature optimization @wking.

philips:
@stevvooe @wking lets discuss companion data on https://github.com/opencontainers/image-spec/issues/129

philips:
The only thing blocking this is the continued discussion on some sort of prefix sharding: https://github.com/opencontainers/image-spec/pull/94#discussion_r65265938

I personally feel like we should punt on the sharding. But, if someone can come up with some benchmarks or other evidence that it is necessary for performance or other reasons I am happy to add the language.

stevvooe:
> “Content-addressable violation” warning lights started flashing all
over at my house once I read “companion data” ;).

Yep. That is somewhat the point.

Let's say that you want to put something next to the target:

```
sha256/aa/aaaaaa/
   data
```

This allows you to later place something next to it:

```
sha256/aa/aaaaaa/
   data
   references
```

Its unverified in the default access path, but it can be handy for optimizations. Leaving the extra path component leaves space to add something here in the future.

The best argument against doing this is that most of this should be in the descriptor or other secured data.

> I personally feel like we should punt on the sharding. But, if someone can come up with some benchmarks or other evidence that it is necessary for performance or other reasons I am happy to add the language.

This is a best practice. I do admit, there may be a cargo-cult aspect to it.

wking:
On Wed, Jun 08, 2016 at 06:51:59PM -0700, Brandon Philips wrote:
> The only thing blocking this is the continued discussion on some
> sort of prefix sharding…

I also think it should be rebased around #111 so it can link to the
new descriptor docs.

stevvooe:
@philips Let's not block on the directory layout.

LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/image-spec/pull-request/94/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/image-spec)

vbatts:
LGTM
Though I feel like the structure will come up again

PR #95, opened 2016-05-27
serialization: remove combined spec

philips:
The combined spec isn't really part of the image format and introduces
new objects that we don't have a schema for. Largely I think that it
should be replaced by something like the image layout.

https://github.com/opencontainers/image-spec/pull/94


philips:
This is from a discussion with @stevvooe and @wking that is outlined in #92. 

cc @vbatts  @opencontainers/image-spec-maintainers

wking:
+1 to this (#94 is a nicer way to accomplish the same goal).

How does removing it square with backwards-compat [1]?  Maybe you
should be deprecating instead of removing?

[1]: https://github.com/opencontainers/image-spec/blob/ec271e880f3eec688344ccb479fab6c49a00cf52/media-types.md#applicationvndociimageserializationcombinedv1json

philips:
@wking AFAIK this combined format is never used on the wire anywhere. It is only in `docker save/load`.

wking:
On Fri, May 27, 2016 at 12:23:05PM -0700, Brandon Philips wrote:
> @wking AFAIK this combined format is never used on the wire
> anywhere. It is only in `docker save/load`.

I'm not sure how far image-spec's compatibility goal extends ;).  #94
is doing the same sort of thing, so it seems like the *idea* is in
scope for image-spec.  Supporting something like #94 but *not* the old
Docker equivalent sounds like a broken backwards-compat.

Personally, I'm not really a backwards-compat-in-the-spec guy ;).  As
long as the spec calls for appropriate identifiers, external tooling
can provide the backwards-compat (e.g. an archive unpacker that
recognizes and unpacks both the old Docker format and the new #94).
But I'm not sure where the image-spec maintainers stand on that sort
of thing.

stevvooe:
The implication here is that we lose compatibility with the `docker save` format.

For testing purposes, OCI (tools or here) should have a way of processing `docker save` files into the OCI image format, but I am split on whether or not we want to "lock down" the `docker save` format. The main issue being that it is an organic format, with several subtleties, making a correct specification challenging.

The format in #94 is wildly better, but we really need to leverage the existing tools for bootstrap of OCI.

I do however agree that this doesn't really belong in the image config specification.

philips:
@stevvooe Is there a library or something for parsing all of those formats in docker save/load? We can certainly put them into oci-image-tool. Would you be OK merging this PR if I filed an issue to add that as an oci-image-tool feature?

stevvooe:
@philips The package is [here](https://github.com/docker/docker/blob/b316bc42fe1ad8edc709bf28975fb3a52e766344/image/tarexport/tarexport.go#L38).

> Would you be OK merging this PR if I filed an issue to add that as an oci-image-tool feature?

That sounds reasonable. We may want to add some language about translation to target formats for backwards compat for the packed archive in #94.

philips:
OK, I will add language to #94. Wanna LGTM this?

stevvooe:
LGTM

jonboulle:
LGTM

PR #96, opened 2016-05-31
travis: make the docs in CI

vbatts:
This will likely currently fail, as the pandoc is grumpy with the image references in master, but such is a good test case that it should fail CI if that happens.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,16 @@ language: go
 go:
   - 1.6
 
-sudo: false
+sudo: required
+
+services:
+  - docker
 
 before_script:
   - export PATH=$HOME/gopath/bin:$PATH
 
 before_install:
+  - docker pull vbatts/pandoc
   - go get github.com/vbatts/git-validation
   - go get github.com/alecthomas/gometalinter
   - gometalinter --install --update
@@ -21,3 +25,4 @@ script:
   - make check-license
   - make test
   - make oci-image-tool
+  - make docs


philips:
Good idea! Let me know with an IRC ping once you get it working.

vbatts:
@stevvooe from your blackfriday experiences, is there a creative way to look for `img src="...">` and replace it with alternate content?

vbatts:
god. We could just use blackfriday instead of pandoc ...

vbatts:
though it's not as pretty as latex in pandoc. and would be easier to just provide an html doc. Such a hastle...

vbatts:
ha, and doesn't handle inline html...

stevvooe:
> from your blackfriday experiences, is there a creative way to look for img src="..."> and replace it with alternate content?

You get access to a kind of AST through one of the callbacks.

> ha, and doesn't handle inline html...

That seems like a bug. Markdown is supposed to handle and escape inline html. Are you sure it is not an option?

To tell you the truth, I was going to replace blackfriday with a tokenizer once I had some time...

vbatts:
@stevvooe i would love this markdown to presentation to be smoother. :-\

vbatts:
#97 fixes the `make docs` target

vbatts:
rebasing on #97 to be sure

vbatts:
cool. Passes clean. Rebased back to master.

philips:
LGTM

stevvooe:
LGTM

Ghostman2591 (approved):

PR #78, opened 2016-05-19
*: add license header checking and tweak Makefile

jonboulle:
- move file-finding logic into lint check script
- move lint check script into .tool/ subdirectory (similar to
  opencontainers/runtime-spec)
- add a simple `check-license` script to .tool/ to ensure all Go source
  files have an Apache license header, and wire up to `check-license`
  make target
- invoke this from travis
- add `schema-fs` target to generate schema/fs.go (including header)
- add header everywhere it's missing so far
- add `output/` directory to .gitignore

Signed-off-by: Jonathan Boulle <jonathanboulle@gmail.com>

PR #111, opened 2016-06-01
spec: describe descriptors and digests

### LINKED ISSUE #92: media-type: add a media-type for descriptor

stevvooe:
Add a definition for content descriptors and digests, which are used as
content identifiers in the OCI image specification.

This is a draft of the descriptor specification, which includes the digest calculation.

Before merging:

- [x] Add json schema definition
- [x] Add descriptor examples to unit test suite

Closes #92 

cc @philips @vbatts

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -1,9 +1,3 @@
-<!--[metadata]>
-+++
-draft = true
-+++
-<![end-metadata]-->
-
 # OpenContainers Content Descriptors
 
 OCI have several components that come to together to describe an image.
@@ -77,7 +71,7 @@ digest                                                                          
 sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b           | Common sha256 based digest                    |
 
 Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the _digest_.
-The size of the content SHOULD be verified, as well, to protect against [Length Extension Attacks](https://en.wikipedia.org/wiki/Length_extension_attack).
+The size of the content SHOULD be verified to reduce hash collision space.
 Heavy processing of before calculating a hash SHOULD be avoided.
 Implementations MAY employ some canonicalization to ensure stable content identifiers.
 
--- a/descriptor.md
+++ b/descriptor.md
@@ -1,0 +1,123 @@
+<!--[metadata]>
++++
+draft = true
++++
+<![end-metadata]-->
+
+# OpenContainers Content Descriptors
+
+OCI have several components that come to together to describe an image.
+References between components form a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
+The references in the _Merkle DAG_ are expressed through _Content Descriptors_.
+A _Content Descriptor_ or _Descriptor_, describes the disposition of targeted content.
+A _Descriptor_ includes the type of content, an independently-verifiable content identifier, known as a "digest" and the byte-size of the raw content.
+
+Descriptors SHOULD be embedded in other formats to securely reference external content.
+
+Other formats SHOULD use descriptors to securely reference external content.
+
+## Properties
+
+The following describe the primary set of properties that make up a _Descriptor_.
+
+- **`mediaType`** *string*
+
+  This REQUIRED property contains the MIME type of the referenced object.
+
+- **`digest`** *string*
+
+  This REQUIRED property is the _digest_ of the targeted content, meeting the requirements outlined in [Digests and Verification](#digests-and—verification).
+  Retrieved content SHOULD be verified against this digest when consumed via untrusted sources.
+
+- **`size`** *int*
+  This REQUIRED property specifies the size in bytes of the blob.
+  This property exists so that a client will have an expected size for the content before validating.
+  If the length of the retrieved content does not match the specified length, the content SHOULD NOT be trusted.
+
+### Reserved
+
+The following are field keys that MUST NOT be used in descriptors specified in other OCI specifications:
+
+- **`urls`** *array*
+
+  This key is RESERVED for future versions of the specification.
+
+- **`data`** *string*
+
+  This key is RESERVED for futures versions of the specification.
+
+All other fields may be included in other OCI specifications.
+Extended _Descriptor_ field additions proposed in other OCI specifications SHOULD first be considered for addition into this specification.
+
+## Digests and Verification
+
+The _digest_ component of a _Descriptor_ acts as a content identifier, employing [content addressability](http://en.wikipedia.org/wiki/Content-addressable_storage) for the OCI image format.
+It uniquely identifies content by taking a collision-resistant hash of the bytes.
+Such an identifier can be independently calculated and verified by selection of a common _algorithm_.
+If such an identifier can be communicated in a secure manner, one can retrieve the content from an insecure source, calculate it independently and be certain that the correct content was obtained.
+Put simply, the identifier is a property of the content.
+
+To disambiguate from other concepts, we call this identifier a _digest_.
+A _digest_ is a serialized hash result, consisting of a _algorithm_ and _hex_ portion.
+The _algorithm_ identifies the methodology used to calculate the digest, which are shared by implementations.
+The _hex_ portion is the hex-encoded result of the hash.
+
+We define a _digest_ string to match the following grammar:
+
+```
+digest      := algorithm ":" hex
+algorithm   := /[A-Fa-f0-9_+.-]+/
+hex         := /[A-Fa-f0-9]+/
+```
+
+Some examples of _digests_ include the following:
+
+digest                                                                            | description                                   |
+----------------------------------------------------------------------------------|------------------------------------------------
+sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b           | Common sha256 based digest                    |
+
+Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the _digest_.
+The size of the content SHOULD be verified, as well, to protect against [Length Extension Attacks](https://en.wikipedia.org/wiki/Length_extension_attack).
+Heavy processing of before calculating a hash SHOULD be avoided.
+Implementations MAY employ some canonicalization to ensure stable content identifiers.
+
+### Algorithms
+
+While the _algorithm_ does allow one to implement a wide variety of algorithms, compliant implementations SHOULD use [SHA-256](#SHA-256).
+
+Let's use a simple example in pseudo-code to demonstrate a digest calculation:
+A _digest_ is calculated by the following pseudo-code, where `H` is the selected hash algorithm, identified by string `<alg>`:
+```
+let ID(C) = Descriptor.digest
+let C = <bytes>
+let D = '<alg>:' + EncodeHex(H(C))
+let verified = ID(C) == D
+```
+Above, we define the content identifier as `ID(C)`, extracted from the `Descriptor.digest` field.
+Content `C` is a string of bytes.
+Function `H` returns a the hashs of `C` in bytes and is passed to function `EncodeHex` to obtain the _digest_.
+The result `verified` is true if `ID(C)` is equal to `D`, confirming that `C` is the content identified by `D`.
+After verification, the following is true:
+
+```
+D == ID(C) == '<alg>:' + EncodeHex(H(C))
+```
+
+The _digest_ is confirmed as the content identifier by independently calculating the _digest_.
+
+#### SHA-256
+
+[SHA-256](https://tools.ietf.org/html/rfc4634#page-7) is a collision-resistant hash function, chosen for ubiquity, reasonable size and secure characteristics.
+Implementations MUST implement SHA-256 digest verification for use in descriptors.
+
+## Examples
+
+The following example describes a [_Manifest_](manifest.md#image-manifest) with a content identifier of "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270", of size 7682 bytes:
+
+```json,title=Content%20Descriptor&mediatype=application/vnd.oci.descriptor.v1%2Bjson
+{
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
+  "size": 7682,
+  "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"
+}
+```
--- a/schema/image-manifest-schema.json
+++ b/schema/image-manifest-schema.json
@@ -14,12 +14,12 @@
       "$ref": "defs-image.json#/definitions/mediaType"
     },
     "config": {
-      "$ref": "defs-image.json#/definitions/descriptor"
+      "$ref": "content-descriptor.json"
     },
     "layers": {
       "type": "array",
       "items": {
-        "$ref": "defs-image.json#/definitions/descriptor"
+        "$ref": "content-descriptor.json"
       }
     },
     "annotations": {
--- a/media-types.md
+++ b/media-types.md
@@ -2,6 +2,7 @@
 
 The following `mediaType` MIME types are used by the formats described here, and the resources they reference:
 
+- `application/vnd.oci.descriptor.v1+json`: [Content Descriptor](descriptor.md)
 - `application/vnd.oci.image.manifest.list.v1+json`: [Manifest list](manifest.md#manifest-list)
 - `application/vnd.oci.image.manifest.v1+json`: [Image manifest format](manifest.md#image-manifest)
 - `application/vnd.oci.image.serialization.rootfs.tar.gzip`: ["Layer", as a gzipped tar archive](serialization.md#creating-an-image-filesystem-changeset)
--- a/schema/content-descriptor.json
+++ b/schema/content-descriptor.json
@@ -1,0 +1,24 @@
+{
+  "description": "OpenContainer Content Descriptor Specification",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://opencontainers.org/schema/descriptor",
+  "type": "object",
+  "properties": {
+    "mediaType": {
+      "description": "the mediatype of the referenced object",
+      "$ref": "defs-image.json#definitions/mediaType"
+    },
+    "size": {
+      "description": "the size in bytes of the referenced object",
+      "type": "integer"
+    },
+    "digest": {
+      "$ref": "defs-image.json#definitions/digest"
+    }
+  },
+  "required": [
+    "mediaType",
+    "size",
+    "digest"
+  ]
+}
--- a/schema/defs-image.json
+++ b/schema/defs-image.json
@@ -11,28 +11,6 @@
       "type": "string",
       "pattern": "^[a-z0-9]+:[a-fA-F0-9]+$"
     },
-    "descriptor": {
-      "id": "https://opencontainers.org/schema/image/descriptor",
-      "type": "object",
-      "required": [
-        "mediaType",
-        "size",
-        "digest"
-      ],
-      "properties": {
-        "mediaType": {
-          "description": "the mediatype of the referenced object",
-          "$ref": "#definitions/mediaType"
-        },
-        "size": {
-          "description": "the size in bytes of the referenced object",
-          "type": "integer"
-        },
-        "digest": {
-          "$ref": "#definitions/digest"
-        }
-      }
-    },
     "manifestDescriptor": {
       "id": "https://opencontainers.org/schema/image/manifestDescriptor",
       "type": "object",
--- a/schema/fs.go
+++ b/schema/fs.go
@@ -203,7 +203,7 @@ var _escData = map[string]*_escFile{
 	"/config-schema.json": {
 		local:   "config-schema.json",
 		size:    707,
-		modtime: 1463700693,
+		modtime: 1464139547,
 		compressed: `
 H4sIAAAJbogA/5SRPW7DMAyF5/oUhpOxjjp0ytoDdOgJVJmKGcCiQDJDUPju1U/c2kvhLobx+L73JOqr
 adtuAHGMUZFCd2679wjhjYJaDMBt+vN4aT8iOPTobHE9Z+woboTJZmRUjWdjrkKhr+qJ+GIGtl77l1dT
@@ -213,10 +213,23 @@ T4aE9IoTdGU2V0tnbzoS/xG1dbMbUdPhbgx7GZK9zscuVu4jgy+HBy99HZ/yKxxMUjBgfi1ZdrjJYiL1
 `,
 	},
 
+	"/content-descriptor.json": {
+		local:   "content-descriptor.json",
+		size:    616,
+		modtime: 1464914956,
+		compressed: `
+H4sIAAAJbogA/4yRMVPDMAyF9/wKXdqR1gxMXWFngI1jcG0lVe9iG1kMhet/x4oTSGGgW/L8vvck+7MB
+aD1mx5SEYmh30D4mDPcxiKWADPqFQeBhMkWGp4SOOnJ2JG40Yp3dAQer+EEk7Yw55hg2Vd1G7o1n28nm
+9s5UbVU58jOSCxNLs5ub84hVt/Hf7ZWTU0Il4/6ITqqWuPAshLmc6GJFG9CTfa7mKv3dVw4Io09DIXag
+AmOHXKZBD4uOEV+XM+U8dnlDg+1xq8uuyj8F0tRsfnpH6lzhNtPHf5OoBSjA/iSYr5hmvgkqz9QjX/Z5
+6jHLsvGa4SeqJjVTWsv49k6M+mAvvy93ud5ldfl5bc7NVwAAAP//Zc2MR2gCAAA=
+`,
+	},
+
 	"/defs-config.json": {
 		local:   "defs-config.json",
 		size:    1755,
-		modtime: 1463700704,
+		modtime: 1464139547,
 		compressed: `
 H4sIAAAJbogA/7xUzc7TMBA8N09hBY6BXhAHri1HVKQIOFZusm63xF5rvQEi1HfHSauS/qRKKV8PVeOx
 Z2bXY/t3olRaQigYvSC59INK52DQYTsKymsWLOpKsxJSCw9uRk40OmAVvwyuVe6hQIOF7vjZXvCoEAVb
@@ -230,27 +243,27 @@ W1SwXppll8oQfUVUgXaDqSTtb5f8CQAA//8Cok052wYAAA==
 
 	"/defs-image.json": {
 		local:   "defs-image.json",
-		size:    3100,
-		modtime: 1462965360,
+		size:    2528,
+		modtime: 1464914958,
 		compressed: `
-H4sIAAAJbogA/+xWy27bMBC86ysIJUAOfqjXGkGAoEGBnlIgPdVQgY20kphKpLqkgzqB/r3U+x3Xidte
-erK55A5nhmPSzxZjto/KI55qLoW9YfYNBlzwfKRYCqS5t4uBmJbsNkXxQQoNXCCxTwmEyO5S9HjAPSja
-lyVeA2Dw8i1MMUGfw5d9ik3JFLmfbxhpnaqN40gD79Xwai0pdJQXYQIOz7dyWohlDaBLQFtp4iJs6ylo
-jVTI+baF1ZO7cLbvVu/Nt+vV1/XCXZzbxdKs7LB9HqLSXWoDU3SEzKN9qmVIkEbcY4aZ913tElb2Mhmw
-fJG8f0BPLxkXxbAiwi4uI1DR1eYywp/gG8sSiKvOq4vj9Rgt7mJjvgXXq4/FYCiooi/p9X53MEYES5lt
-nfDHjhPm+Nuq1jv0ZVtU/Kk3rryvCm6rmQxBEz9UHQkzSZo7smJtzrk+HsIAychGnw0kFBDnZj7vPetk
-uJO7Zmk21HOYSr4sT8X9XqP6TTq121xoDJGm9x9ld15J32oDY3U/6+wkIHhg1t2cIEMTWH8hS51KGoMO
-JCX/8/Uv8jV1EAOg4/LUoEzqmNI4maZiBsiLuDYNO8Jej5mTyu4U3B7iTHDGmMPZV6t1XqA6fjV609lY
-2OloGbA3klk/mh3KFJ+OVAP6VnIBQu74aS1rUWfpARHsx9MmAckUlwPC2nt+WugjEAcx/IUf7ddLZv0x
-YSMo8P3iMoL4c/dnGkCs0JrzJDvwIoIQUkP/H+3REeiCNI+QFHgb9K6myVvWXLJq/aCkOHN6Lwekd4Uv
-dwN3Or48T12UYhfH478BrlWPMiuzfgUAAP//VjUNyBwMAAA=
+H4sIAAAJbogA/7RWy27bMBC8+ysIJUAOfqjXGkGAoEGBnlIgPTVQgY20kphKpLqkgzqB/r2k3k+nrt0b
+ueQOZ4Zjym8LxpwAlU8801wKZ8ucOwy54HamWAakub9LgJiW7D5D8UkKDVwgsS8pRMgeMvR5yH0o2lcl
+XgNg8OwRpphiwOHbPsOmZIo8sAfGWmdq67rSwPs1vNpIilzlx5iCy+1RbguxqgF0CegoTVxEbT0DrZEK
+OT8eYf3qLd3HD+uPZnS7/r5ZestLp9ialx1OwCNUukttYIqOkfm0z7SMCLKY+8ww83+qXcrKXiZDZjfJ
+p2f09YpxUUwrIuzqOgYV32yvY/wNgbEshaTqvLk6Xo/R4i23ZhTerj8Xk4GgFAQPDfhdJUPSCb6PsUaE
+S9ltnfDXjhPacx6rWi8Eq7ao+GtvXt1Fp5IloENJqVOVvNYXMuRNRFF15M2kbe5ai71WR32FhCGSsQQD
+NpBVQFyaddt70cl5J5vN1nyo8X0qdptNztNeo/pLOvUNcKExQpo+f5TveSXV1kmY5iIGQMflqUGZ1DGl
+cTJNxQqQH3NtGnaEvR6zJpXTKXg9xJngjDGHq/+s1j1AdfzL7y3nY2Hno2XATiSzeTEnlCk+H6kG9FRy
+IYJ1/LyWtaiz9IAI9uNlk4B0iss7woy0g0JfgDiI4S/8aL8OmfXfhI2gIAiKxwiSr92faQiJwsWcJ+24
+HuW9LyIIITX0/5UcHYEuSPMRkgLvw97TNPnKmkdWbZ6VFBdu78sB2UPhy8PAnY4vb1MPpdgliTMS7S3q
+Wb7IF38CAAD//wKthPngCQAA
 `,
 	},
 
 	"/defs.json": {
 		local:   "defs.json",
 		size:    3193,
-		modtime: 1463700693,
+		modtime: 1464139547,
 		compressed: `
 H4sIAAAJbogA/7RWTXPaMBC98ys8tEfa2PIX9NYp/cghAzOZnjo9uGYBtSCpstxpmuG/VzLGWPZiMKWH
 JPau9r23T6tYzwPHGS4gSyUVinI2fOMMp7CkjJq3zMkzWDhqLXm+WvNc6UdwZgLYO85UQhlI51FASpc0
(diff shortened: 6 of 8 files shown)


descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+
+# OpenContainers Content Descriptors
+
+OCI have several components that come to together to describe an image. 
+References between components form a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
+The references in the _Merkle DAG_ are expressed through _Content Descriptors_.
+A _Content Descriptor_ or _Descriptor_, describes the disposition of targeted content.
+A _Descriptor_ includes the type of content, an independently-verifiable content identifier, known as a "digest" and the byte-size of the raw content.
+
+Descriptors SHOULD be embedded in other formats to securely reference external content.
```
wking: I think this should be turned around to get “Other formats SHOULD use descriptors to securely reference external content”.

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+
+Descriptors MAY be used as an independent object to specify content.
+
+Implementations MUST verify fetched content against
```
wking: “against…”?  Should probably be “ȧgainst the digest”.

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Properties
+
+- **`mediaType`** *string*
+
+  This REQUIRED property contains the MIME type of the referenced object.
+
+- **`digest`** *string*
+
+  This REQUIRED parameter is the digest of the content, meeting the requirements outlined in [Digests and Verification](#digests-and—verification).
```
wking: You use “property” for the other structures and the section header, but “parameter” here.  Both terms are fine with me, but it's probably best to pick one for this spec and stick with it.

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+  Retrieved content SHOULD be verified against this digest before processing.
```
wking: This SHOULD conflicts with your earlier “Implementations MUST verify…”.  I like SHOULD better, because if you know a separate local layer has already verified the hash, there's no sense in having your layer do it again.  We should probably just drop the earlier line.

descriptor.md
wking: Actually, I think we should just [drop this line][1].

[1]: https://github.com/opencontainers/image-spec/pull/111/files#r65460586

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+
+- **`size`** *int*
+  This REQUIRED property specifies the size in bytes of the object.
+  This property exists so that a client will have an expected size for the content before validating.
+  If the length of the retrieved content does not match the specified length, the content should not be trusted.
```
wking: “should” → “MUST”, or at least “SHOULD”.  Why require a size field and then allow for size mismatches?

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+
+  This REQUIRED property contains the MIME type of the referenced object.
+
+- **`digest`** *string*
+
+  This REQUIRED parameter is the digest of the content, meeting the requirements outlined in [Digests and Verification](#digests-and—verification).
+  Retrieved content SHOULD be verified against this digest before processing.
+
+- **`size`** *int*
+  This REQUIRED property specifies the size in bytes of the object.
```
wking: [“object”→ “blob”][1]? ;)

[1]: https://github.com/opencontainers/image-spec/pull/94#discussion_r64955841

descriptor.md
stevvooe: This should probably say MUST.

> if you know a separate local layer has already verified the hash, there's no sense in having your layer do it again.

If you already have it locally, the requirement does not apply. I'll add a blurb about this in the section on calculation saying that it is okay to cache calculation.

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+hex         := /[A-Fa-f0-9]+/
+```
+
+Some examples of _digests_ include the following:
+
+digest                                                                            | description                                   |
+----------------------------------------------------------------------------------|------------------------------------------------
+sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b           | Common sha256 based digest                    |
+
+Before consuming content targeted by a descriptor, the byte content MUST be verified against the _digest_.
```
wking: This is another MUST that I'd weaken to [SHOULD][1].

[1]: https://github.com/opencontainers/image-spec/pull/111/files#r65460586

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+Heavy processing of before calculating a hash SHOULD be avoided.
```
wking: “processing of before” → “processing before”.  And maybe explain that this suggestion is denial-of-service protection.

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+Implementations MAY employ some canonicalization to ensure stable content identifiers.
```
wking: This content-production suggestion seems out-of-scope for a digest spec.

descriptor.md
stevvooe: How can we have a secure merkle DAG if an implementation can ignore a digest?

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Algorithms
+
+While the _algorithm_ does allow one to implement a wide variety of algorithms, compliant implementations SHOULD use [SHA-256](#SHA-256). 
+
+Let's use a simple example in pseudo-code to demonstrate a digest calculation:
```
wking: This example is generic “Digests and Verification” stuff, so I think we should move it out of the “Algorithms” subsection.

descriptor.md
stevvooe: This is a fairly important consideration when building content-addressable systems. This is saying the practice is acceptable but discouraged.

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+```
+D == ID(C) == '<alg>:' + EncodeHex(H(C))
+```
+
+The _digest_ is confirmed as the content identifier by independently calculating the _digest_.
+
+#### SHA-256
+
+[SHA-256](https://tools.ietf.org/html/rfc4634#page-7) is a collision-resistant hash function, chosen for ubiquity, reasonable size and secure characteristics.
+Implementations MUST implement SHA256 digest verification for use in descriptors.
```
wking: “SHA256” → “SHA-256” for consistency with your surrounding text.  And you should mention somewhere that the identifier for this hash is `sha256`.

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Examples
+
+The following example describes a _Manifest_ with a content identifier of "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270", of size 7682 bytes:
```
wking: Link `_Manifest_` to `manifest.md#image-manifest`?

descriptor.md
wking: On Wed, Jun 01, 2016 at 04:30:58PM -0700, Stephen Day wrote:
> > if you know a separate local layer has already verified the hash,
> > there's no sense in having your layer do it again.
> 
> If you already have it locally, the requirement does not apply.

That sounds like SHOULD to me (i.e. “good luck defining ‘local’” ;).
Verifying the hash is a trust-based decision (do you already trust the
source?  Do you care if the blob is valid?).  I don't see a need to
REQUIRE every layer in every stack to make a new trust check (but I'm
fine making a strong suggestion that they SHOULD).

descriptor.md
wking: On Wed, Jun 01, 2016 at 04:37:46PM -0700, Stephen Day wrote:
> > +Implementations MAY employ some canonicalization to ensure stable
> > content identifiers.
>
> This is a fairly important consideration when building
> content-addressable systems. This is saying the practice is
> acceptable but discouraged.

Wait why discouraged?  I think it's a *good* idea.  I'm just saying
that blob authors are unlikely to be perusing the descriptor docs for
those kind of hints.

descriptor.md
philips: +1 on turning it around as this format will stand on its own in #94

philips: Well, +1 on this change but I see below it says they can be independent

descriptor.md
philips: +1 on SHOULD. We can add language that it SHOULD be verified if it is coming over an untrusted transport.

descriptor.md
philips: +1 on drop

descriptor.md
philips: I think the earlier language is sufficient.

descriptor.md
philips: +1

descriptor.md
philips: @stevvooe can you provide a concrete example of canonicalization?

stevvooe: > Wait why discouraged? 

Processing unverified content is problematic. For example, one should not process canonicalize a tar file before calculating the content digest. If this is done, it should only be done on the generation side.

I'll add some language here to clear that up.

>  I'm just saying
that blob authors are unlikely to be perusing the descriptor docs for
those kind of hints.

And that will likely result in hash stability problems.

philips:
LGTM after minor nit cleanups.

descriptor.md
stevvooe: I think we define this to be when crossing a trust boundary, but what is the point of making verification of digests optional?

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+Some examples of _digests_ include the following:
+
+digest                                                                            | description                                   |
+----------------------------------------------------------------------------------|------------------------------------------------
+sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b           | Common sha256 based digest                    |
+
+Before consuming content targeted by a descriptor, the byte content MUST be verified against the _digest_.
+Heavy processing of before calculating a hash SHOULD be avoided.
+Implementations MAY employ some canonicalization to ensure stable content identifiers.
+
```
stevvooe: Include mention of size verification to avoid [Length Extension Attack](https://en.wikipedia.org/wiki/Length_extension_attack).

descriptor.md
```diff
@@ -0,0 +1,113 @@
...
+  This REQUIRED parameter is the digest of the content, meeting the requirements outlined in [Digests and Verification](#digests-and—verification).
+  Retrieved content SHOULD be verified against this digest before processing.
+
+- **`size`** *int*
+  This REQUIRED property specifies the size in bytes of the object.
+  This property exists so that a client will have an expected size for the content before validating.
+  If the length of the retrieved content does not match the specified length, the content should not be trusted.
+
+### Reserved
```
stevvooe: @philips @vbatts @wking Do we need to also call out fields that are used by other specs that are included on the descriptor? I think we have `platform` elsewhere.

It might be sufficient to say we allow extensions of descriptor, but reserve a few (ie, `data`).

descriptor.md
wking: On Wed, Jun 01, 2016 at 05:05:58PM -0700, Stephen Day wrote:
> > +  Retrieved content SHOULD be verified against this digest before processing.
> 
> I think we define this to be when crossing a trust boundary, but
> what is the point of making verification of digests optional?

How do you certify implementations if the definition involves
“crossing a trust boundary”?  What's your fuctional definition for
that?

descriptor.md
wking: On Wed, Jun 01, 2016 at 05:04:43PM -0700, Stephen Day wrote:
> > I'm just saying that blob authors are unlikely to be perusing the
> > descriptor docs for those kind of hints.
> 
> And that will likely result in hash stability problems.

We should definitely be talking about hash stability in image-spec,
just not *in this file*.

stevvooe: > @stevvooe can you provide a concrete example of canonicalization?

In the spec or in response to your comment?

descriptor.md
philips: @stevvooe I don't follow. Are you saying some other type is polymorphic? Which one?

stevvooe: In general, we want to allow extension of descriptors but reserve a set of fields that may not be supplanted by other specifications.

An example is the `urls` field, which was recently added to descriptor in Docker. We want to reserve definition of the `urls` field for this specification and not other specifications that may inherit these base properties.

Another example is `data`, which would may include the content directly.

wking: On Thu, Jun 02, 2016 at 01:11:54PM -0700, Stephen Day wrote:
> > +### Reserved
>
> In general, we want to allow extension of descriptors but reserve a
> set of fields that may not be supplanted by other specifications.

*cough* LSON-LD *cough* [1] ;).

[1]: https://github.com/opencontainers/runtime-spec/pull/371#issuecomment-209684002

stevvooe: I don't see how JSON-LD is related at all. Seems to be overcomplicated and unfocused in the problems it solves. What benefit do you see in bringing the complexity of XML to JSON?

Do you have examples?

wking: On Thu, Jun 02, 2016 at 01:47:53PM -0700, Stephen Day wrote:
> > +### Reserved
> 
> I don't see how JSON-LD is related at all.

It allows you to unambiguously declare semantics at a per-property
level (e.g. “when I say ‘urls’, I mean
http://opencontainers.org/image/descriptor/urls” or “when I say
‘urls’, I mean http://example.com/overload/urls”)

> Do you have examples?

Plain vanilla:

  {
    "@context": "http://opencontainers.org/image/descriptor/1.0"
    "size": 7682,
  }

Vendor ‘urls’:

  {
    "@context": "http://opencontainers.org/image/descriptor/1.0",
    "size": 7682,
    "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270",
    "urls": {
      "@type": "http://example.com/overload/urls",
      "@value": ["a.example.com"]
    }
  }

v1.1 OCI descriptor with URLs:

  {
    "@context": "http://opencontainers.org/image/descriptor/1.1"
    "size": 7682,
    "urls": [
      "https://a.example.com/cas",
      "https://b.example.com/blobs"
    ]
  }

stevvooe: Seems overly complex, when we can just call it out in the specification.

wking: On Thu, Jun 02, 2016 at 02:28:46PM -0700, Stephen Day wrote:
> Seems overly complex, when we can just call it out in the
> specification.

But that approach requires you (and anyone adding their own fields,
e.g. vendor A and vendor B) to declare, up front, any namespace they
want to claim.  While with JSON-LD, you only need to declare the bits
you actually need, and spec-consumers can cleanly mix and match as
they see fit.

And it looks like my vendor example should have been [1]:

  {
    "@context": [
      "http://opencontainers.org/image/descriptor/1.0",
      {
        "urls": "http://example.com/overload/urls"
      }
    ],
    "size": 7682,
    "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270",
    "urls": ["a.example.com"]
  }

[1]: https://www.w3.org/TR/json-ld/#advanced-context-usage

stevvooe: Sure, but what problem does this complexity actually solve?

wking: On Thu, Jun 02, 2016 at 02:58:34PM -0700, Stephen Day wrote:
> > +### Reserved
> 
> Sure, but what problem does this complexity actually solve?

You get to drop this section and sleep soundly without wondering if
you should have claimed ‘urls’ (or ‘data’ [1], or …).  Vendors can add
their own optimizations without worrying about ambiguous
interpretations.  I can put an entry in my manifest saying “if you
don't want to bother with the usual layout, you can find this
directory at /ipfs/QmTkzDw…”, and clients who understand that type can
take advantage of it, while clients who don't can fall back to the
image-spec tarball layers.  Clients who are compatible with v1.0 of
the spec can *still* process v1.1 configs if that config only uses
parameters that were spec'ed in v1.0 (because the v1.1 context will
link the same definitions for those parameters).  If you want me to
keep adding to this list, let me know ;).

[1]: https://github.com/opencontainers/image-spec/pull/111/files/03d6c78ea37021b9030cf56fa97e0a0f214f18d8#r65468737

stevvooe: I'm not sure I could sleep soundly unleashing JSON-LD on implementors.

Either way, this clause wouldn't protect against vendor-added fields in a descriptor. That is simply not supported. If one drops a random field into a descriptor, it may be overwritten by future versions of the specification. We already have annotations to cover this use case.

This statement would only reserve it from use in other OCI specifications that use descriptors.

wking: On Thu, Jun 02, 2016 at 03:33:16PM -0700, Stephen Day wrote:
> Either way, this clause wouldn't protect against vendor-added fields
> in a descriptor. That is simply not supported. If one drops a random
> field into a descriptor, it may be overwritten by future versions of
> the specification.

That *is* supported by JSON-LD.  Here's how you use both at once:

  {
    "@context": [
      "http://opencontainers.org/image/descriptor/1.0",
      {
        "vendorUrls": "http://example.com/overload/urls"
      }
    ],
    "size": 7682,
    "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270",
    "urls": [
      "https://a.example.com/cas",
      "https://b.example.com/blobs"
    ],
    "vendorUrls": ["a.example.com"]
  }

With JSON-LD, the property name (e.g. ‘urls’, ‘vendorUrls’, …) isn't
the semantic identifier; it's just a link to the @context entry where
you assign semantic IRIs
(e.g. ‘http://opencontainers.org/image/descriptor/urls’,
http://example.com/overload/urls’, …)

descriptor.md
stevvooe: This is to clarify the algorithm for using `H` to get a _digest_ from `C`.

stevvooe:
Most of the feedback has been serviced. Schema definitions and validation of examples have been added.

Note that the schema layout probably isn't ideal, but we may want to revisit in another PR.

descriptor.md
```diff
@@ -0,0 +1,123 @@
+<!--[metadata]>
++++
+draft = true
++++
+<![end-metadata]-->
```
philips: what is this stuff?

philips:
LGTM

nit: There is some metadata at the header of this file. I don't know what it is for but otherwise looks great.

philips:
@vbatts do you know how we force a rebuild?

Also, @opencontainers/image-spec-maintainers can we get an LGTM? One more LGTM required to merge this.

descriptor.md
```diff
@@ -0,0 +1,123 @@
...
+  Retrieved content SHOULD be verified against this digest when consumed via untrusted sources.
+
+- **`size`** *int*
+  This REQUIRED property specifies the size in bytes of the blob.
+  This property exists so that a client will have an expected size for the content before validating.
+  If the length of the retrieved content does not match the specified length, the content SHOULD NOT be trusted.
+
+### Reserved
+
+The following are field keys that MUST NOT be used in descriptors specified in other OCI specifications:
```
wking: You're restricting your reservation to “other OCI specifications”?  I expect you can trust OCI maintainers to PR this spec if they want a new descriptor field.  The folks you want to warn off are *external* implementations and specs.

wking: And because it doesn't hurt to *also* warn off other OCI specs, I'd use something generic like “MUST NOT be defined outside this specificiation”.

descriptor.md
```diff
@@ -0,0 +1,123 @@
...
+- **`urls`** *array*
+
+  This key is RESERVED for future versions of the specification.
+
+- **`data`** *string*
+
+  This key is RESERVED for futures versions of the specification.
+
+All other fields may be included in other OCI specifications.
+Extended _Descriptor_ field additions proposed in other OCI specifications SHOULD first be considered for addition into this specification.
```
wking: I'd [drop OCI][1] from both of these lines.

[1]: https://github.com/opencontainers/image-spec/pull/111/files#r65659259

descriptor.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+Some examples of _digests_ include the following:
+
+digest                                                                            | description                                   |
+----------------------------------------------------------------------------------|------------------------------------------------
+sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b           | Common sha256 based digest                    |
+
+Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the _digest_.
+The size of the content SHOULD be verified, as well, to protect against [Length Extension Attacks](https://en.wikipedia.org/wiki/Length_extension_attack).
```
wking: Can we drop the length-extension reference and say something like:

> The size of the content SHOULD be verified because it's cheaper than hashing.  You SHOULD verify the _digest_ if and only if the content length matches the expected value.

Length extension attacks allow Malory to generate a new hash without having access to the initial content, but I don't see how that is a problem here.  The hash of the extended content will still be different from the original hash.

The attack would be a problem if we were specifying a signature algorithm, but I'm really hoping this spec allows existing, off-the-shelf crypto for signing (e.g. OpenPGP, X.509, …).

Makefile
```diff
@@ -68,7 +68,7 @@ oci-image-tool:
 
 schema-fs:
 	@echo "generating schema fs"
-	@cd schema && echo -e "$$(cat ../.header)\n\n$$(go generate)" > fs.go
+	@cd schema && echo "$$(cat ../.header)\n\n$$(go generate)" > fs.go
```
wking: If you remove `-e`, you'll get literal `\n`:

    $ echo 'a\n\nb'
    a\n\nb
    $ echo -e 'a\n\nb'
    a

    b

If want POSIX compliance, you probably want to use [printf][1] here.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/printf.html

descriptor.md
vbatts: i'm not sure

schema/image-manifest-schema.json
```diff
@@ -14,12 +14,12 @@
...
     },
     "config": {
-      "$ref": "defs-image.json#/definitions/descriptor"
+      "$ref": "content-descriptor.json"
     },
     "layers": {
       "type": "array",
       "items": {
-        "$ref": "defs-image.json#/definitions/descriptor"
+        "$ref": "content-descriptor.json"
```
vbatts: I reckon this is fine, but seems just as well to make it an object within a defs*.json file, rather than adding a new file just for a single definition.

vbatts: ah, but it can have it's own schema file to be validated against ... hrm

descriptor.md
```diff
@@ -0,0 +1,123 @@
...
+draft = true
++++
+<![end-metadata]-->
+
+# OpenContainers Content Descriptors
+
+OCI have several components that come to together to describe an image.
+References between components form a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
+The references in the _Merkle DAG_ are expressed through _Content Descriptors_.
+A _Content Descriptor_ or _Descriptor_, describes the disposition of targeted content.
```
vbatts: disposition? seems like an ambiguous word here.

descriptor.md
```diff
@@ -0,0 +1,123 @@
...
+  This REQUIRED property is the _digest_ of the targeted content, meeting the requirements outlined in [Digests and Verification](#digests-and—verification).
+  Retrieved content SHOULD be verified against this digest when consumed via untrusted sources.
+
+- **`size`** *int*
+  This REQUIRED property specifies the size in bytes of the blob.
+  This property exists so that a client will have an expected size for the content before validating.
+  If the length of the retrieved content does not match the specified length, the content SHOULD NOT be trusted.
+
+### Reserved
+
```
vbatts: So why are these reserved? I'm surprising unclear on this. Is it because docker is using them?

schema/image-manifest-schema.json
stevvooe: I suspect we'll have to shuffle the organization a bit, but, yes, this allows individual validation in the current test setup.

Makefile
stevvooe: `-e` isn't supported on BSD userspace tools. Removing it didn't change the output.

descriptor.md
stevvooe: "disposition" precisely means "the way in which something is placed or arranged, especially in relation to other things".

However, "describes the disposition" stutters, as well. I'll tweak this. Any suggestions?

descriptor.md
stevvooe: Docker is introducing a `urls` field in https://github.com/docker/docker/pull/22866 and it is already part of the latest manifest specification. We have not yet pushed it in OCI since we are unsure of the correct considerations in practice, but it should be introduced.

`data` is reserved for possibly directly embedding the binary data for a resource, avoiding a fetch. We haven't found a convincing use case, but we should hold it here for consideration.

Makefile
wking: On Fri, Jun 03, 2016 at 11:36:46AM -0700, Stephen Day wrote:
> > -	@cd schema && echo -e "$$(cat ../.header)\n\n$$(go generate)" > fs.go
> > +	@cd schema && echo "$$(cat ../.header)\n\n$$(go generate)" > fs.go
> 
> `-e` isn't supported on BSD userspace tools. Removing it didn't change the output.

Maybe not for your echo implementation.  But the \n bits are not in
POSIX, so whether or not they're supported is probably undefined.  On
my system:

  $ command -V echo
  echo is a shell builtin
  $ echo $SHELL
  /bin/bash
  $ echo $BASH_VERSION
  4.3.42(1)-release
  $ echo 'a\n\nb'
  a\n\nb

descriptor.md
wking: On Fri, Jun 03, 2016 at 11:38:09AM -0700, Stephen Day wrote:
> > +The references in the _Merkle DAG_ are expressed through _Content Descriptors_.
> > +A _Content Descriptor_ or _Descriptor_, describes the disposition of targeted content.
> 
> "disposition" precisely means "the way in which something is placed
> or arranged, especially in relation to other things".
> 
> However, "describes the disposition" stutters, as well. I'll tweak
> this. Any suggestions?

I'd just drop the line.  The previous line and rest of the docs make
the idea clear enough.

descriptor.md
stevvooe: It was in the top of the other spec. Figured we were using it.

Was this copied over from docker? This looks like part of Docker's docs build system.

Makefile
stevvooe: You're welcome to submit a PR.

descriptor.md
wking: On Fri, Jun 03, 2016 at 12:14:44PM -0700, Stephen Day wrote:
> Checking size protects against replacing content `C` with malicious
> content `M`, such that, `H(C) == H(M)`? Ensuring that len(C) is
> constant makes it much harder to generate an `M` that matches
> H(`C`). It effectively reduces the space available for hash
> collisions defined by `H(C) == H(M)` to `H(C) == H(M) | len(C) ==
> len(M)`.
> 
> What is that attack called?

A collision attack [1].  And if you're at all worried about
hash-collisions, you should be choosing a stronger hash, or using
multiple hashes with different algorithms.

[1]: https://en.wikipedia.org/wiki/Collision_attack

Makefile
wking: On Fri, Jun 03, 2016 at 12:06:45PM -0700, Stephen Day wrote:
> > @@ -68,7 +68,7 @@ oci-image-tool:
> >  
> >  schema-fs:
> >  	@echo "generating schema fs"
> > -	@cd schema && echo -e "$$(cat ../.header)\n\n$$(go generate)" > fs.go
> > +	@cd schema && echo "$$(cat ../.header)\n\n$$(go generate)" > fs.go
> 
> You're welcome to submit a PR.

Sure.  Until I do (tonight?), can we drop the drive-by change here?

descriptor.md
vbatts: perhaps. It's seems fine to drop.

Makefile
stevvooe: This is not a drive by change. This make target is broken with `-e` and removing it did not change the output.

descriptor.md
vbatts: right on. Maybe there could be a notion of scoping? `docker.urls` or `x.urls`?

Makefile
wking: On Fri, Jun 03, 2016 at 12:57:12PM -0700, Stephen Day wrote:
> >  schema-fs:
> >  	@echo "generating schema fs"
> > -	@cd schema && echo -e "$$(cat ../.header)\n\n$$(go generate)" > fs.go
> > +	@cd schema && echo "$$(cat ../.header)\n\n$$(go generate)" > fs.go
>
> This is not a drive by change. This make target is broken with `-e`
> and removing it did not change the output.

This target is broken with -e *for you*, and removing it breaks the
output *for me*.  The current form also presumably works for
@jonboulle (who submitted it in #78).  So rather than changing the
folks who it's broken for, I'd like to either leave it alone, or fix
it so it works for everyone.

descriptor.md
stevvooe: > Maybe there could be a notion of scoping? docker.urls or x.urls?

Namespacing is a solid option, as is allowing annotations, but we really need to balance this against the simplicity and compatibility of descriptors.

The manifest list descriptor uses `platform` as an extension, over the more flexible labels approach. I suspect that these kind of "curated extensions" are best for cross-vendor compatibility.

Either way, let's see how `urls` support pans out in Docker and work from there. 👀

philips: I am fine leaving these as reserved. Honestly, I would rather just define them here first but reserving more stuff rather than less in the interest of caution is fine.

Overtime, my hope is that whatever discussion lead to urls being added to the Docker media type happens in OCI post-v1.0 as we all focus on a single shared spec.

philips:
@stevvooe I think this just needs a couple of trivial nits cleaned up and a rebase for the Makefile printf stuff merged in #116.

After that we will need an LGTM from someone else.

philips:
@vbatts maybe you can give an LGTM on this one?

wking:
On Tue, Jun 07, 2016 at 02:47:17PM -0700, Brandon Philips wrote:
> @vbatts maybe you can give an LGTM on this one?

Probably want to rebase it around #116 before collecting LGTMs.

stevvooe:
Most of the larger nits are addressed. Other minor things can be cleaned up in another pass.

descriptor.md
```diff
@@ -77,7 +71,7 @@ digest
 sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b           | Common sha256 based digest                    |
 
 Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the _digest_.
-The size of the content SHOULD be verified, as well, to protect against [Length Extension Attacks](https://en.wikipedia.org/wiki/Length_extension_attack).
+The size of the content SHOULD be verified to reduce hash collision space.
```
wking: This (and the `<!--[metadata]…-->` removal) should have probably been squashed into the previous commit.  Or maybe make the whole branch a single commit?

descriptor.md
```diff
@@ -0,0 +1,117 @@
+# OpenContainers Content Descriptors
+
+OCI have several components that come to together to describe an image.
+References between components form a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
+The references in the _Merkle DAG_ are expressed through _Content Descriptors_.
+A _Content Descriptor_ or _Descriptor_, describes the disposition of targeted content.
+A _Descriptor_ includes the type of content, an independently-verifiable content identifier, known as a "digest" and the byte-size of the raw content.
```
wking: The “, known as a "digest"” bit makes the three-part content listing hard for me to read.  Can we drop this sentence, since we have a detailed “Properties” section below?

descriptor.md
wking: On Wed, Jun 01, 2016 at 04:59:36PM -0700, Brandon Philips wrote:
> > +Descriptors SHOULD be embedded in other formats to securely reference external content.
> 
> Well, +1 on this change but I see below it says they can be independent

We got the turned-around “Other formats SHOULD use descriptors to
securely reference external content.” line, but this is still here.
Can we drop this line?

wking:
On Tue, Jun 07, 2016 at 03:23:10PM -0700, Stephen Day wrote:
> Most of the larger nits are addressed. Other minor things can be
> cleaned up in another pass.

Yeah.  I don't see anything major left, and I'm happy to file PRs for
my remaining suggestions.

philips:
LGTM

Do I have to do this again? I don't understand pullapprove yet.

philips:
OK, looks like that worked.

Need an additional LGTM from an @opencontainers/image-spec-maintainers

wking:
On Tue, Jun 07, 2016 at 04:30:06PM -0700, Brandon Philips wrote:
> LGTM
> 
> Do I have to do this again?

After each push [1].

[1]: https://github.com/opencontainers/image-spec/blob/a4abdf47da40e49d6fec803278f35c65b8c467eb/.pullapprove.yml#L4

philips:
@wking ick. I will file an issue with @caniszczyk to see if we can only have it decrement by 1 or something. This makes reviewing this stuff super annoying.

wking:
On Tue, Jun 07, 2016 at 04:35:42PM -0700, Brandon Philips wrote:
> This makes reviewing this stuff super annoying.

Just don't bother stamping approval on things until they settle down
;).

philips:
@wking I just don't like the system implied mistrust. Like I trust Stephen to just fix nits after I give an LGTM and not require a full re-review of the PR. There is some middle ground here that more accurately reflects the trust model of a collaborative project.

wking:
On Tue, Jun 07, 2016 at 08:29:13PM -0700, Brandon Philips wrote:
> @wking I just don't like the system implied mistrust.

It's you're project, so you can obviously pick the rules that make
sense to you.  But “$OTHER_MAINTAINER is probably still fine with this
PR after the changes” seems like something you'd only have to assume
if $OTHER_MAINTAINER wasn't paying enough attention to explicitly say
“yeah, that still LGTM”.

philips:
I think this is ready to merge. Can other @opencontainers/image-spec-maintainers take a look and given an LGTM?

stevvooe:
LGTM

Apparently, we can LGTM ourselves. Feel free to wait for another maintainer to LGTM before merging.

vbatts:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/image-spec/pull-request/111/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/image-spec)

Issue #87, opened 2016-05-25
Convert a serialization config JSON to a OCI runtime configuration.


s-urbaniak:
As per https://github.com/opencontainers/image-spec/pull/82#issuecomment-221450375: Currently, the `oci-image-tool` supports a simple unpacking mechanism to inspect deflated images and validation.

More concretely the config/image JSON https://github.com/opencontainers/image-spec/blob/master/serialization.md#image-json-description has to be converted to a https://github.com/opencontainers/runtime-spec/blob/master/config.md#container-configuration-file.

wking:
On Tue, May 24, 2016 at 06:44:43PM -0700, Sergiusz Urbaniak wrote:
> More concretely the config/image JSON
> https://github.com/opencontainers/image-spec/blob/master/serialization.md#image-json-description
> has to be converted to a
> https://github.com/opencontainers/runtime-spec/blob/master/config.md#container-configuration-file.

I don't think this is possible.  For example, the image-spec schema
currently has ‘ExposedPorts’, but networking is out-of-scope for the
runtime config [1].  You'd have to select a particular network manager
(that isn't covered by an OCI spec) and setup hooks to drive that
network manager.

Stepping back, I see no reason for the image-spec repo to care about
what is in the *runtime* config (see also @stevvooe's “decouple”
wording in [2]).  I think imaging tools need to know “this is a
directory that I'm (de)serializing”.  Then the image spec can add on
distribution-oriented layers like “distribute that directory using
layered tarballs” or “sign name ↔ directory associations”.  But baking
knowledge of the filesystem directory (e.g. there's a ‘config.json’
specified by runtime-spec, there may be a ‘rootfs’ directory
referenced from ‘root.path’ in that config.json, …) into the image
spec seems like it's breaking a perfectly good abstraction boundary
for no purpose.

If the “we don't care what's inside the directory” approach from my
last paragraph gains traction, there will still be a need for some
conversion logic for backwards compatibility with existing images.
But I think we want the image spec to get out of that business (and
let the runtimes handle it).

[1]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/3XqoyLqtA6I
     Subject: Do we have any plan for network and storage specs?
     Date: Fri, 14 Aug 2015 02:36:15 -0700 (PDT)
     Message-Id: <c056925e-6ecf-4457-8c16-7ab34d279604@opencontainers.org>
[...]

PR #114, opened 2016-06-02
oci-image-tool: implement create-runtime-bundle

s-urbaniak:
This is a first shot towards a `oci-image-tool create-runtime-bundle` subcommand.

I found that converting towards the runtime bundle we have (obviously) quite some impedance mismatch here and there so I guess I surely missed something.

The most important questions for me are:

- What `ociVersion` to I have to set?
- How to map image `Volumes` to bundle `mounts`?
- Is there anything else obvious I am missing?
- We have redundant declarations of layers, one in the manifest, one in the image JSON, it is not clear to me which layer declaration is the one I should pick for unpacking to the bundle. This also applies to unpack. I opened https://github.com/opencontainers/image-spec/issues/115 for this.

Generally I have the impression that I can only fill up only a small subset of the bundle fields, hence I ask for advice how to proceed here.

### DIFF
--- a/cmd/oci-image-tool/create_runtime_bundle.go
+++ b/cmd/oci-image-tool/create_runtime_bundle.go
@@ -24,29 +24,30 @@ import (
 	"github.com/spf13/cobra"
 )
 
-// supported unpack types
-var unpackTypes = []string{
+// supported bundle types
+var bundleTypes = []string{
 	typeImageLayout,
 	typeImage,
 }
 
-type unpackCmd struct {
+type bundleCmd struct {
 	stdout *log.Logger
 	stderr *log.Logger
-	typ    string // the type to unpack, can be empty string
+	typ    string // the type to bundle, can be empty string
 	ref    string
+	root   string
 }
 
-func newUnpackCmd(stdout, stderr *log.Logger) *cobra.Command {
-	v := &unpackCmd{
+func newBundleCmd(stdout, stderr *log.Logger) *cobra.Command {
+	v := &bundleCmd{
 		stdout: stdout,
 		stderr: stderr,
 	}
 
 	cmd := &cobra.Command{
-		Use:   "unpack [src] [dest]",
-		Short: "Unpack an image or image source layout",
-		Long:  `Unpack the OCI image .tar file or OCI image layout directory present at [src] to the destination directory [dest].`,
+		Use:   "create-runtime-bundle [src] [dest]",
+		Short: "Create an OCI image runtime bundle",
+		Long:  `Creates an OCI image runtime bundle at the destination directory [dest] from an OCI image present at [src].`,
 		Run:   v.Run,
 	}
 
@@ -54,19 +55,25 @@ func newUnpackCmd(stdout, stderr *log.Logger) *cobra.Command {
 		&v.typ, "type", "",
 		fmt.Sprintf(
 			`Type of the file to unpack. If unset, oci-image-tool will try to auto-detect the type. One of "%s"`,
-			strings.Join(unpackTypes, ","),
+			strings.Join(bundleTypes, ","),
 		),
 	)
 
 	cmd.Flags().StringVar(
 		&v.ref, "ref", "v1.0",
-		`The ref pointing to the manifest to be unpacked. This must be present in the "refs" subdirectory of the image.`,
+		`The ref pointing to the manifest of the OCI image. This must be present in the "refs" subdirectory of the image.`,
+	)
+
+	cmd.Flags().StringVar(
+		&v.root, "rootfs", "rootfs",
+		`A directory representing the root filesystem of the container in the OCI runtime bundle.
+It is strongly recommended to keep the default value.`,
 	)
 
 	return cmd
 }
 
-func (v *unpackCmd) Run(cmd *cobra.Command, args []string) {
+func (v *bundleCmd) Run(cmd *cobra.Command, args []string) {
 	if len(args) != 2 {
 		v.stderr.Print("both src and dest must be provided")
 		if err := cmd.Usage(); err != nil {
@@ -87,10 +94,10 @@ func (v *unpackCmd) Run(cmd *cobra.Command, args []string) {
 	var err error
 	switch v.typ {
 	case typeImageLayout:
-		err = image.UnpackLayout(args[0], args[1], v.ref)
+		err = image.CreateRuntimeBundleLayout(args[0], args[1], v.ref, v.root)
 
 	case typeImage:
-		err = image.Unpack(args[0], args[1], v.ref)
+		err = image.CreateRuntimeBundle(args[0], args[1], v.ref, v.root)
 	}
 
 	if err != nil {
--- a/image/config.go
+++ b/image/config.go
@@ -1,0 +1,156 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package image
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/opencontainers/image-spec/schema"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
+type cfg struct {
+	User         string
+	Memory       int64
+	MemorySwap   int64
+	CPUShares    int64 `json:"CpuShares"`
+	ExposedPorts map[string]struct{}
+	Env          []string
+	Entrypoint   []string
+	Cmd          []string
+	Volumes      map[string]struct{}
+	WorkingDir   string
+}
+
+type config struct {
+	Architecture string `json:"architecture"`
+	OS           string `json:"os"`
+	Config       cfg    `json:"config"`
+}
+
+func findConfig(w walker, d *descriptor) (*config, error) {
+	var c config
+	cpath := filepath.Join("blobs", d.Digest)
+
+	f := func(path string, info os.FileInfo, r io.Reader) error {
+		if info.IsDir() {
+			return nil
+		}
+
+		if filepath.Clean(path) != cpath {
+			return nil
+		}
+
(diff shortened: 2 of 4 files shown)


s-urbaniak:

wking:
On Thu, Jun 02, 2016 at 07:07:44AM -0700, Sergiusz Urbaniak wrote:
> - What `ociVersion` to I have to set?

For now, you probably want to set the tip ociVersion (0.5.0 at the
moment, but 1.0 rcs are coming out soon).  Once we cut 1.0 and start
maintaining backwards compatibility, you should pick the lowest
runtime version that has everything you need (e.g. stay with 1.0
unless 1.1 lands a new feature you need).

image/config.go
```diff
@@ -0,0 +1,138 @@
...
+	default:
+		return nil, err
+	}
+
+	return &c, nil
+}
+
+func (c *config) runtimeSpec(rootfs string) specs.Spec {
+	var s specs.Spec
+	s.Version = "v0.5.0"
```
wking: You don't want the `v` prefix (see [this example][1]).

[1]: https://github.com/opencontainers/runtime-spec/blob/v0.5.0/config.md#specification-version

image/config.go
```diff
@@ -0,0 +1,138 @@
...
+	s.Process.Cwd = c.Config.WorkingDir
+	s.Process.Env = append([]string(nil), c.Config.Env...)
+	s.Process.Args = append([]string(nil), c.Config.Entrypoint...)
+	s.Process.Args = append(s.Process.Args, c.Config.Cmd...)
+
+	if uid, err := strconv.Atoi(c.Config.User); err == nil {
+		s.Process.User.UID = uint32(uid)
+	}
+
+	if ug := strings.Split(c.Config.User, ":"); len(ug) == 2 {
```
wking: The UID-only case should probably be an `else` of this block:

```go
if ug := strings.Split(c.Config.User, ":"); len(ug) == 2 {
  …
} else if uid, err := strconv.Atoi(c.Config.User); err == nil {
  …
}
```

(guessing at Go's else-if phrasing ;).

image/config.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/pkg/errors"
+)
+
+type cfg struct {
+	User         string
+	Memory       int64
+	MemorySwap   int64
+	CPUShares    int64 `json:"CpuShares"`
+	ExposedPorts map[string]struct{}
```
wking: You're currently not translating this to anything (because the runtime spec [doesn't have knobs for Linux networking][1]).  @philips, can you explain how you expect [higher-levels notice and apply this configuration][2]?

[1]: https://github.com/opencontainers/image-spec/issues/87#issuecomment-221463049
[2]: https://github.com/opencontainers/image-spec/issues/87#issuecomment-221623046

wking:
On Thu, Jun 02, 2016 at 07:07:44AM -0700, Sergiusz Urbaniak wrote:
> - How to map image `Volumes` to bundle `mounts`?

I'd fill in bind-mount entries and just leave ‘source’ entry null.
For example:

  "mounts": [
    {
      "destination": "${path_from_image_config}",
      "type": "bind",
      "source": null,
      "options": ["rbind"]
    }
  ]

Then the runtime can fill in (or remove) the null-source entries
depending on ‘docker run -v a:b …’ etc.

image/config.go
s-urbaniak: ha, good catch, thanks!

image/config.go
s-urbaniak: ack :-)

image/config.go
```diff
@@ -0,0 +1,138 @@
...
+	s.Process.Args = append(s.Process.Args, c.Config.Cmd...)
+
+	if uid, err := strconv.Atoi(c.Config.User); err == nil {
+		s.Process.User.UID = uint32(uid)
+	}
+
+	if ug := strings.Split(c.Config.User, ":"); len(ug) == 2 {
+		if uid, err := strconv.Atoi(ug[0]); err == nil {
+			s.Process.User.UID = uint32(uid)
+		}
```
wking: These should probably be throwing “not supported yet” errors if it can't make the conversion until you add code to spin up a container with all the mounts and use getent or similar (see opencontainers/runtime-spec#38).  Docker may be doing something like this already, so maybe you can grab their code.  And this is hideous, ugly, and dangerous, which is why the runtime spec [punted it to config authors][1].

[1]: https://github.com/opencontainers/runtime-spec/pull/191#issuecomment-171473348

image/config.go
vbatts: But i'm sure that other extentsions will come along and use such. Also, having a golang structure for the data will allow folks to import and use it for their own use cases.

wking: On Thu, Jun 02, 2016 at 11:49:50AM -0700, Vincent Batts wrote:
> > +	ExposedPorts map[string]struct{}
> 
> But i'm sure that other extentsions will come along and use such.

How do they find it?  If this command unpacks $IMAGE into $BUNDLE, but
does not leave a pointer in $BUNDLE to say “this was unpacked from
$IMAGE”, I don't see how tooling invoked on the $BUNDLE would be able
to find this image-specific data.  So to support that additional
tooling, you either need:

a. A generic “this was unpacked from $IMAGE” reference in the bundle
   directory somewhere, or
b. To copy all of the “unused” information like ExposedPorts over to
   some well-known location inside the bundle.

image/config.go
s-urbaniak: fixed locally

image/config.go
s-urbaniak: fixed locally

image/config.go
s-urbaniak: ack, I'll return an error in this case, thanks for the suggestion :+1:

s-urbaniak:
@wking addressed review comments, PTAL and thanks for the review!

image/config.go
```diff
@@ -0,0 +1,154 @@
...
+		s.Process.User.UID = uint32(uid)
+		s.Process.User.GID = uint32(gid)
+	} else {
+		return nil, errors.New("config.User: unsupported format")
+	}
+
+	s.Platform.OS = c.OS
+	s.Platform.Arch = c.Architecture
+
+	if c.OS == "linux" {
```
stevvooe: Should this return an error when run on an unsupported OS?

cmd/oci-image-tool/create_runtime_bundle.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/opencontainers/image-spec/image"
+	"github.com/spf13/cobra"
+)
+
+// supported unpack types
```
s-urbaniak: Note to myself: `supported bundle types`

cmd/oci-image-tool/create_runtime_bundle.go
```diff
@@ -0,0 +1,109 @@
...
+var bundleTypes = []string{
+	typeImageLayout,
+	typeImage,
+}
+
+type bundleCmd struct {
+	stdout *log.Logger
+	stderr *log.Logger
+	typ    string // the type to validate, can be empty string
```
s-urbaniak: Note to myself: "the type to bundle"

wking:
On Fri, Jun 03, 2016 at 05:42:07AM -0700, Sergiusz Urbaniak wrote:
> @wking addressed review comments, PTAL and thanks for the review!

I haven't gone over the whole thing again, but 5a38e9f → cfd87a7 looks
good to me.

image/config.go
wking: On Fri, Jun 03, 2016 at 12:03:08PM -0700, Stephen Day wrote:
> > +	if c.OS == "linux" {
> 
> Should this return an error when run on an unsupported OS?

I don't think so.  Perhaps the user is unpacking an image on one
machine that they intend to copy (via a shared filesystem?) and run on
another.  It's probably not going to be a core use-case, but if the
image tool is *capable* of unpacking the image (regardless of its
target OS), I think it should unpack it.

The runtime, on the other hand, must die for unsupported platforms.
An explicit statement to that effect landed two weeks ago via
opencontainers/runtime-spec#441 [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/441/files#diff-c9c91c29b41257aea3a3403cc606ad99R198

stevvooe: To be compatible, it needs to have a conversion for every single OS and needs to fail if it doesn't support that particular OS, as it may drop relevant data.

wking: On Fri, Jun 03, 2016 at 02:51:22PM -0700, Stephen Day wrote:
> > +	if c.OS == "linux" {
>
> To be compatible, it needs to have a conversion for every single OS…

I don't think you need that, if you have this:

> … and needs to fail if it doesn't support that particular OS, as it
> may drop relevant data.

So “die if you are asked to convert a format you can't handle
correctly”, not “die if you are asked to convert a format targetting
another platform”.

s-urbaniak: As far as I see it I am "dying if I am asked to convert a format I can't handle correctly", so does this looks good to you?

image/config.go
s-urbaniak: @vbatts the `cfg` structure currently is private; here in `image-spec` we don't have redundantly declared public Golang structs, only the JSON schema files. If we want to go the route as https://github.com/opencontainers/runtime-spec/tree/master/specs-go, I'd have to refactor; please advise.

image/config.go
wking: On Mon, Jun 06, 2016 at 04:12:27AM -0700, Sergiusz Urbaniak wrote:
> > +	if c.OS == "linux" {
> 
> As far as I see it I am "dying if I am asked to convert a format I
> can't handle correctly", so does this looks good to you?

Yeah.  This sub-thread was just clarifying @stevvooe's concern about a
blanket supported-OS check [1], and we both agree that you don't need
that blanket check [2].

[1]: https://github.com/opencontainers/image-spec/pull/114#discussion_r65759735
[2]: https://github.com/opencontainers/image-spec/pull/114#discussion_r65781903

stevvooe: > As far as I see it I am "dying if I am asked to convert a format I can't handle correctly", so does this looks good to you?

Where is this happening? It looks like it checks for linux and then moves on if it doesn't understand the OS.

s-urbaniak: @stevvooe ok, I think I misunderstood ;-)

I'm merely dying when validation failed (https://github.com/opencontainers/image-spec/pull/114/files#diff-da07f373046a0a24e67b18699dc15a70R142)

I'll change the code to return an err if I get an OS != "linux".

image/config.go
philips: @wking commented on #87 about networking.

philips:
LGTM overall

The ExposedPorts thing is an issue but I don't think we should bike shed over it and instead discuss on #87.

philips:
Can anyone else review and give an LGTM on this one? @opencontainers/image-spec-maintainers 

LGTM

vbatts:
LGTM

stevvooe:
Whoops, sorry.

LGTM, here.

PR #117, opened 2016-06-05
schema: add line/col info in case of JSON err

s-urbaniak:
Previously no line/col information was provided if JSON parsing failed.
This fixes it.

Unfortunately we still cannot determine line/col information in case of
schema validation because the original JSON decoderState is lost.

### DIFF
--- a/schema/error.go
+++ b/schema/error.go
@@ -1,0 +1,44 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema
+
+import (
+	"encoding/json"
+	"io"
+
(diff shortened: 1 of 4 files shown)


schema/error.go
```diff
@@ -0,0 +1,22 @@
...
+	"go4.org/errorutil"
+)
+
+// WrapSyntaxError checks whether the given error is a json.Syntax error
+// and returns a wrapped error containing line/col information using the given reader.
+func WrapSyntaxError(r io.Reader, err error) error {
+	if serr, ok := err.(*json.SyntaxError); ok {
+		line, col, _ := errorutil.HighlightBytePosition(r, serr.Offset)
+
+		return errors.Wrapf(err, "line %d col %d", line, col)
```
s-urbaniak: TODO(sur): instead of putting this information here, investigate whether we can unwrap wrapped `*json.SyntaxError` errs and produce a nice machine-readable output using the "standard" `file:line:col: msg` format.

philips:
LGTM

vbatts:
LGTM

schema/error.go
```diff
@@ -0,0 +1,44 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema
+
+import (
+	"encoding/json"
+	"io"
+
+	"go4.org/errorutil"
```
stevvooe: Do we need to pull in a whole dependency for this?

s-urbaniak: I can implement the necessary bits here.

stevvooe: Cheers. This dep just has that "leftpad feel".

I think we can wait till another PR.

stevvooe:
LGTM

PR #118, opened 2016-06-06
schema: add more tests

### LINKED ISSUE #83: schema: define "negative" schema examples
We have several examples of well-formed JSON objects that pass schema validation, but no examples that fail. We need several for the following reason:

1. Ensure that schema validation is actually working.
2. Clarify aspects of the specification.

Ideally, these could be part of the specification and marked as bad, in congruence with the approach of having the examples validated.

Please see https://github.com/opencontainers/image-spec/pull/81#discussion_r64127544.

cc @s-urbaniak

s-urbaniak:
Currently samples in the specification are being tested. This adds
additional tests.

Closes #83.

### DIFF
--- a/schema/config_test.go
+++ b/schema/config_test.go
@@ -1,0 +1,165 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema_test
+
+import (
+	"strings"
+	"testing"
+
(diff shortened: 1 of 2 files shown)


philips:
It would be nice to have a doc explaining why each of the fail cases should fail to help future people understand.

stevvooe:
> It would be nice to have a doc explaining why each of the fail cases should fail to help future people understand.

Agreed. Typically, I find these kinds of undocumented fixtures to be disorienting when they are introduced without context.

We can use the existing markdown parser. Add a `mustfail` flag to the markdown example extractor next to the mediatype.

vbatts:
Perhaps those comments and such would be easier if the job blobs were just
byte array variables in a *_test.go file?

stevvooe:
@vbatts `[]byte/string` is fine. We just need a little context to these (not these examples are wildly complex).

stevvooe:
Edit: added that this closes #83.

philips:
@s-urbaniak can you fixup these tests to include metadata on why they fail?

philips:
*human readable metadata.

s-urbaniak:
@philips @stevvooe @vbatts (travelling, having limited access to the internets) PTAL

philips:
LGTM, thanks!

philips:
I think this is ready to merge. Can other @opencontainers/image-spec-maintainers take a look and give an LGTM?

vbatts:
LGTM

PR #131, opened 2016-06-13
Makefile: Enable native pandoc builds

grahamwhaley:
Enable building the documents with a native pandoc.
Code pretty much taken directly from the similar series in the
runtime-specificaiton Makefile, such as in:
2e2073b (Makefile: Fix native pandoc invocation)

Signed-off-by: Graham Whaley <graham.whaley@linux.intel.com>

---
Note, I don't have the ability to check the docker run based build, just the pandoc one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ DOC_FILES := \
 FIGURE_FILES := \
 	img/media-types.png
 
-OUTPUT		?= output/
+OUTPUT_DIRNAME		?= output/
 DOC_FILENAME	?= oci-image-spec
 
 EPOCH_TEST_COMMIT ?= v0.2.0
@@ -24,7 +24,7 @@ default: help
 help:
 	@echo "Usage: make <target>"
 	@echo
-	@echo " * 'docs' - produce document in the $(OUTPUT) directory"
+	@echo " * 'docs' - produce document in the $(OUTPUT_DIRNAME) directory"
 	@echo " * 'fmt' - format the json with indentation"
 	@echo " * 'validate-examples' - validate the examples in the specification markdown files"
 	@echo " * 'oci-image-tool' - build the oci-image-tool binary"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,5 +1,22 @@
 
-DOCKER ?= $(shell which docker)
+EPOCH_TEST_COMMIT ?= v0.2.0
+OUTPUT_DIRNAME	?= output/
+DOC_FILENAME	?= oci-image-spec
+DOCKER		?= $(shell which docker)
```
vbatts: may as well make this `$(shell command -v docker 2>/dev/null)`

Makefile
```diff
@@ -12,50 +29,35 @@ DOC_FILES := \
...
-	vbatts/pandoc -f markdown_github -t latex -o /$(OUTPUT)/$(DOC_FILENAME).pdf $(patsubst %,/input/%,$(DOC_FILES)) && \
+docs: $(OUTPUT_DIRNAME)/$(DOC_FILENAME).pdf $(OUTPUT_DIRNAME)/$(DOC_FILENAME).html
+
+ifeq "$(strip $(PANDOC))" ''
+$(OUTPUT_DIRNAME)/$(DOC_FILENAME).pdf $(OUTPUT_DIRNAME)/$(DOC_FILENAME).html:
+	$(error cannot build $@ without either pandoc or docker)
+else
+$(OUTPUT_DIRNAME)/$(DOC_FILENAME).pdf: $(DOC_FILES) $(FIGURE_FILES)
+	@mkdir -p $(OUTPUT_DIRNAME)/ && \
+	$(PANDOC) -f markdown_github -t latex -o $(PANDOC_DST)$@ $(patsubst %,$(PANDOC_SRC)%,$(DOC_FILES))
```
vbatts: this is missing the `cp ...`, but even when i add it locally, it fails to find the image files. I'm not sure yet what's happening.

Makefile
grahamwhaley: Yep, for sure. Presume you won't need another pull for that?

philips:
@grahamwhaley @vbatts where are we at on this one?

grahamwhaley:
Hi @philips @vbatts  I'm aware this stalled. I'm probably in a position where I can try to diagnose the docker failure case now, but it has been getting pushed down my todo list :-(
@vbatts - did you get any further with the debug than above do you remember? Any more details here might help if/when we eventually get back to it.

grahamwhaley:
I think I've found the problem - I think my patch accidentally dropped the '--workdir' argument to the docker command line. I've tested locally, but it is late and this machine is not set up for repo pushes. I'll see if I can get to push a new patch early next week if you don't get to it before me @vbatts ;-)

grahamwhaley:
Hi. I pushed a new path, but now see it needs a rebase. I've done the local fixup (pretty trival), but want to stare at the patch before pushing a new version. Thus, you can wait for me to push a new version if you don't want to clean it up.

grahamwhaley:
Hi.
 Have updated. Split the patches down so it is a little clearer, and easier for me to check, and helped with the rebase. Have tested with both docker and native pandoc, and for when neither is found. afaict this is now working.

vbatts:
Oh good.

grahamwhaley:
Fix last commit subject: s/Makeile/Makefile/

Makefile
vbatts: this is still missing the `cp` of the img directory

grahamwhaley: I am afk for the next week, will look when I get back. The cp is present for the html generation is it not? Is the cp also necessary for the pdf?, as the images will be incorporated in the pdf file will they not?, or am I missing something?

vbatts: I'll double check this morning

vbatts:
LGTM

stevvooe:
LGTM

PR #136, opened 2016-06-14
image-layout: clarification

### LINKED ISSUE #132: image-layout: are `refs` following semver?
basically $SUBJECT - or can we name refs inside `./refs/` the way we want? Can I name refs using _digests_ also?

/cc @vbatts @philips

### LINKED ISSUE #134: image-layout: clarification about ./blobs/ directory content
It isn't (wasn't for me) immediately clear that files under `./blobs/` can be either manifests or actual tar blobs - the example at the end shows only a manifest so this made me ask where acutal data (tars) was. I think this should be clarified.

runcom:
Fix #132 
Fix #134

/cc @vbatts @philips

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -26,7 +26,7 @@ $ find .
 ./oci-layout
 ./refs
 ./refs/v1.0
-./refs/v1.1
+./refs/stable-release
 ```
 
 Blobs are named by their contents:
@@ -36,12 +36,13 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the refs and blobs MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
+Object names in the `refs` and `blobs` MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
 Hash algorithm identifiers containing the colon `:` will be converted to the hyphen `-`.
 For example `sha256:5b` will map to the layout `blobs/sha256-5b`.
 The blobs directory MAY contain blobs which are not referenced by any of the refs.
 The blobs directory MAY be missing referenced blobs, in which case the missing blobs SHOULD be fulfilled by an external blob store.
 
+No semantic restriction is given for object names in the `refs` subdirectory.
 Each object in the `refs` subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
 In general the `mediatype` of this descriptor object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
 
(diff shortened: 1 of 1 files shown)


image-layout.md
```diff
@@ -26,7 +26,7 @@ $ find .
 ./oci-layout
 ./refs
 ./refs/v1.0
-./refs/v1.1
+./refs/another-ref
```
vbatts: Good point to have a generic reference. Perhaps it could be something like `./refs/stable-release`?

philips:
besides the feedback from @vbatts this LGTM

image-layout.md
runcom: I can name it as you wish

vbatts:
one nit, but LGTM

image-layout.md
runcom: Fixed :)

vbatts:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/image-spec/pull-request/136/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/image-spec)

philips:
LGTM

image-layout.md
```diff
@@ -36,27 +36,46 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the refs and blobs MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
+Object names in the `refs` and `blobs` MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
```
wking: A few comments on this line:

* We should include the digits `0` through `9`.
* Aren't the blob filenames already completely specified?  `{algo}-{hex-digest}` with a defined list of algorithms cannot exceed this set unless we explicitly assign an algorithm an odd tag.

image-layout.md
```diff
@@ -36,27 +36,46 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
...
-{"size": 4096, "digest": "sha256:afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51", "mediatype": "application/vnd.oci.image.manifest.list.v1+json"}
+{"size": 4096, "digest": "sha256:afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51", "mediatype": "application/vnd.oci.image.manifest.v1+json"}
 ```
 ```
 $ cat ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 {
   "schemaVersion": 2,
-  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
-  "manifests": [
```
wking: I liked the `manifest.list` example better because that's [how we expect multiple platforms and manifest versions to be supported][1].

[1]: https://github.com/opencontainers/image-spec/pull/82#issuecomment-222003638

runcom: That'd require going to down to a tar stream eventually to make this clear as it is now

wking: On Tue, Jun 14, 2016 at 12:30:22PM -0700, Antonio Murdaca wrote:
> >  ```
> >  ```
> >  $ cat ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
> >  {
> >    "schemaVersion": 2,
> > -  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
> > -  "manifests": [
> > +  "mediaType": "application/vnd.oci.image.manifest.v1+json",
>
> That'd require going to down to a tar stream eventually to make this
> clear as it is now

I'm fine with that.  It gives us space to explain both the utility of
the manifest.list reference and the location of the tarballs, and both
of those have been non-obvious.

PR #137, opened 2016-06-15
image-layout.md: fixes

runcom:
/cc @vbatts @philips see individual commits messages for context

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -36,7 +36,7 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the `refs` and `blobs` MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
+Object names in the `refs` and `blobs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
 Hash algorithm identifiers containing the colon `:` will be converted to the hyphen `-`.
 For example `sha256:5b` will map to the layout `blobs/sha256-5b`.
 The blobs directory MAY contain blobs which are not referenced by any of the refs.
--- a/image-layout.md
+++ b/image-layout.md
@@ -7,7 +7,7 @@ Given an image layout a tool can convert a given ref into a runnable OCI Image F
 The image layout has two top level directories:
 
 - "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque.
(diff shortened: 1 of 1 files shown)


image-layout.md
```diff
@@ -7,7 +7,7 @@ Given an image layout a tool can convert a given ref into a runnable OCI Image F
 The image layout has two top level directories:
 
 - "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque.
-- "refs" contains descriptors pointing to an image manifest list
+- "refs" contains descriptors pointing to an image manifest
```
wking: I'd make this more generic, and just say that it contains descriptors.  There is the later:

> The blobs directory MAY be missing referenced blobs, in which case the missing blobs SHOULD be fulfilled by an external blob store.

which means the referenced blob may not even be in the blob store (and I think that's fine).

wking:
I left one minor nit comment inline, but otherwise 346ca185 looks good
to me (and b0996ba and 790e2d3 look good to me as they stand).

image-layout.md
runcom: I don't get this, referenced blobs may be missing from `./blob/` but it's a thing that `./refs` contains  manifest (either single or list)

philips:
LGTM

image-layout.md
```diff
@@ -36,7 +37,7 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the `refs` and `blobs` MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", the hyphen `-`, the dot `.`, and the underscore `_`.
+Object names in the `refs` and `blobs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
```
vbatts: good catch

image-layout.md
vbatts: refs are going to be a descriptor, though they ought to point to a manifest or manifest list

vbatts:
LGTM

image-layout.md
wking: On Wed, Jun 15, 2016 at 08:22:22AM -0700, Vincent Batts wrote:
> > -- "refs" contains descriptors pointing to an image manifest list
> > +- "refs" contains descriptors pointing to an image manifest
> 
> refs are going to be a descriptor, though they ought to point to a
> manifest or manifest list

refs (being descriptors) *tell* you what they point to.  Folks
interested in having oci-image-tool (or other tool that just
understands the current version of this spec) be able to unpack their
ref should point them at
application/vnd.oci.image.manifest.list.v1+json or
application/vnd.oci.image.manifest.v1+json.  But there's no problem
with image-layout if they have a ref that points at a different media
type (e.g. maybe they have an extention, or are pointing at
application/vnd.oci.image.manifest.list.v2+json).  The image-layout
format (and contents of oci-layout) can still be the same.

Issue #130, opened 2016-06-10
Describe format and application of whiteout files


stevvooe:
Whiteout files allow layers higher in the stack to control how base layer content appears in a container filesystem. Much of the whiteout format comes from the AUFS implementation, but certain files are not transmitted. There are many details about format and application that are currently missing from the specification.

Ensuring that these whiteouts are handled consistently is critical to a portable image format.

The docker implementation can inform this specification. Several details on the whiteout files are available [here](https://github.com/docker/docker/blob/3d13fddd2bc4d679f0eaa68b0be877e5a816ad53/pkg/archive/whiteouts.go). Change calculation is [here](https://github.com/docker/docker/blob/3d13fddd2bc4d679f0eaa68b0be877e5a816ad53/pkg/archive/changes.go#L83).

It would be good to examine other implementations to inform this, as well.

Supercedes #127

philips:
Are you going to take on the task of writing this up? Can I assign this to you?

vbatts:
This is specifically referring to the `*.wh.*` entries in a tar archive? or also a whiteout list?

philips:
ping @stevvooe

stevvooe:
@philips Sorry. I have been busy with a release.

We need to specify the format of the `*.wh.*` files. There are also security considerations that we may want to address in the specification.

stevvooe:
It may be sufficient to include language to this affect, from http://aufs.sourceforge.net/aufs.html:

> The whiteout is for hiding files on lower branches. Also it is applied to stop readdir going lower branches. The latter case is called ’opaque directory.’ Any whiteout is an empty file, it means whiteout is just an mark. In the case of hiding lower files, the name of whiteout is ’.wh.<filename>.’ And in the case of stopping readdir, the name is ’.wh..wh..opq’ or ’.wh.__dir_opaque.’ The name depends upon your compile configuration CONFIG_AUFS_COMPAT. All whiteouts are hardlinked, including ’<writable branch top dir>/.wh..wh.aufs.’

The overlay documentation contains s imilar section: https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt.

The main issue is that we have an example based specification. Adding an introduction to the changeset model will cover this.

PR #141, opened 2016-06-15
serialization: add language about whiteout

philips:
This fixes https://github.com/opencontainers/image-spec/issues/130

### DIFF
--- a/serialization.md
+++ b/serialization.md
@@ -442,7 +442,7 @@ The `f60c56784b83` directory then looks like this:
 ```
 f60c56784b83/
     etc/
-        .wh.my-app.cfg
+        .wh.my-app-config
         my-app.d/
             default.cfg
     bin/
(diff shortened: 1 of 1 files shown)


stevvooe:
I think the only open item is the handling of opaque directories: https://github.com/docker/docker/blob/22aca92ee3503f811f813be4a1349ad8e67fbbdb/daemon/graphdriver/aufs/aufs.go#L369. From there, it looks like they are actually omitted.

In particular, we have the following case, assuming `/foo/bar` exists and has entries:

```
delete  dir  /foo/bar
add     dir  /foo/bar
add     file /foo/bar/data
```

Based on the specification, we get the following in the top-level tar:

```
/foo/.wh.bar
/foo/bar
/foo/bar/data
```

From my understanding of aufs/overlay, this is handled via opq, but these aren't transmitted in layers, currently.

philips:
@stevvooe If the opaque directories aren't transmitted/handled what needs to be documented?

stevvooe:
@philips We need to clarify how to represent that particular situation. I'll have to track this down.

philips:
@stevvooe Can we merge this improvement and I will file a separate opaque issue for you to hunt down?

stevvooe:
LGTM

jonboulle:
LGTM

Issue #133, opened 2016-06-13
export types for implementation


runcom:
are types under `image/` exportable for implementation rather than defining our own in pkgs?

wking:
On Mon, Jun 13, 2016 at 03:32:49PM -0700, Antonio Murdaca wrote:
> are types under `image/` exportable for implementation instead of
> defining our own in pkgs?

I'd expect so, but either way it's probably worth documenting.
runtime-spec references its Go types as bindings since
opencontainers/runtime-spec#342, and we may want similar language
here.

vbatts:
I was mulling on this as well.

philips:
@vbatts what is your hesitation?

vbatts:
nothing more than a clean layout of the source. Reference structs in one import path, and validation in another.

stevvooe:
@vbatts @runcom @philips Ultimately, I think this is a good idea. I do share @vbatts hesitation, but would like to see Go implementations using common implementations to avoid fracture.

https://godoc.org/github.com/docker/distribution/manifest is the package structure used in the Docker implementations. There is schema back-registration in https://godoc.org/github.com/docker/distribution#RegisterManifestSchema. Its not perfect, but could inform the implementation and help avoid mistakes.

That said, the biggest problems that arise are dependency coupling. It might be easier to deal with backwards compatibility if the algorithms are exported, rather than the structures themselves.

vbatts:
@runcom do you want to try your hand at this?

runcom:
@vbatts sure, what's in docker/distribution is a nice start as well

wking:
Fixed by #172?

philips:
I think this is fixed by #172 so I am closing it. Let me know if this isn't true @runcom

PR #147, opened 2016-06-16
Export main types for v1

runcom:
Fix https://github.com/opencontainers/image-spec/issues/133

/cc @stevvooe @vbatts @philips 

This is basically how I expect those types to be exported and used in other tools.


philips:
Two questions:

1) Can we add tests against the json schema to ensure these things can serialize and unserialize that correctly?
2) Are there any code gen programs that can generate structs from json schema?

runcom:
1) my plan was to firstly agree on these two commits, then do what you're saying (and even adapt current code already present in the code base to use these types) 

2) I'm not sure, will have to look around

runcom:
2) bis, I basically did what's in runtime-spec atm and docker is using as well

philips:
@runcom understood. A trivial review of the structs LGTM

vbatts:
2) I looked all over for such a creature and wish there were a code-gen from json-schema. It would like be a tailored solution.

jonboulle:
This seems totally crazy. What's the utility of json-schema if you can't
generate code?!? pah

philips:
cc @casualjim know of any json schema to Go struct generators? Our schema files are here: https://github.com/opencontainers/image-spec/tree/master/schema

casualjim:
You can generate code from it with go-swagger. There is only one construct in use that go-swagger currently doesn't support. 

The oneOf here is currently not supported but we support  a slightly different syntax to get that same effect (x-isnullable).
https://github.com/opencontainers/image-spec/blob/master/schema/defs-image.json#L76-L86

This should get the generation process going: 

```
swagger generate model --help
```

runcom:
Cool, we can work with that and maybe add a makefile target as well

vbatts:
@casualjim thanks for the info

runcom:
last commit does a simple generation of the defs in the ./schema dir (just a test so everyone can see how they look like)

specs/array_of_g_ids.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	strfmt "github.com/go-openapi/strfmt"
+
+	"github.com/go-openapi/errors"
+)
+
+/*ArrayOfGIds array of g ids
+
+swagger:model ArrayOfGIDs
```
stevvooe: Aren't we already using jsonschema? Do we need to introduce yet another thing in here?

runcom: @stevvooe please refer to comments above about using swagger.. (this is just a test commit to discuss if this is the way we would go with this generation)

specs/mount.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/go-openapi/validate"
+)
+
+/*Mount mount
+
+swagger:model Mount
+*/
+type Mount struct {
+
+	/* destination
```
stevvooe: These should follow Go style guidelines. I'm not sure if we have that policy in this project, but we should probably adopt it.

specs/mount.go
```diff
@@ -0,0 +1,92 @@
...
+
+	/* type
+
+	Required: true
+	*/
+	Type *string `json:"type"`
+}
+
+// Validate validates this mount
+func (m *Mount) Validate(formats strfmt.Registry) error {
```
stevvooe: Let's decouple validation from the types.

specs/mount.go
runcom: that's the generation tool used - swagger - I would not go there and fix this. But agree we should follow Go style..

specs/mount.go
runcom: there's a flag not to generate validation at all - not sure how to generate _only_ validation, but really, should we go ahead with this way of generating code with swagger?!

stevvooe: There are a lot of problems with coupling the types and validation.

runcom: Right, I'll see if swagger can just output validations to different files at least

runcom: unfortunately I cannot generate w/o validation - the `--skip-validator` seems not to work

specs/mount.go
casualjim: What's missing for it to be compliant with the style guide. The goals are that the generated code passes golint etc checks.

specs/mount.go
casualjim: I'd love to learn about _a lot of problems_ you can find me on gophers slack for example to teach me what is correct.

specs/mount.go
stevvooe: This particular comment does not comply with `godoc` style.

specs/mount.go
runcom: thanks @casualjim for https://github.com/go-swagger/go-swagger/pull/569 gonna regenerate models w/o validations.

Though, @casualjim is there any way to output validations only?

runcom:
so this is what swagger generates w/o the validation - https://github.com/runcom/image-spec/commit/5e320779471f7341954ada5beb99fddaa423e2c1 - I still have to find a way to generate validation only go files (because `--skip-struct` does not seem to generate validation only, it just does nothing). Just wanted to gather acks on that commit and the try and add validation

vbatts:
@runcom what's the latest here?

runcom:
@vbatts basically types are generated in https://github.com/runcom/image-spec/commit/5e320779471f7341954ada5beb99fddaa423e2c1. Validation isn't coming with that PR because `--skip-struct` in swagger does not work correctly so I'm unsure on what to do next (given we have just the generated types..)

casualjim:
Can you help me understand what is required? You want to split up the generated files in a file with the type definition and a file with the validation?  I'd love to understand the reasons for it but let's start first with confirming intent.

Would it be possible to submit an issue when you run into a blocker with swagger?

wking:
On Mon, Jun 27, 2016 at 04:57:46PM -0700, Ivan Porto Carrero wrote:
> You want to split up the generated files in a file with the type
> definition and a file with the validation?

Extrapolating from [1], I think the idea is to decouple the types
(which should be very stable) from validation code (which probably
doesn't need to be as stable).

It seems like that's what --skip-validator would be about, but we had
trouble figuring out what that did (or if it was wired up to anything
at all) [2].

[1]: https://github.com/opencontainers/image-spec/pull/147#discussion_r67387694
[2]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-06-16.log.html#t2016-06-16T18:41:38

casualjim:
I've added support for `--skip-struct` to the `generate model` command

runcom:
thx @casualjim - I've opened an Issue in swagger for validations

runcom:
Updated @vbatts @philips @stevvooe PTAL

philips:
LGTM overall, two questions:

1) Should we be running go fmt over the generated code?
2) Is x-nullable sufficiently "standard" enough that other language generators will work?

runcom:
need a bunch of test fixing

runcom:
actually looking at https://github.com/go-swagger/go-swagger/issues/578

runcom:
blocking again at https://github.com/sgotti/glide-vc/pull/16 (found a bug)

casualjim:
@philips x-nullable comes from the swagger folks as the recommended work-around, I've seen both x-isnullable and x-nullable in use so we support both. I hope that for oas 3.0 there will be full support for json schema instead of the cherry-picking that was done for 2.0

stevvooe:
I'm still fairly uncomfortable with this PR. The generated code and validation are of low quality and we don't end up with very solid godoc output. Without language-specific directives polluting the schema, I am not sure there is a resolution to this.

It feels like this could have been done by now if we just did it by hand and validated against the schema after generation. Even with the generated code, we're still going to have to manually compare the specification with the generated implementations. If we can't get `go-swagger` to work in those cases, we'll end up 

As far as I see it, there are a few options:

1. Hand generated structs and validators.
  * Must auto-generate test cases and validate them against schema.
  * Pain-staking.
  * Easy to introduce inconsistencies
  * By far, will be the best result
2. Fix go-swagger.
  * May require backwards-incompatible changes to go-swagger.
  * Language-specific formatting directives (see gogo/protobuf for a taste of that nightmare)
3. Write our own generator.
  * May be a lot of work.
  * Easy to special case generation
  * Probably faster to write them by hand.

I'm sorry if this is not a majority opinion. In general, I agree with the sentiment of having this in place, so thank you @runcom for your work in this area.

Let me know if I can help in any way to make this go along faster.

specs-go/annotations.go
```diff
@@ -0,0 +1,25 @@
...
+
+
+/*Annotations annotations
+
+swagger:model annotations
+*/
+type Annotations struct {
+  
+  
+  MapStringString
```
stevvooe: I know this is generated code, but this should be `type Annotations map[string]string`.

runcom:
I tend to agree with @stevvooe here as I'm not totally fond about going with go-swagger here for the same reasons he told us above. 

If anyone is interested I have the hand-written types in these patches:

- https://paste.fedoraproject.org/388805/30275146
- https://paste.fedoraproject.org/388806/30306146

stevvooe:
@runcom Those patches look great!

vbatts:
What are the next steps here? Wait on using code generation, and pull in the hand done structs/types/etc?

runcom:
> What are the next steps here? Wait on using code generation, and pull in the hand done structs/types/etc?

I'd go for my hand written one as runtime-spec is already doing due to the lack of an appropriate and clean code generation tool

vbatts:
Sounds good.

PR #155, opened 2016-06-17
manifest: explain schemaVersion == 2

### LINKED ISSUE #121: manifest: Why use schemaVersion == 2
It might be good to have some additional guidance around schemaVersion
> schemaVersion  int
> This REQUIRED property specifies the image manifest schema version. This schema uses the version  2 .

For example, is this intended to align with major and minor versions?  i.e. v0.1 ==1, v0.2 ==2, etc.?

runcom:
Fix #121 

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -23,7 +23,7 @@ A client will distinguish a manifest list from an image manifest based on the Co
 - **`schemaVersion`** *int*
 
   This REQUIRED property specifies the image manifest schema version.
-  This schema uses the version `2`.
+  For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker. The value of this field will not change. This field MAY be removed in a future version of the specification.
 
 - **`mediaType`** *string*
 
@@ -139,7 +139,7 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 - **`schemaVersion`** *int*
 
   This REQUIRED property specifies the image manifest schema version.
-  For this version of the specification, this MUST be `2`.
+  For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker. The value of this field will not change. This field MAY be removed in a future version of the specification.
 
 - **`mediaType`** *string*
 


runcom:
@stevvooe should we mention schemaVersion might be removed in future schema version?

jonboulle:
Yeah, I would like to clarify that it's only necessary for backwards-compat and will be deprecated in future

runcom:
@jonboulle added a note

manifest.md
```diff
@@ -139,7 +139,7 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 - **`schemaVersion`** *int*
 
   This REQUIRED property specifies the image manifest schema version.
-  For this version of the specification, this MUST be `2`.
+  For this version of the specification, this MUST be `2` to ensure backward compatibility with older version of Docker. The schema version reflects a sub-versioning over the representation of the schema over time. This field MAY be removed in future version of the specification.
```
jonboulle: Can we make this note the same in both places?

runcom: fixed

jonboulle:
@runcom sorry for the back and forth, but I think we should actually omit the bit about subversioning, it's not complete enough context. My suggestion:

>  For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker. The value of this field will not change. This field MAY be removed in a future version of the specification.

runcom:
> @runcom sorry for the back and forth, but I think we should actually omit the bit about subversioning, it's not complete enough context. My suggestion:

right, wasn't really sure about that bit as well, updated

jonboulle:
thanks! LGTM

jonboulle:
LGTM
pull approve--

philips:
LGTM 

I hope the discussion on https://groups.google.com/a/opencontainers.org/d/msg/dev/Rcfn4xVLn7s/m9BbPXjjAQAJ clarifies the meaning of compatibility with Docker Engine/Distribution in the near future.

stevvooe:
LGTM

PR #164, opened 2016-06-26
Add some text about extensions

duglin:
Need to be kept in-sync with: https://github.com/opencontainers/runtime-spec/pull/510

Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -61,17 +61,22 @@ A client will distinguish a manifest list from an image manifest based on the Co
 
         This OPTIONAL property specifies an array of strings, each specifying a mandatory CPU feature (for example `sse4` or `aes`).
 
-- **`annotations`** *string-string hashmap*
+- **`annotations`** *string-string map*
 
     This OPTIONAL property contains arbitrary metadata for the manifest list.
-    Annotations is a key-value, unordered hashmap.
-    Keys are unique, and best practice is to namespace the keys.
-    Common annotation keys include:
-    * **created** date on which the image was built (string, timestamps type)
-    * **authors** contact details of the people or organization responsible for the image (freeform string)
-    * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
-    * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
+    Annotations MUST be a key-value map where both the key and value MUST be strings.
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique within this map, and best practice is to namespace the keys.
+    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications.
+    If there are no annotations then this property MUST either be absent or be an empty map.
+    Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
 
+    See [Pre-Defined Annotation Keys](#pre-defined-annotation-keys).
+
+### Extensibility
+Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown property.
+Instead they MUST ignore unknown properties.
 
 ## Example Manifest List
 
@@ -104,8 +109,8 @@ A client will distinguish a manifest list from an image manifest based on the Co
     }
   ],
   "annotations": {
-    "key1": "value1",
-    "key2": "value2"
+    "com.example.key1": "value1",
+    "com.example.key2": "value2"
   }
 }
 ```
@@ -139,18 +144,22 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
     Subsequent layers MUST then follow in the order in which they are to be layered on top of each other.
     The algorithm to create the final unpacked filesystem layout MUST be to first unpack the layer at index 0, then index 1, and so on.
 
-- **`annotations`** *hashmap*
+- **`annotations`** *string-string map*
 
-    This OPTIONAL property contains arbitrary metadata for the manifest list.
-    Annotations is a key-value, unordered hashmap.
-    Keys MUST be unique within the hashmap, and best practice is to namespace the keys.
-    Common annotation keys include:
-    * **created** date on which the image was built (string, timestamps type)
-    * **authors** contact details of the people or organization responsible for the image (freeform string)
-    * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
-    * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
-    * **source** URL to get the source code for the binary files in the image (string, must be a URL with scheme HTTP or HTTPS)
+    This OPTIONAL property contains arbitrary metadata for the image manifest.
+    Annotations MUST be a key-value map where both the key and value MUST be strings.
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique within this map, and best practice is to namespace the keys.
+    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications.
+    If there are no annotations then this property MUST either be absent or be an empty map.
+    Implementations that are reading/processing the image manifest MUST NOT generate an error if they encounter an unknown annotation key.
 
+    See [Pre-Defined Annotation Keys](#pre-defined-annotation-keys).
+
+### Extensibility
+Implementations that are reading/processing image manifests MUST NOT generate an error if they encounter an unknown property.
+Instead they MUST ignore unknown properties.
 
 ## Example Image Manifest
 
@@ -182,8 +191,15 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
     }
   ],
   "annotations": {
-    "key1": "value1",
-    "key2": "value2"
+    "com.example.key1": "value1",
+    "com.example.key2": "value2"
   }
 }
 ```
+
+# Pre-Defined Annotation Keys
+This specification defines the following annotation keys, which MAY be used by manifest list and image manifest authors:
+* **org.opencontainers.created** date on which the image was built (string, date-time as defined by [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6)).
+* **org.opencontainers.authors** contact details of the people or organization responsible for the image (freeform string)
+* **org.opencontainers.homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
+* **org.opencontainers.documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)


manifest.md
```diff
@@ -85,13 +85,18 @@ A client will distinguish a manifest list from an image manifest based on the Co
 
     This OPTIONAL property contains arbitrary metadata for the manifest list.
     Annotations is a key-value, unordered hashmap.
-    Keys are unique, and best practice is to namespace the keys.
+    Keys MUST be unique, and best practice is to namespace the keys.
+    If there are no annotations then this property MAY either be absent or be an empty hashmap.
+    Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation.
```
vbatts: s/unknown annotation/unknown annotation key/ ?

vbatts:
SGTM, also, there are two annotations sections in this doc, but your PR only covers one of them.

manifest.md
duglin: fixed

duglin:
oops - got the other one now

manifest.md
```diff
@@ -85,13 +85,18 @@ A client will distinguish a manifest list from an image manifest based on the Co
 
     This OPTIONAL property contains arbitrary metadata for the manifest list.
     Annotations is a key-value, unordered hashmap.
-    Keys are unique, and best practice is to namespace the keys.
+    Keys MUST be unique, and best practice is to namespace the keys.
+    If there are no annotations then this property MAY either be absent or be an empty hashmap.
+    Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
```
vbatts: do you want to see this same sentence on the other section too?

duglin: darn- I think I moved instead of copied.  fixed -thanks!

manifest.md
```diff
@@ -85,13 +85,18 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
     Common annotation keys include:
     * **created** date on which the image was built (string, timestamps type)
     * **authors** contact details of the people or organization responsible for the image (freeform string)
     * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
     * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
 
+### Extensibility
+The `annotations` property MAY be used as an extensibility point to include additional information that is not defined as part of this specification.
```
stevvooe: "extension point"

duglin:
updated based on today's conf call

manifest.md
```diff
@@ -85,13 +85,20 @@ A client will distinguish a manifest list from an image manifest based on the Co
 
     This OPTIONAL property contains arbitrary metadata for the manifest list.
     Annotations is a key-value, unordered hashmap.
-    Keys are unique, and best practice is to namespace the keys.
+    Keys MUST be unique, and best practice is to namespace the keys.
+	Keys starting with the `opencontainers.org` namespace are reserved and MUST NOT be used.
```
vbatts: Mostly I like it. I'm wondering if just saying "... are reserved." would be enough? If not to be used, then should this cover when the are used?

stevvooe: Is this not `org.opencontainers` or is that not how it is done here?

vbatts: that ought to be clarified. I had in mind that it would be namespaced more
like an URL, rather than a java package.

wking: On Mon, Jul 18, 2016 at 10:20:04AM -0700, Vincent Batts wrote:
> > -    Keys are unique, and best practice is to namespace the keys.
> > +    Keys MUST be unique, and best practice is to namespace the keys.
> > +	Keys starting with the `opencontainers.org` namespace are reserved and MUST NOT be used.
> 
> that ought to be clarified. I had in mind that it would be namespaced more
> like an URL, rather than a java package.

Does it matter?  And as long as folks don't have actual hosts like
net.example.com, folks could use both.  If we pick one ordering we
don't even have to worry about that, but I don't see any technical
difference between the two orderings.

vbatts: for all the specifics of clarification, it's interesting you don't see a
difference in the two

wking: On Mon, Jul 18, 2016 at 10:31:59AM -0700, Vincent Batts wrote:
> > @@ -85,13 +85,20 @@ A client will distinguish a manifest list from an image manifest based on the Co
> >  
> >      This OPTIONAL property contains arbitrary metadata for the manifest list.
> >      Annotations is a key-value, unordered hashmap.
> > -    Keys are unique, and best practice is to namespace the keys.
> > +    Keys MUST be unique, and best practice is to namespace the keys.
> > +	Keys starting with the `opencontainers.org` namespace are reserved and MUST NOT be used.
> 
> for all the specifics of clarification, it's interesting you don't
> see a difference in the two

I'm all for picking one ordering to recommend, but I'm happy to flip a
coin to pick ;).  Does anyone have a reason to *not* choose via
coin-flip?

It also looks like there's inconsistent whitespace on the indent here,
with the final new line using a tab while everyone else uses spaces.

vbatts:
do y'all think we can clear up this conversation on this feature? like maybe for v0.4.0?

I'm for this feature. Perhaps just with clarity around the prefixing?

duglin:
I looked at both Docker and Kubernetes for examples and Docker seems to use the reverse notation (org.docker.*) while Kubernetes actually uses both based on (I guess) who wrote the code.  Averaging it out, that pushes it a bit towards reverse domain notation,  so I switched it to that and added a bit of text pushing people to follow that pattern.

vbatts:
@duglin good thoughts, though to me that seems we should cover both use cases.

duglin:
you want to reserve org.opencontainers.* and *.opencontainers.org ?

I guess we could, but I have to admit, the inconsistency I saw in Kube's docs on this, while not a show-stopper, is kind of weird - so my preference would be to pick one and stick with it.

vbatts:
@brendandburns  (or @thockin or @smarterclayton) have opinions on annotations prefixes?

wking:
On Thu, Jul 21, 2016 at 05:53:43AM -0700, Doug Davis wrote:
> you want to reserve org.opencontainers.* and *.opencontainers.org ?
>
> I guess we could, but I have to admit, the inconsistency I saw in
> Kube's docs on this, while not a show-stopper, is kind of weird - so
> my preference would be to pick one and stick with it.

I agree on picking one to reserve, and recommending other folks use
the same order (although they can obviously use the other order if
they have a strong preference and are ok with the collision risk).

thockin:
I don't think there's much value in reserving both the forward and backward domains.  The backward domain is pretty annoying, frankly, and I don't understand why anyone would prefer it.  It's human-hostile, and we don't tend to use annotations as a hierarchy (like Java does).

Decide which way is standard and run with it.  I happen to think what is documented here is wrong, but Docker already chose that answer.

vbatts:
the other benefit to having an `opencontainers.org/` style, is that it is directly translatable to what could be a resolving path for folks looking for more info on the particular key. (much in the same vein of XML namespacing and validation)

wking:
On Thu, Jul 21, 2016 at 10:57:50AM -0700, Vincent Batts wrote:
> the other benefit to having an `opencontainers.org/` style, is that
> it is directly translatable to what could be a resolving path for
> folks looking for more info on the particular key.

That sounds like a reason to not choose via coin flip to me ;).

stevvooe:
With the reverse name model, `com.example.foo` and `com.example.bar` would both be related to `com.example`.

We can say a set of annotations is related if they share a common prefix. This cannot be done without more careful parsing when using the `example.com/` model.

The other thing to note is that they aren't actually domain names. Even the use of `.` is a convention, but the prefix model doesn't need to know this, as it just matches prefixes.

I'm not saying I prefer the reverse domain name order but they aren't actually domain names. The properties above probably weren't even considered when selecting a convention.

It's just a convention.

thockin:
Any standard is better than no standard, though it's unfortunate that the
two major systems that use annotations chose different syntaxes.

vbatts:
ugh. I see the value and drawback in both.

philips:
The advantage of the forward domains is you can visit the URL in a browser and potentially get some docs.

Honestly, I don't think we are going to solve much by trying to make people use the prefixing scheme "the project prefers". Lets just reserve "opencontainers.org" and "org.opencontainers" and tell users they should choose one for their own use cases in either style but they need to stick with whatever prefixing scheme they choose.

wking:
On Wed, Jul 27, 2016 at 02:57:31PM -0700, Brandon Philips wrote:
> Lets just reserve "opencontainers.org" and "org.opencontainers"…

If you don't add a termination character, it's hard to know what else
is reserved.  For example, say I own example.com and want to use
opencontainers.org.example.com.  Or if I have opencontainers-rock.org
and want to use org.opencontainers-rock.  If we claim
opencontainers.org/ and org.opencontainers/ we should be safe.

philips:
@wking I think it would be reserving `opencontainers.org/` and `org.opencontainers.` prefixes

philips:
But, yes +1

wking:
On Wed, Aug 03, 2016 at 01:18:07PM -0700, Brandon Philips wrote:
> @wking I think it would be reserving `opencontainers.org/` and
> `org.opencontainers.` prefixes

opencontainers/runtime-spec#510 is currently using
com.example/gpu-cores in an example, but I'm fine with
org.opencontainers. too.

wking:
On Wed, Aug 03, 2016 at 01:33:53PM -0700, W. Trevor King wrote:
> Wed, Aug 03, 2016 at 01:18:07PM -0700, Brandon Philips:
> > @wking I think it would be reserving `opencontainers.org/` and
> > `org.opencontainers.` prefixes
> 
> opencontainers/runtime-spec#510 is currently using
> com.example/gpu-cores in an example, but I'm fine with
> org.opencontainers. too.

Actually no, it has to be slash in both cases.  Otherwise you block
the org.opencontainers.example.com folks from using
org.opencontainers.example.com/….

philips:
Having a slash in the middle of reverse dns notation isn't helpful either as that is unexpected. I am OK telling users they can't use org.opencontainers as a prefix by fiat forever.

wking:
On Wed, Aug 03, 2016 at 02:49:33PM -0700, Brandon Philips wrote:
> Having a slash in the middle of reverse dns notation isn't helpful
> either as that is unexpected.

Maybe I just haven't looked at enough reverse DNS names, but I don't
have any expectations violated by seeing a slash after them.

> I am OK telling users they can't use org.opencontainers as a prefix
> by fiat forever.

“We don't have a sane delimiter for it” (assuming / is not sane)
sounds like an argument against reverse DNS to me ;).

thockin:
org.opencontainers.example.com translates to a DNS name
com.example.opencontainers.org.  Nobody who is not authoritatively part of
opencontainers.org should use that (reverse) prefix.  The point of using
any tie-in to DNS is that we have a well-understood, global delegation
system.  It is relatively easy to prove that you are allowed to use
foo.opencontainers.org

wking:
On Wed, Aug 03, 2016 at 10:12:10PM -0700, Tim Hockin wrote:
> org.opencontainers.example.com translates to a DNS name
> com.example.opencontainers.org.

If you know that it is reversed.  If we support both forward and
reverse names, how can you tell that it's not a forward version of
org.opencontainers.example.com?

thockin:
because in forward form it would have a slash and a suffix?

"com.example.foo" and "example.com/foo" are semantically the same in the
two systems.  We do not allow multiple slashes in kubernetes, and AFAIK
docker assumes the last token is the "name".  Neither system applies
"meaning" to the structure nor decomposes it further, AFAIK.  It's more
about how humans parse it.

wking:
On Thu, Aug 04, 2016 at 10:25:21AM -0700, Tim Hockin wrote:
> because in forward form it would have a slash and a suffix?

So as regular expressions we claim:

  org\.opencontainers\.[^/]*
  opencontainers\.org/.*

That sounds fine to me.  The general pattern for DNS-based namespace
delegation is then:

* Reverse-order keys must not use a slash anywhere in the key.
* Forward-order keys must use a slash immediately after the domain
  name.

Although we're ok with folks violating these rules if they want, so
long as they don't step into the opencontainers.org space that the
rules mark out.

philips:
@wking sure, making it explicit via a regex works for me. Thoughts @duglin ?

duglin:
I would really prefer if we picked one so that our users had a consistent pattern to follow, and I like the reverse DNS format for the reason @stevvooe gave -  its easier to find groups when they're of the form `com.example/xxx` or `com.example.xxx`.  Its harder when its `xxx.example.com` because then someone could do `xxx.example.com/yyy` - not much harder but enough to sway me.

But I defer to the folks with the power to LGTM - if you guys vote to allow for both and want it expressed as a regex I can make that change.

One thing though, and this gets to @thockin's comment, I do think we should make it clear that these are just strings to the spec and the use of `.` vs `/` is of no consequence to us - its the user of this data that might infer meaning from that info.  Additionally, for the purpose of searching we should be clear whether names are case sensitive or not, and I would opt for 'yes'.  Inconsistency here between impls could be a nightmare for interop. I can add text around this once we get agreement on which direction to go.

wking:
On Wed, Aug 10, 2016 at 10:17:41AM -0700, Doug Davis wrote:
> Additionally, for the purpose of searching we should be clear
> whether names are case sensitive or not, and I would opt for 'yes'.

Do we need to specify that?  JSON object keys are case-sensitive, so
annotation keys are case-sensitive just like all other JSON object
keys.

I'm fine adding language to make it clear that the only purpose of the
DNS-based namespacing is to reduce the likelyhood of multiple vendors
using the same annotation-key for different purposes (so it's a
caution to external spec-writers).  The code consuming the annotations
isn't writing specs, so it should just be treating the annotation keys
as opaque, case-sensitive tokens.

philips:
@duglin I would prefer if we picked one or the other as well. But, honestly, the shipped has sailed on trying to choose the one true key format for annotations. And, largely, it is of zero consequence to this specification at this point.

duglin:
True, but if we're going to say users can't use XXX because we might, and we have total control over XXX, I don't think we need to grab it forwards and backwards.  Unless we want to be consistent too :-)

I'll remove the `Keys SHOULD be named using a reverse domain notation` but I personally don't see a problem with the recommendation - note its not a `MUST`.

wking:
On Thu, Aug 11, 2016 at 06:02:48AM -0700, Doug Davis wrote:
> True, but if we're going to say users can't use XXX because we
> might, and we have total control over XXX, I don't think we need to
> grab it forwards and backwards.  Unless we want to be consistent too
> :-)

If we're going to involve DNS at all, I think we want SHOULD-level
rules like [1] for all domains.  Then the a.example.com can go to the
b.example.com folks and say, “please stop squatting on my
com.example.a.[^/]*.  Here's where image-spec SHOULDs that to me.”
Folks without a domain can use something like
{group-name}/{parameter}, e.g. ccon/magic.  With that in place, we
don't need to call out the opencontainers.org instances explicitly.

I'm also fine dropping DNS and just advising spec-writers to pick keys
they think are unlikely to collide ;).

[1]: https://github.com/opencontainers/image-spec/pull/164#issuecomment-237625793

philips:
In person discussion at OCI f2f. We are going to go with reverse in image-spec and will do the same in the runtime-spec.

philips:
@caniszczyk can you add @duglin so I can assign this to him?

duglin:
FYI the PR was written to use reverse-DNS as was agreed in the last f2f meeting

manifest.md
```diff
@@ -84,14 +84,22 @@ A client will distinguish a manifest list from an image manifest based on the Co
 - **`annotations`** *string-string hashmap*
 
     This OPTIONAL property contains arbitrary metadata for the manifest list.
-    Annotations is a key-value, unordered hashmap.
-    Keys are unique, and best practice is to namespace the keys.
+    Annotations MUST BE a key-value, unordered hashmap where both the key and value MUST be strings.
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique, and best practice is to namespace the keys.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used.
```
wking: I think we want to claim this [with a trailing period][1] and also SHOULD a [general rule for mapping a DNS name to the key namespace][2].

[1]: https://github.com/opencontainers/image-spec/pull/164#issuecomment-237625793
[2]: https://github.com/opencontainers/image-spec/pull/164#issuecomment-239182187

manifest.md
```diff
@@ -84,14 +84,22 @@ A client will distinguish a manifest list from an image manifest based on the Co
 - **`annotations`** *string-string hashmap*
 
     This OPTIONAL property contains arbitrary metadata for the manifest list.
-    Annotations is a key-value, unordered hashmap.
-    Keys are unique, and best practice is to namespace the keys.
+    Annotations MUST BE a key-value, unordered hashmap where both the key and value MUST be strings.
```
wking: Can we avoid duplicate entries for manifest and manifest-list?  The manifest is more fundamental, so I'd rather replace this list entry with:

    - **`annotations`** (string → string map, optional) as defined for [manifest](#image-manifest).

manifest.md
```diff
@@ -193,16 +201,24 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
...
     Keys MUST be unique within the hashmap, and best practice is to namespace the keys.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used.
+    If there are no annotations then this property MAY either be absent or be an empty hashmap.
+    Implementations that are reading/processing the image manifest MUST NOT generate an error if they encounter an unknown annotation key.
     Common annotation keys include:
     * **created** date on which the image was built (string, timestamps type)
     * **authors** contact details of the people or organization responsible for the image (freeform string)
     * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
     * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
     * **source** URL to get the source code for the binary files in the image (string, must be a URL with scheme HTTP or HTTPS)
```
wking: It feels strange to claim the `org.opencontainers.` namespace and then define “common” keys.  Are these part of the spec?  Are other folks free to interpret these keys with different semantics?  I'd rather put them in our namespace (following the SHOULD recommendation you removed with 100dcce8f → cf0fcb6b2 and the “best practice” recommendation that remains) and use`org.opencontainers.created`, `org.opencontainers.authors`, etc.

wking:
This PR currently adds “Extensibility” sections for manifests and manifest-lists.  I think we also want to cover configurations (and really, any JSON files defined by this spec).  Can we move that Extensibility section somewhere else so it can cover all the JSON at once?  One way to do this would be to push the definition as low as possible (the serialization config format?) and [link to it from higher-level structures][1].

[1]: https://github.com/opencontainers/image-spec/pull/164/files#r76831354

duglin:
updated

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
-    Common annotation keys include:
-    * **created** date on which the image was built (string, timestamps type)
-    * **authors** contact details of the people or organization responsible for the image (freeform string)
-    * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
-    * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
-
+    Annotations MUST BE a key-value map where both the key and value MUST be strings.
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique within this map, and best practice is to namespace the keys.
+	Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
```
wking: Inconsistent indent?  Likely involving a tab ;).

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent.
```
wking: Still missing a trailing dot.

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
+    If there are no annotations then this property MAY either be absent or be an empty map.
+    Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
+
+    Pre-defined annotation keys which MAY be used by manifest list authors:
```
wking: “Pre-defined” reads strangely to me.  Hopefully *all* annotation keys are defined somewhere before someone uses them in an actual manifest.  How about:

> This specification defines the following keys, which MAY be used by manifest-list authors:

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
+    * **org.opencontainers.created** date on which the image was built (string, timestamps type)
```
wking: Are timestamps defined anywhere?  I'd rather have this be:

> date-time on which the image was build (string, [`date-time` as defined by RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6)).

wking:
On Thu, Sep 01, 2016 at 07:18:32PM -0700, Doug Davis wrote:
> updated

I'd still like to cover the config in serialization.md (it's strange
that that doesn't have annotations already.  Oversight?) [1].  And I'd
like to DRY this up by definining annotations and extensions for one
of the JSON formats, and then linking to that definition from the
other JSON formats [1].

[1]: https://github.com/opencontainers/image-spec/pull/164#issuecomment-243952187

manifest.md
wking: Also you have “by subsequent.”.  See [here][1] for my suggested rewording.

[1]: https://github.com/opencontainers/runtime-spec/pull/510/files#r77291491

wking:
opencontainers/runtime-spec#510 [also namespaces the example keys][1], but here we still have:

    $ git grep -A2 '"annotations"' origin/pr/164 -- manifest.md
    origin/pr/164:manifest.md:  "annotations": {
    origin/pr/164:manifest.md-    "key1": "value1",
    origin/pr/164:manifest.md-    "key2": "value2"
    --
    origin/pr/164:manifest.md:  "annotations": {
    origin/pr/164:manifest.md-    "key1": "value1",
    origin/pr/164:manifest.md-    "key2": "value2"

[1]: https://github.com/opencontainers/runtime-spec/pull/510/files#diff-c9c91c29b41257aea3a3403cc606ad99L686

manifest.md
duglin: Fixed

manifest.md
duglin: I'd prefer to stick with the same wording and namespace we have for the runtime spec, which was merged yesterday. We can use a new PR to tweak if needed but we should do them both the same way.

manifest.md
duglin: done

manifest.md
duglin: done

duglin:
Updated

manifest.md
```diff
@@ -137,18 +146,27 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
...
+    This OPTIONAL property contains arbitrary metadata for the image manifest.
+    Annotations MUST BE a key-value map where both the key and value MUST be strings.
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique within this map, and best practice is to namespace the keys.
+    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
+    If there are no annotations then this property MAY either be absent or be an empty map.
+    Implementations that are reading/processing the image manifest MUST NOT generate an error if they encounter an unknown annotation key.
+
+	This specification defines the following key, which MAY be used by image manifest authors:
```
wking: “key” → “keys”.  And this line has a tab indent, but should probably be indented with spaces.

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
-
+    Annotations MUST BE a key-value map where both the key and value MUST be strings.
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique within this map, and best practice is to namespace the keys.
+    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
+    If there are no annotations then this property MAY either be absent or be an empty map.
+    Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
+
+	This specification defines the following keys, which MAY be used by manifest list authors:
```
wking: This line had a tab-indent [too][1].  And if we went with a [DRY approach][2] you'd only have one place to fix ;).

[1]: https://github.com/opencontainers/image-spec/pull/164/files#r77847153
[2]: https://github.com/opencontainers/image-spec/pull/164#issuecomment-243952187

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
     This OPTIONAL property contains arbitrary metadata for the manifest list.
-    Annotations is a key-value, unordered hashmap.
-    Keys are unique, and best practice is to namespace the keys.
-    Common annotation keys include:
-    * **created** date on which the image was built (string, timestamps type)
-    * **authors** contact details of the people or organization responsible for the image (freeform string)
-    * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
-    * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
-
+    Annotations MUST BE a key-value map where both the key and value MUST be strings.
```
jonboulle: MUST BE -> MUST be

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique within this map, and best practice is to namespace the keys.
+    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
```
jonboulle: what do you mean by subsequent specifications? how about just "other"? (wouldn't we expect that subsequent versions of this specification would honour the same ones?)

manifest.md
```diff
@@ -60,17 +60,26 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
+    If there are no annotations then this property MAY either be absent or be an empty map.
```
jonboulle: "MAY either be absent or an empty map" is not really any different from "MUST either be absent or an empty map"

manifest.md
jonboulle: Do we really need this list duplicated in the document? could we put it in a common section and reference it from both the manifest list and the image manifest? e.g. "well known opencontainers keys" or something

manifest.md
wking: On Wed, Sep 07, 2016 at 08:36:09AM -0700, Jonathan Boulle wrote:
> >  
> >      This OPTIONAL property contains arbitrary metadata for the manifest list.
> > -    Annotations is a key-value, unordered hashmap.
> > -    Keys are unique, and best practice is to namespace the keys.
> > -    Common annotation keys include:
> > -    * **created** date on which the image was built (string, timestamps type)
> > -    * **authors** contact details of the people or organization responsible for the image (freeform string)
> > -    * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
> > -    * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
> > -
> > +    Annotations MUST BE a key-value map where both the key and value MUST be strings.
> > +    While the value MUST be present, it MAY be an empty string.
> > +    Keys MUST be unique within this map, and best practice is to namespace the keys.
> > +    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
> > +    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
>
> what do you mean by subsequent specifications? how about just
> "other"? (wouldn't we expect that subsequent versions of this
> specification would honour the same ones?)

+1 to avoiding “subsequent specifications”.  I recommended using “…
are reserved for future versions of the specification.” [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/510#discussion_r77291491

manifest.md
wking: On Wed, Sep 07, 2016 at 08:36:40AM -0700, Jonathan Boulle wrote:
> could we put it in a common section and reference it from both the
> manifest list and the image manifest? e.g. "well known
> opencontainers keys" or something

I think we want to put it in the lowest type (manifest?  config?) and
have other types link there [1].

[1]: https://github.com/opencontainers/image-spec/pull/164#issuecomment-243952187

manifest.md
duglin: darn - thought I got both of those... guess not. fixed - thanks!
-Doug

manifest.md
duglin: to BE or not to BE !  :-0
fixed

manifest.md
duglin: done but I have to admit, it felt weird typing it

duglin:
updated

manifest.md
```diff
@@ -60,17 +60,22 @@ A client will distinguish a manifest list from an image manifest based on the Co
...
-    * **created** date on which the image was built (string, timestamps type)
-    * **authors** contact details of the people or organization responsible for the image (freeform string)
-    * **homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
-    * **documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
+    Annotations MUST be a key-value map where both the key and value MUST be strings.
+    While the value MUST be present, it MAY be an empty string.
+    Keys MUST be unique within this map, and best practice is to namespace the keys.
+    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications.
+    If there are no annotations then this property MUST either be absent or be an empty map.
```
wking: This is an odd requirement.  I think “absent” is covered by the earlier “This OPTIONAL property…”.  I think “empty map” is covered by there not being any required keys.

And I can't think of a way to test this.  You'd need to do something like “make a manifest without annotations” and check for an unset or empty-map `annotations`.  But the only way I can think of saying “make a make a manifest without annotations” is to say “`annotations` should be unset (or an empty map)”, so my test would be tautological.

If we don't have a clear idea of a better test for this line, can we drop it?

manifest.md
```diff
@@ -180,8 +189,16 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
...
   "annotations": {
-    "key1": "value1",
-    "key2": "value2"
+    "com.example.key1": "value1",
+    "com.example.key2": "value2"
   }
 }
 ```
+
+# Pre-Defined Annotation Keys
```
wking: This DRYs up the pre-defined annotation keys, which is good.  But there's still a lot of duplicate language about extensibility and the `annotations` field itself.  Can we [collect *all* of that into one location and link to it from other consumers][1]?

[1]: https://github.com/opencontainers/image-spec/pull/164#issuecomment-243952187

manifest.md
duglin: the point of this was to make it clear that if you don't have annotation you can still have the "annotations" property but it needs to be empty. What I was trying to avoid was someone claiming an impl is not compliant because they put an empty map instead of dropping the property all together.   OPTIONAL only covers the ability to leave it off entirely, not for it to be empty.

wking: On Wed, Sep 07, 2016 at 11:10:55AM -0700, Doug Davis wrote:
> > +    If there are no annotations then this property MUST either be absent or be an empty map.
>
> … What I was trying to avoid was someone claiming an impl is not
> compliant because they put an empty map instead of dropping the
> property all together.

Under what grounds would they make that claim?  There is no “MUST
contain at least one key” requirement.

> OPTIONAL only covers the ability to leave it off entirely, not for
> it to be empty.

Agreed.  The empty map case is covered by there not being any required
keys [1], because the spec'ed type is a JSON object, and an empty
object is still an object.  Array/object/string properties allow the
empty value by default unless the spec explicitly forbids it.

[1]: https://github.com/opencontainers/image-spec/pull/164/files/7350d6ca92e444636f45b5c669076cb83dee6756#r77871484

duglin: The points isn't that from a pure geeky perspective whether or not an empty set is allowed, it is. The problem is when humans get involved and they don't fully understand all of the specs when they come together. I've seen this type of interop problem before and I'm trying to avoid it.  As an example, I've seen it in Docker where two different people were treating foo:null and the absence of "foo" differently - one thought they were the semantic equivalent and the other did not.  Let's just be explicit and avoid it.

wking: On Wed, Sep 07, 2016 at 11:47:09AM -0700, Doug Davis wrote:
> As an example, I've seen it in Docker where two different people
> were treating foo:null and the absence of "foo" differently - one
> thought they were the semantic equivalent and the other did not.
> Let's just be explicit and avoid it.

I'm ok with explicitly saying that unset and null have the same
semantics as an empty array/map/string.  But I think we should be
making that point at a spec-wide level, and not just for the
annotations property.  Array/map/string properties for which that
general rule does not hold can override it explicitly on a
per-property level (to pick an example from runtime-spec, an unset or
null process.capabilities presumably means “use the platform default
capabilities” while an empty array means “drop all capabilities” [1]).

[1]: https://github.com/opencontainers/runtime-spec/blob/22c29c18eddc75ccee9f6b202f61524f29653c5f/config.md#process-configuration

duglin: Can we save that for another PR so we can get this behind us?

wking: On Wed, Sep 07, 2016 at 02:44:03PM -0700, Doug Davis wrote:
> > +    If there are no annotations then this property MUST either be absent or be an empty map.
> 
> Can we save that for another PR so we can get this behind us?

If you prefer.  In that case, I'd rather remove this line with the
understanding that we'll be adding more broadly-scoped language soon.
But I'm ok with this landing and then getting pulled back out when the
more broadly-scoped language lands.

stevvooe:
LGTM

manifest.md
jonboulle: bump, I'd still prefer this so there's less chance of getting out of sync later

stevvooe:
@duglin There are some whitespace errors that is choking the CI. Want to fix those so we can re-LGTM and get this merged?

duglin:
@stevvooe CI is fixed now

stevvooe:
LGTM

philips:
LGTM

Issue #88, opened 2016-05-25
documentation: explain the download workflow in practice


philips:
It is likely that the OCI Image Spec [won't have an optional transport layer](https://groups.google.com/a/opencontainers.org/d/msg/tob/oaEza4cu25M/gBLeHm5NFAAJ) for a few more months while the TOB discusses and comes to a conclusion. But, as people start to use/implement the spec they will want advice on how to use the format with existing transport mechanisms.

The first topic to cover is how to use the OCI Image Format spec with the [Docker v2 Registry API](https://docs.docker.com/registry/spec/api/#pulling-an-image). This document will need to explain the entire process from a name to extracted on disk. @wking drew this diagram of the basic process:

![oci-unpack](https://cloud.githubusercontent.com/assets/33544/15545982/0abcaf22-2253-11e6-9411-3f38dfc196e1.png)

We should probably create a helper `oci-image-docker-registry-fetch` tool to make it all concrete as well.

runcom:
@philips is the end result of this downloader-from-docker-registryv2 an image-layout as defined here in the spec or something else?

wking:
On Mon, Jun 13, 2016 at 07:20:10PM -0700, Antonio Murdaca wrote:
> @philips is the end result of this downloader-from-docker-registryv2
> an image-layout as defined here in the spec or something else?

The CAS fetching (e.g. pulling stuff from ./blobs) is part of this
download workflow.  I was expecting it to setup a generic directory
(with whatever content the image author had setup, possibly with the
intention of some pre-runtime processing) [1].  But if the image-spec
repository really sees a benefit to maintaining its own required
container config independent of the runtime config (#87), then the
graphic probably needs another “translate image-spec container config
to runtime-spec config” step.

[1]: https://github.com/opencontainers/image-spec/issues/87#issuecomment-221695266

vbatts:
Closing this as https://github.com/opencontainers/distribution-spec is fully in motion.

PR #165, opened 2016-06-27
image: convert colon to hypen for path digests.

sgotti:
Noticed trying to validate and unpack a (I think) valid oci image layout.

@philips @s-urbaniak Are there some tests for oci-image-tool in the work? I haven't written a test because probably a generic tests could have covered this.

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -51,7 +51,7 @@ type config struct {
 
 func findConfig(w walker, d *descriptor) (*config, error) {
 	var c config
-	cpath := filepath.Join("blobs", d.Digest)
+	cpath := filepath.Join("blobs", pathDigest(d.Digest))
 
 	f := func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() {
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -65,13 +65,14 @@ func findDescriptor(w walker, name string) (*descriptor, error) {
 }
 
 func (d *descriptor) validate(w walker) error {
+	pd := pathDigest(d.Digest)
 	f := func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() {
 			return nil
 		}
 
 		digest, err := filepath.Rel("blobs", filepath.Clean(path))
-		if err != nil || d.Digest != digest {
+		if err != nil || pd != digest {
 			return nil // ignore
 		}
 
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -38,7 +38,7 @@ type manifest struct {
 
 func findManifest(w walker, d *descriptor) (*manifest, error) {
 	var m manifest
-	mpath := filepath.Join("blobs", d.Digest)
+	mpath := filepath.Join("blobs", pathDigest(d.Digest))
 
 	f := func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() {
@@ -101,13 +101,15 @@ func (m *manifest) unpack(w walker, dest string) error {
 			continue
 		}
 
+		pd := pathDigest(d.Digest)
+
 		f := func(path string, info os.FileInfo, r io.Reader) error {
 			if info.IsDir() {
 				return nil
 			}
 
 			dd, err := filepath.Rel("blobs", filepath.Clean(path))
-			if err != nil || d.Digest != dd {
+			if err != nil || pd != dd {
 				return nil // ignore
 			}
 
--- a/image/util.go
+++ b/image/util.go
@@ -1,0 +1,21 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package image
+
+import "strings"
+
+func pathDigest(digest string) string {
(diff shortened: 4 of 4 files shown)


wking:
On Sun, Jun 26, 2016 at 11:43:10PM -0700, Simone Gotti wrote:
>   * image: convert colon to hypen for path digests.

I'd rather address this with #159 (which converts to colons here [1]).
I was waiting on “yeah, the cas/ref engines approach sounds good”
feedback before investing time in a directory-based backend and tests.

[1]: https://github.com/opencontainers/image-spec/pull/159/files#diff-b11cc21c1cf82254929b0e2e98cbd695R56

sgotti:
@wking Sorry I lost that PR. I'll take a deep look at it (since I'm implementing an OCI store derived from the rkt one that is very similar to your interface and adds all the advanced features).

Feel free to close this.

Thanks!

wking:
On Mon, Jun 27, 2016 at 02:02:21PM -0700, Simone Gotti wrote:
> … I'm implementing an OCI store derived from the rkt one that is
> very similar to your interface and adds all the advanced features…

I was curious about your advanced features, so I went hunting for the
similar rkt interface.  The closest thing I can find is [1], but it
seems to require a local disk to back the store and it has a much more
structured view of the stored content than #159's generic Engine
interfaces.

As it stands now, #159 only provides tar-backed implementations of
those interfaces, but if the general idea is approved I intend to add
interfaces for more of the transports listed in [2] (zip, directory,
HTTP, …).  I'm not sure I'm motivated enough to implement Engines
backed by the Docker registry API [3] (I'm not sure it's CAS-y
enough), but the goal with #159 was to separate the engine
implementation from the consumers as thoroughly as possible.

It's good not to go too far in that direction though, so if you have
consumer use-cases or Engine implementation targets that aren't
supported by the generic Engine interfaces I'm adding in #159, please
comment there and tell me ;).

[1]: https://godoc.org/github.com/coreos/rkt/store#Store
[2]: https://github.com/opencontainers/image-spec/blob/ca58cdfc7c8200c6c62287e677fadb0074cc29cc/image-layout.md#open-container-initiative-image-layout-specification
[3]: https://docs.docker.com/registry/spec/api/

sgotti:
@wking yeah, Sorry but I wasn't arguing that your implementation needs advanced features :smile: but that, since I'm writing an oci-store that have advanced features required for an ondisk image cache (https://github.com/sgotti/oci-store) to be used by projects like rkt, with your proposal in #159 it'll be useful, also only as an exercise, to write an additional engine that uses it (so you can do an `oci-image-tool verify` on top it). The unique differences is that the image-layout looks tied to just one image per layout (the ref can be 'latest' 'v1.0.0' etc...) while the oci store is meant to be a multi application store so the ref will be something like `busybox:latest` so the engine should assume a 'default' application name.

If you're interested, as suggested by @runcom and @philips, I opened a topic here: https://groups.google.com/forum/#!topic/containers-dev/HPzM3KrRGbI

> [1]: https://godoc.org/github.com/coreos/rkt/store#Store

Yes that's the current rkt stores but it's very very tied to the appc/spec, that's the reason I'm trying to create a base one to be used by specialized stores like to oci and perhaps appc (v2) store.

wking:
On Thu, Jun 30, 2016 at 01:32:38AM -0700, Simone Gotti wrote:
> The unique differences is that the image-layout looks tied to just
> one image per layout (the ref can be 'latest' 'v1.0.0' etc...) while
> the oci store is meant to be a multi application store so the ref
> will be something like `busybox:latest` so the engine should assume
> a 'default' application name.

The only thing stopping you from using ‘busybox:latest’ as a ref name
in image-layout formats is the charset restriction [1].  You could
certainly put busybox-latest and debian-latest refs in the same
image-layout archive.

And while compatability with image-layout is useful, there's no reason
for all Engine backends to be based on image-layout.  You're free to
use Redis for a refs.Engine, a custom HTTP API for a cas.Engine, etc.

> If you're interested, as suggested by @runcom and @philips, I opened
> a topic here:
> https://groups.google.com/forum/#!topic/containers-dev/HPzM3KrRGbI

Ah, the graphic that you use in [2] looks similar to [3] :).  I hadn't
included caching, and I wasn't aware of the “tree store” concept, but
now I'd suggest the following groupings:

1. Name → descriptor lookup (via “registry” in [3], or via refs.Engine
   in #159).  This mutable stuff is probably not worth caching.

2. Descriptor → CAS drilling (via “content addressable storage” in
   [3], or via cas.Engine in #159).  This is easily cachable, and you
   can layer on as many caching algorithms as you like with something
   like:

     directEngine, err := docker.NewEngine("docker://registry-1.docker.io/")
     cachePath := "/var/lib/cas"
     engine, err := cache.NewEngine(cachePath, directEngine)
     reader, err := engine.Get("sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270")

   where the cache engine will check it's local cachePath for the blob
   first, and only hit directEngine on misses.

3. Unpacking/rendering (“unpack” in [3], “render” in [2]).  The
   interface here can be Unpack(casEngine, destination), although
   pulling that public function out of image/unpack will probably be a
   post-#159 PR.  This can use a tree store to transparently cache
   unpacked trees just like (2) can use cache.NewEngine to
   transparently cache CAS objects.

[1]: https://github.com/opencontainers/image-spec/blob/ca58cdfc7c8200c6c62287e677fadb0074cc29cc/image-layout.md#open-container-initiative-image-layout-specification
[2]: https://docs.google.com/document/d/1dd19t2bMdnRnpdP-eYLzbhPO1NYYP1M1Xo6YQfhAxRw/edit
[3]: https://github.com/opencontainers/image-spec/issues/88#issue-156781487

sgotti:
@wking Thanks for your detailed answer, I have some doubts that I'd like to clarify. Can we move this to the containers-dev@googlegroups.com topic (or the github #88 issue)? It seems to me the better place.

wking:
On Thu, Jun 30, 2016 at 12:03:59PM -0700, Simone Gotti wrote:
> Can we move this to the containers-dev@googlegroups.com topic…

Done.

PR #143, opened 2016-06-15
manifest: use descriptor doc as canonical source

philips:
Right now manifest is relying on the external Docker docs to explain the
digest format. Instead use the OCI Descriptor docs.

Signed-off-by: Brandon Philips <brandon.philips@coreos.com>

PR #167, opened 2016-07-02
manifest: Punt mediaType, size, and digest to descriptor docs

wking:
1149faca (manifest: use descriptor doc as canonical source, 2016-06-15, #143) was a step in this direction.  But we can DRY things up a bit more by punting those three properties entirely and just saying "it's a descriptor", and then listing extensions, if any.


wking:
Also a drive-by tab fix.

vbatts:
seems fine to me

stevvooe:
LGTM

brendandburns:
LGTM

PR #171, opened 2016-07-07
serialization: describe the role of opaque whiteout files

### LINKED ISSUE #144: serialization: clarify situation around opaque
During the discussion on whiteout files in #141 we hit upon the issue of opaques. From @stevvooe's explanation:

I think the only open item is the handling of opaque directories: https://github.com/docker/docker/blob/22aca92ee3503f811f813be4a1349ad8e67fbbdb/daemon/graphdriver/aufs/aufs.go#L369. From there, it looks like they are actually omitted.

In particular, we have the following case, assuming `/foo/bar` exists and has entries:

```
delete  dir  /foo/bar
add     dir  /foo/bar
add     file /foo/bar/data
```

Based on the specification, we get the following in the top-level tar:

```
/foo/.wh.bar
/foo/bar
/foo/bar/data
```

From my understanding of aufs/overlay, this is handled via opq, but these aren't transmitted in layers, currently.

stevvooe:
This is a pass at explaining the concept of opaque files and their role in the specification. Provides recommendations that implementations should generate explicit format, but accept opaques.

cc @vbatts @runcom @philips 

Closes #144

### DIFF
--- a/serialization.md
+++ b/serialization.md
@@ -477,4 +477,76 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
 /etc/.wh.my-app-config
 ```
 
+Whiteout files MUST only apply to resources in lower layers.
+Files that are present in the same layer as a whiteout file can only be hidden by whiteout files in subsequent layers.
+The following is a base layer with several resources:
+
+```
+a/
+a/b/
+a/b/c/
+a/b/c/bar
+```
+
+When the next layer is created, the original `a/b` directory is deleted and recreated with `a/b/c/foo`:
+
+```
+a/
+a/.wh..wh..opq
+a/b/
+a/b/c/
+a/b/c/foo
+```
+
+When processing the second layer, `a/.wh..wh..opq` is applied first, before creating the new version of `a/b, regardless of the ordering in which the whiteout file was encountered.
+For example, the following layer is equivalent to the layer above:
+
+```
+a/
+a/b/
+a/b/c/
+a/b/c/foo
+a/.wh..wh..opq
+```
+
+Implementations SHOULD generate layers such that the whiteout files appear before sibling directory entries.
+
+In addition to expressing that a single entry should be removed from a lower layer, layers may remove all of the children using an opaque whiteout entry.
+An opaque whiteout entry is a file with the name `.wh..wh..opq` indicating that all siblings are hidden in the lower layer.
+Let's take the following base layer as an example:
+
+```
+etc/
+	my-app-config
+bin/
+	my-app-binary
+	my-app-tools
+	tools/
+		my-app-tool-one
+```
+
+If all children of `bin/` are removed, the next layer would have the following:
+
+```
+bin/
+	.wh..wh..opq
+```
+
+This is called _opaque whiteout_ format.
+An _opaque whiteout_ file hides _all_ children of the `bin/` including sub-directories and all descendents.
+Using _explicit whiteout_ files, this would be equivalent to the following:
+
+```
+bin/
+	.wh.my-app-binary
+	.wh.my-app-tools
+	.wh.tools
+```
+
+In this case, a unique whiteout file is generated for each entry.
+If there were more children of `bin/` in the base layer, there would be an entry for each.
+Note that this opaque file will apply to _all_ children, including sub-directories, other resources and all descendents.
+
+Implementations SHOULD generate layers using _explicit whiteout_ files, but MUST accept both.
+
 Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.


serialization.md
```diff
@@ -472,4 +472,37 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
 /etc/.wh.my-app-config
 ```
 
+In addition to expressing that a single entry should be removed from a lower layer, layers may remove all of the children using and opaque whiteout entry.
```
runcom: s/using and opaque/using an opaque/ ?

serialization.md
```diff
@@ -472,4 +472,37 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
...
+Using _explicit whiteout_ files, this would be equivalent to the following:
+
+```
+bin/
+	.wh.my-app-binary
+	.wh.my-app-tools
+```
+
+In this case, a unique whiteout file is generated for each entry.
+If there were more children of `bin/` in the base layer, there would be an entry for each.
```
runcom: what if a child is a directory itself? gonna try that out but would be nice to provide an example if it's not clear instantly (at least to me for now)

stevvooe: Want to try it and dump the example in #144?

My guess is that it is treated as any child.

runcom: if `d` is an empty sub-dir of `/a/b/c` from #144 then we have:
```
drwxr-xr-x 0/0               0 2016-07-08 00:42 a/
drwxr-xr-x 0/0               0 2016-07-08 00:42 a/b/
drwxr-xr-x 0/0               0 2016-07-08 00:42 a/b/c/
---------- 0/0               0 2016-07-08 00:42 a/b/c/.wh.bak
---------- 0/0               0 2016-07-08 00:42 a/b/c/.wh.baz
---------- 0/0               0 2016-07-08 00:42 a/b/c/.wh.boo
---------- 0/0               0 2016-07-08 00:42 a/b/c/.wh.d
-rw-r--r-- 0/0               0 2016-07-08 00:42 a/b/c/foo
```

runcom: but if `d` it's not empty:
```
drwxr-xr-x 0/0               0 2016-07-08 00:45 a/
drwxr-xr-x 0/0               0 2016-07-08 00:45 a/b/
drwxr-xr-x 0/0               0 2016-07-08 00:45 a/b/c/
---------- 0/0               0 2016-07-08 00:45 a/b/c/.wh.bak
---------- 0/0               0 2016-07-08 00:45 a/b/c/.wh.baz
---------- 0/0               0 2016-07-08 00:45 a/b/c/.wh.boo
---------- 0/0               0 2016-07-08 00:45 a/b/c/.wh.d
-rw-r--r-- 0/0               0 2016-07-08 00:45 a/b/c/foo
```
This is the Dockerfile:
```
FROM alpine

RUN mkdir -p /a/b/c && touch /a/b/c/baz  /a/b/c/boo  /a/b/c/bak && mkdir /a/b/c/d && touch /a/b/c/d/ciao
RUN rm -rf /a/b/c && mkdir -p /a/b/c && touch /a/b/c/foo
```

runcom: The same as above applies if any children of `d` is either a file or a directory

runcom: @stevvooe I believe it's clear - just could you provide an example of a contained dir inside `/bin` with a children and notice that only `d` is whiteout'ed if it makes sense to you also.

runcom:
:+1: (IANAM)

serialization.md
stevvooe: Confirmed with AUFS:

```
drwxr-xr-x  0 0      0           0 Jul  7 16:01 a/
drwxr-xr-x  0 0      0           0 Jul  7 16:01 a/b/
drwxr-xr-x  0 0      0           0 Jul  7 16:01 a/b/c/
-r--r--r--  0 0      0           0 Jul  7 16:01 a/b/c/.wh..wh..opq
-rw-r--r--  0 0      0           0 Jul  7 16:01 a/b/c/foo
```

serialization.md
```diff
@@ -472,4 +472,40 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
...
+An opaque whiteout entry is a file with the name `.wh..wh..opq` indicating that all siblings are hidden in the lower layer.
+Let's take the following base layer as an example:
+
+```
+etc/
+	my-app-config
+bin/
+	my-app-binary
+	my-app-tools
+	tools/
```
runcom: mind telling (or actually adding a file under `tools/`) that if `tools/` is not-empty we get the opaque only for the topdir `/bin` (this was my misunderstanding also)

stevvooe: Confirmed the behavior and updated.

runcom: thanks for bearing with me @stevvooe

serialization.md
```diff
@@ -472,4 +472,41 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
...
+Using _explicit whiteout_ files, this would be equivalent to the following:
+
+```
+bin/
+	.wh.my-app-binary
+	.wh.my-app-tools
+	.wh.tools
+```
+
+In this case, a unique whiteout file is generated for each entry.
```
runcom: this can be in misleading with the example above https://github.com/opencontainers/image-spec/pull/171/files#diff-9f7e4d8b174e71a2ac5d4d28c9548f5cR497

runcom: and the line below as well

runcom: no, nvm It's referring to the same example

serialization.md
stevvooe: Thanks for the review. This is a poorly documented topic and we need to get it right.

serialization.md
stevvooe: Could be quote the suspect line or suggest something different?

runcom: I would move `Note that this opaque file will apply to _all_ children, including sub-directories, other resources and all descendents.` before line 510 to make it clear it's valid for both of the whiteout strategies

serialization.md
```diff
@@ -472,4 +472,41 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
...
+```
+
+If we remove all of the children of `bin/` in the next layer, we would have the following:
+
+```
+bin/
+	.wh..wh..opq
+```
+
+We'll call this the _opaque whiteout_ format.
```
vbatts: perhaps the pronoun we and we'll sounds a bit too informal or personal?

serialization.md
stevvooe: I'm adding it in both places. 😎

serialization.md
stevvooe: Pronouns being replaced with passive voice.

stevvooe:

serialization.md
```diff
@@ -472,4 +472,42 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
 /etc/.wh.my-app-config
 ```
 
+In addition to expressing that a single entry should be removed from a lower layer, layers may remove all of the children using an opaque whiteout entry.
```
vbatts: so, perhaps this might more clearly say something like: if the entry is a directory ...  
What do you think? Because the use of the word 'children' only applies if that is the case.
Also, I'm curious (and need to test this), but what would happen if the entry is a symlink to a directory? does this imply that only the symlink be removed?

stevvooe: Symlinks don't have children, by definition, but feel free to test it out.

From the paragraph above, it is fairly clear we are talking about directories. If you insist, I can add something to clarify it.

vbatts: right. And after testing on AUFS, it does not change the resulting opaque file

stevvooe: Great!

stevvooe:
@vbatts @philips Was there any more language to clear up here? If possible, it would be great to get this merged. If we need to make revisions, a secondary PR would be reasonable.

vbatts:
after reviewing this, i see that line 445 is wrong. But seems right in master. Perhaps that's a rebase thing.

stevvooe:
Rebased.

> after reviewing this, i see that line 445 is wrong.

Could you check this and clarify?

vbatts:
ugh. some linter is not being installed correctly? 
https://travis-ci.org/opencontainers/image-spec/builds/148376653#L223
vs. 
https://travis-ci.org/opencontainers/image-spec/builds/148376653#L248

vbatts:
@stevvooe thanks. 

I'm reading over it again.

serialization.md
```diff
@@ -477,4 +477,42 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
...
+	.wh.my-app-binary
+	.wh.my-app-tools
+	.wh.tools
+```
+
+In this case, a unique whiteout file is generated for each entry.
+If there were more children of `bin/` in the base layer, there would be an entry for each.
+Note that this opaque file will apply to _all_ children, including sub-directories, other resources and all descendents.
+
+Implementations SHOULD generate layers using _explicit whiteout_ files but MUST accept both.
```
vbatts: comma after "files"

serialization.md
```diff
@@ -477,4 +477,42 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
...
+
+If all children of `bin/` are removed, the next layer would have the following:
+
+```
+bin/
+	.wh..wh..opq
+```
+
+This is called _opaque whiteout_ format.
+An _opaque whiteout_ file hides _all_ children of the `bin/`. including sub-directories and all descendents.
```
vbatts: unneeded period after `bin/`

vbatts:
one thing not clear here, is the ordering of the entries in a tar archive. Since the first sentence of the doc says an image is an ordered collection, and then the diff_ids is a bottom to top ordering, but here in the whiteouts, it doesn't address ordering.

In a example archive of
```
drwxr-xr-x 0/0               0 2016-07-29 14:08 a/
drwxr-xr-x 0/0               0 2016-07-29 14:08 a/b/
drwxr-xr-x 0/0               0 2016-07-29 14:08 a/b/c/
-rw-r--r-- 0/0               0 2016-07-29 14:08 a/b/c/bar
-rw-r--r-- 0/0               0 2016-07-29 14:08 a/b/c/foo
```

and the next layer is:
```
drwxr-xr-x 0/0               0 2016-07-29 14:08 a/
-r--r--r-- 0/0               0 2016-07-29 14:08 a/.wh..wh..opq
drwxr-xr-x 0/0               0 2016-07-29 14:08 a/b/
drwxr-xr-x 0/0               0 2016-07-29 14:08 a/b/c/
-rw-r--r-- 0/0               0 2016-07-29 14:08 a/b/c/foo
```

that opaque whiteout, must come before the new `a/b/c/foo` entry, otherwise it's removed.
Perhaps we should add an explanation or even example of such?

vbatts:
(also, opened issue to track the travis failure https://github.com/opencontainers/image-spec/issues/185)

stevvooe:
> that opaque whiteout, must come before the new a/b/c/foo entry, otherwise it's removed. Perhaps we should add an explanation or even example of such?

In your example, the whiteout only hides the files _below_ and doesn't apply to siblings. `a/.wh..wh..opq` means hide `a/b` from the last layer but has no bearing on the resources in the _same_ layer.

I've added a section clarifying this, since it is relevant to this PR.

serialization.md
```diff
@@ -477,4 +477,76 @@ The resulting Tar archive for `f60c56784b83` has the following entries:
...
+```
+a/
+a/.wh..wh..opq
+a/b/
+a/b/c/
+a/b/c/foo
+```
+
+When processing the second layer, `a/.wh..wh..opq` is applied first, before creating the new version of `a/b, regardless of the ordering in which the whiteout file was encountered.
+For example, the following layer is equivalent to the layer above:
```
vbatts: woah, really? that is expensive! I would expect that the example below would remove `a/` even after creating the previous file entries

stevvooe: @vbatts So, if you're assembling an image from a set of layers into a directory layout, it must be done this way to replicate the behavior of a union filesystem. I'll need to double check the implementations.

This is cheap for a union filesystem, since you just unpack everything and the union filesystem will apply the modifications correctly.

Double check with @dmcgowan.

dmcgowan: This is correct. An opaque directory in one layer should only hide the lower layer and not apply to the currently layer. If the intent is just to remove the directory, an opaque directory is not needed and the directory can just be marked deleted. The usage of an opaque marker specifies that the intent is to hide the lower layer and allow for files in the current layer.

vbatts:
I didn't get to review before leaving town. Having read through Docker's impl, I'm surprised that I never notices that reconciliation. It seems illogical to allow for non-ordered entries in the tar archive, regardless of how docker's `archive.UnpackLayer()` works.
It's not an impossible problem to solve, just adding complexity and inconsistent expectations for a negative gain.

I other @opencontainers/image-spec-maintainers are fine with this, then LGTM.

We still need #189 merged for tests to pass.

vbatts:
Please rebase

stevvooe:
rebased!

vbatts:
LGTM

philips:
LGTM

philips:
@stevvooe Thank you for the work! And @runcom and @vbatts for the review :)

PR #177, opened 2016-07-22
Fix broken oci-image-tool

runcom:
```sh
$ mkdir fedora-oci
$ skopeo copy docker://fedora oci:fedora-oci
$ mkdir fedora-bundle
$ oci-image-tool create-runtime-bundle --ref latest fedora-oci fedora-bundle
# get most of the errors fixed in this PR
```
Eventually the extraction fails because of permissions but this is expected because `fedora` has `/root` populated while for instance `busybox` doesn't have it so the only option is to unpack as root:
```sh
$ oci-image-tool create-runtime-bundle --ref latest fedora-oci fedora-bundle
[...]
unpacking failed: error extracting layer: unable to open file: open fedora-bundle/rootfs/root/.bash_logout: permission denied
```

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -51,7 +51,7 @@ type config struct {
 
 func findConfig(w walker, d *descriptor) (*config, error) {
 	var c config
-	cpath := filepath.Join("blobs", d.Digest)
+	cpath := filepath.Join("blobs", d.getDigest())
 
 	f := func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() {
--- a/image/config.go
+++ b/image/config.go
@@ -51,7 +51,7 @@ type config struct {
 
 func findConfig(w walker, d *descriptor) (*config, error) {
 	var c config
-	cpath := filepath.Join("blobs", d.getDigest())
+	cpath := filepath.Join("blobs", d.normalizeDigest())
 
 	f := func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() {
@@ -97,13 +97,15 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 
 	var s specs.Spec
 	s.Version = "0.5.0"
+	// we should at least apply the default spec, otherwise this is totally useless
+	s.Process.Terminal = true
 	s.Root.Path = rootfs
 	s.Process.Cwd = "/"
 	if c.Config.WorkingDir != "" {
 		s.Process.Cwd = c.Config.WorkingDir
 	}
 	s.Process.Env = append(s.Process.Env, c.Config.Env...)
-	s.Process.Args = append(s.Process.Env, c.Config.Entrypoint...)
+	s.Process.Args = append(s.Process.Args, c.Config.Entrypoint...)
 	s.Process.Args = append(s.Process.Args, c.Config.Cmd...)
 
 	if len(s.Process.Args) == 0 {
--- a/image/config.go
+++ b/image/config.go
@@ -98,10 +98,17 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 	var s specs.Spec
 	s.Version = "0.5.0"
 	s.Root.Path = rootfs
-	s.Process.Cwd = c.Config.WorkingDir
-	s.Process.Env = append([]string(nil), c.Config.Env...)
-	s.Process.Args = append([]string(nil), c.Config.Entrypoint...)
+	s.Process.Cwd = "/"
+	if c.Config.WorkingDir != "" {
+		s.Process.Cwd = c.Config.WorkingDir
+	}
+	s.Process.Env = append(s.Process.Env, c.Config.Env...)
+	s.Process.Args = append(s.Process.Env, c.Config.Entrypoint...)
 	s.Process.Args = append(s.Process.Args, c.Config.Cmd...)
+
+	if len(s.Process.Args) == 0 {
+		s.Process.Args = append(s.Process.Args, "sh")
+	}
 
 	if uid, err := strconv.Atoi(c.Config.User); err == nil {
 		s.Process.User.UID = uint32(uid)
@@ -118,7 +125,7 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
 
 		s.Process.User.UID = uint32(uid)
 		s.Process.User.GID = uint32(gid)
-	} else {
+	} else if c.Config.User != "" {
 		return nil, errors.New("config.User: unsupported format")
 	}
 
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -33,7 +33,7 @@ type descriptor struct {
 	Size      int64  `json:"size"`
 }
 
-func (d *descriptor) getDigest() string {
+func (d *descriptor) normalizeDigest() string {
 	return strings.Replace(d.Digest, ":", "-", -1)
 }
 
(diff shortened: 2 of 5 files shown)


runcom:
ping @s-urbaniak

runcom:
also the tar extraction is copied from docker `createTarFile` in `pkg/archive` why don't we use that instead?

runcom:
funny go lint failure...

image/config.go
```diff
@@ -51,7 +51,7 @@ type config struct {
 
 func findConfig(w walker, d *descriptor) (*config, error) {
 	var c config
-	cpath := filepath.Join("blobs", d.Digest)
+	cpath := filepath.Join("blobs", d.getDigest())
```
s-urbaniak: let's not do getters

image/descriptor.go
```diff
@@ -32,6 +33,10 @@ type descriptor struct {
 	Size      int64  `json:"size"`
 }
 
+func (d *descriptor) getDigest() string {
```
s-urbaniak: see above, let's avoid getters, see https://golang.org/doc/effective_go.html#Getters

image/descriptor.go
```diff
@@ -32,6 +33,10 @@ type descriptor struct {
...
+	return strings.Replace(d.Digest, ":", "-", -1)
```
s-urbaniak: I'd rethink the method name, maybe `normalizeDigest`.

image/config.go
```diff
@@ -97,12 +97,21 @@ func (c *config) runtimeSpec(rootfs string) (*specs.Spec, error) {
...
+	s.Process.Cwd = "/"
+	if c.Config.WorkingDir != "" {
+		s.Process.Cwd = c.Config.WorkingDir
+	}
+	s.Process.Env = append(s.Process.Env, c.Config.Env...)
+	s.Process.Args = append(s.Process.Args, c.Config.Entrypoint...)
 	s.Process.Args = append(s.Process.Args, c.Config.Cmd...)
 
+	if len(s.Process.Args) == 0 {
+		s.Process.Args = append(s.Process.Args, "sh")
```
s-urbaniak: is `sh` spec'ed as a default?

image/manifest.go
```diff
@@ -148,8 +149,30 @@ loop:
 			return errors.Wrapf(err, "error advancing tar stream")
 		}
 
-		path := filepath.Join(dest, filepath.Clean(hdr.Name))
+		hdr.Name = filepath.Clean(hdr.Name)
+		// After calling filepath.Clean(hdr.Name) above, hdr.Name will now be in
+		// the filepath format for the OS on which the daemon is running. Hence
```
s-urbaniak: what daemon?

s-urbaniak:
Just a couple of nits, thanks :+1: back then I didn't have a reference tool to build those images.

image/descriptor.go
s-urbaniak: this actually doesn't need to be attached as a receiver method on `*descriptor`, but just accept a `string`.

runcom: I like this way since it's something on the descriptor itself...

image/config.go
runcom: yes

runcom:
yet, the fedora image doesn't unpack correctly - should we just use the Docker's `Unpack` from `pkg/archive` instead of this copy-pasted fork which isn't really working?

s-urbaniak:
@runcom sure, if it helps

runcom:
I won't hold this PR though - the `fedora` image fails with `permission denied` because its `/root` is populated while, for instance, `busybox` does not populate it. I think this is good to go for now.

s-urbaniak:
lgtm, once green

image/config.go
runcom: ocitools is also merging a library which will ease all of this and painlessly merge the image's config and the default one.

runcom:
```
image/manifest.go:130::warning: cyclomatic complexity 31 of function unpackLayer() is high (> 20) (gocyclo)
```
can we relax gocyclo? seems like that function is that complex because it needs to be (it's the same in Docker)

image/descriptor.go
```diff
@@ -32,6 +33,10 @@ type descriptor struct {
 	Size      int64  `json:"size"`
 }
 
+func (d *descriptor) normalizeDigest() string {
+	return strings.Replace(d.Digest, ":", "-", -1)
+}
```
wking: This is a dup of #165, and I'd still rather fix it by getting #159 landed ;).

runcom: I see, but `oci-image-tool` isn't working at all w/o this fix and your #159 it's a rather bigger PR with a broader scope, I'd love to have this merged so people can use this tool instead of waiting on #159

s-urbaniak:
@runcom sure, do you mind adding an additional commit, and bump `--cyclo-over=20` in `.tool/lint`?

runcom:
> @runcom sure, do you mind adding an additional commit, and bump --cyclo-over=20 in .tool/lint?

done, thanks @s-urbaniak

image/descriptor.go
wking: On Fri, Jul 22, 2016 at 07:18:32AM -0700, Antonio Murdaca wrote:
> > +func (d *descriptor) normalizeDigest() string {
> > +	return strings.Replace(d.Digest, ":", "-", -1)
> > +}
>
> I see, but `oci-image-tool` isn't working at all w/o this fix and
> your #159 it's a rather bigger PR with a broader scope, I'd love to
> have this merged so people can use this tool instead of waiting on
> #159

I'm not sure what's holding #159 up, it's been out there for a while
;).  So I'm not sure what timeframe to expect before people can use it
(although folks are certainly free to use it now, without waiting on
the branch to land in this repo ;).

runcom: > I'm not sure what's holding #159 up, it's been out there for a while
> ;).  So I'm not sure what timeframe to expect before people can use it
> (although folks are certainly free to use it now, without waiting on
> the branch to land in this repo ;).

I don't know either - just that you PR is 1000+ loc changed - while this one is just pulling in 10 additions (https://github.com/opencontainers/image-spec/pull/177/commits/0e8f74b634b84f56e6fcb46b68971219fd3d9e55) to fix a fatal bug.

I believe this should go in asap. Potentially #159 can take a lot more time to be reviewed (not maintainers reviewed it yet)

wking: On Fri, Jul 22, 2016 at 07:52:32AM -0700, Antonio Murdaca wrote:
> I believe this should go in asap. Potentially #159 can take a lot
> more time to be reviewed (not maintainers reviewed it yet)

I think it would be more efficient to put better abstractions into
place as soon as possible, instead of iterating on an approach we
expect to replace.  But that's a throughput vs. latency issue for the
maintainers to figure out ;).

runcom: the point now is that, unfortunately this tool is broken and it's broken since a lot now probably (https://github.com/opencontainers/image-spec/pull/165 is 28 days old why was that closed)
Anyhow I'm ok either way...

wking: On Fri, Jul 22, 2016 at 08:09:47AM -0700, Antonio Murdaca wrote:
> the point now is that, unfortunately this tool is broken and it's
> broken since a lot now…

It's been broken (on this semicolon point) since it was born, because
the implementation was merged in #82 well before the spec landed in
#94.  That means that nobody has ever used it for spec-compliant
unpacking, which takes the pressure off (in my opinion) for fixing it
ASAP.

vbatts:
LGTM

philips:
LGTM

PR #189, opened 2016-08-01
image: reduce permissions on mkdir

vbatts:
This was causing a lint failure https://travis-ci.org/opencontainers/image-spec/builds/148403656#L247

### DIFF
--- a/.tool/lint
+++ b/.tool/lint
@@ -5,7 +5,7 @@ set -o nounset
 set -o pipefail
 
 if [ ! $(command -v gometalinter) ]; then
-	go get github.com/alecthomas/gometalinter
+	go get -u github.com/alecthomas/gometalinter
 	gometalinter --update --install
 fi
 
@@ -17,6 +17,7 @@ for d in $(find . -type d -not -iwholename '*.git*' -a -not -iname '.tool' -a -n
 		--exclude='schema/fs.go' \
 		--disable=aligncheck \
 		--disable=gotype \
+		--disable=gas \
 		--cyclo-over=35 \
 		--tests \
 		--deadline=10s "${d}"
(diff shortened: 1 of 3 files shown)


vbatts:
ping @opencontainers/image-spec-maintainers so we can clear up travis failures

stevvooe:
@vbatts Shouldn't these honor umask?

vbatts:
that is beside the point. The linter is failing because os.Mkdir of 0777 is
too open

stevvooe:
@vbatts My local umask, for example, is 022. `0777 & (~022) == 0755`. The whole point of `umask` is leave the permissions of created resources up to the user.

Directory permissions levels are based on the design need. In this case, we defer to `0777` to say the user has control with `umask`.

Is this a broken linter or a too permissive directory permissions?

vbatts:
i get umasks. perhaps the linter is strict here, but there is no reason to have a `0777` permission directory.

vbatts:
the gometalinter is using the https://github.com/HewlettPackard/gas tool, which a perms check.
Not sure how to just loosen up that one rule.
@jonboulle @s-urbaniak ideas?

stevvooe:
> i get umasks.

I know. Sorry if it came off that way.

`0755` is probably okay here, but `0700` is way too locked down. There are some use cases where user may want to create `0755` by setting mask to `0022` to allow group read, which is impossible with `0700` base mask for the creation.

Either way, we should be selecting permissions based on use case. A linter isn't the right way to enforce this.

vbatts:
agreed

philips:
@vbatts LGTM after you 0755 the thing; which is the conclusion I think you and @stevvooe came to.

vbatts:
OK, updated. PTAL so we can clear the CI failures.

vbatts:
weird CI failure. Now green.

stevvooe:
LGTM

jonboulle:
lgtm

PR #192, opened 2016-08-04
Typo: remove duplicate given

coolljt0725:
What is the double `given` mean? I'm not sure if it's a typo :)


vbatts:
It is funky English. I think either are technically correct, but a single given is less stuttering.

LGTM

brendandburns:
LGTM

PR #199, opened 2016-08-22
correct blob digest set

xiekeyang:
blob digest seems should be hex format based on sha256.

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -38,7 +38,8 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the `refs` and `blobs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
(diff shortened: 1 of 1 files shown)


image-layout.md
```diff
@@ -38,7 +38,7 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the `refs` and `blobs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
+Object names in the `refs` and `blobs` subdirectories MUST NOT include characters outside of the set of "A" to "F", "a" to "f", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
```
vbatts: then perhaps we ought to have two separate sentences. Because the `blobs/` files will have a _suffix_ that is hexidecimal.
But for the `refs/` they can be the full letter set.

xiekeyang:
@vbatts  Thanks a lot! I separate their description. Please take a look.

vbatts:
LGTM

philips:
LGTM

stevvooe:
Don't we specify this elsewhere?

xiekeyang:
@stevvooe It seems that this is the only specification for refs name and blob name.
Are you suggesting that we should clarify it elsewhere, or should set up a individual chapter to specify it?

stevvooe:
@xiekeyang I thought we could use the digest character set here, from descriptors.md, but this clarification is the right place, since they don't have the same constraints.

Everything looks good here.

PR #200, opened 2016-08-22
correct example in bundle man

xiekeyang:
runc seems should be executed as `run` here.


cyphar:
You might want to make it `runc run` instead.

xiekeyang:
@cyphar You are right! I changed the command to `run`

vbatts:
LGTM

philips:
LGTM

PR #201, opened 2016-08-23
improve walk function on returning result

xiekeyang:
1. walk functions had better to return `nil` in the case of no-error.
2. walk functions may not be defined as variable, but as function argument directly.

### DIFF
--- a/image/config.go
+++ b/image/config.go
@@ -53,15 +53,10 @@ func findConfig(w walker, d *descriptor) (*config, error) {
 	var c config
 	cpath := filepath.Join("blobs", d.normalizeDigest())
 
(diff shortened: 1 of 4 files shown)


philips:
cc @s-urbaniak can you please review?

wking:
Or we can drop the walker-based approach.  #159 has [explicit errors for “blob/ref not found”][1].

[1]: https://github.com/opencontainers/image-spec/pull/159/files#diff-a372a96ac413817b9f8f567e310b60bcR37

s-urbaniak:
lgtm, thanks! It introduces a bit of redundancy, but that is fine with me.

vbatts:
LGTM

philips:
LGTM

PR #206, opened 2016-08-24
spec: split serialization into image and layer

### LINKED ISSUE #128: Split serialization.md into layers and config documents
From the [mailing-list thread on v1.0.0](https://groups.google.com/a/opencontainers.org/d/msg/dev/NPBMTrniG8Q/BS6adHTaAgAJ)

### LINKED ISSUE #196: serialization: clarify order of application of tar entries
We need to clarify that when encountering conflicting tar entries, the latest entry in the layer should be used. This has always been implied but we need to clarify this in the context of opaque layers, as well.

cc @vbatts

stevvooe:

Open after merging this:

- [ ] Introduction to layer format is insufficient (starts with an example) #220
- [ ] Name "serialization" is too general (#212)
- [ ] Media types imply unnecessary hierarchical relationship (#212)
- [ ] Fixup specification validators (part of other PRs) (#212)

Closes #128, ~~Depends on #212~~


serialization.md
```diff
@@ -1,7 +1,7 @@
 # OCI Image Serialization
 
 An *Image* is an ordered collection of root filesystem changes and the corresponding execution parameters for use within a container runtime.
-This specification outlines the format of these filesystem changes and corresponding parameters and describes how to create and use them for use with a container runtime and execution tool.
+This specification outlines the JSON format describing images for use with a container runtime and execution tool and its relationship to filesystem change sets, described in [Layers](layer.md).
 
 ## Terminology
```
wking: While we're splitting things up, do we want to move the terminology into a glossary.md?

serialization.md
```diff
@@ -1,7 +1,7 @@
 # OCI Image Serialization
 
 An *Image* is an ordered collection of root filesystem changes and the corresponding execution parameters for use within a container runtime.
-This specification outlines the format of these filesystem changes and corresponding parameters and describes how to create and use them for use with a container runtime and execution tool.
+This specification outlines the JSON format describing images for use with a container runtime and execution tool and its relationship to filesystem change sets, described in [Layers](layer.md).
```
wking: This paragraph seems like it's either covered in the terminology section below or [in the README][1].  Can we drop the paragraph, [move the terminology to a glossary][2], and get straight into the serialization schema/example?  If that's too much to bite off in this PR, I'm happy to file it in a follow-up PR.

[1]: https://github.com/opencontainers/image-spec/blame/v0.4.0/README.md#L5
[2]: https://github.com/opencontainers/image-spec/pull/206/files#r76134009

serialization.md
```diff
@@ -1,7 +1,7 @@
 # OCI Image Serialization
```
stevvooe: Change to "OCI Image Config".

serialization.md
stevvooe: This is terminology that is specific to the context of this format. Pulling it up would imply a larger context, which is both confusing and incorrect.

serialization.md
stevvooe: Context is king.

serialization.md
wking: On Thu, Aug 25, 2016 at 07:44:39AM -0700, Stephen Day wrote:
> >  ## Terminology
> 
> This is terminology that is specific to the context of this format.

“Layer” is more general than this format, since it's now covered in a
separate file (layer.md).  “Tag” and “Repository” seem like they're
waiting for some future naming spec, and don't seem to be specific to
this file's serialization spec.  If we want to keep a terminology
section for the serialization spec, we should at least restrict it to
serialization-specific terms.

stevvooe: @wking Layer has very specific to this version of the specification and config. For the most part, the meaning of a layer at the higher-level is largely agnostic. At this level, their meaning is narrow and part of their relationship with the image configuration.

Read through it again.

Pulling this detail up doesn't help for contextual understanding or clarity.

Either way, this is massively outside the scope of this PR. You may submit another PR that removes, but I won't LGTM, as it creates confusion, ambiguity and indirection that makes meaning impossible to interpret.

wking: On Fri, Aug 26, 2016 at 03:22:13PM -0700, Stephen Day wrote:
> Either way, this is massively outside the scope of this PR.

I have no problem punting to a follow-up PR.  No need to slow this one
down with the terminology move being contentious.

philips:
Can we make this not depend on #212? Getting this split merged quickly without any major changes is important to ensure this PR doesn't block other stuff.

stevvooe:
@philips They could probably be one PR. If you want me to do a simple split, then PR the fixups listed above, that would be okay, but I really don't want this to be "half-done".

philips:
@stevvooe Your call. It just depends on how much rebasing you want to do. I think moving the files around with zero changes is something we should do first and quickly which will fix any rebasing issues.

stevvooe:
@philips This PR, currently, is as close to zero-changes as possible, so let's merge this one, as is. I'll create issues from the other items here.

Dropped WIP

philips:
LGTM

jonboulle:
lgtm (definitely in need of various improvements per the comments, but let's split this out and follow up)

PR #207, opened 2016-08-24
[ReadMe] Adds boilerplate MUST, MAY, etc. language

RobDolinMS:
Signed-off-by: Rob Dolin <robdolin@microsoft.com>


stevvooe:
LGTM

jonboulle:
lgtm

duglin:
We should probably also include the other boilerplate text to goes with it:
```
An implementation is not compliant if it fails to satisfy one or more of the MUST or REQUIRED requirements for the protocols it implements. An implementation that satisfies all the MUST or REQUIRED and all the SHOULD requirements for its protocols is said to be "unconditionally compliant".
```

vbatts:
@duglin i like that. Open a PR?

vbatts:
LGTM

RobDolinMS:
@duglin The language you mention above is going to be edited in the runtime spec so there is one tier of compliance instead of two.  Once that is done, I recommend the image-spec pick it up.

duglin:
@RobDolinMS okie dokie - just let me know if you want me to do anything - after all, @vbatts wanted a PR :-)

wking:
The runtime-spec compliance language is also being touched by the
in-flight opencontainers/runtime-spec#527, if the plan is to let
runtime-spec settle before adding compliance language here.

Issue #208, opened 2016-08-24
image layout: move to slash separation for blob storage layout


stevvooe:
Currently, blobs are stored with the follow format:

```
blobs/<digest>-<hex bytes>
```

This has caused confusion in the context of the `:`-separated hash format.

To make this more clear, the blobs should be stored in the following format:

```
blobs/sha256/<hex bytes>
```

cc @vbatts @philips

wking:
Or we could use `<algo>-<hex-digest>` for both the blob names and the blob filenames, in which case there would be even less confusion.

I'm ok with using a path separator, but Windows folks would have to be careful to use the appropriate separator for the engine (e.g. `/` for image-layout backed by HTTPS, tar, … and `\` for image-layout backed by local directories).

philips:
+1 to this concept.

glestaris:
That would be nice. We can store multiple images under the same directory structure.

stevvooe:
@xiekeyang @glestaris We have discussed the convention of using a portion of the digest to partition the CAS files but have been unable to find any recent examples of where this actually speeds up retrieval.

Do you have any data supporting this design? What is the actual performance impact?

wking:
On Thu, Aug 25, 2016 at 07:06:12AM -0700, Stephen Day wrote:
> @xiekeyang @glestaris We have discussed the convention of using a
> portion of the digest to partition the CAS files but have been
> unable to find any recent examples of where this actually speeds up
> retrieval.

Previous discussion of hash sharding starting here [1].

[1]: https://github.com/opencontainers/image-spec/pull/94#discussion_r65265938

xiekeyang:
@stevvooe @wking  Got it, thanks a lot!

philips:
@runcom @s-urbaniak could you work-up a patch to make this change in the tooling? And then @stevvooe or I can make the english language change.

stevvooe:
@philips I can take this on!

philips:
@stevvooe are you going to do the code, english or both?

philips:
@runcom volunteered to do the code side of this one.

runcom:
created https://github.com/opencontainers/image-spec/pull/230 for the code side

PR #210, opened 2016-08-24
image-layout: More explicit blob filenames

wking:
There's no reason to allow a blacklisted range of separators; just
require a hyphen.  Similarly, hex has a well defined charset that is
within the previous restrictions without us having to reiterate here.

The current spec wording around digest hash algorithm's isn't all that
strong, but if we have a charset restriction on hash algorithms that
should be documented in the descriptor file.  I've updated that part
of these docs to link to the descriptor file's table of hash algorithm
IDs (currently just 'sha256').

Also replace "Object names" with "filenames", to make the distinction
between blob names like sha256:5b... and filenames like sha256-5b...

This language was last touched by #199, if folks want to look at the
earlier discussion.

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -38,9 +38,8 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the `refs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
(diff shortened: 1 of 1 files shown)


stevvooe:
In general, I like this change.

Do you mind tackling #208 in the process? They are generally related.

wking:
On Fri, Aug 26, 2016 at 04:00:28PM -0700, Stephen Day wrote:
> Do you mind tackling #208 in the process? They are generally related.

My preferred solution to the #208 problem (that mapping <algo>:<hash>
↔ blobs/<algo>-<hash> is confusing) is to use <algo>-<hash> for both
the blob name and the filename [1].  I'm less excited about rolling in
#208's suggested shift to <algo>:<hash> ↔ blobs/<algo>/<hash>, but
while I don't see how blobs/<algo>/<hash> helps reduce confusion, I
don't think it will increase confusion either, so I'm fine rolling it
in.

However, changing the mapping will require patching the Go tooling.
I'd rather patch that tooling after #159 lands, and keep this PR
focused tighter wording for the existing mapping.  If #159 happens to
land before this one, I'll happily roll in either mapping shift.  If
you want to land the mapping shift before #159, I'd rather have the
mapping shift handled by a separate PR.

[1]: https://github.com/opencontainers/image-spec/issues/208#issuecomment-242202032

wking:
The digest-separator ambiguity I was addressing with this PR has been removed by #232.

PR #211, opened 2016-08-26
serialization: remove windows-specific layers+base rootfs

### LINKED ISSUE #186: Remove windows-specific rootfs type from serialization
The following PR is removing support for machine-located windows base layers:

https://github.com/docker/docker/pull/25090/files#diff-b1247b61dc69edc892f3be4ea4f300a6R373

This was a hack that allowed the local container runtime to supply filesystem layers. This needs to be removed from OCI, as well, as this was never intended for long term support and problem won't be supported in OCI.

The use case may be valid, but OCI support should be done very differently.

stevvooe:
The `rootfs.type` field previously allowed a `type` of `layers+base` to
signify that there the layers must be applied to a base image. This was
required to support windows containers in Docker initially but has since
been replaced with a more compatible mechanism. It has no relevence for
OCI.

Closes #186.

### DIFF
--- a/serialization.md
+++ b/serialization.md
@@ -327,7 +327,6 @@ Note: whitespace has been added to this example for clarity. Whitespace is OPTIO
         <ul>
           <li>
             <code>type</code> is usually set to <code>layers</code>.
-	    There is also a Windows-specific value <code>layers+base</code> that allows a base layer to be specified in a field of <code>rootfs</code> called <code>base_layer</code>.
           </li>
           <li>
             <code>diff_ids</code> is an array of layer content hashes (<code>DiffIDs</code>), in order from bottom-most to top-most.


serialization.md
```diff
@@ -327,7 +327,6 @@ Note: whitespace has been added to this example for clarity. Whitespace is OPTIO
         <ul>
           <li>
             <code>type</code> is usually set to <code>layers</code>.
-	    There is also a Windows-specific value <code>layers+base</code> that allows a base layer to be specified in a field of <code>rootfs</code> called <code>base_layer</code>.
```
wking: Are you leaving the “usually” language to allow extensions here to still be valid images?  Do you have a plan for validating unknown values?  It seems like a single option should be:

> `type` MUST be set to `layers`.

That gives you space to extend the type in the future.

But if there's only a single legal value, it seems like a waste of space.  Can't users look at the versioning information (the v1 in `application/vnd.oci.image.serialization.config.v1+json`?) and know that the rootfs type would be “layers” without having to specify it in the config?  Or do you expect future versions of the v1 serliazation spec to add support for additional types?

RobDolinMS:
Tagging @jstarks for a look / FYI

serialization.md
stevvooe: @wking Could you please focus your review of the PR on what can actually be changed? This is not the time for reviewing the docker image format, that is the baseline of the specification and those decisions have already be made.

This field has no relationship to content descriptors present in the manifest layer. This is the internal image structure represented by this particular format. The presence of these field allows it to internally evolve without much effort.

Please keep the scope of your comments focused.

wking: On Mon, Aug 29, 2016 at 04:11:52PM -0700, Stephen Day wrote:
> @wking Could you please focus your review of the PR on what can
> actually be changed?

If the intention here is to only allow the layers-type rootfses, I
don't see a problem with dropping the field from the spec.  You could
handle the drop/restoration with the same code that translates Docker
↔ OCI media types.  Is there a compatibility issue with that that I'm
missing?

> The presence of these field allows it to internally evolve without
> much effort.

So the idea is to provide space for alternative types that still use
the application/vnd.oci.image.serialization.config.v1+json type?
That's fine with me, but I'd rather you changed the current “usually
set” language to explain how users are supposed to handle configs that
use different values.  MUST they error out if asked to unpack an image
whose rootfs.type they don't understand (I expect so)?  Is there any
way to determine if your image-spec implementation is modern enough to
unpack a particular image (I expect not by comparing version numbers,
but you could probably do this by using your tooling to validate the
image)?

stevvooe: > If the intention here is to only allow the layers-type rootfses, I
don't see a problem with dropping the field from the spec.

I'm not sure that you have full knowledge to evaluate the usage of this field, nor is this PR the forum for that discussion.

These are two _completely_ different components operating at separate layers of the specification. I don't know what you want me to say.

philips:
LGTM

serialization.md
wking: On Mon, Aug 29, 2016 at 04:31:23PM -0700, Stephen Day wrote:
> > If the intention here is to only allow the layers-type rootfses, I
> > don't see a problem with dropping the field from the spec.
>
> I'm not sure that you have full knowledge to evaluate the usage of
> this field, nor is this PR the forum for that discussion.

I'm sure you're right and I don't have sufficient background to
evaluate the usage of this field.  I would expect that this spec
(since it defines the field) would provide sufficient background on
how it is expected to be used, and what it means is the value is
‘layers’, or unset, or some other string.  But if that discussion is
supposed to happen somewhere else, I'm happy to have the discussion
there.  Can you point me at the appropriate forum?

stevvooe: @wking This field was added as part of the modifications that came out with Docker 1.10 to address long standing image format issues. The `rootfs` provide image provenance over the root filesystem components that make up a given image. The `type` field provides some measure of algorithmic agility. Removing this field probably has unintended consequences.

Recently, this value for the field was removed from the docker landscape due to a better solution for identifying base image layers that could not be distributed. The field value would never have been seen in production ready versions of docker and has no place in OCI.

The scope of this PR was to remove the value for the field and bring it in line with real world usage. Any discussion around changes to the meaning of the field or whether the field exists at all are outside the scope of this PR.

Keeping the discussion focused will ensure that PRs move more quickly and don't end up in a state where the path forward is unclear.

wking: On Mon, Aug 29, 2016 at 05:41:40PM -0700, Stephen Day wrote:
> Removing this field probably has unintended consequences.

“Unintended consequences” are hard to discuss ;).  Can you float an
example of the sort of consequences you're concerned about?

> Recently, this value for the field was removed from the docker
> landscape due to a better solution for identifying base image layers
> that could not be distributed. The field value would never have been
> seen in production ready versions of docker and has no place in OCI.

Yeah, I think “we want to remove layers+base” is something that we
agree on.

> The scope of this PR was to remove the value for the field and bring
> it in line with real world usage. Any discussion around changes to
> the meaning of the field or whether the field exists at all are
> outside the scope of this PR.
>
> Keeping the discussion focused will ensure that PRs move more
> quickly and don't end up in a state where the path forward is
> unclear.

Touching/removing a particular line seems like a good opportunity to
consider the sense of the surrounding material with less maintainer
context-switching.  But if getting to the bottom of the rootfs.type
field as a whole is too much to bite off in this PR, I'm happy to punt
everything after dropping layers+base to follow-up PRs.  It's possible
that a general increase in formality like that proposed in #220 will
also help clarify rootfs.type as well.

jonboulle:
LGTM

PR #212, opened 2016-08-26
serialization: remove extra "hierarchy" from mediatypes

stevvooe:
The fact that we are serializing is implied. This PR removes the extra
bits in the media type that only serve to confuse and waste bytes. The
new media types better reflect the reality for the usage of each type
and how they relate.

This must be followed up by a split of the serialization document, but
this change _should_ be applied first.

This change is backwards incompatible with previous versions of the OCI
specification. Since we are before 1.0, this is acceptable.

Related to #206.


image-layout.md
```diff
@@ -78,13 +78,13 @@ $ cat ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1
...
   "config": [
-    "mediaType": "application/vnd.oci.image.serialization.config.v1+json",
+    "mediaType": "application/vnd.oci.image.config.v1+json",
     "size": 7023,
     "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"
   },
   "layers": [
     {
-      "mediaType": "application/vnd.oci.image.serialization.rootfs.tar.gzip",
+      "mediaType": "application/vnd.oci.image.layer.tar+gzip",
```
wking: It looks like the `+{suffix}` format is from [RFC 6839][1].  That doesn't have any examples of nested formats, but it doesn't specify a formal syntax for suffixes either.  Since the layer idea doesn't seem particularly tied to tar *or* gzip, I'd rather use `application/vnd.oci.image.layer+tar+gzip` to make it easy to also support `+zip` or `+tar+xz`, etc.

[1]: https://tools.ietf.org/html/rfc6839

wking: Or just `+tar` (with no compression, see [here][1]).

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-08-16.log.html#t2016-08-16T23:52:38

vbatts:
OHMAN GPG SIGNED COMMIT

vbatts:
While i'm not terribly opposed, I'm not sure what harm or confusion having the name stay is?
The biggest beneficial piece of this seems to be the `+gzip` such that it's an optional representation of the artifact.

stevvooe:
@vbatts I'd like to see cruft remove, if we can.

It would be great to get an LGTM here so I can carry on with the split.

philips:
On the call no one really objected to making this change on the call. Saving a few bytes isn't a huge deal in json land but I don't object.

LGTM
(rebase needed, I will LGTM again)

stevvooe:
@philips @vbatts rebase, PTAL

vbatts:
LGTM

philips:
LGTM

Issue #173, opened 2016-07-21
Misleading language in image-layout.md


runcom:
Three issues:

First:
```
Object names in the refs and blobs subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen -, the dot ., and the underscore _.
```
This sentence is wrong because I can have colons `:` into the refs subdirectory to say `./refs/vbatts-app:latest`

Second: 
```
Hash algorithm identifiers containing the colon : will be converted to the hyphen -. For example sha256:5b will map to the layout blobs/sha256-5b.
```
these two sentences don't play nice together. What are we talking about? `refs` naming? both? I can have `./refs/sha256:5b58496` but what does that mean that _the name_ will map to `blobs/sha256-5b58496`? AFAICT the content of that refs which is a descriptor will give us the location in the blobs dirs, not the ref file name.

Third:
```
$ cd example.com/app/
$ find .
```
The above (with `app/`) isn't making clear that an image-layout can carry more than just one app with:
```
./refs/vbatts-app:latest
./refs/alpine:1.0
./refs/ubuntu:sha256-5bfdjsldf
``` 

/cc @vbatts @stevvooe @philips

runcom:
I believe we need to split the `refs` and `blobs` naming paragraph into two pieces - one for `refs` and one for `blobs` because clearly the two have different naming restrictions.

wking:
On Thu, Jul 21, 2016 at 06:44:19AM -0700, Antonio Murdaca wrote:
> This sentence is wrong because I can have colons `:` into the refs
> subdirectory to say `./refs/vbatts-app:latest`

Colons are not portable, and we map them to hyphens [1].  If folks
want more characters in the ref names, I suggest we have Unicode names
that are encoded to UTF-8 and then to filesystem-safe base64 [2].

> ```
> Hash algorithm identifiers containing the colon : will be converted to the hyphen -. For example sha256:5b will map to the layout blobs/sha256-5b.
> ```
> these two sentences don't play nice together. What are we talking
> about? `refs` naming? both?

This is just about the blob filename.  Splitting it into its own
paragraph would help clarify that.

> AFAICT the content of that refs which is a descriptor will give us
> the location in the blobs dirs, not the ref file name.

This is correct.

> ```
> $ cd example.com/app/
> $ find .
> ```
> The above (with `app/`) isn't making clear that an image-layout can carry more than just one app with:
> ```
[...]

Issue #176, opened 2016-07-22
Signed name-assertion objects


wking:
Spun off from #173:

On Thu, Jul 21, 2016 at 04:38:55PM -0700, Stephen Day [wrote](https://github.com/opencontainers/image-spec/issues/173#issuecomment-234416091):

> 1. The trust model for naming must be clarified (how does one trust
>    the names in a tar file?). File names aren't a very good
>    authority if you can't verify the integrity of the underlying
>    storage.

I have some notes on modeling this [here](https://github.com/opencontainers/image-spec/issues/23#issuecomment-212651871), [here](https://github.com/opencontainers/image-spec/issues/23#issuecomment-221432834), and [here](https://groups.google.com/d/msg/containers-dev/HPzM3KrRGbI/QEshLiVSAgAJ), but I think we need the following types:
- `application/vnd.oci.image.named.blob.v1+json`, with properties for a `name` string and a `blob` descriptor, asserting that that name applies to that descriptor.
- `application/vnd.oci.image.signed.blob.v1+json`, with properties for a `blob` descriptor and `signatures` array of descriptors.

The signature payloads would use existing media types like [application/pgp-signature](http://tools.ietf.org/html/rfc3156#section-9) and [application/jose+json](https://tools.ietf.org/html/rfc7515#section-9.2.1).

I think that's enough to get started, and we can talk about [blobs for public keys](https://github.com/ipfs/specs/tree/5173085ad35a5b6121008da61600f85027dc3035/keychain#diagrams), [signing algorithms](https://github.com/ipfs/specs/tree/5173085ad35a5b6121008da61600f85027dc3035/keychain#abstract), [validity schemes](https://github.com/ipfs/specs/tree/5173085ad35a5b6121008da61600f85027dc3035/iprs-interplanetary-record-system#the-objects), etc. later if there is demand for carrying them in-band.

vbatts:
I don't see this conversation having traction or a common goal.

wking:
> I don't see this conversation having traction...

Yeah, over two years without a response sounds like "no traction" to me ;).

> ... or a common goal.

I think havong a standardized way to sign images is a pretty clear goal.  This proposal is one way to do that, and it has the benefit of allowing signed name-assertions to live in CAS where they are easy to mirror and distribute.  #22 has discussion on many alternatives as well.  It would be nice if image-spec maintainers eventually picked one approach and specified it.

Issue #174, opened 2016-07-21
Base64 UTF-8 for image-layout ref names?


wking:
Spun off from #173.

We currently have a [charset restriction on ref names][1], but folks [may][2] [want][3] to use colons, etc.  I don't think folks should care about human-readability of the `refs` filenames, so I'd like to change the mapping from refs to filenames from “they're the same thing” to “the filename is the NFC Unicode that is UTF-8 encoded and then [filesystem-safe Base 64][4] encoded”.  Then folks can put whatever they want in the ref name.

As an example in Python 3:

    >>> import base64
    >>> import unicodedata
    >>> name = 'Ζεύς'
    >>> name_nfc = unicodedata.normalize('NFC', name)
    >>> name_utf8 = name_nfc.encode('UTF-8')
    >>> name_base64 = base64.b64encode(name_utf8, altchars=b'-_')
    >>> print((name, name_nfc, name_utf8, name_base64))
    ('Ζεύς', 'Ζεύς', b'\xce\x96\xce\xb5\xcf\x8d\xcf\x82', b'zpbOtc-Nz4I=')

So the `Ζεύς` ref would be stored in `refs/zpbOtc-Nz4I=`.

There's previous Base 64 discussion (in the context of container IDs) in opencontainers/runc#675 and opencontainers/runc#676.

[1]: https://github.com/opencontainers/image-spec/blame/45e80364014b03a0876aa2bd53e4fddecaead0d7/image-layout.md#L41
[2]: https://github.com/opencontainers/image-spec/issues/173#issue-166822797
[3]: https://groups.google.com/d/msg/containers-dev/HPzM3KrRGbI/-vKVy9tAAgAJ
[4]: https://tools.ietf.org/html/rfc3548.html#section-4

runcom:
> I don't think folks should care about human-readability of the refs filenames

I think naming should be expressed in some way - how to you expect people to talk about a given _image_? "HEY I'm using `./refs/busybox:latest`" over "HEY I'm using `./refs/djEuMA==`"
Also, I'm not sure every tool out there is ready to change to UTF8 for something like image names (but that can be just me) 

I'm not saying I'm against changing the set of allowed chars in `refs` to introduce colons - I'm saying going to human readable to robot-only isn't really what I expect cause I think I'm losing the way to communicate an image name. Is there any other way to say this ref is "busybox:latest" w/o having to name the ref like that?

wking:
[...]

PR #219, opened 2016-08-29
image-layout: move `refs` to `refs/tags`

philips:
Based on the confusion from Antonio in the image-layout in
https://github.com/opencontainers/image-spec/issues/173 we are going to
move everything that was in `refs` to `refs/tags`. This does a few
things:

1. It makes it clearer that this layout is for a single "image name" instead of allowing for lots of different refs.
2. It matches the layout of git so it likely matches existing expectations.

This was also discussed during the OCI F2F and decided to be a
reasonable idea.

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -1,13 +1,13 @@
 ## Open Container Initiative Image Layout Specification
 
-The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
+The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) reference tags (tags).
 This layout MAY be used in a variety of different transport mechanisms: archive formats (e.g. tar, zip), shared filesystem environments (e.g. nfs), or networked file fetching (e.g. http, ftp, rsync).
 Given an image layout a tool can convert a given ref into a runnable OCI Image Format by finding an appopriate manifest from the manifest list, unpacking the filesystem serializations in the correct order, and then converting the image configuration into an OCI Runtime config.json.
 
-The image layout has two top level directories:
+The image layout has two directory structures:
 
 - "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque.
-- "refs" contains [descriptors][descriptors]. Commonly pointing to an image manifest.
+- "refs/tags" contains [descriptors][descriptors]. Commonly pointing to an image manifest. Other subdirectories under `refs/` are reserved for future use.
 
 
 It also contains a file that is used to identify the layout version:
@@ -26,9 +26,9 @@ $ find .
 ./blobs/sha256-5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270
 ./blobs/sha256-e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
 ./oci-layout
-./refs
-./refs/v1.0
-./refs/stable-release
+./refs/tags
+./refs/tags/v1.0
+./refs/tags/stable-release
 ```
 
 Blobs are named by their contents:
@@ -38,21 +38,21 @@ $ shasum -a 256 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935e
 afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256-afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51
 ```
 
-Object names in the `refs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
-Object names in the `blobs` subdirectories are composed of hash algorithm, hash digest and separator. Hash digest MUST NOT include characters outside of the set of "A" to "F", "a" to "f", "0" to "9". Separator MUST NOT include characters outside of the hyphen `-`, the dot `.`, and the underscore `_`.
+Object names in the `refs/tags` directory MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
+Object names in the `blobs` directory are composed of hash algorithm, hash digest and separator. Hash digest MUST NOT include characters outside of the set of "A" to "F", "a" to "f", "0" to "9". Separator MUST NOT include characters outside of the hyphen `-`, the dot `.`, and the underscore `_`.
 Hash algorithm identifiers containing the colon `:` will be converted to the hyphen `-`.
 For example `sha256:5b` will map to the layout `blobs/sha256-5b`.
-The blobs directory MAY contain blobs which are not referenced by any of the refs.
+The blobs directory MAY contain blobs which are not referenced by any of the tags.
 The blobs directory MAY be missing referenced blobs, in which case the missing blobs SHOULD be fulfilled by an external blob store.
 
-No semantic restriction is given for object names in the `refs` subdirectory.
-Each object in the `refs` subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
+No semantic restriction is given for object names in the `refs/tags` subdirectory.
+Each object in the `refs/tags` subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
 In general the `mediatype` of this [descriptor][descriptors] object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
 
 This illustrates the expected contents of a given ref, the manifest list it points to and the blobs the manifest references.
 
 ```
-$ cat ./refs/v1.0
+$ cat ./refs/tags/v1.0
 {"size": 4096, "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f", "mediatype": "application/vnd.oci.image.manifest.list.v1+json"}
 ```
 ```
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -39,7 +39,7 @@ func (d *descriptor) normalizeDigest() string {
 
 func findDescriptor(w walker, name string) (*descriptor, error) {
 	var d descriptor
-	dpath := filepath.Join("refs", name)
+	dpath := filepath.Join("refs", "tags", name)
 
 	f := func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() {


runcom:
Great! I like `refs/tags` also, thx @philips. LGTM (IANAM)

wking:
On Mon, Aug 29, 2016 at 09:44:58AM -0700, Brandon Philips wrote:
> Based on the confusion from Antonio in the image-layout in
> https://github.com/opencontainers/image-spec/issues/173 we are going to
> move everything that was in `refs` to `refs/tags`. This does a few
> things:
> 
> 1. It makes it clearer that this layout is for a single "image name"
>    instead of allowing for lots of different refs.

Previous discussion of this around [1].  And the current spec has (and
I like) [3]:

  No semantic restriction is given for object names in the refs
  subdirectory.

I see no reason to back away from that at the image-layout level.  And
I don't see why there would be a problem with folks keeping *all*
their refs in a single image-layout (although they probably wouldn't
want to back it with tar ;).  For example:

  https://oci.example.com/refs/…
  https://oci.example.com/blobs/…

could have multiple debian:… refs, multiple ubuntu:… refs, multiple
python:… refs, etc.

Folks with special per-blob auth rules might choose to use per-"image
name" image-layout instances, but and there's no reason you *couldn't*
do that now.  But I think we want to keep support for multi-"image
name" image-layout instances too.

> 2. It matches the layout of git so it likely matches existing
>    expectations.

Git also has refs/heads and refs/remotes.  Do we expect image-layout
to need those, or other kinds of refs?  I don't see a use case for
them, excepting possibly a refs/HEAD working like Git's .git/HEAD to
provide a default entry point for folks running an image with multiple
refs but not specifying a ref directly.

[1]: https://github.com/opencontainers/image-spec/issues/173#issuecomment-234407153
     Subject: Misleading language in image-layout.md
[3]: https://github.com/opencontainers/image-spec/blame/09fc80cdf1a538cee333476b5d426becfc88dfa8/image-layout.md#L48

sgotti:
I was also thinking that the oci image layout could also work for multiple images and the ref is not tied to a name (or just a tag) but it was a level below image name. For example you can just create a simple image "registry" using an oci image layout using a filesystem backed by http, nfs, local fs access etc...
Instead retrieving which location and ref refers to `example.com/app:tag` will be an upper layer that does image discovery.

I also tried to explain some possibilities of this approach in the Distribution point concept for rkt (coreos/rkt#2953) where there are two different distribution types: a direct distribution point called oci-image-layout and one or more theoretical (not currently covered by the image spec) indirect (that will do location and ref discovery starting from an image name) oci distributions.

vbatts:
LGTM

philips:
@wking the problem is that we don't allow slashes or colons in ref names to ensure we don't confuse filesystems. Without those separators it is strange to imagine inventing some sort of serialization from the common `debian/base:v0.3.0` pattern or `gcr.io/foo/bar:v1.0` into the refs directory.

wking:
On Mon, Aug 29, 2016 at 01:52:07PM -0700, Brandon Philips wrote:
> @wking the problem is that we don't allow slashes or colons in ref
> names to ensure we don't confuse filesystems. Without those
> separators it is strange to imagine inventing some sort of
> serialization from the common `debian/base:v0.3.0` pattern or
> `gcr.io/foo/bar:v1.0` into the refs directory.

There's a straightforward fix for that in #174.  Even without that
fix, there are still reasonably straightforward ways to put multiple
names in one image-layout (docker/docker#25779).

But regardless, “we're not sure how to name refs when there are
multiple image names in one image-layout” doesn't seem like it
directly motivates “let's add a tags/ directory”.  It sounds more like
motivation for “let's add semantic restrictions to refs”, and I don't
think image-layout is the place to add those restrictions.  I'm fine
with a higher-level policy section in image-spec that adds these
restrictions.

Folks who can find an ref naming scheme that they're comfortable with
([1], #174, docker/docker#25779) should be allowed to pack several
image-names into a single image-layout without breaking image-layout
compliance (although they may break image-spec compliance).

image-layout.md
```diff
@@ -1,13 +1,13 @@
 ## Open Container Initiative Image Layout Specification
 
-The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
+The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) reference tags (tags).
```
stevvooe: Does this mean we are dropping the concept of generic `refs`? I thought we were just calling out a "Reserved" structure within `refs`?

wking: On Mon, Aug 29, 2016 at 02:07:10PM -0700, Stephen Day wrote:
> > @@ -1,13 +1,13 @@
> >  ## Open Container Initiative Image Layout Specification
> >  
> > -The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
> > +The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) reference tags (tags).
>
> Does this mean we are dropping the concept of generic `refs`? I
> thought we were just calling out a "Reserved" structure within
> `refs`?

Ugh.  In order of my decreasing preference:

a. Keep our current generic refs [1].
b. Replace generic refs with refs that have some required tag
   semantics (this PR?).
c. Support *both* generic refs and refs with required tag semantics
   (what you seem to be suggesting with the “Reserved structure”
   comment).

Any type information about refs *should* go in the descriptor media
type field.  I see no reason to bake it into the bucket that holds the
descriptors.

[1]: https://github.com/opencontainers/image-spec/pull/219#issuecomment-243197974

philips: What is the use case of "generic refs"? I thought we were just trying to clear up the use case of the refs directory by making the tags use case explicit through naming, as git does.

The alternative proposal we discussed was just renaming "refs/" to "tags/".

wking: On Mon, Aug 29, 2016 at 02:24:21PM -0700, Brandon Philips wrote:
> What is the use case of "generic refs"?

It allows image-layout users to link link other content if they see a
use case for it (vs. requiring refs to point to a manifest(-list) with
some standard name mapping).  For example, Git has a tagged blob [1],
because that was a convenient way to distribute Junio's public key.

Having multiple image-names in one image-layout is a similar use-case
that I think we want to allow at the imae-spec level.

[1]: https://git.kernel.org/cgit/git/git.git/tag/?h=junio-gpg-pub

philips: How is calling this a "tag" stopping you from having multiple image-names? I view this as similar to how GitHub has the gh-pages branch. We aren't restricting the use cases but trying to clearup the intended use case.

wking: On Mon, Aug 29, 2016 at 02:47:15PM -0700, Brandon Philips wrote:
> How is calling this a "tag" stopping you from having multiple
> image-names?

Ah.  If that's the intention, then I'd suggest the refs/ → tags/
rename [1] if you need to do anything at all.  Introducing
“references” and “tags” as two ways of talking about the same thing
(location-addressed Descriptors) seems like an unnecessary source of
confusion.

And it means we keep the “allowing for lots of different refs” case
you call out in [2] but don't address in the bd370ef2 diff.

I don't see much point in churning the name, but I'm ok with name
churn as long as image-layout keeps the position that refs (or tags,
or whatever you want to call the mutable, location-addressable
entries) are have no semantics outside of what's carried by the
Descriptor.

[1]: https://github.com/opencontainers/image-spec/pull/219/files#r76690262
[2]: https://github.com/opencontainers/image-spec/pull/219#issue-173818111

philips: I don't know. I simply don't understand why having a directory called refs/tags is confusing. Honestly, I am fine with the refs name too; this was mostly to help clear up @runcom's confusion and it seemed reasonable.

What is your use case for having lots of refs pointing to different images in a single image-layout?

The desired UX for this stuff, I think, is like:

```
rkt image import --name mystuff.com/foo/bar /path/to/oci-image-layout
docker import --name mystuff.com/foo/bar /path/to/oci-image-layout
```

In these cases the name and tags make sense. What is your use case?

wking: On Mon, Aug 29, 2016 at 03:02:46PM -0700, Brandon Philips wrote:
> I don't know. I simply don't understand why having a directory
> called refs/tags is confusing.

You'll have people talking about refs and people talking about tags
and people wondering if there's some distinction between them (which
it sounds like there isn't).

> What is your use case for having lots of refs pointing to different
> images in a single image-layout?

I'm publishing a bunch of images on example.com, and don't want to be
bothered with maintaining multiple image-layout instances [1].  Or I
want a convenient way to distribute my public key along with the
images it signs [2].  And supporting generic refs protects
image-layout implementations from having to worry about scope creep
and such if a more opinionated refs spec decided that it needed to
support something different from the narrow use case it initially
picked.

> rkt image import --name mystuff.com/foo/bar /path/to/oci-image-layout
> docker import --name mystuff.com/foo/bar /path/to/oci-image-layout
> 
> In these cases the name and tags make sense. What is your use case?

I'd rather keep both explicit:

  $ my-tool import LAYOUT-PATH REF [NAME]

unless we decide to support something like HEAD [1], in which case REF
would also be optional.  If name was not specified on the command
line, the name used for the imported object would come from some
verifiable name (e.g. using the Merkle objects discussed in #176, see
[3]).

[1]: https://github.com/opencontainers/image-spec/pull/219#issuecomment-243197974
[2]: https://github.com/opencontainers/image-spec/pull/219#discussion_r76691546
[3]: https://github.com/opencontainers/image-spec/issues/173#issuecomment-234640781

philips: On Mon, Aug 29, 2016 at 3:16 PM W. Trevor King <notifications@github.com>
wrote:

> I'd rather keep both explicit: $ my-tool import LAYOUT-PATH REF [NAME]
> unless we decide to support something like HEAD [1], in which case REF
> would also be optional. If name was not specified on the command line, the
> name used for the imported object would come from some verifiable name
> (e.g. using the Merkle objects discussed in #176
> <https://github.com/opencontainers/image-spec/issues/176>, see [3]). [1]: #219
> (comment)
> <https://github.com/opencontainers/image-spec/pull/219#issuecomment-243197974>
> [2]: #219 (comment)
> <https://github.com/opencontainers/image-spec/pull/219#discussion_r76691546>
> [3]: #173 (comment)
> <https://github.com/opencontainers/image-spec/issues/173#issuecomment-234640781>
>

Eh, I think that it is going to be more common for someone to download "an
entire image named foo" then wanting to import the whole thing and expect
the tags to show back up rather than forcing the user to specify every ref.
For example you would actually need to do:

my-tool import /path/to/layout my-img-foo-bar-v1.0.0 my.img/foo/bar v1.0.0
my-tool run my.img/foo/bar:v1.0.0

vs

my-tool import /path/to/layout  my.img/foo/bar
my-tool run my.img/foo/bar:v1.0.0

I feel there is a usability cost to supporting lots of imgs in a single
layout on this common import UX.

wking: On Mon, Aug 29, 2016 at 03:24:32PM -0700, Brandon Philips wrote:
> Eh, I think that it is going to be more common for someone to download "an
> entire image named foo" then wanting to import the whole thing and expect
> the tags to show back up rather than forcing the user to specify every ref.
> For example you would actually need to do:
> 
> my-tool import /path/to/layout my-img-foo-bar-v1.0.0 my.img/foo/bar v1.0.0
> my-tool run my.img/foo/bar:v1.0.0
> 
> vs
> 
> my-tool import /path/to/layout  my.img/foo/bar
> my-tool run my.img/foo/bar:v1.0.0
> 
> I feel there is a usability cost to supporting lots of imgs in a single
> layout on this common import UX.

So add an:

  $ my-tool import-all LAYOUT-PATH

that imports all the refs (still getting names from the CAS blobs
[1]).  Or, without a HEAD spec, make:

  $ my-tool import LAYOUT-PATH

(without an explicit REF) import all the refs.

Specifying an unverified name on the command line or extracting
information from *the ref name* (vs. the referenced CAS content)
should not be part of your main UX.

[1]: https://github.com/opencontainers/image-spec/pull/219#discussion_r76698336

philips: How would you get a name from a CAS blob?

wking: On Mon, Aug 29, 2016 at 04:19:33PM -0700, Brandon Philips wrote:
> How would you get a name from a CAS blob?

See #176 and [1].

[1]: https://github.com/opencontainers/image-spec/issues/173#issuecomment-234640781

philips:
I was perfectly fine with the status quo of having refs be flat with no implied use case. If we can just add some clarifying language about them mapping to tags in a general case for import/export from docker engines would that be sufficient @runcom ?

runcom:
@philips, yes that was my plan at the beginning (just a language issue)

philips:
I am going to close this out and open something that tries to clear up the language issue instead.

vbatts:
@philips was a new issue opened for this yet?

wking:
On Thu, Sep 15, 2016 at 07:00:39AM -0700, Vincent Batts wrote:
> @philips was a new issue opened for this yet?

#225, since merged.

vbatts:
nah. the use of `./refs/tags/foo`

wking:
On Thu, Sep 15, 2016 at 07:30:55AM -0700, Vincent Batts wrote:
> nah. the use of `./refs/tags/foo`

Why would you do that?  Can't higher levels pick a convention like
tags-foo just like [1] if they're worried about mixing different sorts
of refs?

[1]: https://github.com/opencontainers/image-spec/issues/174#issuecomment-247168427

vbatts:
like has been mentioned before, for the same reason git does it. a tag is just one reference type. there could be others. this is the same thing i said yesterday on the call.

wking:
On Thu, Sep 15, 2016 at 07:49:49AM -0700, Vincent Batts wrote:
> like has been mentioned before, for the same reason git does it. a
> tag is just one reference type. there could be others.

And higher levels can namespace as many as they like using the
existing set of allowed characters.  Why special case a slash for this
one class of refs?

vbatts:
sometimes i'm not even sure the point you're trying to make. Honestly.

wking:
On Thu, Sep 15, 2016 at 07:57:28AM -0700, Vincent Batts wrote:
> sometimes i'm not even sure the point you're trying to make. Honestly.

We currently have one class for all refs.  Higher levels which want to
divide that into multiple classs can already namespace however they
wish as suggested here [1].  That approach works for “I want a
subclass for just base-64-encoded refs” (you prefix with
‘org.prefix.base64-’ or ‘base64-’ or whatever).  It also works for “I
want a subclass for just tags” (you prefix with ‘org.prefix.tag-’ or
‘tag-’ or whatever).  This is the same approach taken by RFC 5890 with
A labels (prefixed with ‘xn--’) [2].

If you instead bless a subclass with a slashed name (like ‘tags/’)
associated tooling and APIs need to grow to handle the additional
class.  That seems like more complication then it's worth, when you
could accomplish the same goal by defining a ‘tag-’ prefix and make it
easy for tooling to handle your refs as opaque bytes without having to
grow special slash handling.

So I'm suggesting we either leave the spec as it is on this point or
define a ‘tag-’ prefix, but that we don't define a ‘tag/’ prefix.

[1]: https://github.com/opencontainers/image-spec/issues/174#issuecomment-247168427
[2]: https://tools.ietf.org/html/rfc5890#section-2.3.1

PR #223, opened 2016-08-30
serialization: Drop "Tag" and "Repository" terms

wking:
These are not used anywhere else in this repository at the moment, and we can always dig them back out of the Git history if/when we need them for naming.


philips:
LGTM

stevvooe:
LGTM

PR #225, opened 2016-08-31
image-layout: add explanation of the refs directory use case for tags

philips:


philips:
cc @runcom

wking:
79d0c7f5 looks fine to me.

runcom:
LGTM (IANAM)

vbatts:
LGTM

stevvooe:
@philips Are we not standardizing on the `refs/tags` convention?

philips:
@stevvooe If you want to try and do that. I couldn't articulate a good reason for refs/tags besides "future use". So, I am fine with this as-is as long as an explanation is clear enough for @runcom who originally was confused around the use-case of refs.

stevvooe:
LGTM

PR #227, opened 2016-08-31
fix tag regexp in serialization.md

x1022as:
tag should match 'a-z', 'A-Z', '0-9', '.' and '-'

Signed-off-by: Deng Guangxing <dengguangxing@huawei.com>


philips:
I think we should just drop tags instead https://github.com/opencontainers/image-spec/pull/223

x1022as:
@philips sorry for missing 223, will close this :)

PR #233, opened 2016-09-01
layers: describe policy enforcement for foreign layer media type

stevvooe:

Supersedes #216, Requires #169.

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -270,12 +270,12 @@ Implementations SHOULD generate layers using _explicit whiteout_ files, but MUST
 
 Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.
 
-# Foreign Layers
+# Non-Distributable Layers
 
 Certain layers, due to legal requirements, may not be regularly distributable.
 Typically, such layers are downloaded directly from a distributor but are never uploaded.
 
-Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.foreign.tar+gzip`.
+Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.nondistributable.tar+gzip`.
 [Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
 It is implementation defined whether or not implementations upload layers tagged with this media type.
 
--- a/layer.md
+++ b/layer.md
@@ -270,4 +270,13 @@ Implementations SHOULD generate layers using _explicit whiteout_ files, but MUST
 
 Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.
 
+# Foreign Layers
+
+Certain layers, due to legal requirements, may not be regularly distributable.
+Typically, such layers are downloaded directly from a distributor but are never uploaded.
+
+Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.foreign.tar+gzip`.
+[Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
+It is implementation defined whether or not implementations upload layers tagged with this media type.
+
 [tar-archive]: https://en.wikipedia.org/wiki/Tar_(computing)
--- a/media-types.md
+++ b/media-types.md
@@ -7,7 +7,7 @@ The following `mediaType` MIME types are used by the formats described here, and
 - `application/vnd.oci.image.manifest.v1+json`: [Image manifest format](manifest.md#image-manifest)
 - `application/vnd.oci.image.config.v1+json`: [Container config JSON](serialization.md)
 - `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
-- `application/vnd.oci.image.layer.foreign.tar+gzip`: ["Layer", as a gzipped tar that has distribution restrictions](layer.md#foreign-layers)
+- `application/vnd.oci.image.layer.nondistributable.tar+gzip`: ["Layer", as a gzipped tar that has distribution restrictions](layer.md#non-distributable-layers)
 
 ## Compatibility Matrix
 
--- a/media-types.md
+++ b/media-types.md
@@ -5,8 +5,9 @@ The following `mediaType` MIME types are used by the formats described here, and
 - `application/vnd.oci.descriptor.v1+json`: [Content Descriptor](descriptor.md)
 - `application/vnd.oci.image.manifest.list.v1+json`: [Manifest list](manifest.md#manifest-list)
 - `application/vnd.oci.image.manifest.v1+json`: [Image manifest format](manifest.md#image-manifest)
-- `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
 - `application/vnd.oci.image.config.v1+json`: [Container config JSON](serialization.md)
+- `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
+- `application/vnd.oci.image.layer.foreign.tar+gzip`: ["Layer", as a gzipped tar that has distribution restrictions](layer.md#foreign-layers)
 
 ## Compatibility Matrix
 


philips:
@stephenrwalli Can you take a look at this language? I am still unclear if "MUST NOT distribute" is something we can define usefully in a spec.

I totally get it as a human being with context on copyrights for proprietary software about the usefullness. I just don't know how to codify this. It is like the evil bit RFC: https://www.ietf.org/rfc/rfc3514.txt

layer.md
```diff
@@ -156,3 +156,13 @@ Note that this opaque file will apply to _all_ children, including sub-directori
...
 
 Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.
+
+# Foreign Layers
+
+Certain layers, due to legal requirements, may not be regularly distributable.
+Typically, one can download such layers but they should never be uploaded.
+
+Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.serialization.rootfs.foreign.tar.gzip`.
+[Descriptors][descriptor.md] referencing these layers MAY include `urls` for downloading these layers.
```
wking: `[Descriptors][descriptor.md]` → `[Descriptors](descriptor.md)`.

layer.md
```diff
@@ -156,3 +156,13 @@ Note that this opaque file will apply to _all_ children, including sub-directori
 Implementations SHOULD generate layers using _explicit whiteout_ files, but MUST accept both.
 
 Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.
+
+# Foreign Layers
+
+Certain layers, due to legal requirements, may not be regularly distributable.
+Typically, one can download such layers but they should never be uploaded.
+
+Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.serialization.rootfs.foreign.tar.gzip`.
```
wking: Pre-#212 media type pattern?

stevvooe: Yes, good catch!

media-types.md
```diff
@@ -5,8 +5,9 @@ The following `mediaType` MIME types are used by the formats described here, and
 - `application/vnd.oci.descriptor.v1+json`: [Content Descriptor](descriptor.md)
 - `application/vnd.oci.image.manifest.list.v1+json`: [Manifest list](manifest.md#manifest-list)
 - `application/vnd.oci.image.manifest.v1+json`: [Image manifest format](manifest.md#image-manifest)
-- `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
 - `application/vnd.oci.image.config.v1+json`: [Container config JSON](serialization.md)
+- `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
+- `application/vnd.oci.image.layer.foreign.tar+gzip`: ["Layer", as a gzipped tar that should never be pushed](layer.md#foreign-layers)
```
jlb13: Since a foreign layer must not be pushed per the layer spec, might this media type be better defined as "must not be pushed" rather than "should never be pushed"?

Possibly better to redefine "as a non-redistributable gzipped tar archive", or similar.

jonboulle:
-1 on the underspecified use of the "push" verb, for the same reasons captured in https://github.com/opencontainers/image-spec/pull/216#issuecomment-243845627

stevvooe:
@jonboulle It is sufficient to comment inline on the specific language that needs correcting rather than saying "-1". It is much more constructive, unless you intend to block the PR.

I've replaced the term "push" with the more generic "distribute".

stephenrwalli:
@philips I understand your example. It's not necessarily how you would codify it, so much as it gives the certification process the language to revoke a certification based on non-compliance. While it might be difficult to write a test to prove the MUST NOT, if an implementation was demonstrated to distribute such a layer, it clearly doesn't do what the specification says, and would not be able to claim conformance/compliance. Likewise, for a certification process, one can design the process to manual test such a thing or to use the condition to revoke a certification until the implementation complies. Make sense?  (Sorry for the tardy reply.)

wking:
On Tue, Sep 06, 2016 at 01:58:50PM -0700, Stephen Walli wrote:
> While it might be difficult to write a test to prove the MUST NOT,
> if an implementation was demonstrated to distribute such a layer, it
> clearly doesn't do what the specification says, and would not be
> able to claim conformance/compliance. Likewise, for a certification
> process, one can design the process to manual test such a thing or
> to use the condition to revoke a certification until the
> implementation complies.

“Yeah, automatically testing that would be hard, but we can manually
test it” is a marker for a wishy definition [1].  I think the problem
here is that “distributing” is not something that one instance can do
by itself.  If it's pushing a blob, someone else must be receiving it.
In some cases (e.g. pushing to a local disk on a single-user system),
that probably doesn't count as “distributing” in a legal sense.  In
others (e.g. pushing to a public repo on hub.docker.com) it does.
Where does caching to disk on a multi-user system fall in that
continuum?  Are you distributing it by saving to a system-readabale
cache?  Or is it only distributing if another user on the system reads
it out of that cache?  If that system-readable cache happens to live
under your home directory, are you being robbed (and not distributing)
when the other user reads the file?

I imagine there are consensus opinions on at least some of that
(although I'm not aware of them).  I don't think we want to make it
the business of spec implementers and validators to independently
figure out what those opinions are.  If there are conditions under
which you would certify (or not) an implementation, I'd like to see
those conditions spelled out in the spec in enough detail for them to
be objectively tested without appealing to “The average person,
applying local community standards, looking at the work in its
entirety” and such.  In some cases, that wishyness is unavoidable.
But we need something more solid if we're expecting implementation
software to act consistently based on a ‘foreign’ boolean.

One option would be to have a CAS engine API (e.g. #159) where the
public/private-ness of the engine was part of its initialization.
E.g.:

  casEngine, err := NewEngine(ctx, path, public)

Then the engine is non-compliant if you can successfully:

  casEngine.Put(ctx, reader, mediaType)

with a foreign mediaType (this is testable by the OCI).  And the
engine-initializer is breaking the law if they create a CAS engine
pointing at a public hub.docker.com repo and claiming it is not public
(this is between the CAS engine configurer and the copyright holders
of the pushed content, and not something the OCI has to certify).

[1]: https://en.wikipedia.org/wiki/I_know_it_when_I_see_it
[2]: https://github.com/opencontainers/image-spec/pull/216#issuecomment-243866928

philips:
cc @runcom @vbatts ?

runcom:
I'm not sure either on this one - is it an hard requirement for this spec to outline the legal matters of foreign layers? I (sorry I may be wrong) don't think so. I think as foreign layers just layers that are downloaded through a list of urls in a descriptor. Do we really need to clarify that those layers can't be distributed? I understand this was born in Docker for this exact legal reason but I don't think the spec should enforce this restriction (not to push) - if people end up pushing the layer it's them breaking any legal terms, I'm not sure what the spec has to do with this. (I believe it's perfectly possible to upload the Windows base layer by taking it and uploading it, implementers (Docker) enforce this, can we just add an implementor note?)

(sorry if any of the above makes no sense...)

stephenrwalli:
You make good sense, @runcom. So a way to call attention to such things, that I've seen in the past, is to explicitly call something out as implementation defined. This is basically a warning to spec implementers that are not part of the immediate spec community to pay attention in the space before naively doing something in their implementation. It also allows for the behaviors that already exist in implementations. So in this case, wording becomes: "It is implementation defined whether or not implementations distribute layers tagged with this media type."

jonboulle:
Such a phrasing would definitely assuage my concerns about the vagueness of "pushing", "distributing", "distributable", etc. (My wariness is largely around combining such terms with the keyword MUST without being _very_ explicit about what they mean.)

runcom:
@stephenrwalli that sounds perfect to me if people agree

layer.md
```diff
@@ -156,3 +156,13 @@ Note that this opaque file will apply to _all_ children, including sub-directori
 Implementations SHOULD generate layers using _explicit whiteout_ files, but MUST accept both.
 
 Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.
+
+# Foreign Layers
+
+Certain layers, due to legal requirements, may not be regularly distributable.
+Typically, one can download such layers but they should never be uploaded.
```
wking: @RobDolinMS is going to quibble with this non-SHOULD should (e.g. opencontainers/runtime-spec#536 ;).  How about dropping this line, since the idea is covered in the rest of this section?

wking:
On Wed, Sep 07, 2016 at 10:23:08AM -0700, Stephen Walli wrote:
> So in this case, wording becomes: "It is implementation defined
> whether or not implementations distribute layers tagged with this
> media type."

This works for me too.

layer.md
stevvooe: This is an important clarification and frames the stronger language below. I can capitalize the SHOULD or re-phrase it.

wking: On Wed, Sep 07, 2016 at 11:43:09AM -0700, Stephen Day wrote:
> > @@ -156,3 +156,13 @@ Note that this opaque file will apply to _all_ children, including sub-directori
> >  Implementations SHOULD generate layers using _explicit whiteout_ files, but MUST accept both.
> >  
> >  Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.
> > +
> > +# Foreign Layers
> > +
> > +Certain layers, due to legal requirements, may not be regularly distributable.
> > +Typically, one can download such layers but they should never be uploaded.
>
> This is an important clarification and frames the stronger language
> below.

What is this line adding beyond the earlier:

  Certain layers, due to legal requirements, may not be regularly distributable.

and the planned [1]:

  It is implementation defined whether or not implementations
  distribute layers tagged with this media type.

Can you pin down the information the other lines are missing that
you're trying to convey here?

[1]: https://github.com/opencontainers/image-spec/pull/233#issuecomment-245354663

runcom: I agree with @wking to drop this line and favor the new one proposed

wking:
On Wed, Sep 07, 2016 at 10:23:08AM -0700, Stephen Walli wrote:
> So in this case, wording becomes: "It is implementation defined
> whether or not implementations distribute layers tagged with this
> media type."

It feels awkard to repeat “implementation”.  How about:

  It is unspecified whether or not implementations distribute layers
  tagged with this media type.

?

stephenrwalli:
I believe the suggested text becomes: 

```
Foreign Layers
Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.foreign.tar+gzip`.
 [Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
It is implementation defined whether or not implementations upload layers tagged with this media type.
```

stephenrwalli:
@wking I appreciate the awkwardness of language, but I'm using very specific phrasing. "Unspecified" says you can do anything. "Implementation defined" is that bit stronger saying implementations may already do things here and you should pay attention to your choices.

philips:
LGTM

wking:
On Wed, Sep 07, 2016 at 01:56:09PM -0700, Stephen Walli wrote:
> @wking I appreciate the awkwardness of language, but I'm using very
> specific phrasing. "Unspecified" says you can do
> anything. "Implementation defined" is that bit stronger saying
> implementations may already do things here and you should pay
> attention to your choices.

Interesting.  It looks like C defines those terms [1,2].  If we are
going to split those hairs (and it sounds like a good idea to me), I'd
recommend we put something like that in our spec or reference another
spec to define them for us.  Ideally the latter.  Is there an
equivalent to RFC 2119 for these terms?

[1]: http://stackoverflow.com/questions/2397984/undefined-unspecified-and-implementation-defined-behavior/14487077#14487077
[2]: http://www.open-std.org/jtc1/sc22/wg14/www/C99RationaleV5.10.pdf
     Page 18 of the PDF, page 11 of the document.

runcom:
Are we sticking with the current language?

stevvooe:
Updated with the new language!

layer.md
```diff
@@ -156,3 +156,13 @@ Note that this opaque file will apply to _all_ children, including sub-directori
...
+
+Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.foreign.tar+gzip`.
+[Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
+It is implementation defined whether or not implementations upload layers tagged with this media type.
```
runcom: s/upload/distribute?

stevvooe: @runcom I think there were concerns about the meaning of distribute. I'll change it to that if we can re-obtain consensus.

@stephenrwalli CC

runcom: @stevvooe sorry for the noise, I re-read and you're right..

wking:
On Wed, Sep 07, 2016 at 02:40:52PM -0700, Stephen Day wrote:
> Updated with the new language!

I still think we want to drop the “should never be uploaded” line [1],
and would like to define “implementation defined” [2].  But those can
be addressed in follow-up work if they're too contentious to handle
now, and the rest of fbc6ce4d looks good to me.

[1]: https://github.com/opencontainers/image-spec/pull/233/files#r77873139
[2]: https://github.com/opencontainers/image-spec/pull/233#issuecomment-245418703

stevvooe:
@wking The use of "should" in that sentence applies to generalizing the use case scenarios and does not apply to the implementations of the specification.

wking:
On Wed, Sep 07, 2016 at 03:45:22PM -0700, Stephen Day wrote:
> @wking The use of "should" in that sentence applies to generalizing
> the use case scenarios and does not apply to the implementations of
> the specification.

azurestandard/beehive#536 did not touch any shoulds applying to
implementations either.  And if we aren't clear on what the line is
adding that's not covered somewhere else in that section [1], I don't
see a reason to add it.

[1]: https://github.com/opencontainers/image-spec/pull/233#discussion_r77884942

philips:
LGTM

philips:
@opencontainers/image-spec-maintainers please review and LGTM

wking:
fbc6ce4 → edddc0a9 keeps the never-uploaded line, which addresses the
“should” issue [1] without dropping the line.  I'd still rather drop
the line, but am ok with the current phrasing.

And I'd still like to define “implementation defined” [2], but can
address that in follow up work.

So edddc0a9 looks good enough to me.

[1]: https://github.com/opencontainers/image-spec/pull/233#discussion_r77873139
[2]: https://github.com/opencontainers/image-spec/pull/233#issuecomment-245418703

stevvooe:
Pushed e29e122 to deal with whitespace CI errors.

wking:
On Thu, Sep 08, 2016 at 12:45:32PM -0700, Stephen Day wrote:
> Pushed e29e122 to deal with whitespace CI errors.

This looks good enough to me too [1].

philips:
LGTM

[Approved with PullApprove](https://pullapprove.com/opencontainers/image-spec/pull-request/233/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/image-spec)

jonboulle:
after sleeping on this - I really don't like the term `foreign` - at best it's non-obvious what it means and at worst I think it implies the layer _is_ from a "remote"/unknown source. can someone explain it to me (and spec readers generally) in a way that's more obvious? otherwise, how about any of `local`, `internal`, `private`, `nondistributable`, `unpublishable`, ...

wking:
On Fri, Sep 09, 2016 at 12:39:32AM -0700, Jonathan Boulle wrote:
> … `nondistributable`, …

I like this one, and it echos the phrasing we're adding in the lead-in
line here [1].

[1]: https://github.com/opencontainers/image-spec/pull/233/commits/e29e1225e6347e359622c1707998b491f13a75e6#diff-dc2bb3b4eddcffe96e682193926e8ff6R162

media-types.md
```diff
@@ -5,8 +5,9 @@ The following `mediaType` MIME types are used by the formats described here, and
 - `application/vnd.oci.descriptor.v1+json`: [Content Descriptor](descriptor.md)
 - `application/vnd.oci.image.manifest.list.v1+json`: [Manifest list](manifest.md#manifest-list)
 - `application/vnd.oci.image.manifest.v1+json`: [Image manifest format](manifest.md#image-manifest)
-- `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
 - `application/vnd.oci.image.config.v1+json`: [Container config JSON](serialization.md)
+- `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
+- `application/vnd.oci.image.layer.foreign.tar+gzip`: ["Layer", as a gzipped tar that MUST NOT be distributed](layer.md#foreign-layers)
```
jonboulle: The MUST here seems to be contradicting line 167 of the other doc. Can we remove it or maybe change it to MAY?

wking: On Fri, Sep 09, 2016 at 01:13:31AM -0700, Jonathan Boulle wrote:
> >  - `application/vnd.oci.image.config.v1+json`: [Container config JSON](serialization.md)
> > +- `application/vnd.oci.image.layer.tar+gzip`: ["Layer", as a gzipped tar archive](layer.md)
> > +- `application/vnd.oci.image.layer.foreign.tar+gzip`: ["Layer", as a gzipped tar that MUST NOT be distributed](layer.md#foreign-layers)
>
> The MUST here seems to be contradicting line 167 of the other
> doc. Can we remove it or maybe change it to MAY?

I don't think we want MAY here.  Can we end both this and the other section with:

  … has distribution restrictions.

jonboulle: SGTM

stevvooe:
@jonboulle The term "foreign" came out of the notion that these layers are foreign to the primary distribution process. Originally, these were tightly coupled with the `urls` field on layer descriptors. Once this coupling was removed, their role as "foreign layers" was more about name, as it is really a layer that you can fetch by url and cannot distribute. Neither component of this feature is "foreign" and it only becomes so when used together. It is probably not the best term (I would favor `nondistributable`), but it is already in place in docker.

If you think bike shedding is important, we can change it, but I am not sure what the value that provides.

media-types.md
stevvooe: Fixed.

wking:
On Fri, Sep 09, 2016 at 01:03:24PM -0700, Stephen Day wrote:
> It is probably not the best term (I would favor `nondistributable`),
> but it is already in place in docker.

All of our media types are already different from Docker, so I don't
see a backwards-compat issue here.  We're just tweaking a key/value
pair in the media type compatibility map.

> If you think bike shedding is important, we can change it, but I am
> not sure what the value that provides.

More expressive naming will hopefully leave fewer people scratching
their heads wondering what ‘foreign’ means ;).

jonboulle:
@stevvooe c'mon, this is not about bike-shedding - I want progress just as much as anyone else. But I don't understand why we should stick to nonobvious language when we're already changing media types. The value it provides is clear: lowering the cognitive burden to people approaching the spec.

stevvooe:
@jonboulle I fully agree with the goal of reducing cognitive burden but consistency is also a valid goal. Docker calls this `application/vnd.docker.image.rootfs.foreign.diff.tar.gzip` and the concepts are isomorphic. Do you think these having different names helps to reduce cognitive burden?

How about we get this merged based on content and concept do another PR where we discuss the name of the media type? Would that suffice?

wking:
On Fri, Sep 09, 2016 at 02:07:49PM -0700, Stephen Day wrote:
> Do you think these having different names helps to reduce cognitive
> burden?

When the new OCI name is more clear?  Yes.  If we didn't think so, we
wouldn't have landed #212.

> How about we get this merged based on content and concept do another
> PR where we discuss the name of the media type? Would that suffice?

I don't see how splitting this discussion out into a new PR will help
resolve it (there's not a lot of other distracting stuff happing in
this one), but if folks feel like that's the case, I'm happy to
discuss foreign → nondistributable in a follow-up PR.

stevvooe:
> When the new OCI name is more clear?  Yes.  If we didn't think so, we
wouldn't have landed #212.

#212 actually brings it closer to https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#media-types. Before, we had an extra level of indirection in OCI.

I'll say that I am as much of a stickler for naming as anyone, but I'm really just trying to move things forward. I think we have agreement on the concept, but I don't want to get it held up on naming. Let's get the concept in. If we agree on a name change, then we can merge that, as well.

wking:
On Fri, Sep 09, 2016 at 02:39:34PM -0700, Stephen Day wrote:
> Let's get the concept in. If we agree on a name change, then we can
> merge that, as well.

Fair.  I'm just not sure why we don't *already* have an agreement on
changing the name ;).

wking:
I still have the concerns from [1], which I'm ok ignoring or punting
on, but otherwise e48e595f looks good to me.

[1]: https://github.com/opencontainers/image-spec/pull/233#issuecomment-245667367

jonboulle:
LGTM now, thanks @stevvooe 

I do agree with wking's comments in https://github.com/opencontainers/image-spec/pull/233#issuecomment-245667367 but we can follow up for that.

philips:
LGTM

stevvooe:
🎊

runcom:
:tada:

PR #232, opened 2016-08-31
image-layout: use slash separated hierarchy for blob storage

stevvooe:
Signed-off-by: Stephen J Day <stephen.day@docker.com>

Closes #208

PR #230, opened 2016-08-31
image: slash separation between algo and hash

runcom:
Fix up oci-image-tool to look for hashes under `blobs/{algo}` as per https://github.com/opencontainers/image-spec/issues/208

This isn't addressing any language change required

```bash
$ tree 
.
├── blobs
│   └── sha256
│       ├── 086acbc3376eedf5a3d9c969a0d446e803600091f133288e8f5643a46e801976
│       ├── 98b338c114d348c417532b8a3bafd364f5d0c8f3541c01d0652e7c043d7d9056
│       └── bc6eb737e26b988486a6e8d5ac29167ae7aaaae7ba98fac406e4722a4eb219bb
├── oci-layout
└── refs
    └── latest

3 directories, 5 files

$ oci-image-tool validate --ref latest .                                                             
.: OK # nevermind here, I just noticed it and I have plan to fix this up :) net is that tests pass
```

Signed-off-by: Antonio Murdaca <runcom@redhat.com>

PR #235, opened 2016-09-01
image tool: use slash separated hierarchy for blob storage

xiekeyang:
It is relevant to #232 .
I've run the new code for `create`, `unpack` and `validate`, and they are passed.


coolljt0725:
maybe duplicate with #230 ?

xiekeyang:
Sorry I miss that, and will close it.

PR #237, opened 2016-09-01
Fix id_desc link in layer.md

coolljt0725:

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -3,7 +3,7 @@
 An example of creating an Image Filesystem Changeset follows.
 
 An image root filesystem is first created as an empty directory.
-Here is the initial empty directory structure for a changeset using the randomly-generated directory name `c3167915dc9d` ([actual layer DiffIDs are generated based on the content](#id_desc)).
+Here is the initial empty directory structure for a changeset using the randomly-generated directory name `c3167915dc9d` ([actual layer DiffIDs are generated based on the content](serialization.md#terminology)).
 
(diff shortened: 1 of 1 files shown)


philips:
id_desc doesn't exist either.

layer.md
```diff
@@ -3,7 +3,7 @@
 An example of creating an Image Filesystem Changeset follows.
 
 An image root filesystem is first created as an empty directory.
-Here is the initial empty directory structure for the a changeset using the randomly-generated directory name `c3167915dc9d` ([actual layer DiffIDs are generated based on the content](#id_desc)).
+Here is the initial empty directory structure for the a changeset using the randomly-generated directory name `c3167915dc9d` ([actual layer DiffIDs are generated based on the content](#serialization.md#terminology)).
```
wking: `#serialization.md#terminology` → `serialization.md#terminology`

wking:
On Thu, Sep 01, 2016 at 05:56:19AM -0700, Brandon Philips wrote:
> id_desc doesn't exist either.

The serialization file's “Terminology” section does define “Layer
DiffID”, so I think linking to that section is fine.

coolljt0725:
@philips  `serialization.md#terminology` has description of `layer DiffIDs`, so I think it's ok to link to there

jonboulle:
This is addressed in #255 which is being actively worked on now so I think we should just defer to there.

PR #238, opened 2016-09-01
panic fix

xiekeyang:
Here the program will happen panic if the target path not found. So, it had better to add a condition to avoid that. I check the other `info` point which should be safe.

### DIFF
--- a/image/walker.go
+++ b/image/walker.go
@@ -92,6 +92,12 @@ func newPathWalker(root string) walker {
 
 func (w *pathWalker) walk(f walkFunc) error {
 	return filepath.Walk(w.root, func(path string, info os.FileInfo, err error) error {
+		// MUST check error value, to make sure the `os.FileInfo` is available.
+		// Otherwise panic risk will exist.
+		if err != nil {
+			return errors.Wrap(err, "error walking path")
+		}
+
 		rel, err := filepath.Rel(w.root, path)
 		if err != nil {
 			return errors.Wrap(err, "error walking path") // err from filepath.Walk includes path name


image/walker.go
```diff
@@ -96,7 +96,9 @@ func (w *pathWalker) walk(f walkFunc) error {
 		if err != nil {
 			return errors.Wrap(err, "error walking path") // err from filepath.Walk includes path name
 		}
-
+		if info == nil {
+			return fmt.Errorf("the path %s not found", path)
```
runcom: Can you use `errors.Wrap` instead of `fmt`?

xiekeyang: OK

runcom:
BTW, do you have a way to reproduce this panic? I cannot seem to be able to reproduce it and afaict if `filepath.Walk` enters `WalkFn` then `path` isn't empty and `info` is not `nil` (there's no doc about this in golang's doc though)

xiekeyang:
@runcom 
in case `busybox-oci` doesn't under `mydir` folder:

```
mydir$ oci-image-tool validate --type imageLayout --ref latest busybox-oci
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0x52d8ed]

goroutine 1 [running]:
panic(0x86cf80, 0xc8200120c0)
	/home/kxie/go/src/runtime/panic.go:481 +0x3e6
github.com/opencontainers/image-spec/image.(*pathWalker).walk.func1(0x7ffffefaf441, 0xb, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/image/walker.go:100 +0x10d
path/filepath.Walk(0x7ffffefaf441, 0xb, 0xc8200c1440, 0x0, 0x0)
	/home/kxie/go/src/path/filepath/path.go:394 +0xa5
github.com/opencontainers/image-spec/image.(*pathWalker).walk(0xc8201307b0, 0xc8200c1420, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/image/walker.go:111 +0x90
github.com/opencontainers/image-spec/image.findDescriptor(0x7feae1d3aa88, 0xc8201307b0, 0x7ffffefaf43a, 0x6, 0x412dbd, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/image/descriptor.go:44 +0x182
github.com/opencontainers/image-spec/image.validate(0x7feae1d3aa88, 0xc8201307b0, 0x7ffffefaf43a, 0x6, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/image/image.go:46 +0x52
github.com/opencontainers/image-spec/image.ValidateLayout(0x7ffffefaf441, 0xb, 0x7ffffefaf43a, 0x6, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/image/image.go:29 +0xd5
main.(*validateCmd).validatePath(0xc820011380, 0x7ffffefaf441, 0xb, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/cmd/oci-image-tool/validate.go:126 +0x26d
main.(*validateCmd).Run(0xc820011380, 0xc820087200, 0xc82001c690, 0x1, 0x5)
	/home/kxie/work/src/github.com/opencontainers/image-spec/cmd/oci-image-tool/validate.go:84 +0x236
main.(*validateCmd).Run-fm(0xc820087200, 0xc82001c690, 0x1, 0x5)
	/home/kxie/work/src/github.com/opencontainers/image-spec/cmd/oci-image-tool/validate.go:54 +0x48
github.com/opencontainers/image-spec/vendor/github.com/spf13/cobra.(*Command).execute(0xc820087200, 0xc82001c640, 0x5, 0x5, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/vendor/github.com/spf13/cobra/command.go:603 +0x896
github.com/opencontainers/image-spec/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0xc820086fc0, 0xc820087200, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/vendor/github.com/spf13/cobra/command.go:689 +0x55c
github.com/opencontainers/image-spec/vendor/github.com/spf13/cobra.(*Command).Execute(0xc820086fc0, 0x0, 0x0)
	/home/kxie/work/src/github.com/opencontainers/image-spec/vendor/github.com/spf13/cobra/command.go:648 +0x2d
main.main()
	/home/kxie/work/src/github.com/opencontainers/image-spec/cmd/oci-image-tool/main.go:37 +0x3e0
```	

panic happen at:
/home/kxie/work/src/github.com/opencontainers/image-spec/image/walker.go:100 +0x10d

runcom:
@xiekeyang great, should we instead check if the path exists before even going ahead and try to walk the path?

xiekeyang:
@runcom all right, let me find a more graceful way.

xiekeyang:
@runcom , I'd place a check ahead.
However, I still feel that `func(path string, info os.FileInfo, err error) error` is a local function block.
`info` is a func arg which may cause panic. It likely we should reserve `if info == nil` judgement inside.
What you think? reserve here or not?

runcom:
> @runcom , I'd place a check ahead.
> However, I still feel that func(path string, info os.FileInfo, err error) error is a local function block.
> info is a func arg which may cause panic. It likely we should reserve if info == nil judgement inside.
> What you think? reserve here or not?

yeah keep it, I'm fine :) thx

xiekeyang:
@runcom modified, PTAL , thanks so much.

image/walker.go
```diff
@@ -91,12 +91,17 @@ func newPathWalker(root string) walker {
 }
 
 func (w *pathWalker) walk(f walkFunc) error {
+	if _, err := os.Stat(w.root); err != nil {
```
runcom: does it make sense to have this check before even calling/creating the walker?

xiekeyang:
@runcom I modified in `ValidateLayout`. Other new walker points have check the path status, so they should be safe. PTAL, thanks.

runcom:
Sgtm ping @s-urbaniak

image/walker.go
```diff
@@ -92,11 +92,13 @@ func newPathWalker(root string) walker {
 
 func (w *pathWalker) walk(f walkFunc) error {
 	return filepath.Walk(w.root, func(path string, info os.FileInfo, err error) error {
+		if info == nil {
```
stevvooe: Need to check `err` here. It is the error returned by the stat call that populates this `os.FileInfo`.

xiekeyang:
@stevvooe @runcom @s-urbaniak ,
It is modified, PTAL, thanks.

xiekeyang:
Updated.

image/walker.go
```diff
@@ -92,11 +92,16 @@ func newPathWalker(root string) walker {
 
 func (w *pathWalker) walk(f walkFunc) error {
 	return filepath.Walk(w.root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return errors.Wrap(err, "error walking path")
+		}
+		if info == nil {
```
runcom: Is info guaranteed to be not nil after the error check? (maybe @stevooee knows)

xiekeyang: @runcom @stevvooe 
I tested `filepath.Walk` for any possible cases. I did not really find the result that error == nil and FileInfo == nil. 
Maybe `if info == nil` should be removed safely. Waiting for your proposal.

xiekeyang:
Unfortunately 
[golang spec](https://golang.org/pkg/path/filepath/#WalkFunc) doesn't explain that.

xiekeyang:
@stevvooe @runcom ,
I check the source code inside function `filepath.Walk()`, `info` must be evaluated on each nil-error returned point. So, let's safely say that we needn't  check `info` arg, instead check `err` arg only.
Here I update the PR, PTAL.
Thanks for your review and suggestion very much.

runcom:
SGTM

image/walker.go
```diff
@@ -92,11 +92,13 @@ func newPathWalker(root string) walker {
...
 func (w *pathWalker) walk(f walkFunc) error {
 	return filepath.Walk(w.root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return errors.Wrap(err, "error walking path")
+		}
 		rel, err := filepath.Rel(w.root, path)
 		if err != nil {
 			return errors.Wrap(err, "error walking path") // err from filepath.Walk includes path name
 		}
-
```
stevvooe: nit: please don't remove extra white space. This helps the code to be more readable.

image/walker.go
```diff
@@ -92,11 +92,13 @@ func newPathWalker(root string) walker {
 
 func (w *pathWalker) walk(f walkFunc) error {
 	return filepath.Walk(w.root, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return errors.Wrap(err, "error walking path")
+		}
 		rel, err := filepath.Rel(w.root, path)
```
stevvooe: nit: need a new line about this statement.

stevvooe:
LGTM + a few whitespace nits

xiekeyang:
OK, updated.

xiekeyang:
@runcom Is it OK?

s-urbaniak:
LGTM

xiekeyang:
@opencontainers/image-spec-maintainers Could this be merged? THX.

image/image.go
```diff
@@ -26,6 +26,9 @@ import (
 // validates the manifest pointed to by the given ref
 // or returns an error if the validation failed.
 func ValidateLayout(src, ref string) error {
+	if _, err := os.Stat(src); err != nil {
```
jonboulle: This is unnecessary (and slightly racy) given the validate->pathwalker code does appropriate error handling anyway.

xiekeyang: @jonboulle excuse me, I don't fully understand what you mean. This check should be done because that `no file exist case` should be checked and returned as early as possible, to avoid the future redundant program being executed.

jonboulle: The code that is executed after this MUST be able to deal with the same situation (it's possible, even if unlikely, that that path is removed after this check but before being opened), which makes the check redundant. What's the benefit in returning early here?

xiekeyang: My thought is that we should check function's legal or illegal argument in its self. It seems to be inappropriate to throw the checking out to other functions. If other functions work out to remove the checking, it is hard to be aware, and the bug will be recurrent.

> it's possible, even if unlikely, that that path is removed after this check but before being opened
Right, so we check it in `(w *pathWalker) walk()` which can protect the execution from the the case.

If my comment fully cover your idea? Of course hopefully we can present a nice solution. Thanks so much.

/cc @stevvooe @runcom @s-urbaniak

jonboulle: > If other functions work out to remove the checking, it is hard to be aware, and the bug will be recurrent.

This argument doesn't hold up; the "other functions" (in this case the validation code) absolutely need to be able to deal with such a case - if they can't, then it's a serious bug that we shouldn't be trying to mask here. 

If you want to make the argument that it's cleaner/more consistent for end users, that's a different story

stevvooe: This check is already an implicit property of the walk. It is not necessary.

If you want to wrap the error, wrap the error from validate, as well.

xiekeyang:
fixed, PTAL.
cc @runcom @jonboulle @s-urbaniak

xiekeyang:
Any more suggestion?

image/walker.go
jonboulle: I'm not sure the error message here is meaningful. I think we should omit this check (perhaps you could also file a bug upstream on golang to clarify the guarantees here)

xiekeyang:
@jonboulle Yeah, sorry it should have been removed, as my pre comment.
> xiekeyang commented 6 days ago
> I check the source code inside function filepath.Walk(), info must be evaluated on each nil-error  
> returned point. So, let's safely say that we needn't check info arg, instead check err arg only.

Updated. PTAL again, thanks.

jonboulle:
LGTM (at last!)

wking:
c7158ec6 looks fine to me, but @stevvooe still has some whitespace
nits that would be easy to address [1,2].

[1]: https://github.com/opencontainers/image-spec/pull/238/commits/c7158ec643e71d4d2766fff41728f21d7bfc4aa2#r77421829
[2]: https://github.com/opencontainers/image-spec/pull/238/commits/c7158ec643e71d4d2766fff41728f21d7bfc4aa2#r77421797

xiekeyang:
Had done, but unfortunately lost.
I supplement it now.
/cc @wking

image/walker.go
stevvooe: bump.

stevvooe: https://github.com/xiekeyang/image-spec/pull/1

stevvooe:
LGTM

I've made https://github.com/xiekeyang/image-spec/pull/1. Please submit as PR and we'll get this code fix merged.

stevvooe:
LGTM

xiekeyang:
ping @stevvooe 
fix `sign-off` failed, and keep the precious whitespace:
https://github.com/xiekeyang/image-spec/blob/3f2490d8abad6e7d9088022027228a3ed4b9ed59/image/walker.go#L105

stevvooe:
LGTM

wking:
3f2490d looks good to me.

jonboulle:
lgtm

PR #250, opened 2016-09-02
oci-image-tool: Cleanup partially-unpacked directories on failures

### LINKED ISSUE #242: oci-image-tool: Cleanup partially-unpacked directories on failures
We might want a switch to turn this off, but I think the default should be to remove the unpack directory if we fail partway through the unpack.

Spun off from [here][1].

[1]: https://github.com/opencontainers/image-spec/pull/240#issuecomment-244138589

coolljt0725:

fixes #242
cc @runcom  @wking

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -87,7 +87,23 @@ func (m *manifest) validate(w walker) error {
 	return nil
 }
 
-func (m *manifest) unpack(w walker, dest string) error {
+func (m *manifest) unpack(w walker, dest string) (retErr error) {
+	// error out if the dest directory is not empty
+	s, err := ioutil.ReadDir(dest)
+	if err != nil && !os.IsNotExist(err) {
+		return errors.Wrap(err, "unable to open file") // err contains dest
+	}
+	if len(s) > 0 {
+		return fmt.Errorf("%s is not empty", dest)
+	}
+	defer func() {
+		// if we encounter error during unpacking
+		// clean up partially-unpacked destination
+		if retErr != nil {
+			os.RemoveAll(dest)
+		}
+	}()
+
 	for _, d := range m.Layers {
 		if d.MediaType != string(schema.MediaTypeImageConfig) {
 			continue
--- a/image/manifest_test.go
+++ b/image/manifest_test.go
@@ -1,3 +1,17 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package image
 
 import (
--- a/image/manifest_test.go
+++ b/image/manifest_test.go
@@ -4,6 +4,8 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"crypto/sha256"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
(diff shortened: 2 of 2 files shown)


runcom:
does it make sense to unpack to a temp dir and then os.Rename? to have something more _atomic_ somehow (on error just throw off the temp)

jonboulle:
aiyeee, I really don't like this code. There is (probably?) a breakout if [this](https://github.com/coolljt0725/image-spec/blob/63f50bb1e2049f0971474cceaf247a6619070e6a/image/manifest.go#L181) is hit (removing a path outside the tarball) - and if I'm wrong about that one I'd bet there are other corner cases not covered yet.

Can we use some existing, well-tested tar extracting code instead?

runcom:
This is already coming from Docker code which is imo well tested. Not sure about other options.

wking:
On Fri, Sep 02, 2016 at 07:18:51AM -0700, Jonathan Boulle wrote:
> There is (probably?) a breakout if
> [this](https://github.com/coolljt0725/image-spec/blob/63f50bb1e2049f0971474cceaf247a6619070e6a/image/manifest.go#L181)
> is hit (removing a path outside the tarball) - and if I'm wrong
> about that one I'd bet there are other corner cases not covered yet.

Yeah, that looks like a bug to me.  You could fix it by moving:

  entries[path] = true

after the check.

But I think this per-layer unwinding approach is trying too hard.  And
there will be some things (whiteout removal, file clobbers) that you
can't unwind easily.  I'd rather *not* try to cleanup inside
unpackLayer and instead address #242 by (conditionally) catching
unpack errors and using:

  os.RemoveAll(dest)

in *unpack* if any of the unpackLayer calls failed.

coolljt0725:
@jonboulle  
>aiyeee, I really don't like this code. There is (probably?) a breakout if this is hit (removing a path outside the tarball) - and if I'm wrong about that one I'd bet there are other corner cases not covered yet.

it's does a bug here, thank you for pointing out:)


@wking  I has thought about `catching unpack errors and using: os.RemoveAll(dest)`, but I think if there are already some files in `dest` before unpacking, it would make mistakes. So I just want to delete what had been unpacked. 
But this approach of this pr only remove the current layer if encounter error,  still need to some work to remove already unpacked layer.

@runcom  Do you mean we unpack the layer to a tmp file and then rename to `dest` if no error encounter?  In this way, I think it would make mistake if there are already files in`dest`

Thank you all for reviewing, I'll work on this tomorrow :)
`

wking:
On Fri, Sep 02, 2016 at 09:43:44AM -0700, Lei Jitang wrote:
> @wking I has thought about `catching unpack errors and using:
> os.RemoveAll(dest)`, but I think if there are already some files in
> `dest` before unpacking, it would make mistakes.

I think it should be an error if there is anything in ‘dest’ before
you start unpacking.

coolljt0725:
@wking that's make sense. 
updated

image/manifest.go
```diff
@@ -87,7 +87,22 @@ func (m *manifest) validate(w walker) error {
 	return nil
 }
 
-func (m *manifest) unpack(w walker, dest string) error {
+func (m *manifest) unpack(w walker, dest string) (retErr error) {
+	// error out if the dest directory is not empty
+	if _, err := os.Stat(dest); err == nil {
+		s, _ := ioutil.ReadDir(dest)
```
wking: You don't need both `Stat` and `ReadDir`.  If you want to error out if there is an entry at that path, just use `Stat`.  If you want to error out unless the path has no entry or contains an empty directory, just use `ReadDir`.  The latter sounds better to me (you can use it with `mktemp -d`), but I'm ok with either.

wking: And if you end up using `ReadDir`, you should be checking for errors that it returns and returning those, instead of returning your own “is not empty” error in those cases (e.g. maybe the directory is not empty, but you lack read permission).

image/manifest.go
```diff
@@ -87,7 +87,22 @@ func (m *manifest) validate(w walker) error {
...
+		if len(s) > 0 {
+			return fmt.Errorf("dest rootfs directory '%s' is not empty", dest)
+		}
+	}
+	defer func() {
+		// if we encounter error during unpacking
+		// clean up partically-unpacked directories
```
wking: “partically” → “partially” and “directories” → “destination”

image/manifest.go
coolljt0725: @wking  Thanks, will update

coolljt0725:
@wking  updated

image/manifest.go
```diff
@@ -87,7 +87,22 @@ func (m *manifest) validate(w walker) error {
 	return nil
 }
 
-func (m *manifest) unpack(w walker, dest string) error {
+func (m *manifest) unpack(w walker, dest string) (retErr error) {
+	// error out if the dest directory is not empty
+	if s, err := ioutil.ReadDir(dest); err != nil && !os.IsNotExist(err) || len(s) > 0 {
+		if err != nil {
+			return fmt.Errorf("failed to open %s: %v", dest, err)
```
wking: Use [`errors.Wrap`][1] instead of `fmt.Errorf`?  There is a lot of existing use of `errors.Wrap` in this package.

[1]: https://godoc.org/github.com/pkg/errors#hdr-Adding_context_to_an_error

image/manifest.go
```diff
@@ -87,7 +87,22 @@ func (m *manifest) validate(w walker) error {
...
+		}
+		return fmt.Errorf("%s is not empty", dest)
```
wking: Can you flatten the current:

    if … err != nil && !os.IsNotExist(err) || len(s) > 0 {
      if err != nil {
        …a…
      }
      …b…
    }

into:

    if err != nil && !os.IsNotExist(err) {
        …a…
    }
    if len(s) > 0 {
      …b…
    }

coolljt0725: @wking  I just want the if statement in a single block, but I'm ok with your change

coolljt0725:
@wking updated, thanks :)

wking:
7684889 looks good to me.

jonboulle:
Looks good but can you add a test for this please?

coolljt0725:
@jbouzane  test added, but the test block by #257 
with #257, make test paas
comment the codes added in image/manifest.go and make test again test will failed with 
````
FAIL: TestUnpackLayerRemovePartialyUnpackedFile (0.01s)
        manifest_test.go:113: Execpt partialy unpacked file has been removed

````

wking:
On Wed, Sep 07, 2016 at 06:47:25AM -0700, Lei Jitang wrote:
> test added, but the test block by #257

Rebase now that #257 has landed?

coolljt0725:
@wking  Should we continue this after splitting tool to a separate repo?

wking:
On Wed, Sep 14, 2016 at 12:32:40AM -0700, Lei Jitang wrote:
> @wking Should we continue this after splitting tool to a separate
> repo?

If you like.  I'm a maintainer of neither, but I see no reason to
pause development while we wait for the shift.

PR #243, opened 2016-09-01
serialization: Remove "changeset" from terminology

wking:
On Wed, Aug 31, 2016 at 06:53:45PM -0700, Brandon Philips [wrote][1]:
> @wking I think changeset is the logical concept, layer is the
> serialization of that changeset into a tar. I really don't have a
> strong opnion about having different nouns for these things but
> @stevvooe seems to.

On Thu, Sep 01, 2016 at 10:29:16AM -0700, Stephen Day [wrote][2]:
> A "changeset" is just an english word for a set of changes. We
> represent a set of changes with a layer.

I also removed the "Image" prefix, since all of these terms are image-specific.

[1]: https://github.com/opencontainers/image-spec/pull/231#issuecomment-243954395
[2]: https://github.com/opencontainers/image-spec/pull/231#issuecomment-244151555

PR #255, opened 2016-09-04
layer: revamp for readability

vbatts:
When adding verbiage for restriction on duplicate entries in the tar
archive, it was not immediately clear where this information would be
organized.

This revamp seeks to make the document on the whole more suitable for
implementors and be interpretable for compliance.

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -1,6 +1,12 @@
-# Creating an Image Filesystem Changeset
+# OpenContainers Image Layers Specification
 
-An example of creating an Image Filesystem Changeset follows.
+This specification describes how to serialize a filesystem and filesystem changes like removed files into a blob called a layer.
+To create a complete filesystem one or more layers are ordered on top of each other.
+This document will use a concrete example to illustrate how to create and consume these filesystem layers.
+
+## Creating an Image Layers Based on Filesystem Changes
+
+An example of creating image filesystem layers follows.
 
 An image root filesystem is first created as an empty directory.
 Here is the initial empty directory structure for a changeset using the randomly-generated directory name `c3167915dc9d` ([actual layer DiffIDs are generated based on the content](#id_desc)).
@@ -28,8 +34,9 @@ bin/my-app-binary
 bin/my-app-tools
 ```
 
-To make changes to the filesystem of this container image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
-NOTE: a copy-on-write or union filesystem can make this very efficient:
+To make changes to the filesystem of this image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
+
+**Implementor's Note:** a copy-on-write or union filesystem can make this very efficient.
 
 ```
 f60c56784b83/
--- a/layer.md
+++ b/layer.md
@@ -1,24 +1,74 @@
-# OpenContainers Image Layers Specification
+# Image Layer Filesystem Changeset
 
-This specification describes how to serialize a filesystem and filesystem changes like removed files into a blob called a layer.
-To create a complete filesystem one or more layers are ordered on top of each other.
+This document describes how to serialize a filesystem and filesystem changes like removed files into a blob called a layer.
+One or more layers are ordered on top of each other to create a complete filesystem.
 This document will use a concrete example to illustrate how to create and consume these filesystem layers.
 
-## Creating an Image Layers Based on Filesystem Changes
+## Distributable Format
 
-An example of creating image filesystem layers follows.
+Layer Changesets for the [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` MUST be packaged in [tar archive][tar-archive].
+Layer Changesets for the [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` MUST NOT include duplicate entries for file paths in the resulting [tar archive][tar-archive].
 
-An image root filesystem is first created as an empty directory.
-Here is the initial empty directory structure for a changeset using the randomly-generated directory name `c3167915dc9d` ([actual layer DiffIDs are generated based on the content](#id_desc)).
+## Change Types
+
+Types of changes that can occur in a changeset are:
+
+* Additions
+* Modifications
+* Removals
+
+Additions and Modifications are represented the same in the changeset tar archive.
+
+Removals are represented using "[whiteout](#whiteouts)" file entries (See [Representing Changes](#representing-changes)).
+
+### File Types
+
+Throughout this document section, the use of word "files" includes:
+
+* regular files
+* directories
+* sockets
+* symbolic links
+* block devices
+* character devices
+* FIFOs
+
+### File Attributes
+
+Where supported, MUST include file attributes for Additions and Modifications include:
+
+* Modification Time (`mtime`)
+* User ID (`uid`)
+    * User Name (`uname`) *secondary to `uid`*
+* Group ID (`gid `)
+    * Group Name (`gname`) *secondary to `gid`*
+* Mode (`mode`)
+* Extended Attributes (`xattrs`)
+* Symlink reference (`linkname`)
+
+[Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they lack consistent support across tar implementations.
+
+## Creating
+
+### Initial Root Filesystem
+
+The initial root filesystem is the base or parent layer.
+
+For this example, an image root filesystem has an initial state as an empty directory.
+The name of the directory is not relevant to the layer itself, only for the purpose of producing comparisons.
+
+Here is an initial empty directory structure for a changeset, with a unique directory name `rootfs-c9d-v1`.
 
 ```
-c3167915dc9d/
+rootfs-c9d-v1/
 ```
+
+### Populate Initial Filesystem
 
 Files and directories are then created:
 
 ```
-c3167915dc9d/
+rootfs-c9d-v1/
     etc/
         my-app-config
     bin/
@@ -26,20 +76,37 @@ c3167915dc9d/
         my-app-tools
 ```
 
-The `c3167915dc9d` directory is then committed as a plain Tar archive with entries for the following files:
+The `rootfs-c9d-v1` directory is then created as a plain [tar archive][tar-archive] with relative path to `rootfs-c9d-v1`.
+Entries for the following files:
 
 ```
-etc/my-app-config
-bin/my-app-binary
-bin/my-app-tools
+./
+./etc/
+./etc/my-app-config
+./bin/
+./bin/my-app-binary
+./bin/my-app-tools
 ```
 
-To make changes to the filesystem of this image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
+### Populate a Comparison Filesystem
 
-**Implementor's Note:** a copy-on-write or union filesystem can make this very efficient.
+Create a new directory and initialize it with an copy or snapshot of the prior root filesystem.
+Example commands that can preserve [file attributes](#file-attributes) to make this copy are:
+* [cp(1)](http://linux.die.net/man/1/cp): `cp -a rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
+* [rsync(1)](http://linux.die.net/man/1/rsync):  `rsync -aHAX rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
+* [tar(1)](http://linux.die.net/man/1/tar): `mkdir rootfs-c9d-v1.s1 && tar --acls --xattrs -C rootfs-c9d-v1/ -c . | tar -C rootfs-c9d-v1.s1/ --acls --xattrs -x` (including `--selinux` where supported)
+
+Any [changes](#change-types) to the snapshot MUST NOT change or affect the directory it was copied from.
+
+For example `rootfs-c9d-v1.s1` is an identical snapshot of `rootfs-c9d-v1`.
+In this way `rootfs-c9d-v1.s1` is prepared for updates and alterations.
+
+**Implementor's Note**: *a copy-on-write or union filesystem can efficiently make directory snapshots*
+
+Initial layout of the snapshot:
 
 ```
-f60c56784b83/
+rootfs-c9d-v1.s1/
     etc/
         my-app-config
     bin/
@@ -47,14 +114,16 @@ f60c56784b83/
         my-app-tools
 ```
 
-This example change is going to add a configuration directory at `/etc/my-app.d` which contains a default config file.
-There's also a change to the `my-app-tools` binary to handle the config layout change.
-The `f60c56784b83` directory then looks like this:
+See [Change Types](#change-types) for more details on changes.
+
+For example, add a directory at `/etc/my-app.d` containing a default config file, removing the existing config file.
+Also a change (in attribute or file content) to `./bin/my-app-tools` binary to handle the config layout change.
+
+Following these changes, the representation of the `rootfs-c9d-v1.s1` directory:
 
 ```
-f60c56784b83/
+rootfs-c9d-v1.s1/
     etc/
-        .wh.my-app-config
         my-app.d/
             default.cfg
     bin/
@@ -62,35 +131,70 @@ f60c56784b83/
         my-app-tools
 ```
 
-This reflects the removal of `/etc/my-app-config` and creation of a file and directory at `/etc/my-app.d/default.cfg`.
-`/bin/my-app-tools` has also been replaced with an updated version.
-Before committing this directory to a changeset, because it has a parent image, it is first compared with the directory tree of the parent snapshot, `f60c56784b83`, looking for files and directories that have been added, modified, or removed.
+### Determining Changes
+
+When two directories are compared, the relative root is the top-level directory.
+The directories are compared, looking for files that have been [added, modified, or removed](#change-types).
+
+For this example, `rootfs-c9d-v1/` and `rootfs-c9d-v1.s1/` are recursively compared, each as relative root path.
+
 The following changeset is found:
 
 ```
+Added:      /etc/my-app.d/
 Added:      /etc/my-app.d/default.cfg
 Modified:   /bin/my-app-tools
 Deleted:    /etc/my-app-config
 ```
 
-A Tar Archive is then created which contains *only* this changeset:
+This reflects the removal of `/etc/my-app-config` and creation of a file and directory at `/etc/my-app.d/default.cfg`.
+`/bin/my-app-tools` has also been replaced with an updated version.
+
+### Representing Changes
+
+A [tar archive][tar-archive] is then created which contains *only* this changeset:
 
 - Added and modified files and directories in their entirety
-- Deleted files or directory marked with a whiteout file
+- Deleted files or directories marked with a [whiteout file](#whiteouts)
 
-A whiteout file is an empty file that prefixes the deleted paths basename `.wh.`.
-When a whiteout is found in the upper changeset of a filesystem, any matching name in the lower changeset is ignored, and the whiteout itself is also hidden.
-As files prefixed with `.wh.` are special whiteout tombstones it is not possible to create a filesystem which has a file or directory with a name beginning with `.wh.`.
-
-The resulting Tar archive for `f60c56784b83` has the following entries:
+The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
 
 ```
-/etc/my-app.d/default.cfg
-/bin/my-app-tools
-/etc/.wh.my-app-config
+./etc/my-app.d/
+./etc/my-app.d/default.cfg
+./bin/my-app-tools
+./etc/.wh.my-app-config
 ```
 
-Whiteout files MUST only apply to resources in lower layers.
+Where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
+
+## Applying
+
+Layer Changesets of [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` are applied rather than strictly extracted in normal fashion for tar archives.
+
+Applying a layer changeset requires consideration for the [whiteout](#whiteouts) files.
+In the absence of any [whiteout](#whiteouts) files in a layer changeset, the archive is extracted like a regular tar archive.
+
+
+### Changeset over existing files
+
+This section covers applying an entry in a layer changeset, if the file path already exists.
+
+If the file path is a directory, then the existing path just has it's attribute set from the layer changeset for that filepath.
+If the file path is any other file type (regular file, FIFO, etc), then the:
+* file path is unlinked (See [`unlink(2)`](http://linux.die.net/man/2/unlink))
+* create the file
+    * If a regular file then content written.
+* set attributes on the filepath
+
+## Whiteouts
+
+A whiteout file is an empty file with a special filename that signifies a path should be deleted.
+A whiteout filename consists of the prefix .wh. plus the basename of the path to be deleted.
+As files prefixed with `.wh.` are special whiteout markers, it is not possible to create a filesystem which has a file or directory with a name beginning with `.wh.`.
+
+Once a whiteout is applied, the whiteout itself MUST also be hidden.
+Whiteout files MUST only apply to resources in lower/parent layers.
 Files that are present in the same layer as a whiteout file can only be hidden by whiteout files in subsequent layers.
 The following is a base layer with several resources:
 
@@ -124,6 +228,8 @@ a/.wh..wh..opq
 
 Implementations SHOULD generate layers such that the whiteout files appear before sibling directory entries.
 
+### Opaque Whiteout
+
 In addition to expressing that a single entry should be removed from a lower layer, layers may remove all of the children using an opaque whiteout entry.
 An opaque whiteout entry is a file with the name `.wh..wh..opq` indicating that all siblings are hidden in the lower layer.
 Let's take the following base layer as an example:
@@ -146,7 +252,7 @@ bin/
 ```
 
 This is called _opaque whiteout_ format.
-An _opaque whiteout_ file hides _all_ children of the `bin/` including sub-directories and all descendents.
+An _opaque whiteout_ file hides _all_ children of the `bin/` including sub-directories and all descendants.
 Using _explicit whiteout_ files, this would be equivalent to the following:
 
 ```
@@ -158,8 +264,10 @@ bin/
 
 In this case, a unique whiteout file is generated for each entry.
 If there were more children of `bin/` in the base layer, there would be an entry for each.
-Note that this opaque file will apply to _all_ children, including sub-directories, other resources and all descendents.
+Note that this opaque file will apply to _all_ children, including sub-directories, other resources and all descendants.
 
 Implementations SHOULD generate layers using _explicit whiteout_ files, but MUST accept both.
 
 Any given image is likely to be composed of several of these Image Filesystem Changeset tar archives.
+
+[tar-archive]: https://en.wikipedia.org/wiki/Tar_(computing)


vbatts:
Fixes #196 

For convenience, the view or the document is https://github.com/vbatts/oci-image-spec/blob/clarify_tar_entry_order/layer.md

layer.md
```diff
@@ -1,63 +1,109 @@
...
+./bin/
+./bin/my-app-binary
+./bin/my-app-tools
 ```
 
-To make changes to the filesystem of this container image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
-NOTE: a copy-on-write or union filesystem can make this very efficient:
+### Populate a Comparison Filesystem
+
+Create a new directory and initialize it with an identical copy or snapshot of the prior root filesystem.
```
coolljt0725: Should we make `an identical copy` more clear here or add a example ?  if we just `cp -r rootfs-c9d-v1 rootfs-c9d-v1.s1`, this is not an identical copy ,the timestamp is changed and I think the timestamp is important to generate diff.  To make a copy identical use `cp`, we should along with `-p` flag.

vbatts: I wondered about that. Something like `cp -pav <src> <dst>` or `rsync -avPHS <src> <dst>` or so on. These are all very Linux centric, which may be fine for starters, but this line is more of the rule, rather than the example.
Perhaps, having a statement of the files, their contents, and all available attributes (timestamps, mode, owner, group, xattrs, etc)?

layer.md
```diff
@@ -1,63 +1,109 @@
...
         my-app-config
     bin/
         my-app-binary
         my-app-tools
 ```
 
-This example change is going to add a configuration directory at `/etc/my-app.d` which contains a default config file.
-There's also a change to the `my-app-tools` binary to handle the config layout change.
-The `f60c56784b83` directory then looks like this:
+Changes include new files, file and path removals, and updates to file attributes (mtime, uid, gid, mode, xattrs, etc.).
```
vbatts: @coolljt0725 well this line addresses that. Any changes to these attributes will be seen as a change. Does this suffice?

layer.md
```diff
@@ -1,63 +1,109 @@
...
+## Distributable Format
+
+Layer Changesets for the [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` MUST be packaged in [tar archive][tar-archive].
+Layer Changesets for the [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` MUST NOT include duplicate entries for file paths in the resulting [tar archive][tar-archive].
+
+## Change Types
+
+Types of changes that can occur in a changeset are:
+
+* Added
```
jonboulle: additions, modifications, removals

layer.md
```diff
@@ -1,63 +1,109 @@
...
+Additions and Modifications are reflected the same in the changeset tar archive.
+
+Removals are reflected using "[whiteout](#whiteouts)" file entries (See [Representing Changes](#representing-changes)).
+
+## Creating
+
+### Initial Root Filesystem
+
+An image root filesystem has an initial state as an empty directory.
+The name of this not relevant to the layer itself, only for the purpose of producing comparisons.
```
jonboulle: the name of the directory is not relevant to the layer itself

layer.md
vbatts: k

layer.md
vbatts: k

layer.md
```diff
@@ -1,63 +1,109 @@
...
 
-The `c3167915dc9d` directory is then committed as a plain Tar archive with entries for the following files:
+The `rootfs-c9d-v1` directory is then committed as a plain [Tar archive][tar-archive] with relative path to `rootfs-c9d-v1`.
+Entries for the following files:
 
 ```
-etc/my-app-config
-bin/my-app-binary
-bin/my-app-tools
+./
```
jonboulle: why now with ./?

layer.md
```diff
@@ -1,63 +1,109 @@
...
+./bin/my-app-binary
+./bin/my-app-tools
 ```
 
-To make changes to the filesystem of this container image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
-NOTE: a copy-on-write or union filesystem can make this very efficient:
+### Populate a Comparison Filesystem
+
+Create a new directory and initialize it with an identical copy or snapshot of the prior root filesystem.
+Any changes to the snapshot MUST not change or affect the directory it was copied from.
```
jonboulle: MUST NOT

layer.md
jonboulle: Yes, identical should be more thoroughly defined

layer.md
vbatts: not a hard requirement, just more accurate, no?

```bash
[vbatts@bananaboat] {master *} ~/v1$ find
.
./etc
./etc/my-app-config
./bin
./bin/my-app-tools
./bin/my-app-binary
[vbatts@bananaboat] {master *} ~/v1$ tar -c . | tar tv
drwxr-xr-x vbatts/users      0 2016-09-02 10:45 ./
drwxr-xr-x vbatts/users      0 2016-09-02 10:45 ./etc/
-rw-r--r-- vbatts/users      0 2016-09-02 10:45 ./etc/my-app-config
drwxr-xr-x vbatts/users      0 2016-09-02 10:46 ./bin/
-rw-r--r-- vbatts/users      0 2016-09-02 10:46 ./bin/my-app-tools
-rw-r--r-- vbatts/users      0 2016-09-02 10:46 ./bin/my-app-binary
```

layer.md
vbatts: ah yeah. got it

layer.md
```diff
@@ -66,23 +112,33 @@ Modified:   /bin/my-app-tools
...
+### Representing Changes
 
-A whiteout file is an empty file that prefixes the deleted paths basename `.wh.`.
-When a whiteout is found in the upper changeset of a filesystem, any matching name in the lower changeset is ignored, and the whiteout itself is also hidden.
-As files prefixed with `.wh.` are special whiteout tombstones it is not possible to create a filesystem which has a file or directory with a name beginning with `.wh.`.
+A [tar archive][tar-archive] is then created which contains *only* this changeset:
 
-The resulting Tar archive for `f60c56784b83` has the following entries:
+- Added and modified files and directories in their entirety
+- Deleted files or directory marked with a [whiteout file](#whiteouts)
```
jonboulle: directories

layer.md
```diff
@@ -66,23 +112,33 @@ Modified:   /bin/my-app-tools
...
+`/bin/my-app-tools` has also been replaced with an updated version.
 
-- Added and modified files and directories in their entirety
-- Deleted files or directory marked with a whiteout file
+### Representing Changes
 
-A whiteout file is an empty file that prefixes the deleted paths basename `.wh.`.
-When a whiteout is found in the upper changeset of a filesystem, any matching name in the lower changeset is ignored, and the whiteout itself is also hidden.
-As files prefixed with `.wh.` are special whiteout tombstones it is not possible to create a filesystem which has a file or directory with a name beginning with `.wh.`.
+A [tar archive][tar-archive] is then created which contains *only* this changeset:
```
jonboulle: tar archive vs Tar archive in other places

layer.md
vbatts: @jonboulle so, then provide the reader with example commands to accomplish this?
I reckon "identical" could get into inode discussions and otherwise. Though just a couple lines down the attributes are mentioned. Perhaps this ought to itemize the important attributes when making a copy?

layer.md
```diff
@@ -66,23 +112,33 @@ Modified:   /bin/my-app-tools
...
+./etc/my-app.d/default.cfg
+./bin/my-app-tools
+./etc/.wh.my-app-config
 ```
 
+## Applying
+
+## Whiteouts
+
+A whiteout file is an empty file that prefixes the deleted paths basename with `.wh.`.
```
jonboulle: A whiteout file an empty file with a special filename that signifies a path should be deleted. A whiteout filename consists of the prefix `.wh.` plus the basename of the path to be deleted.

vbatts: fixed

layer.md
vbatts: i down cased all of them. Perhaps it is not _so_ proper to be capitalized.

layer.md
jonboulle: If we are defining tar as the distributable format, then can't we lean on that?

vbatts: @jonboulle fair, though most tar usages are through the golang library and up to an implementer.
Something like a tar command would be `tar --xattrs --selinux --acls -c .` where selinux is supported and step back from there?

vbatts: updated @coolljt0725 @jonboulle how it that?

layer.md
```diff
@@ -1,63 +1,129 @@
-# Creating an Image Filesystem Changeset
+# Filesystem Layer Changeset
 
-An example of creating an Image Filesystem Changeset follows.
+A layer changeset is the distributable difference between filesystems.
```
stevvooe: Maybe something like this:

> A layer encodes a distributable changeset between filesystems.

layer.md
stevvooe: @vbatts Are layers generated with the top-level entry? I'm not sure if attributes can applied safely in this manner.

layer.md
```diff
@@ -1,63 +1,129 @@
...
+rootfs-c9d-v1/
     etc/
         my-app-config
     bin/
         my-app-binary
         my-app-tools
 ```
 
-The `c3167915dc9d` directory is then committed as a plain Tar archive with entries for the following files:
+The `rootfs-c9d-v1` directory is then committed as a plain [tar archive][tar-archive] with relative path to `rootfs-c9d-v1`.
```
amitsaha: (Sorry,  I am coming from nowhere) Is committed supposed to have a special meaning here?

layer.md
coolljt0725: @vbatts  This seems good to me

layer.md
```diff
@@ -1,63 +1,129 @@
-# Creating an Image Filesystem Changeset
+# Filesystem Layer Changeset
 
-An example of creating an Image Filesystem Changeset follows.
+A layer changeset is the distributable difference between filesystems.
+A layer changeset is not directly extracted, but applied to a filesystem.
```
wking: I think this is “don't extract this with generic tar tooling”.  I'd rather drop this line to avoid drawing a distinction between the undefined “directly extracted” and the undefined “applied” at this level.  You say pretty much the same thing [here][1], and at that point are actually defining application.

[1]: https://github.com/opencontainers/image-spec/pull/255/files#diff-dc2bb3b4eddcffe96e682193926e8ff6R156

layer.md
```diff
@@ -1,63 +1,129 @@
...
+
+## Change Types
+
+Types of changes that can occur in a changeset are:
+
+* Additions
+* Modifications
+* Removals
+
+Additions and Modifications are reflected the same in the changeset tar archive.
```
wking: Maybe:

> Additions and Modifications have the same representation in the layer tar archive.

And I'd rather see “Layer Changeset” or “Layer” and not “Changeset” (#243).

layer.md
```diff
@@ -1,63 +1,129 @@
...
+
+Removals are reflected using "[whiteout](#whiteouts)" file entries (See [Representing Changes](#representing-changes)).
```
wking: “reflected” → “represented”?

layer.md
```diff
@@ -1,63 +1,129 @@
...
+
+### File Attributes
+
+File attributes that are provided for Additions and Modifications include:
+
+* Modification Time (`mtime`)
+* User ID (`uid`)
+ * User Name (`uname`) is secondary to `uid`
+* Group ID (`gid `)
```
wking: Four-space indents for sub-lists?  opencontainers/runtime-spec#495

layer.md
```diff
@@ -1,63 +1,129 @@
...
+
+Removals are reflected using "[whiteout](#whiteouts)" file entries (See [Representing Changes](#representing-changes)).
+
+### File Attributes
+
+File attributes that are provided for Additions and Modifications include:
+
+* Modification Time (`mtime`)
+* User ID (`uid`)
+ * User Name (`uname`) is secondary to `uid`
```
wking: Four-space indents for sub-lists?  opencontainers/runtime-spec#495

layer.md
```diff
@@ -1,63 +1,129 @@
...
+* Group ID (`gid `)
+ * Group Name (`gname`) is secondary to `gid`
+* Mode (`mode`)
+* Extended Attributes (`xattrs`)
+
+[Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used.
+Due to lack of consistent support will appear to mutate regular files on creating or applying of a layer changeset.
```
wking: This sentence does not parse for me.  I think you mean “you SHOULD NOT insert sparse files as entries in layer tarballs because some filesystems lack sparse support and you could fill the whole disk extracting such a file on those filesystems”.

layer.md
```diff
@@ -1,63 +1,129 @@
...
+
+## Creating
+
+### Initial Root Filesystem
+
+An image root filesystem has an initial state as an empty directory.
```
wking: This is not a *layer* requirement, it is a config/manifest requirement.  I suggest moving it to the [`rootfs.diff_ids`][1] and [`layers`][2] docs, since that's where the validation tests for this requirement will apply.

[1]: https://github.com/opencontainers/image-spec/blob/f3a8be2fffca47fcb15c0fd20786bafc5928761a/serialization.md#image-json-field-descriptions
[2]: https://github.com/opencontainers/image-spec/blob/f3a8be2fffca47fcb15c0fd20786bafc5928761a/manifest.md#image-manifest-property-descriptions

layer.md
```diff
@@ -1,63 +1,129 @@
...
+The name of the directory is not relevant to the layer itself, only for the purpose of producing comparisons.
+
+Here is an initial empty directory structure for a changeset, with a unique directory name `rootfs-c9d-v1`.
 
 ```
-c3167915dc9d/
+rootfs-c9d-v1/
```
wking: What is `c9d` about?  Can't we just call this `rootfs-v1`?

layer.md
wking: +1 to avoiding “committed”.  How about “The base layer is a plain [tar archive][tar-archive] of the contents of `rootfs-c9d-v1` (or [whatever we call the initial directory][1]).

[1]: https://github.com/opencontainers/image-spec/pull/255/files#r77778699

layer.md
wking: On Tue, Sep 06, 2016 at 04:44:12PM -0700, Stephen Day wrote:
> >  
> >  ```
> > -etc/my-app-config
> > -bin/my-app-binary
> > -bin/my-app-tools
> > +./
>
> @vbatts Are layers generated with the top-level entry? I'm not sure
> if attributes can applied safely in this manner.

That top-level entry will be the root of the container filesystem.
Why wouldn't its attributes be under the control of the image author?
I don't see how this is any more sensitive than the more deeply-nested
layer entries.

layer.md
```diff
@@ -1,63 +1,129 @@
...
+
+File attributes that are provided for Additions and Modifications include:
+
+* Modification Time (`mtime`)
+* User ID (`uid`)
+ * User Name (`uname`) is secondary to `uid`
+* Group ID (`gid `)
+ * Group Name (`gname`) is secondary to `gid`
+* Mode (`mode`)
+* Extended Attributes (`xattrs`)
```
wking: I don't see [`xattrs`][1] in [tar(5)][1].  Can we unpack this into something from a tar spec?  Although maybe [there is no current tar spec][2]?  I'm a bit jumpy about leaning on tar without defining tar ;).

[1]: https://github.com/libarchive/libarchive/wiki/ManPageTar5
[2]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#standards

layer.md
```diff
@@ -1,63 +1,129 @@
...
+./bin/my-app-tools
 ```
 
-To make changes to the filesystem of this container image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
-NOTE: a copy-on-write or union filesystem can make this very efficient:
+### Populate a Comparison Filesystem
+
+Create a new directory and initialize it with an identical copy or snapshot of the prior root filesystem.
+Example commands that can preserve file attributes to make this copy are:
+* [cp(1)](http://linux.die.net/man/1/cp): `cp -pav rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
```
wking: [`-p` is in POSIX][1], and `-v` is unnecessary.  `-a` depends on [how `xattr` breaks down][2], and with `-a`, `-p` is unnecessary.  So with `xattr`, I'd prefer:

    [cp(1)](http://www.man7.org/linux/man-pages/man1/cp.1.html): `cp -a …`

And without `xattr` I'd prefer:

    [cp(1)](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html): `cp -p …`

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html
[2]: https://github.com/opencontainers/image-spec/pull/255/files#r77781172

layer.md
```diff
@@ -1,63 +1,129 @@
...
+### File Attributes
+
+File attributes that are provided for Additions and Modifications include:
+
+* Modification Time (`mtime`)
+* User ID (`uid`)
+ * User Name (`uname`) is secondary to `uid`
+* Group ID (`gid `)
+ * Group Name (`gname`) is secondary to `gid`
+* Mode (`mode`)
```
wking: Do we need to break `mode` down into permissions, symlink-ness, etc.?  `tar` allows you to dereference symlinks (and hard links).

layer.md
```diff
@@ -1,63 +1,129 @@
...
 ```
 
-To make changes to the filesystem of this container image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
-NOTE: a copy-on-write or union filesystem can make this very efficient:
+### Populate a Comparison Filesystem
+
+Create a new directory and initialize it with an identical copy or snapshot of the prior root filesystem.
+Example commands that can preserve file attributes to make this copy are:
+* [cp(1)](http://linux.die.net/man/1/cp): `cp -pav rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
+* [rsync(1)](http://linux.die.net/man/1/rsync):  `rsync -avHAX rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
```
wking: Why `-A` or `-H`?  Neither ACLs nor hard-links are listed in your “File Attributes” section.  `-a` includes more stuff you don't list in your “File Attributes” section, including `--devices --specials` (via `-D`).

layer.md
```diff
@@ -1,63 +1,129 @@
...
         my-app-tools
 ```
 
-This example change is going to add a configuration directory at `/etc/my-app.d` which contains a default config file.
-There's also a change to the `my-app-tools` binary to handle the config layout change.
-The `f60c56784b83` directory then looks like this:
+Changes include new files, file and path removals, and updates to file attributes (mtime, uid, gid, mode, xattrs, etc.).
+
+For example, add a directory at `/etc/my-app.d` containing a default config file, removing the existing config file.
+Also a change to the `./bin/my-app-tools` binary to handle the config layout change.
```
wking: “change to the” → “change the”?

layer.md
wking: On Tue, Sep 06, 2016 at 07:48:55PM -0700, Lei Jitang wrote:
> > +Changes include new files, file and path removals, and updates to file attributes (mtime, uid, gid, mode, xattrs, etc.).

Instead of listing attributes, can we have:

  … updates to [file attributes](#file-attributes).

?

layer.md
```diff
@@ -1,63 +1,129 @@
...
 ```
 
-This reflects the removal of `/etc/my-app-config` and creation of a file and directory at `/etc/my-app.d/default.cfg`.
-`/bin/my-app-tools` has also been replaced with an updated version.
-Before committing this directory to a changeset, because it has a parent image, it is first compared with the directory tree of the parent snapshot, `f60c56784b83`, looking for files and directories that have been added, modified, or removed.
+### Determining Changes
+
+When two directories are compared, the relative root is the top-level directory.
+The directories are compared, looking for files that have been added, modified, or removed.
+Where "files" includes regular files, directories, sockets, symbolic links, block devices, character devices and FIFOs.
```
wking: This listing of file types should probably be in the “File Attributes” section (as a [breakdown of mode][1]?  At least in part.  I'm not sure if major/minor are stored in mode).

[1]: https://github.com/opencontainers/image-spec/pull/255/files#r77782597

layer.md
```diff
@@ -1,63 +1,129 @@
...
+* Modifications
+* Removals
+
+Additions and Modifications are reflected the same in the changeset tar archive.
+
+Removals are reflected using "[whiteout](#whiteouts)" file entries (See [Representing Changes](#representing-changes)).
+
+### File Attributes
+
+File attributes that are provided for Additions and Modifications include:
```
wking: Is the idea that these are attributes which a portable implementation MUST support?  If so, I'd rephrase to include a MUST.

Is there a limit on extensibility?  If some user's fancy “tar” supports some crazy attribute and adds it to a layer, is that layer compliant (even though that user may be the only one who can extract it back out)?

layer.md
```diff
@@ -66,23 +132,51 @@ Modified:   /bin/my-app-tools
...
+- Added and modified files and directories in their entirety
+- Deleted files or directories marked with a [whiteout file](#whiteouts)
+
+The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
 
 ```
-/etc/my-app.d/default.cfg
-/bin/my-app-tools
-/etc/.wh.my-app-config
+./etc/my-app.d/
```
wking: If we're going to list this directory here, we'll want to explain how it was modified above.

layer.md
```diff
@@ -66,23 +132,51 @@ Modified:   /bin/my-app-tools
...
+./etc/my-app.d/default.cfg
+./bin/my-app-tools
+./etc/.wh.my-app-config
 ```
 
+## Applying
+
+Layer Changesets of [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` are applied rather than strictly extracted in normal fashion for tar archives.
```
wking: I think the following two lines cover this idea.  Can we drop this line?  You mention the media type early on in the “Distributable Format” section.

layer.md
```diff
@@ -66,23 +132,51 @@ Modified:   /bin/my-app-tools
...
+* create the file
+ * If a regular file then content written.
+* set attributes on the filepath
+
+## Whiteouts
+
+A whiteout file is an empty file with a special filename that signifies a path should be deleted.
+A whiteout filename consists of the prefix .wh. plus the basename of the path to be deleted.
+As files prefixed with `.wh.` are special whiteout markers, it is not possible to create a filesystem which has a file or directory with a name beginning with `.wh.`.
+When a whiteout is found in the upper changeset of a filesystem, any matching name in the lower changeset is ignored, and the whiteout itself is also hidden.
```
wking: I think we can drop this line (the idea seems to be covered more clearly in the next paragraph).  But it's descended from #141, so I'm happy taking that up in another PR.

layer.md
vbatts: got it.

layer.md
vbatts: @amitsaha good point

layer.md
vbatts: got it

philips:
@vbatts I had another PR trying to add intro text. Want to use any of it? https://github.com/opencontainers/image-spec/pull/231

layer.md
vbatts: I'm not inclined to just stick with layer. As Layer is generic enough to refer to the on disk representation, where changeset is the specific mechanics being described.

layer.md
vbatts: sure

layer.md
vbatts: oh right.

layer.md
vbatts: fair, though didnt' say filesystems, because it isn't the filesystems. Golang `archive/tar` does not preserve the sparse-ness, as much `tar` implementations do not, largely only GNU tar does.
tar-split does not support sparse files either.
Most filesystems I know of do support sparse files.

layer.md
vbatts: there is no MUST language here, so this is not a requirement.
I can update the wording to be more about being an example.

layer.md
vbatts: uniqueness. We could call it `fart-jar/`.

jonboulle: What makes you think that's unique, Vincent?

layer.md
vbatts: that's why i did not link the libarchive man page, because it's support is varying.
xattrs support in GNU tar was carried patches for a long time by various linux distros, but now it is in upstream GNU tar. Reflected in http://linux.die.net/man/1/tar
There is no particular tar spec.

layer.md
vbatts: fair, `fart-jar` is as arbitrary as including `-c9d-`

layer.md
vbatts: not sure what to say here. Most folks are on GNU, but if POSIX is the bar then `cp` may not be an option.

layer.md
vbatts: well, mode is not the same as the file type. Perhaps file type ought to be a field, because if `foo` changes from a regular file into a FIFO, that is a modification.

As for hardlink, perhaps but could be limited by underlying fs support. As for symlink, yes, this reference is effectively the content of the file.

layer.md
vbatts: `-A` is acls, which are stored in xattrs, not directly implied via `-X`.
`-H` makes a difference, regardless of whether it's listed as a required attribute.
`-a` does make a difference for file type. that is considered a modification.

layer.md
vbatts: Yeah, this sentence I wrote first, so I'll just reference the other section from here.

layer.md
vbatts: done

layer.md
vbatts: @stevvooe the image would have say over the perms and owner of `/`, wouldn't it? Regardless of user-namespace, etc., many use-cases would ensure the `root` or who ever own `/`, otherwise it is implied by the choice of base image being built on. It is up to the extractor/consumer of the image layer to ensure that the directory being extracted to is confined and not vulnerable.

vbatts: "vulnerable" in the sense of potentially being world-writable and having policy not allowing such images, but the image layer has the facility declare its need for such.

layer.md
```diff
@@ -1,89 +1,199 @@
...
         my-app-tools
 ```
 
-This example change is going to add a configuration directory at `/etc/my-app.d` which contains a default config file.
-There's also a change to the `my-app-tools` binary to handle the config layout change.
-The `f60c56784b83` directory then looks like this:
+Changes include new files, file and path removals, and updates to file attributes (mtime, uid, gid, mode, xattrs, etc.).
+
+For example, add a directory at `/etc/my-app.d` containing a default config file, removing the existing config file.
+Also a change (be it change in attribute or file content) to `./bin/my-app-tools` binary to handle the config layout change.
```
wking: The previous sentence is imperative, telling the reader to add or remove.  I think this one should be as well, and tell the reader to [change the binary][1].

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77783475

vbatts:
@philips yeah for sure. I can carry-over your commit.

layer.md
wking: On Wed, Sep 07, 2016 at 06:31:22AM -0700, Vincent Batts wrote:
> > +Additions and Modifications are reflected the same in the changeset tar archive.
>
> I'm not inclined to just stick with layer. As Layer is generic
> enough to refer to the on disk representation, where changeset is
> the specific mechanics being described.

If you don't like “Layer tar archive”, how about following the pattern
you use in the rest of this file and just saying “tar archive”?

And if you see an important distinction between “layer (changeset)”
and “changeset”, you should probably file a PR rolling back #243 and
explaining the distinction.

layer.md
```diff
@@ -1,89 +1,199 @@
...
+* User ID (`uid`)
+    * User Name (`uname`) is secondary to `uid`
+* Group ID (`gid `)
+    * Group Name (`gname`) is secondary to `gid`
+* Mode (`mode`)
+* Extended Attributes (`xattrs`)
+* Symlink reference (`linkname`)
+
+[Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used.
+Sparse files lack consistent support (various tar implementations and libraries) and will therefore appear to mutate regular files on creating or applying of a layer changeset.
```
wking: I'm still not clear on the mutation bit.  Can we just say:

> [Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they lack consistent support across tar implementations.

And if it's a library thing, it sounds like its in [a similar boat with `xattrs` in the recent past][1].  Can we just punt all of this to user discretion and say “there isn't a current spec for tar, so consider portability before using the fancy stuff” ;).

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77826808

layer.md
wking: On Wed, Sep 07, 2016 at 06:58:28AM -0700, Vincent Batts wrote:
> There is no particular tar spec.

Hmm, it looks like there was a run at trying to get a formal tar media
type in 2014 [1], but I don't see it listed by the IANA [2].  Without
such a spec, validating layer blobs is going to be tricky (Alice: “My
tar library cannot parse your layer!”  Bob: “So what?”).  In the
absence of a tar spec, I see two choices:

a. Punt the specifics to users [3].  Lots of freedom, but
   application/vnd.oci.image.layer.tar+gzip becomes unenforceble and
   there are likely portablility edge cases (e.g. sparse files [4]).

b. Require a specified format like pax [5].  Implementers have to
   figure out how to get their “tar” library to write pax, or have to
   use a pax-specific library.  But
   application/vnd.oci.image.layer.pax+gzip is enforcable and portable
   between all image-spec implementations.

[1]: https://www.ietf.org/mail-archive/web/media-types/current/msg00592.html
[2]: http://www.iana.org/assignments/media-types/media-types.xhtml#application
[3]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77852725
[4]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77822800
[5]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13

layer.md
wking: On Wed, Sep 07, 2016 at 06:55:36AM -0700, Vincent Batts wrote:
> >  
> >  ```
> > -c3167915dc9d/
> > +rootfs-c9d-v1/
> 
> uniqueness. We could call it `fart-jar/`. 

I don't have and of rootfs, rootfs-v1, rootfs-c9d-v1, or fart-jar in
my home directory ;).  I think rootfs (where you have rootfs-c9d-v1)
and rootfs-v1 (where you have rootfs-c9d-v1.s1) are best because:

a. They include ‘rootfs’, which is where layers will be unpacked by
   default [1].
b. They distinguish between the mutable working directory (rootfs) and
   a snapshot of it (rootfs-v1) to make it easy to talk about changes.

[1]: https://github.com/opencontainers/image-spec/blame/d3cfb991ee3ff9e7c9570027cb9696213cfcee00/cmd/oci-image-tool/man/oci-image-tool-create-runtime-bundle.1.md#L24

layer.md
wking: On Wed, Sep 07, 2016 at 08:34:34AM -0700, Vincent Batts wrote:
> `-A` is acls, which are stored in xattrs, not directly implied via `-X`.

So when you say ‘xattrs’, you mean a specific tar header field?  Not
what rsync considers extended attributes (and covers by -X)?

> `-H` makes a difference, regardless of whether it's listed as a
> required attribute.

I agree that these “make a difference”.  I'm just not sure why you
pick one set of attributes to preserve here while listing another set
in the “File Attributes” section.  Is “File Attributes” a required
minimum to preserve?  If implementations can extend it to preserve
more attributes, how does that extension work?  This fits in with
“should we specify a tar-ish spec [1]?”.

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77867267

layer.md
wking: On Wed, Sep 07, 2016 at 08:20:43AM -0700, Vincent Batts wrote:
> > +* [cp(1)](http://linux.die.net/man/1/cp): `cp -pav rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
>
> not sure what to say here. Most folks are on GNU, but if POSIX is
> the bar then `cp` may not be an option.

So if xattr is required, then you should be suggesting ‘cp -a …’ or
‘cp -av’ here.  -p would be redundant.

vbatts:
Closes #231

vbatts:
Fixes #220

vbatts:
now git show --check fails on travis? but not for me locally... does someone have an ubuntu box to run this on to see if git show --check reports issues for them?

wking:
On Wed, Sep 07, 2016 at 11:53:32AM -0700, Vincent Batts wrote:
> does someone have an ubuntu box to run this on to see if git show
> --check reports issues for them?

  $ git --version
  git version 2.1.0.60.g85f0837
  $ git show --oneline --check 6059128
  6059128 layer: add introductory comments
  config.md:64: trailing whitespace.
  +{  
  config.md:74: trailing whitespace.
  +        "ExposedPorts": {  
  config.md:77: trailing whitespace.
  +        "Env": [

vbatts:
strange. @philips did not touch config.md, and my git did not report any issues with that commit...
```bash
[vbatts@bananaboat] {clarify_tar_entry_order} ~/src/opencontainers/image-spec$ git --version
git version 2.9.0
```

Regardless, I fixed those spaces and pushed an update. PTAL

wking:
On Wed, Sep 07, 2016 at 12:45:31PM -0700, Vincent Batts wrote:
> strange. @philips did not touch config.md…

  $ git show --oneline --stat 6059128a66
  6059128 layer: add introductory comments
   config.md        | 366 +++++++++++++++++++++++++++++++++++++++++++++++++++++++
   layer.md         |  15 ++-
   serialization.md | 366 -------------------------------------------------------
   3 files changed, 377 insertions(+), 370 deletions(-)
  $ git show --oneline --stat -M50 6059128a66
  6059128 layer: add introductory comments
   serialization.md => config.md |  0
   layer.md                      | 15 +++++++++++----
   2 files changed, 11 insertions(+), 4 deletions(-)

It was renamed.  Maybe your Git has automatic rename detection without
needing a hint like -M50.

In any case, while I like the serialization.md → config.md rename (and
it was mentioned in the IRC chat copied in #231 [1]), I think it may
have been accidental in #231 and that we should drop it here.  It
doesn't really impact the layer rewrite, and it's half-baked:

  $ git describe --always origin/pr/255
  v0.4.0-82-gce722f3
  wking@ullr ~/src/opencontainers/image-spec $ git grep serialization.md v0.4.0-82-gce722f3
  v0.4.0-82-gce722f3:Makefile:    serialization.md
  v0.4.0-82-gce722f3:README.md:This specification defines how to create an OCI Image, which will generally be done by a build system, and output an [image manifest](manifest.md), a [filesystem serialization](serialization.md), and an [image configuration](serialization.md#image-json-description). At a high level the image manifest contains metadata about the contents and dependencies of the image including the content-addressable identity of one or more filesystem serialization archives that will be unpacked to make up the final runnable filesystem. The image configuration includes information such as application arguments, environments, etc. The combination of the image manifest, image configuration, and one or more filesystem serializations is called the "OCI Image".
  v0.4.0-82-gce722f3:README.md:* An [image manifest](manifest.md) and [filesystem serialization](serialization.md) (base layer)
  v0.4.0-82-gce722f3:manifest.md:    The referenced configuration object is a JSON blob that the runtime uses to set up the container, see [Image JSON Description](serialization.md#image-json-description).
  v0.4.0-82-gce722f3:media-types.md:- `application/vnd.oci.image.config.v1+json`: [Container config JSON](serialization.md)
  v0.4.0-82-gce722f3:schema/spec_test.go: validate(t, "../serialization.md")

[1]: https://github.com/opencontainers/image-spec/pull/231#issue-174394513

vbatts:
fair

README.md
```diff
@@ -2,7 +2,7 @@
...
+This specification defines how to create an OCI Image, which will generally be done by a build system, and output an [image manifest](manifest.md), a [filesystem serialization](config.md), and an [image configuration](config.md#image-json-description). At a high level the image manifest contains metadata about the contents and dependencies of the image including the content-addressable identity of one or more filesystem serialization archives that will be unpacked to make up the final runnable filesystem. The image configuration includes information such as application arguments, environments, etc. The combination of the image manifest, image configuration, and one or more filesystem serializations is called the "OCI Image".
```
wking: I'd rather handle the `serialization.md` → `config.md` rename in a separate commit, instead of making it an unrelated side-effect of cd62032f (layer: add introductory comments, 2016-08-31).  But either way, if we're handling the rename in this PR I think we want “filesystem serialization” → “image configuration” too.  All of these JSON documents are serializing some portion of the image.  The thing that's unique to application/vnd.oci.image.config.v1+json is that it also contains `User`, `Memory`, etc.

wking: Actually, I think the “filesystem serialization” entry is supposed to point to `layer.md`.

README.md
```diff
@@ -18,7 +18,7 @@ The [Media Types](media-types.md) document is a starting point to understanding
 
 The high level components of the spec include:
 
-* An [image manifest](manifest.md) and [filesystem serialization](serialization.md) (base layer)
+* An [image manifest](manifest.md) and [filesystem serialization](config.md) (base layer)
```
wking: I think *this* “filesystem serialization” [should be “image config”][1].

[1]: https://github.com/opencontainers/image-spec/pull/255/files#r77894629

config.md
```diff
@@ -61,7 +61,7 @@ This specification uses the following terms:
 Here is an example image JSON file:
```
wking: The title for this page (up at line one) should probably be adjusted to follow the rename:

> # Image Configuration

or some such.

vbatts: I did not intend to deal with renaming. But to pull in Brandon's commit,
which only partially handled the renaming.

Now it is here and can be solved in one go. Otherwise I can drop he commit
and we can continue with the overlapping PRs.

wking: On Wed, Sep 07, 2016 at 01:22:18PM -0700, Vincent Batts wrote:
> Otherwise I can drop the commit and we can continue with the
> overlapping PRs.

@philips, can we just drop the renaming portion of your #231 (and spin
that out into a separate PR)?

layer.md
philips: I agree with @wking's new wording.

config.md
philips: @wking @vbatts dropped the renaming https://github.com/opencontainers/image-spec/pull/231/files

vbatts:
rebased on the updated #231, which removes the file rename.

PTAL

vbatts:
(tests pass)

philips:
LGTM

layer.md
```diff
@@ -1,89 +1,201 @@
-# Creating an Image Filesystem Changeset
+# Image Filesystem Changeset
```
wking: Can we put the word “Layer” in here?  ;)

philips: +1

philips:
@vbatts also, I didn't realize this change was missed in the rebase https://github.com/opencontainers/image-spec/pull/255/files#r77901622

philips:
REJECT

philips:
Rejected

layer.md
```diff
@@ -1,89 +1,201 @@
...
+* directories
+* sockets
+* symbolic links
+* block devices
+* character devices
+* FIFOs
+
+### File Attributes
+
+File attributes that are provided for Additions and Modifications include:
```
wking: We should replace “are provided” with something testable.  I expect at least:

> Implementations MUST support the following file attributes for Additions and Modifications:

Although I'd [still rather get out of the file-attribute-requiring business or cite a tar-like spec such as pax][1].

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77867267

layer.md
```diff
@@ -1,89 +1,201 @@
...
+./bin/my-app-tools
 ```
 
-To make changes to the filesystem of this container image, create a new directory, such as `f60c56784b83`, and initialize it with a snapshot of the parent image's root filesystem, so that the directory is identical to that of `c3167915dc9d`.
-NOTE: a copy-on-write or union filesystem can make this very efficient:
+### Populate a Comparison Filesystem
+
+Create a new directory and initialize it with an copy or snapshot of the prior root filesystem.
+Example commands that can preserve [file attributes](#file-attributes) to make this copy are:
+* [cp(1)](http://linux.die.net/man/1/cp): `cp -pav rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
```
wking: I [still think `-p` is redundant here][1].

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77869765

layer.md
```diff
@@ -1,89 +1,201 @@
...
+* [rsync(1)](http://linux.die.net/man/1/rsync):  `rsync -avHAX rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
```
wking: I [still don't like copying ACLs with `-A` when we don't mention ACLs as a preserved attribute[1].

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77868887

layer.md
```diff
@@ -1,89 +1,201 @@
...
+* [tar(1)](http://linux.die.net/man/1/tar): `mkdir rootfs-c9d-v1.s1 && tar --acls --xattrs -C rootfs-c9d-v1/ -c . | tar -C rootfs-c9d-v1.s1/ --acls --xattrs -x` (including `--selinux` where supported)
+
+Any [changes](#change-types) to the snapshot MUST NOT change or affect the directory it was copied from.
+
+For example `rootfs-c9d-v1.s1` is an identical snapshot of `rootfs-c9d-v1`.
+In this way `rootfs-c9d-v1.s1` is prepared for updates and alterations.
```
wking: It seems strange to me to be updating the snapshot.  Can we replace `rootfs-c9d-v1.s1` with `rootfs-c9d-v2`?  Or keep `rootfs-c9d-v1.s1` as the snapshot and continue working in the `rootfs-c9d-v1` directory?  If, for some reason, you don't buy [my argument for using `rootfs` and `rootfs-v1`][1] ;).

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77868017

layer.md
vbatts: @wking your chain of links and footnotes _kill_ me.
The options your proposing are getting in the weeds. The requirement are known and have been worked with up to this point. If there are features needed to be added to tools, I'm certain folks will do what they need to do.

layer.md
vbatts: I'll use the new wording, but punting on xattrs as too "fancy" is a blocker, and hardly to wait for another spec to exist just to point at.

layer.md
vbatts: sorry you think it's redundant.

layer.md
vbatts: it's part of the xattrs. the fact that rsync has multiple flags is irrelevant. Just providing the `-X` flag does not include the ACL xattrs when copying.

vbatts:
@philips broken link? what was missed in the rebase?

layer.md
wking: On Wed, Sep 07, 2016 at 06:32:48PM -0700, Vincent Batts wrote:
> > +File attributes that are provided for Additions and Modifications include:
> 
> @wking your chain of links and footnotes _kill_ me.

Sorry, it's hard to preserve the continuity of a sub-thread when
GitHub keeps hiding it on us :p.  Links and footnotes are the best I
can do to make the older conversation discoverable without rehashing
all of it.

> The options your proposing are getting in the weeds. The requirement
> are known and have been worked with up to this point. If there are
> features needed to be added to tools, I'm certain folks will do what
> they need to do.

That sounds like (a) (Punt the specifics to users) from [1], and I'm
ok with that.  However, in that case, I see no particular use to call
out specific file types or attributes.  If a user wants to put some
oddball filetype/attribute into the tarball, more power to them.  If
an implementer wants to only support regular files, directories, and
mode (ignoring or erroring on symlinks, owner, mtime, etc.), more
power to them as well.  Sitting in the middle with a half-done spec
(e.g. requiring ‘xattrs’ but not defining it) doesn't seem helpful.

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77867267

layer.md
wking: On Wed, Sep 07, 2016 at 06:48:09PM -0700, Vincent Batts wrote:
> > +* [cp(1)](http://linux.die.net/man/1/cp): `cp -pav rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
> 
> sorry you think it's redundant.

Heh, is that:

a. You're wrong, it's not redundant or
b. It's redundant, but I like it that way?

Either way, could you elaborate?

layer.md
wking: On Wed, Sep 07, 2016 at 07:02:13PM -0700, Vincent Batts wrote:
> > +* [rsync(1)](http://linux.die.net/man/1/rsync):  `rsync -avHAX rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
> 
> it's part of the xattrs. the fact that rsync has multiple flags is
> irrelevant. Just providing the `-X` flag does not include the ACL
> xattrs when copying.

Which means that, as far as rsync is concerned, ACLs are not a subset
of xattrs.  However, you consider ACLs to be a subset of xattrs.
There's a disconnect there, and if we're going to require xattr
support [1], I think we want to be explicit about what we're requiring
when we say ‘xattrs’.

[1]: https://github.com/opencontainers/image-spec/pull/255/files#diff-dc2bb3b4eddcffe96e682193926e8ff6R38

wking: On Wed, Sep 07, 2016 at 07:51:59PM -0700, W. Trevor King wrote:
> There's a disconnect there, and if we're going to require xattr
> support [1], I think we want to be explicit about what we're requiring
> when we say ‘xattrs’.

Or we can punt and not require xattrs or any other attributes to be
supported, allowing layer authors to use them and implementers to
support them and other tar features as they see fit [1].

[1]: https://github.com/opencontainers/image-spec/pull/255#discussion_r77938673

layer.md
```diff
@@ -1,89 +1,200 @@
...
         my-app-tools
 ```
 
-This example change is going to add a configuration directory at `/etc/my-app.d` which contains a default config file.
-There's also a change to the `my-app-tools` binary to handle the config layout change.
-The `f60c56784b83` directory then looks like this:
+See [Change Types](#change-types) for more details on changes.
+
+For example, add a directory at `/etc/my-app.d` containing a default config file, removing the existing config file.
+Also a change (be it change in attribute or file content) to `./bin/my-app-tools` binary to handle the config layout change.
```
coolljt0725: `be it change` ? typo here? I don't quite understand this from my english

wking: It may be ok as it stands, but I think it would read more clearly if we dropped "be it change".

layer.md
vbatts: The man pages read it to be redundant, though I include `-p` with `-a` because historically I've found that it was needed. I have no proof of such. I'll remove the `-p`, this is an inconsequential helper text.

layer.md
vbatts: what do you mean updating the snapshot?
I do not buy your arguments that are based on the contents of your home directory. Sorry.

philips:
LGTM

@vbatts the link was to a comment that is now hidden because you fixed the issue. Thanks!

layer.md
wking: On Thu, Sep 08, 2016 at 08:55:50AM -0700, Vincent Batts wrote:
> > +* [cp(1)](http://linux.die.net/man/1/cp): `cp -pav rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
>
> … I include `-p` with `-a` because historically I've found that it
> was needed.

If somebody stumbles across a problem because they didn't specify -p,
we can add it back in with steps to reproduce their problem.

layer.md
wking: On Thu, Sep 08, 2016 at 08:57:34AM -0700, Vincent Batts wrote:
> > +For example `rootfs-c9d-v1.s1` is an identical snapshot of `rootfs-c9d-v1`.
> > +In this way `rootfs-c9d-v1.s1` is prepared for updates and alterations.
> 
> what do you mean updating the snapshot?

I want:

  $ mkdir rootfs
  $ populate.sh rootfs
  $ cp -a rootfs rootfs-v1  # freeze out a snapshot for the first layer
  $ tar -C rootfs-v1 -cf layer-0.tar
  $ mutate.sh rootfs
  $ cp -a rootfs rootfs-v2  # freeze out the snapshot for the second layer
  $ generate-layer.sh rootfs-v1 rootfs-v2 >layer-1.tar

So you're working in rootfs, but the rootsf-v* are frozen and
unchanging.

> I do not buy your arguments that are based on the contents of your
> home directory. Sorry.

Where do you expect folks to be working on this that they already have
a ‘rootfs’ directory around?  I'm saying [1] that I don't see a need
to pick something more unique than that.

[1]: https://github.com/opencontainers/image-spec/pull/255/files#r77868017

philips:
@opencontainers/image-spec-maintainers please take a look and LGTM

layer.md
vbatts: i'm not seeing where this example is instructing changes to happen in the original directory after the snapshot is made.
Is that even possible? surely. Then when the changeset is made, the changeset at that moment will reflect the changes in the parent.
I specifically did not say anything to the nature of "once the snapshot/copy is made then the parent is immutable", because it is not.

layer.md
```diff
@@ -1,89 +1,200 @@
...
+* Modifications
+* Removals
+
+Additions and Modifications are represented the same in the changeset tar archive.
+
+Removals are represented using "[whiteout](#whiteouts)" file entries (See [Representing Changes](#representing-changes)).
+
+### File Types
+
+Throughout this document section, the use of word "files" includes:
```
stevvooe: "or entries" maybe?

layer.md
```diff
@@ -1,89 +1,200 @@
...
+Where supported, MUST include file attributes for Additions and Modifications include:
+
+* Modification Time (`mtime`)
+* User ID (`uid`)
+    * User Name (`uname`) *secondary to `uid`*
+* Group ID (`gid `)
+    * Group Name (`gname`) *secondary to `gid`*
+* Mode (`mode`)
+* Extended Attributes (`xattrs`)
+* Symlink reference (`linkname`)
```
stevvooe: Any word here on how hardlinks might be supported?

layer.md
stevvooe: I just submitted a PR to your branch using the edit feature... give that a try?

layer.md
```diff
@@ -1,89 +1,200 @@
...
+rootfs-c9d-v1/
 ```
 
+### Populate Initial Filesystem
+
 Files and directories are then created:
 
 ```
-c3167915dc9d/
+rootfs-c9d-v1/
```
stevvooe: nit: why change the root name?

stevvooe:
LGTM

Nits can be taken care of in later PRs.

Thanks for doing this. This was greatly needed.

layer.md
vbatts: Not yet. It was undefined before, and getting that wording right seems like a further conversation piece.

layer.md
vbatts: ah yeah. Now that this is merge'able, can I carry that to a new PR?

layer.md
stevvooe: Should we file an issue?

PR #257, opened 2016-09-05
Fix media type on unpacking layer

coolljt0725:

The media type on unpacking layer should be `MediaTypeImageLayer`

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -89,7 +89,7 @@ func (m *manifest) validate(w walker) error {
 
 func (m *manifest) unpack(w walker, dest string) error {
 	for _, d := range m.Layers {
-		if d.MediaType != string(schema.MediaTypeImageConfig) {
+		if d.MediaType != string(schema.MediaTypeImageLayer) {
 			continue
 		}
 
--- a/image/manifest_test.go
+++ b/image/manifest_test.go
@@ -4,6 +4,8 @@ import (
 	"archive/tar"
 	"bytes"
 	"compress/gzip"
+	"crypto/sha256"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
@@ -46,5 +48,63 @@ func TestUnpackLayerDuplicateEntries(t *testing.T) {
 	defer os.RemoveAll(tmp2)
 	if err := unpackLayer(tmp2, r); err != nil && !strings.Contains(err.Error(), "duplicate entry for") {
 		t.Fatalf("Expected to fail with duplicate entry, got %v", err)
+	}
+}
+
+func TestUnpackLayer(t *testing.T) {
+	tmp1, err := ioutil.TempDir("", "test-layer")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmp1)
+	err = os.MkdirAll(filepath.Join(tmp1, "blobs", "sha256"), 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tarfile := filepath.Join(tmp1, "blobs", "sha256", "test.tar")
+	f, err := os.Create(tarfile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	gw := gzip.NewWriter(f)
+	tw := tar.NewWriter(gw)
+
+	tw.WriteHeader(&tar.Header{Name: "test", Size: 4, Mode: 0600})
+	io.Copy(tw, bytes.NewReader([]byte("test")))
+	tw.Close()
+	gw.Close()
+	f.Close()
+
+	// generate sha256 hash
+	h := sha256.New()
+	file, err := os.Open(tarfile)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer file.Close()
+	_, err = io.Copy(h, file)
+	if err != nil {
+		t.Fatal(err)
+	}
+	err = os.Rename(tarfile, filepath.Join(tmp1, "blobs", "sha256", fmt.Sprintf("%x", h.Sum(nil))))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	testManifest := manifest{
+		Layers: []descriptor{descriptor{
+			MediaType: "application/vnd.oci.image.layer.tar+gzip",
+			Digest:    fmt.Sprintf("sha256:%s", fmt.Sprintf("%x", h.Sum(nil))),
+		}},
+	}
+	err = testManifest.unpack(newPathWalker(tmp1), filepath.Join(tmp1, "rootfs"))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = os.Stat(filepath.Join(tmp1, "rootfs", "test"))
+	if err != nil {
+		t.Fatal(err)
 	}
 }
--- a/image/manifest_test.go
+++ b/image/manifest_test.go
@@ -1,3 +1,17 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package image
 
 import (


runcom:
would be great to have a test for this :( not sure how this got past CI or manual testing

coolljt0725:
@runcom  Having a test for this is great, I'll try

coolljt0725:
@vbatts  @runcom  @philips  PTAL, if it's ok, can we merge it first and then add test later? This block a lot of tests

image/manifest.go
```diff
@@ -89,7 +89,7 @@ func (m *manifest) validate(w walker) error {
 
 func (m *manifest) unpack(w walker, dest string) error {
 	for _, d := range m.Layers {
-		if d.MediaType != string(schema.MediaTypeImageConfig) {
+		if d.MediaType != string(schema.MediaTypeImageLayer) {
 			continue
```
wking: This should be an error, not a `continue`.  Changing this to an error fixes the “wrong media-type” issue discussed in #261 (but not currently fixed by that PR).

wking: Although the spec [does not currently say that layers MUST have a particular type][1], so this is currently an:

> unimplemented: unpacking layer with media-type %q

I expect the spec should clarify if it places bounds (upper or lower) on manifest layer[] media types, and once that happens we can adjust this to match.

[1]: https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions

coolljt0725: @wking  if the layers MUST have a particular type is not said in the spec currently, I think we should make it clear in the spec first.

wking: On Wed, Sep 07, 2016 at 04:58:52AM -0700, Lei Jitang wrote:
> > @@ -89,7 +89,7 @@ func (m *manifest) validate(w walker) error {
> >  
> >  func (m *manifest) unpack(w walker, dest string) error {
> >  	for _, d := range m.Layers {
> > -		if d.MediaType != string(schema.MediaTypeImageConfig) {
> > +		if d.MediaType != string(schema.MediaTypeImageLayer) {
> >  			continue
>
> @wking if the layers MUST have a particular type is not said in the
> spec currently, I think we should make it clear in the spec first.

If the layers MUST have a particular type, then you should error out
here with “your manifest is broken”.  If the spec does not limit the
allowed types (like now), then you should error out here with
“unimplemented type”.  In no case that I can see should we be silently
ignoring an unrecognized type.

wking:
I nominate this PR for inclusion in 0.5.0.  We should avoid releasing
tooling when we know unpacking is completely broken ;).

image/manifest.go
coolljt0725: @wking   If the spec does not limit the allowed types, I don't think we should error out on unpacking, we should just
unpacking the allowed typed layers.  But I think this is another issue we should discuss more not related to this pr

coolljt0725:
test added

image/manifest_test.go
```diff
@@ -48,3 +64,60 @@ func TestUnpackLayerDuplicateEntries(t *testing.T) {
 		t.Fatalf("Expected to fail with duplicate entry, got %v", err)
 	}
 }
+func TestUnpackLayerRemovePartialyUnpackedFile(t *testing.T) {
```
wking: Missing blank line before this function?

image/manifest.go
wking: On Wed, Sep 07, 2016 at 08:46:36PM -0700, Lei Jitang wrote:
> > @@ -89,7 +89,7 @@ func (m *manifest) validate(w walker) error {
> >  
> >  func (m *manifest) unpack(w walker, dest string) error {
> >  	for _, d := range m.Layers {
> > -		if d.MediaType != string(schema.MediaTypeImageConfig) {
> > +		if d.MediaType != string(schema.MediaTypeImageLayer) {
> >  			continue
> 
> @wking If the spec does not limit the allowed types, I don't think
> we should error out on unpacking, we should just unpacking the
> allowed typed layers.  But I think this is another issue we should
> discuss more not related to this pr

If there's any confusion, I'm fine punting this to another PR.  The
media type fix is critical and obviously right.  The continue→error
fix only matters for oddball manifests, so it's much less important.

image/manifest_test.go
```diff
@@ -48,3 +64,60 @@ func TestUnpackLayerDuplicateEntries(t *testing.T) {
...
+	tmp1, err := ioutil.TempDir("", "test-layer")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmp1)
+	err = os.MkdirAll(filepath.Join(tmp1, "blobs", "sha256"), 0700)
+	if err != nil {
+		t.Fatal(err)
+	}
+	tarfile := filepath.Join(tmp1, "blobs", "sha256", "test.tar")
```
wking: `test.tar` doesn't look like a hex hash ;).  I'd rather build this layer outside of `blobs/sha256` and then rename it into position once we've completed it and can get a content-addressable hash.

And “create image-layout fodder for testing” is likely to be useful stuff in lots of places.  Can we make layer-generator, etc. helpers instead of baking it into `TestUnpackLayerRemovePartialyUnpackedFile`.  I'm missing #159 and 87d6a53 ;).

coolljt0725: @wking  
>test.tar doesn't look like a hex hash ;). I'd rather build this layer outside of blobs/sha256 and then rename it into position once we've completed it and can get a content-addressable hash.

`test.tar` will be rename to a hex hash below.

wking: On Wed, Sep 07, 2016 at 09:15:42PM -0700, Lei Jitang wrote:
> > +	tarfile := filepath.Join(tmp1, "blobs", "sha256", "test.tar")
> >
> > test.tar doesn't look like a hex hash ;). I'd rather build this
> > layer outside of blobs/sha256 and then rename it into position
> > once we've completed it and can get a content-addressable hash.
>
> `test.tar` will be rename to a hex hash below.

I'd rather avoid the temporary invalidity by placing it somewhere else
first.  Can we make a ./tmp directory for scratch space and use:

  tempDir := filepath.Join(tmp1, "tmp")
  err := os.Mkdir(tempDir, 0777)
	if err != nil {
    t.Fatal(err)
  }
  f, err := ioutil.TempFile(tempDir, "blob-")

And then use f.Name() where you were using tarfile?

coolljt0725: @wking 
> I'd rather avoid the temporary invalidity by placing it somewhere else
first.

I don't know if I got your point, I did it similarly like you said at the beginning, put the `tar` file under
`tmp1` and then copy it to `blobs/sha256/` after renaming it to `hex hash`. 
BUt to avoid a unnecessary copy, so I put the `tar` file to the destination dir at the beginning and rename it 
after calculating hash.

wking: On Wed, Sep 07, 2016 at 11:45:29PM -0700, Lei Jitang wrote:
> I don't know if I got your point, I did it similarly like you said
> at the beginning, put the `tar` file under `tmp1` and then copy it
> to `blobs/sha256/` after renaming it to `hex hash`.  BUt to avoid a
> unnecessary copy, so I put the `tar` file to the destination dir at
> the beginning and rename it after calculating hash.

You can use rename no matter what the source and destination are, and
only have to pay for a copy if they are on different filesystems.  In
this case, tmp1/…, {tmp1}/tmp/… and {tmp1}/blobs/sha256/… are all in
the same filesystem, so rename will be cheap in all cases.

coolljt0725:
ping @philips  @vbatts  @stevvooe  @s-urbaniak  PTAL,  we can't use unpack without this

runcom:
> ping @philips @vbatts @stevvooe @s-urbaniak PTAL, we can't use unpack without this

LGTM - received this very same bug report on IRC and was hoping for this to be merged soon(ish), thanks

image/manifest.go
runcom: > Although the spec does not currently say that layers MUST have a particular type, so this is currently an:
> 
> unimplemented: unpacking layer with media-type %q
> I expect the spec should clarify if it places bounds (upper or lower) on manifest layer[] media types, and once that happens we can adjust this to match.

is there an issue for this?

wking: On Fri, Sep 09, 2016 at 03:07:52AM -0700, Antonio Murdaca wrote:
> > @@ -89,7 +89,7 @@ func (m *manifest) validate(w walker) error {
> >  
> >  func (m *manifest) unpack(w walker, dest string) error {
> >  	for _, d := range m.Layers {
> > -		if d.MediaType != string(schema.MediaTypeImageConfig) {
> > +		if d.MediaType != string(schema.MediaTypeImageLayer) {
> >  			continue
> >
> > Although the spec does not currently say that layers MUST have a
> > particular type, so this is currently an:
> > 
> > unimplemented: unpacking layer with media-type %q
> >
> > I expect the spec should clarify if it places bounds (upper or
> > lower) on manifest layer[] media types, and once that happens we
> > can adjust this to match.
> 
> is there an issue for this?

There are ~50 million open validation issues at the moment, so I
haven't opened up one specifically for this ;).  I try and bring it up
where it impacts an existing issue/PR.  Most recently in #286.

philips:
LGTM

We should fix up the test.tar thing and use a tmp dir; but we can do it on a followup.

[Approved with PullApprove](https://pullapprove.com/opencontainers/image-spec/pull-request/257/?utm_source=github-pr&utm_medium=comment-badge&utm_campaign=opencontainers/image-spec)

vbatts:
LGTM

Issue #260, opened 2016-09-05
oci-image-tool: check media types under refs/


runcom:
Came across this and oci-image-tool wasn't warning me when a reference descriptor under refs/ wasn't containing a valid media type (especially because it was the empty string with my testing).

Should oci-image-tool take care of this as well?

/cc @s-urbaniak @vbatts @philips @stevvooe

runcom:
(likewise I think a check of oci-layout to be version == the spec version tested against is required to be added to the tool)

coolljt0725:
I +1 on this, I came across this when I  `oci-image-tool validate` an old image layout which layer media type is still `application/vnd.oci.image.serialization.rootfs.tar.gzip`, I think it should throw out a warning or error

runcom:
I'll fix it if people agree

philips:
Yes, oci-image-tool should error out on this.

runcom:
took care of this at https://github.com/opencontainers/image-spec/issues/262

PR #262, opened 2016-09-06
oci-image-tool: validate descriptors MediaType

runcom:
Fix https://github.com/opencontainers/image-spec/issues/260

/cc @coolljt0725 @s-urbaniak @xiekeyang

### DIFF
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -73,7 +73,17 @@ func findDescriptor(w walker, name string) (*descriptor, error) {
 	}
 }
 
-func (d *descriptor) validate(w walker) error {
+func (d *descriptor) validate(w walker, mts []string) error {
+	var found bool
+	for _, mt := range mts {
+		if d.MediaType == mt {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return fmt.Errorf("invalid descriptor MediaType %q", d.MediaType)
+	}
 	switch err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
 		if info.IsDir() {
 			return nil
--- a/image/config.go
+++ b/image/config.go
@@ -26,28 +26,12 @@ import (
 	"strings"
 
 	"github.com/opencontainers/image-spec/schema"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/pkg/errors"
 )
 
-type cfg struct {
-	User         string
-	Memory       int64
-	MemorySwap   int64
-	CPUShares    int64 `json:"CpuShares"`
-	ExposedPorts map[string]struct{}
-	Env          []string
-	Entrypoint   []string
-	Cmd          []string
-	Volumes      map[string]struct{}
-	WorkingDir   string
-}
-
-type config struct {
-	Architecture string `json:"architecture"`
-	OS           string `json:"os"`
-	Config       cfg    `json:"config"`
-}
+type config v1.Image
 
 func findConfig(w walker, d *descriptor) (*config, error) {
 	var c config
--- a/image/image.go
+++ b/image/image.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 )
 
@@ -43,6 +44,11 @@ func Validate(tarFile string, refs []string, out *log.Logger) error {
 	return validate(newTarWalker(f), refs, out)
 }
 
+var validRefMediaTypes = []string{
+	v1.MediaTypeImageManifest,
+	v1.MediaTypeImageManifestList,
+}
+
 func validate(w walker, refs []string, out *log.Logger) error {
 	for _, r := range refs {
 		ref, err := findDescriptor(w, r)
@@ -50,7 +56,7 @@ func validate(w walker, refs []string, out *log.Logger) error {
 			return err
 		}
 
-		if err = ref.validate(w); err != nil {
+		if err = ref.validate(w, validRefMediaTypes); err != nil {
 			return err
 		}
 
@@ -97,7 +103,7 @@ func unpack(w walker, dest, refName string) error {
 		return err
 	}
 
-	if err = ref.validate(w); err != nil {
+	if err = ref.validate(w, validRefMediaTypes); err != nil {
 		return err
 	}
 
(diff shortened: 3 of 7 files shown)


image/descriptor.go
```diff
@@ -73,7 +73,17 @@ func findDescriptor(w walker, name string) (*descriptor, error) {
...
+	var found bool
+	for _, mt := range mts {
+		if d.MediaType == mt {
+			found = true
+			break
+		}
+	}
+	if !found {
+		return fmt.Errorf("invalid descriptor MediaType %q", d.MediaType)
+	}
```
xiekeyang: if inner func is better here?
```go
if !func(dmt string, mts []string) bool {
		for _, mt := range mts {
			if dmt == mt {
				return true
			}
		}
        return false
}(d.MediaType, mts) {
		return fmt.Errorf("invalid descriptor MediaType %q", d.MediaType)
}
```

runcom: I honestly don't like it

coolljt0725:
I like this :+1:

xiekeyang:
@runcom @coolljt0725 well, it is just my private custom, which bring little benefit.

runcom:
rebased!

philips:
LGTM

image/descriptor.go
stevvooe: You can just exit in line. No need for flag variable.

stevvooe:
LGTM

wking:
On Tue, Sep 06, 2016 at 01:28:02AM -0700, Antonio Murdaca wrote:
>   * *: use types from specs-go/

This commit should also replace the current ‘descriptor’ (in
image/descriptor.go) with Descriptor (from specs-go/descriptor.go).
In #159, I do that in f2dc0e2 (replacing descriptor.validate with
validateDescriptor, although eventually we'll want a public function
for descriptor validation).  That commit is also doing a lot of
walker→engine replacing, but feel free to borrow anything that helps
with descriptor→Descriptor replacing.

image/descriptor.go
```diff
@@ -73,7 +73,17 @@ func findDescriptor(w walker, name string) (*descriptor, error) {
 	}
 }
 
-func (d *descriptor) validate(w walker) error {
+func (d *descriptor) validate(w walker, mts []string) error {
```
wking: I don't think this is the right API.  There interesting descriptor-validation cases for each descriptor property:

a. The referenced blob is exists in (or is missing from) CAS (`digest`).
b. The media type can (or cannot) be validated by this package (`mediaType`).
c. The referenced blob exists in CAS with the right (or wrong) size (`size`).

Those also apply recursively, if you want to walk the Merkle tree of blob ancestors.

If the user wants to validate a descriptor, they should just be able to call `image.ValidateDescriptor(walker, descriptor, recursive)` or [some such][1].  If they *also* want to restrict the media type of the root descriptor, they can do that outside of `ValidateDescriptor` (because there's not a convenient way to specify those restrictions recursively, i.e. “if an ancestor contains a manifest, only allow layer types from {set}”).

If you have an image-layout with a ref with an image/png media type pointing to a valid PNG blob, I `oci-image-tool validate image.tar png-ref` could do any of:

i. Print something like “unrecognized media type 'image/png'" and exit nonzero (“I don't know”).
ii. Confirm that the referenced blob is a PNG of the right size and exit zero (“This is valid”).
iii. Print something like “invalid media type 'image/png'" and exit nonzero (“This is not valid”).

It *is* a valid reference (so (i) and (ii) are ok, but (iii) is not).  The PNG is just not a reference that `oci-image-tool unpack` can handle.  I think the right API for that is to have a separate “unpackable” check bound to a flag.  So the following all pass:

    $ oci-image-tool validate --recursive --unpackable image.tar $VALID_MANIFEST_REF
    $ oci-image-tool validate --unpackable image.tar $BUSTED_ANCESTOR_MANIFEST_REF
    $ oci-image-tool validate --recursive image.tar $VALID_LAYER_REF
    $ oci-image-tool validate --recursive image.tar $VALID_CONFIG_REF
    $ oci-image-tool validate image.tar $VALID_MANIFEST_REF
    $ oci-image-tool validate image.tar $BUSTED_ANCESTOR_MANIFEST_REF
    $ oci-image-tool validate image.tar $VALID_LAYER_REF
    $ oci-image-tool validate image.tar $VALID_CONFIG_REF

and the following all fail:

    $ oci-image-tool validate --recursive --unpackable image.tar $BUSTED_ANCESTOR_MANIFEST_REF
    $ oci-image-tool validate --recursive --unpackable image.tar $VALID_LAYER_REF
    $ oci-image-tool validate --recursive --unpackable image.tar $VALID_CONFIG_REF
    $ oci-image-tool validate --unpackable image.tar $VALID_LAYER_REF
    $ oci-image-tool validate --unpackable image.tar $VALID_CONFIG_REF

[1]: https://github.com/opencontainers/image-spec/pull/262#issuecomment-245187243

runcom: Isn't it better to open another issue tracking this?

wking: On Wed, Sep 07, 2016 at 12:00:38AM -0700, Antonio Murdaca wrote:
> > @@ -73,7 +73,17 @@ func findDescriptor(w walker, name string) (*descriptor, error) {
> >  	}
> >  }
> >  
> > -func (d *descriptor) validate(w walker) error {
> > +func (d *descriptor) validate(w walker, mts []string) error {
> 
> Isn't it better to open another issue tracking this? 

You're adding ‘mts’ here, and I don't think we want it ;).  If folks
feel like the rest of this PR is a step forward, I'm ok with this
landing and me filing a subsequent PR to roll this part back out in
favor of --unpackable.

runcom:
@wking, can we not block PRs if you already took care of using the right descriptor struct in your PR? Also, it's better to have a new PR altogether instead of embedding that in #159.

wking:
On Wed, Sep 07, 2016 at 12:05:13AM -0700, Antonio Murdaca wrote:
> @wking, can we not block PRs if you already took care of using the
> right descriptor struct in your PR?

If you don't feel like covering it in this PR, it can land with #159,
or via some other followup PR, or never ;).  No big deal.  It just
seemed like a natural fit for a “use types from specs-go” commit.

> Also, it's better to have a new PR altogether instead of embedding
> that in #159.

#159 has floated for a lot longer than I'd initial expected.  I may
start splitting off smaller shifts like that, but I really don't like
walkers, so I'm trying to avoid touching them ;).

s-urbaniak:
@runcom LGTM :+1:

vbatts:
LGTM

PR #268, opened 2016-09-06
adds some description re: compat with docker schemas

mikebrow:
Adding a couple links to remove confusion regarding OCI being at v1 and docker at v2 for "fat" manifest lists and distribution (image) manifests.

Signed-off-by: Mike Brown <brownwm@us.ibm.com>

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -21,8 +21,9 @@ A client will distinguish a manifest list from an image manifest based on the Co
 
 - **`mediaType`** *string*
 
-  This REQUIRED property contains the MIME type of the manifest list.
+  This REQUIRED property contains the media type of the manifest list.
   For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.list.v1+json`.
+  For the media type(s) that this is compatible with see the [matrix](media-types.md#compatibility-matrix).
 
 - **`manifests`** *array*
 
@@ -122,8 +123,9 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 
 - **`mediaType`** *string*
 
-    This REQUIRED property contains the MIME type of the image manifest.
+    This REQUIRED property contains the media type of the image manifest.
     For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.v1+json`.
+    For the media type(s) that this is compatible with see the [matrix](media-types.md#compatibility-matrix).
 
 - **`config`** *[descriptor](descriptor.md)*
 


philips:
I would rather not duplicate information from the compatibility matrix. Can you just say "for a list of MIME types this is compatible with refer to the matrix?"

mikebrow:
I suppose I could say "for the MIME type that this is compatible with see the matrix." But this MIME type is only compatible with one other MIME type and, to me, it seems odd to make people jump to the "matrix" on a different page for the big reveal when the big reveal is just that OCI v1 = docker v2 wrt this MIME type. Maybe if more detail was provided in the matrix...

mikebrow:
Updated to address comments.

manifest.md
```diff
@@ -22,7 +22,9 @@ A client will distinguish a manifest list from an image manifest based on the Co
 - **`mediaType`** *string*
 
   This REQUIRED property contains the MIME type of the manifest list.
-  For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.list.v1+json`.
+  For this version of the specification, this MUST be set to
```
jonboulle: The rest of the file is not wrapped but instead one-sentence-per-line (or one-paragraph-per-line), could you do similarly here?

mikebrow: ok.. interesting style, updated to adhere to it. Out of curiosity is there a reason to not break sentences in a .md file?

mikebrow:
updated to adhere to non-wrapped style for sentences.

manifest.md
wking: On Wed, Sep 07, 2016 at 10:50:50AM -0700, Mike Brown wrote:
> > -  For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.list.v1+json`.
> > +  For this version of the specification, this MUST be set to
> 
> ok.. interesting style, updated to adhere to it. Out of curiosity is
> there a reason to not break sentences in a .md file?

There's a small bit of background in opencontainers/runtime-spec#82.

manifest.md
```diff
@@ -23,6 +23,7 @@ A client will distinguish a manifest list from an image manifest based on the Co
 
   This REQUIRED property contains the MIME type of the manifest list.
   For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.list.v1+json`.
+  For the MIME type that this is compatible with see the [matrix](media-types.md#compatibility-matrix).
```
wking: This doesn't have to be singular.  And most of the rest of the spec uses the generic “media type” instead of the email-specific “MIME type”.  So I'd suggest “MIME type” → “media types”.

mikebrow: Yeah MIME was in the prior sentence already so I just ran with it.  Will remove MIME type from the doc.

mikebrow:
Update the document to adhere to using "media type" vs "MIME."

wking:
24c10ea9 looks good to me.

philips:
LGTM

jonboulle:
lgtm

PR #273, opened 2016-09-07
fix bug: media types mismatch in manifest

xiekeyang:
It is passed validation on manifest part, even the config mediatype and layer mediatype inside is mismatch.
It had better to add media type item(config and layer) checking on manifest Validate.
bug case:
```json
{
    "annotations": null,
    "config": {
        "digest": "sha256:2b8fd9751c4c0f5dd266fcae00707e67a2545ef34f9a29354585f93dac906749",
        "mediaType": "application/vnd.oci.image.serialization.config.v1+json",
        "size": 1459
    },  
    "layers": [
        {   
            "digest": "sha256:8ddc19f16526912237dd8af81971d5e4dd0587907234be2b83e249518d5b673f",
            "mediaType": "application/vnd.oci.image.serialization.rootfs.tar.gzip",
            "size": 667590
        }   
    ],  
    "schemaVersion": 2
}
```

validate result is OK. But  "mediaType": "application/vnd.oci.image.serialization.rootfs.tar.gzip" and "mediaType": "application/vnd.oci.image.serialization.config.v1+json" have been unsupported.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -22,11 +22,13 @@ import (
 
 // Media types for the OCI image formats
 const (
-	MediaTypeDescriptor   Validator     = v1.MediaTypeDescriptor
-	MediaTypeManifest     Validator     = v1.MediaTypeImageManifest
-	MediaTypeManifestList Validator     = v1.MediaTypeImageManifestList
-	MediaTypeImageConfig  Validator     = v1.MediaTypeImageConfig
-	MediaTypeImageLayer   unimplemented = v1.MediaTypeImageLayer
+	MediaTypeDescriptor        Validator     = v1.MediaTypeDescriptor
+	MediaTypeManifest          Validator     = v1.MediaTypeImageManifest
+	MediaTypeManifestList      Validator     = v1.MediaTypeImageManifestList
+	MediaTypeImageConfig       Validator     = v1.MediaTypeImageConfig
+	MediaTypeImageLayer        unimplemented = v1.MediaTypeImageLayer
+	MediaTypeDockerImageConfig Validator     = "application/vnd.docker.container.image.v1+json"
+	MediaTypeDockerImageLayer  unimplemented = "application/vnd.docker.image.rootfs.diff.tar.gzip"
 )
 
 var (
--- a/cmd/oci-image-tool/validate.go
+++ b/cmd/oci-image-tool/validate.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"strings"
@@ -138,6 +139,9 @@ func (v *validateCmd) validatePath(name string) error {
 
 	switch typ {
 	case typeManifest:
+		if err := validManifestfile(name); err != nil {
+			return errors.Wrap(err, "invalid manifest format")
+		}
 		return schema.MediaTypeManifest.Validate(f)
 	case typeManifestList:
 		return schema.MediaTypeManifestList.Validate(f)
@@ -146,4 +150,20 @@ func (v *validateCmd) validatePath(name string) error {
 	}
 
 	return fmt.Errorf("type %q unimplemented", typ)
+}
+
+func validManifestfile(name string) error {
+	f, err := os.Open(name)
+	if err != nil {
+		return errors.Wrap(err, "unable to open manifest")
+	}
+	defer f.Close()
+	buf, err := ioutil.ReadAll(f)
+	if err != nil {
+		return errors.Wrap(err, "unable to read manifest")
+	}
+	if err := image.ValidInternalMediaType(buf); err != nil {
+		return errors.Wrapf(err, "%s: manifest validation failed", name)
+	}
+	return nil
 }
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -51,6 +51,10 @@ func findManifest(w walker, d *descriptor) (*manifest, error) {
 			return errors.Wrapf(err, "%s: error reading manifest", path)
 		}
 
+		if err := ValidInternalMediaType(buf); err != nil {
+			return errors.Wrapf(err, "%s: manifest validation failed", path)
+		}
+
 		if err := schema.MediaTypeManifest.Validate(bytes.NewReader(buf)); err != nil {
 			return errors.Wrapf(err, "%s: manifest validation failed", path)
 		}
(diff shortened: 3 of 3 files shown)


schema/validator.go
```diff
@@ -54,6 +55,28 @@ func (v Validator) Validate(src io.Reader) error {
 			"schema %s: unable to validate", v)
 	}
 
+	if v == MediaTypeManifest {
```
wking: This is not part of the JSON Schema validation. Your new logic should live in `manifest.validate` in `image/manifest.go`.

wking: Alternatively, you could adjust `image-manifest-schema.json` to replace:

    "config": {
    },

with a version of descriptor where `mediaType` was an enum listing only `application/vnd.oci.image.config.v1+json`, etc.

Although see [here][1] about the spec not actually defining bounds on the referenced media types.

[1]: https://github.com/opencontainers/image-spec/pull/257/files#r77792057

xiekeyang: @wking `validate --type manifest` type seems not to jump into `manifest.validate`, instead into `Validator.Validate` directly. So, `manifest.validate` can't be effective.

wking: On Wed, Sep 07, 2016 at 04:00:53AM -0700, xiekeyang wrote:
> > +	if v == MediaTypeManifest {
>
> @wking `validate --type manifest` type seems not to jump into
> `manifest.validate`, instead into `Validator.Validate` directly. So,
> `manifest.validate` can't be effective.

?  ‘oci-image-tool validate …’ calls both image.Validate(Layout) [1]
and schema.MediaType(Manifest|ManifestList|Config).Validate [2].  I
expect we should be calling the schema validation from the image
validation functions, but either way, code in the image validation
functions is getting called by ‘oci-image-tool validate …’.

[1]: https://github.com/opencontainers/image-spec/blob/d3cfb991ee3ff9e7c9570027cb9696213cfcee00/cmd/oci-image-tool/validate.go#L126-L131
[2]: https://github.com/opencontainers/image-spec/blob/d3cfb991ee3ff9e7c9570027cb9696213cfcee00/cmd/oci-image-tool/validate.go#L139-L146

> ?  ‘oci-image-tool validate …’ calls both image.Validate(Layout) [1]
> and schema.MediaType(Manifest|ManifestList|Config).Validate [2].  I
> expect we should be calling the schema validation from the image
> validation functions, but either way, code in the image validation
> functions is getting called by ‘oci-image-tool validate …’.

It is some hard to do based on current code.
What you mentioned [1] is on package order main -> image -> schema;
[2] is main -> schema;
I think it had better to add code to schema package, and better to only one place to be called by up layer. For that I figure out the PR. I tried other way, but it seems to be some brittle.

xiekeyang:
WIP

schema/validator.go
wking: On Thu, Sep 08, 2016 at 12:29:29AM -0700, xiekeyang wrote:
> > +	if v == MediaTypeManifest {
> 
> @wking 
> > ?  ‘oci-image-tool validate …’ calls both image.Validate(Layout) [1]
> > and schema.MediaType(Manifest|ManifestList|Config).Validate [2].  I
> > expect we should be calling the schema validation from the image
> > validation functions, but either way, code in the image validation
> > functions is getting called by ‘oci-image-tool validate …’.
> 
> It is some hard to do based on current code.
> What you mentioned [1] is on package order main -> image -> schema;
> [2] is main -> schema;
> I think it had better to add code to schema package, and better to
> only one place to be called by up layer.

I'd rather not tie the JSON Schema too tightly here.  Having the
schemas and schema-only tooling in their own layer makes it easier to
share them (and there is a lot of third-party JSON Schema tooling that
can consume them, opencontainers/validator.opencontainers.org#2).

On the other hand, JSON Schema cannot represent all of our
constraints.  So I'd like another layer (currently in image/*.go) to
do the full validation.  That full-validation layer can save itself a
lot of work by leaning on JSON Schema for part of its validation, but
whether it uses JSON Schema or not is really an internal detail.

So I think we want to use only main → image → schema, but would also
be fine with main → image/validate → schema (if you want to separate
the validation Go code from the unpacking, etc., Go code).  I don't
think we want to push a lot of Go logic into the JSON Schema
directory.

xiekeyang: Yes. So far I find that schema/*.json should not modified because they only present the generally patten and can't cover all constraints. I'd work out on image code layer.

xiekeyang:
Updated， PTAL。
It is some redundant to open and read the manifest file. That is because I follow up the current code blocks, not to refactor them(may to do later).
/cc @runcom @wking @stevvooe

runcom:
We may not want to error out on invalid media types based on https://github.com/opencontainers/image-spec/issues/286 - there should just be something which says _unsupported_ maybe but we need to ease printing to the stdout at various log level with https://github.com/opencontainers/image-spec/issues/288.

xiekeyang:
@runcom I think `unsupported ` warning is inappropriate. media type is standard definition in spec.
Should we pass the validation when it not follow the spec?

runcom:
> @runcom I think unsupported warning is inappropriate. media type is standard definition in spec.
> Should we pass the validation when it not follow the spec?

quoting @wking from https://github.com/opencontainers/image-spec/issues/286#issuecomment-245856987

> The spec does not currently place restrictions on allowed types for
manifest-list.manifests, manifest.layers, or manifest.config

we should first assure the spec defines or not those constraints

xiekeyang:
ping @stevvooe @wking @philips
If the (identifiable media types)[https://github.com/opencontainers/image-spec/blob/master/media-types.md#media-types] are constraints? I think we had better to clarify it.

schema/schema.go
```diff
@@ -22,11 +22,13 @@ import (
...
-	MediaTypeManifest     Validator     = v1.MediaTypeImageManifest
-	MediaTypeManifestList Validator     = v1.MediaTypeImageManifestList
-	MediaTypeImageConfig  Validator     = v1.MediaTypeImageConfig
-	MediaTypeImageLayer   unimplemented = v1.MediaTypeImageLayer
+	MediaTypeDescriptor        Validator     = v1.MediaTypeDescriptor
+	MediaTypeManifest          Validator     = v1.MediaTypeImageManifest
+	MediaTypeManifestList      Validator     = v1.MediaTypeImageManifestList
+	MediaTypeImageConfig       Validator     = v1.MediaTypeImageConfig
+	MediaTypeImageLayer        unimplemented = v1.MediaTypeImageLayer
+	CompatMediaTypeImageConfig Validator     = "application/vnd.docker.container.image.v1+json"
```
stevvooe: `MediaTypeDockerImageConfig`

xiekeyang: Updated

wking:
On Mon, Sep 12, 2016 at 07:08:34PM -0700, xiekeyang wrote:
> If the (identifiable media
> types)[https://github.com/opencontainers/image-spec/blob/master/media-types.md#media-types]
> are constraints? I think we had better to clarify it.

I agree, but that's a big if.  Are we sure we want to require them?
It's a trade-off [1].  And we can place different constraints on
images and implementations.  For example, perhaps an image MUST NOT
use a media type that's not listed in [2] and an implementation MUST
support those types.

  fewer types                     |  media-types.md  |  more types
  --------------------------------+------------------+------------------------
  ← compliant images                                 |  non-compliant images →
  ← non-compliant implementaions  |  compliant images →

But I'm nervous about requiring support for non-OCI media types unless
there is a stable, testable spec we can link for the external types.

[1]: https://github.com/opencontainers/image-spec/issues/286#issuecomment-245856987
[2]: https://github.com/opencontainers/image-spec/blob/master/media-types.md#media-types

xiekeyang:
@wking 

> For example, perhaps an image MUST NOT
use a media type that's not listed in [2] and an implementation MUST
support those types.

For this I think we have announced in spec [Non-Distributable Layers](https://github.com/opencontainers/image-spec/blob/master/layer.md#non-distributable-layers).

As config file, it should be standard format as [Container config JSON](https://github.com/opencontainers/image-spec/blob/master/serialization.md) .

I'm not sure if it covers your implementation case, or can you put some use case? But now any word on `mediaType` field may be passed, that seems not to be standard.

wking:
On Tue, Sep 13, 2016 at 12:12:02AM -0700, xiekeyang wrote:
> > For example, perhaps an image MUST NOT use a media type that's not
> > listed in [2] and an implementation MUST support those types.
>
> For this I think we have announced in spec [Non-Distributable
> Layers](https://github.com/opencontainers/image-spec/blob/master/layer.md#non-distributable-layers).

(Non)distributable is about what you can push.  I'm suggesting wording
around what an implementation can handle.  I've submitted #304 with
explicit wording around my above suggestion (although I've softened
the instance to SHOULD).  If/when something like #304 lands, testing
like you're adding here will be much easier to write (since it will be
“rewrite these requirements in Go” and not “invent requirements and
write them in Go” ;).

xiekeyang:
This is migrated to opencontainers/image-tools/pull/10. Here should be closed.

PR #274, opened 2016-09-07
Fix unpack multiple layers

coolljt0725:

If there are multiple layers, it will only unpack a layer and then return.
This pr fix this.

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -112,7 +112,6 @@ func (m *manifest) unpack(w walker, dest string) error {
 		case nil:
 			return fmt.Errorf("%s: layer not found", dest)
 		case errEOW:
-			return nil
 		default:
 			return err
 		}


image/manifest.go
```diff
@@ -108,14 +108,13 @@ func (m *manifest) unpack(w walker, dest string) error {
...
 
 			return errEOW
-		}); err {
-		case nil:
-			return fmt.Errorf("%s: layer not found", dest)
-		case errEOW:
-			return nil
-		default:
+		})
+		if err != nil && err != errEOW {
```
jonboulle: Can you restore the switch and just change the errEOW branch?

runcom: +1

coolljt0725:
@runcom  @jonboulle  updated

jonboulle:
LGTM

runcom:
LGTM (IANAM)

philips:
LGTM

PR #275, opened 2016-09-07
Makefile: add missing document pages

### LINKED ISSUE #244: Image-layout.md not included in released spec
The text of the [image-layout.md file](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) does not seem to be included in the released HTML version of the spec (as of v0.4.0.)

vbatts:
Closes #244

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,8 +25,11 @@ DOC_FILES := \
 	code-of-conduct.md \
 	project.md \
 	media-types.md \
-	manifest.md \
-	serialization.md
+	descriptor.md \
+	image-layout.md \
+	layer.md \
+	serialization.md \
+	manifest.md
 
 FIGURE_FILES := \
 	img/media-types.png


Makefile
```diff
@@ -26,7 +26,13 @@ DOC_FILES := \
 	project.md \
 	media-types.md \
 	manifest.md \
-	serialization.md
+	descriptor.md \
+	image-layout.md \
+	serialization.md \
+	layer.md \
+	GOVERNANCE.md \
+	RELEASES.md
```
wking: Do we need these last two?  I think they're metadata, and not part of the spec itself.  runtime-spec [does not include them in the generated spec][1] or link to them from any files which are part of the generated spec.

[1]: https://github.com/opencontainers/runtime-spec/blob/22c29c18eddc75ccee9f6b202f61524f29653c5f/Makefile#L23-L38

Makefile
```diff
@@ -26,7 +26,13 @@ DOC_FILES := \
 	project.md \
 	media-types.md \
 	manifest.md \
-	serialization.md
+	descriptor.md \
+	image-layout.md \
+	serialization.md \
+	layer.md \
```
wking: This order seems a bit odd to me.  I'd rather build this up from the ground with something like:

* …
* media-types
* descriptor
* image-layout
* layer
* serialization
* manifest

wking:
Do we want a table of contents (like runtime-spec has [1])?  If so, is
that part of this PR, or a separate issue?

[1]: https://github.com/opencontainers/runtime-spec/blob/22c29c18eddc75ccee9f6b202f61524f29653c5f/README.md#open-container-runtime-specification

Makefile
vbatts: I debated about those two because they are some what boiler plate. I'll
remove them.

Makefile
vbatts: updated

philips:
LGTM

@wking separate issue

wking:
On Wed, Sep 07, 2016 at 01:18:31PM -0700, Brandon Philips wrote:
> @wking separate issue

Sounds good.  In that case, 3f4c1206 looks good to me.

RobDolinMS:
It would be good to add a Table of Contents as @wking suggests, but I recommend doing that as a another issue/PR so this can be merged.  /cc @vbatts

philips:
cc @opencontainers/image-spec-maintainers Please LGTM

stevvooe:
LGTM

Issue #271, opened 2016-09-07
Add yaml config file to image tool command usage


xiekeyang:
@opencontainers/image-spec-maintainers 
Now we are appending more and more flags to image tool commend. For conveniently using, it had better to introduce a yaml config file, like that:

```go
cmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.cobra.yaml)")
```

In which user can fulfill his all flags.
If you allow my idea, I'd like to take on it. Thanks a lot!

/cc @philips @stevvooe @s-urbaniak @runcom @wking

runcom:
I don't like this, I expect the image tool to be just (if possible) a simple tool w/o any state or config saved on disk.

wking:
On Wed, Sep 07, 2016 at 01:54:58AM -0700, Antonio Murdaca wrote:
> I expect the image tool to be just (if possible) a simple tool w/o
> any state or config saved on disk.

I agree.  What sort of config state were you tired of typing out? ;)

xiekeyang:
It is OK. I just feel the command line become long to readable, and seen some config file convenient usage via cobra package.

wking:
On Wed, Sep 07, 2016 at 02:33:58AM -0700, xiekeyang wrote:
> It is OK. I just feel the command line become long to readable…

Is that because create-runtime-bundle is really long?  Or can you paste
an example of a typical long line that you'd like to abbreviate?

jonboulle:
would it suffice if you could use environment variables to specify command line options?

runcom:
> would it suffice if you could use environment variables to specify command line options?

is there any tool out there which duplicates command line opts with env variables? I'm not sure... then the "long line" will be the one specifying the env vars (e.g. `REFS=a,b,c,d,e,f,...` imagine it longer...)

jonboulle:
What do you mean by "duplicates"? if you're asking about a library that'll map env vars to flags, something like e.g. globalconf, or the simpler package we use for CoreOS projects:
https://github.com/rakyll/globalconf
https://github.com/coreos/pkg/blob/master/flagutil/env.go#L10

runcom:
> What do you mean by "duplicates"?

I mean having an env var for a command line option.

I don't see any gain in having env vars because, as explained early, one can have a really long list of references (for instance) to use with $REFS and thus having still a long line to type paste.

[...]

Issue #266, opened 2016-09-06
oci-image-tool: support multiple --ref in unpack and create-runtime-bundle


runcom:
As the title says (validate has been fixed here https://github.com/opencontainers/image-spec/pull/265)

philips:
This issue was moved to opencontainers/image-tools#19

PR #279, opened 2016-09-08
cmd/oci-image-tool: validate all refs by default

runcom:
This is taking care of https://github.com/opencontainers/image-spec/issues/271#issuecomment-245325132 (left a TODO/FIXME for later in case).

I'll adapt everything to be same as per https://github.com/opencontainers/image-spec/issues/266 later

```shell
$ oci-image-tool validate /home/amurdaca/src/github.com/projectatomic/skopeo/busybox-oci2
reference "12.04": OK
reference "3.1": OK
reference "a59906e33509d14c036c8678d687bd4eec81ed7c4b8ce907b888c607f6a1e0e6": OK
reference "latest": OK
reference "notlatest": OK
reference "tag": OK
/home/amurdaca/src/github.com/projectatomic/skopeo/busybox-oci2: OK
```

/cc @wking @vbatts @philips @stevvooe 

### DIFF
--- a/image/descriptor.go
+++ b/image/descriptor.go
@@ -49,6 +49,27 @@ func (d *descriptor) hash() string {
 	return pts[1]
 }
 
+func listReferences(w walker) (map[string]*descriptor, error) {
+	refs := make(map[string]*descriptor)
+
+	if err := w.walk(func(path string, info os.FileInfo, r io.Reader) error {
+		if info.IsDir() || !strings.HasPrefix(path, "refs") {
+			return nil
+		}
+
+		var d descriptor
+		if err := json.NewDecoder(r).Decode(&d); err != nil {
+			return err
+		}
+		refs[info.Name()] = &d
+
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	return refs, nil
+}
+
 func findDescriptor(w walker, name string) (*descriptor, error) {
 	var d descriptor
 	dpath := filepath.Join("refs", name)
--- a/image/image.go
+++ b/image/image.go
@@ -16,6 +16,7 @@ package image
 
 import (
 	"encoding/json"
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
@@ -33,7 +34,7 @@ func ValidateLayout(src string, refs []string, out *log.Logger) error {
 
 // Validate walks through the given .tar file and
 // validates the manifest pointed to by the given refs
-// or returns an error if the validation failed.
+//iiii or returns an error if the validation failed.
 func Validate(tarFile string, refs []string, out *log.Logger) error {
 	f, err := os.Open(tarFile)
 	if err != nil {
@@ -50,17 +51,37 @@ var validRefMediaTypes = []string{
 }
 
 func validate(w walker, refs []string, out *log.Logger) error {
-	for _, r := range refs {
-		ref, err := findDescriptor(w, r)
-		if err != nil {
+	ds, err := listReferences(w)
+	if err != nil {
+		return err
+	}
+	if len(refs) == 0 && len(ds) == 0 {
+		// TODO(runcom): ugly, we'll need a better way and library
+		// to express log levels.
+		// see https://github.com/opencontainers/image-spec/issues/288
+		out.Print("WARNING: no descriptors found")
+	}
+
+	if len(refs) == 0 {
+		for ref := range ds {
+			refs = append(refs, ref)
+		}
+	}
+
+	for _, ref := range refs {
+		d, ok := ds[ref]
+		if !ok {
+			// TODO(runcom):
+			// soften this error to a warning if the user didn't ask for any specific reference
+			// with --ref but she's just validating the whole image.
+			return fmt.Errorf("reference %s not found", ref)
+		}
+
+		if err = d.validate(w, validRefMediaTypes); err != nil {
 			return err
 		}
 
-		if err = ref.validate(w, validRefMediaTypes); err != nil {
-			return err
-		}
-
-		m, err := findManifest(w, ref)
+		m, err := findManifest(w, d)
 		if err != nil {
 			return err
 		}
@@ -69,7 +90,7 @@ func validate(w walker, refs []string, out *log.Logger) error {
 			return err
 		}
 		if out != nil {
-			out.Printf("reference %q: OK", r)
+			out.Printf("reference %q: OK", ref)
 		}
 	}
 	return nil
--- a/cmd/oci-image-tool/validate.go
+++ b/cmd/oci-image-tool/validate.go
@@ -63,7 +63,7 @@ func newValidateCmd(stdout, stderr *log.Logger) *cobra.Command {
 	)
 
 	cmd.Flags().StringSliceVar(
-		&v.refs, "ref", []string{"v1.0"},
+		&v.refs, "ref", nil,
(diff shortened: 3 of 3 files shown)


image/descriptor.go
```diff
@@ -49,6 +49,30 @@ func (d *descriptor) hash() string {
 	return pts[1]
 }
 
+func findDescriptors(w walker) (map[string]*descriptor, error) {
```
wking: Maybe `listReferences`?  In #159, the analogous function is [`refs.Engine.List()`][1].

[1]: https://github.com/opencontainers/image-spec/pull/159/files#diff-d8af84c58f6ae3369e0b5cbca06727dfR43

image/descriptor.go
```diff
@@ -49,6 +49,30 @@ func (d *descriptor) hash() string {
...
+			return err
+		}
+		refs[info.Name()] = &d
+
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	if len(refs) == 0 {
+		return nil, errors.New("no descriptors found")
```
wking: Is this a problem?  I don't think `findDescriptors` ([`listReferences`?]) should error if there aren't any references.  It should just return an empty map.

I think `validate` should *also* succeed if there are no refs in the store and the caller does not ask for any specifically.  There are still no invalid refs in the store.  However, if we want to special-case validate to error out if the caller sets and empty `refs` and the store has no refs, I'm ok with that (in which case we'd move this check over to `validate`).

[1]: https://github.com/opencontainers/image-spec/pull/279/files#r78029573

image/image.go
```diff
@@ -50,17 +50,31 @@ var validRefMediaTypes = []string{
...
+				return true
+			}
 		}
-
-		if err = ref.validate(w, validRefMediaTypes); err != nil {
+		return false
+	}
+	hasRefs := len(refs) != 0
+	for ref, d := range ds {
+		if hasRefs && !inRefs(ref) {
```
wking: I'd rather turn this around and iterate over `refs`, which would allow you to drop `inRefs`.  Something like:

    if len(refs) == 0 {
      for ref, _ := range ds {
        refs = append(refs, ref)
      }
    }
    for _, ref := range refs {
      descriptor, ok := ds[ref]
      if !ok {
        return fmt.Errorf("reference not found: %q", ref)
      }
      err = descriptor.validate(…)
      …
    }

wking:
On Thu, Sep 08, 2016 at 06:52:45AM -0700, Antonio Murdaca wrote:
> -- File Changes --
> 
>     M cmd/oci-image-tool/validate.go (2)
>     M image/descriptor.go (24)
>     M image/image.go (30)

Update the man page too.

image/image.go
runcom: it's equivalent right? you still do an O(N) iterating over refs :(

I guess I'll change as you did if you really wish to

wking: On Thu, Sep 08, 2016 at 08:40:01AM -0700, Antonio Murdaca wrote:
> >  		}
> > -
> > -		if err = ref.validate(w, validRefMediaTypes); err != nil {
> > +		return false
> > +	}
> > +	hasRefs := len(refs) != 0
> > +	for ref, d := range ds {
> > +		if hasRefs && !inRefs(ref) {
> 
> it's equivalent right? you still do an O(N) iterating over refs :(

Iterating over refs is O(len(refs)).  Iterating over ds is O(len(ds)).
‘refs’ is going to be shorter than or the same length as ‘ds’ unless
someone called validate --refs 1.0 --refs 1.0 --refs 1.0 (or some
such).  I don't expect that to happen frequently enough to be worth
deduplicating refs (although iterating over ‘ds’ is one way to
deduplicate ‘refs’).

My main reason for wanting to iterate over refs is that it runs the
validation in the same order that the caller asked for.

runcom: > My main reason for wanting to iterate over refs is that it runs the
> validation in the same order that the caller asked for.

good point, will fix

stevvooe:
@runcom If unknown types are encountered, what is the intended behavior?

I ask because we do need to be able to have references to unknown things but still pass validation. This will make it easier to extend in the future.

wking:
On Thu, Sep 08, 2016 at 12:47:36PM -0700, Stephen Day wrote:
> @runcom If unknown types are encountered, what is the intended behavior?

I think we need to error with “not implemented”-level errors.

> I ask because we do need to be able to have references to unknown
> things but still pass validation.  This will make it easier to extend
> in the future.

That makes sense.  We might want an --ignore-unknown-media-types flag
to avoid erroring out if we hit an unrecognized type.  There's some
more discussion of unrecognized types in [1].

What happens if you recognize the ref type
(e.g. application/vnd.oci.image.manifest.list.v1+json) but not an
ancestor type (e.g. application/vnd.oci.image.layer.zip) and
--ignore-unknown-media-types is set?  Do you print your warning and
return zero?  Or is the error→warning softening only for the root
refs?  We can have a global error→warning softening and still error
for the application/vnd.oci.image.layer.zip case if the specs for our
objects explicitly list allowed descriptor types [2,3].

[1]: https://github.com/opencontainers/image-spec/pull/262/files#r77768227
[2]: https://github.com/opencontainers/image-spec/pull/257/files#r77792057
[3]: https://github.com/opencontainers/image-spec/pull/273#discussion_r77794794

stevvooe:
@wking I was asking @runcom. It's okay to let a conversation develop before commenting.

wking:
On Thu, Sep 08, 2016 at 01:00:21PM -0700, Stephen Day wrote:
> It's okay to let a conversation develop before commenting.

Sure.  I was just linking in previous unrecognized-type discussion to
try and avoid a fragmented discussion of that issue.

runcom:
@stevvooe right now, if the media type is unknown the tool errors out. We should at least soften that to just print a warning. This behavior is in the `validate` function though, I can address this in a follow up and discuss more about errors/warnings if it's ok for you, this PR just makes sure we "find" every reference under refs/ to shorten the command line flags.

wking:
On Thu, Sep 08, 2016 at 01:58:18PM -0700, Antonio Murdaca wrote:
> This behavior is in the `validate` function though, I can address
> this in a follow up and discuss more about errors/warnings if it's
> ok for you, this PR just makes sure we "find" every reference under
> refs/ to shorten the command line flags.

I'm fine punting unrecognized media type issues to follow up work,
since that is an orthogonal problem (this PR just increases the
likelyhood that we find unrecognized types).

stevvooe:
@runcom SGTM. Please open an issue.

image/descriptor.go
runcom: This is a _validation_ question, I opened https://github.com/opencontainers/image-spec/issues/287 to discuss more about this one, I'll move this check in `validate`.

runcom:
@stevvooe @vbatts @philips PTAL

philips:
LGTM.

@s-urbaniak ?

stevvooe:
LGTM

image/image.go
```diff
@@ -33,7 +34,7 @@ func ValidateLayout(src string, refs []string, out *log.Logger) error {
 
 // Validate walks through the given .tar file and
 // validates the manifest pointed to by the given refs
-// or returns an error if the validation failed.
+//iiii or returns an error if the validation failed.
```
s-urbaniak: iiii? ;-)

s-urbaniak:
LGTM, with one small nit

philips:
LGTM

image/image.go
runcom: lol

PR #292, opened 2016-09-09
descriptor: Require lowercase hex

wking:
Now that 75a51ed (fix regular expression of algorithim, 2016-08-30, #221) has made the Markdown and JSON Schema consistent (and required lowercase algorithms), make digest comparisons easier by also requiring lowercase hex.

This also:

* Makes it easier to serve blobs out of a case-insensitive filesystem
  store.
* Avoids having two otherwise-identical descriptor (or
  descriptor-containing) blobs with different hashes because they
  picked differend hex-casing.

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -54,14 +54,14 @@ It uniquely identifies content by taking a [collision-resistant hash](https://en
 If the identifier can be communicated in a secure manner, one can retrieve the content from an insecure source, calculate the digest independently, and be certain that the correct content was obtained.
 
 The value of the digest property, the _digest string_, is a serialized hash result, consisting of an _algorithm_ portion and a _hex_ portion.
-The algorithm identifies the methodology used to calculate the digest; the hex portion is the hex-encoded result of the hash.
+The algorithm identifies the methodology used to calculate the digest; the hex portion is the lowercase hex-encoded result of the hash.
 
-The digest string matches the following grammar:
+The digest string MUST match the following grammar:
 
 ```
 digest      := algorithm ":" hex
 algorithm   := /[a-z0-9_+.-]+/
-hex         := /[A-Fa-f0-9]+/
+hex         := /[a-f0-9]+/
 ```
 
 Some example digest strings include the following:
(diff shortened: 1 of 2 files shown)


stevvooe:
LGTM

runcom:
@stevvooe @wking is this a must for the spec? Should we add validation to oci-image-tool? (I'll fix it if so in a follow up not to block this)

wking:
On Fri, Sep 09, 2016 at 03:40:09PM -0700, Antonio Murdaca wrote:
> is this a must for the spec?

Yes.  I've pushed efb13e3 → 660d130 adding a MUST to make that
clearer.

> Should we add validation to oci-image-tool?

This should be handled automatically via the JSON Schema regex (bumped
in this PR).  We're using that as part of validate already, aren't we?

runcom:
@stevvooe does the MUST look good to you? 

@wking I can't check right now, I hope it's validating using the json schema.

runcom:
@wking @stevvooe unfortunately with this PR, oci-image-tool isn't reporting any _invalid_ digest format.

if you change a digest under refs/ to have some uppercase char you get:
```
oci-image-tool validate --ref latest . 
.: validation failed: sha256:2A2b6168c040d25148a4972cb20108b737adc51685d4dc68a8e01d55416d27a1: not found
```
it isn't saying anything about lower/upper cases..I guess it's the same for blobs/ (notice that I would have had the above to fail with "invalid digest, no uppercase!" or something like this, I know I renamed to digest and now it can't find the correct blob)

stevvooe:
@runcom I don't think a fix for that should be a pre-cursor to a merge here. Sounds like our validation is slightly broken.

Could we still merge this and file a bug?

runcom:
@stevvooe absolutely, this is by any means a blocker, sorry if I intended that, please go ahead and merge this, I'll file a separate issue. 

The fix in oci-image-tool will be done separately and it's part of a big refactor I have in mind about validation steps.

jonboulle:
LGTM

stevvooe:
LGTM

PR #294, opened 2016-09-12
fix inline html/markdown syntax

xiekeyang:


philips:
How is this a typo? markdown allows for intermixing html like this.

xiekeyang:
> How is this a typo? markdown allows for intermixing html like this.
Seems it doesn't.
it shows on page as,   \`application/vnd.oci.image.layer.tar+gzip\` 
We want like,  `application/vnd.oci.image.layer.tar+gzip`

wking:
On Sun, Sep 11, 2016 at 08:58:16PM -0700, Brandon Philips wrote:
> How is this a typo? markdown allows for intermixing html like this.

There are some rules about inline Markdown inside HTML not being
supported.  I haven't dug up a refernce in the Markdown spec, but you
can see the backtics not being converted to code tags in master [1].
So +1 for b0edca4b (although a more specific commit summary would be
nice ;).

[1]: https://github.com/opencontainers/image-spec/blob/d564f8ed01dca4da16fe7d1d0543419deda2f421/serialization.md#terminology

xiekeyang:
Update commit summary.

jonboulle:
lgtm

philips:
LGTM

PR #297, opened 2016-09-12
specs-go/v1: add nondistributable media type

runcom:
Takes care of adding the new media type in the Go types from https://github.com/opencontainers/image-spec/pull/233 - so that can proceed and be merged w/o any other interruption.

/cc @stevvooe @philips @vbatts 

### DIFF
--- a/specs-go/v1/mediatype.go
+++ b/specs-go/v1/mediatype.go
@@ -15,18 +15,22 @@
 package v1
 
 const (
-	// MediaTypeDescriptor specifies the mediaType for a content descriptor.
+	// MediaTypeDescriptor specifies the media type for a content descriptor.
 	MediaTypeDescriptor = "application/vnd.oci.descriptor.v1+json"
 
-	// MediaTypeImageManifest specifies the mediaType for an image manifest.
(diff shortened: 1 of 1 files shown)


jonboulle:
lgtm

wking:
While we're filling in the pieces, img/media-types.dot and one of the
entries in schema/manifest_test.go should probably be updated too.

runcom:
> schema/manifest_test.go

what's wrong with schema/manifest_test.go? just add a test?

wking:
On Mon, Sep 12, 2016 at 11:04:57AM -0700, Antonio Murdaca wrote:
> > schema/manifest_test.go
> 
> what's wrong with schema/manifest_test.go? just add a test?

We should exercise the new type to make sure the image tooling handles
it correctly, which includes not choking on it during validation.

philips:
LGTM

runcom:
> We should exercise the new type to make sure the image tooling handles
> it correctly, which includes not choking on it during validation.

@wking that's the wrong test for sure since it's already a mess with media types like ".serialization." but it  is not failing since that is testing the MANIFEST as whole - not that the media types are actually correct...

PR #299, opened 2016-09-13
Use DiffID instead of ID in config.md

coolljt0725:
We doesn't define  `layer ID`, to avoid confusing
users, this should be a `DiffID`.


wking:
On Mon, Sep 12, 2016 at 07:53:31PM -0700, Lei Jitang wrote:
> We doesn't define  `layer ID`, to avoid confusing
> users, this should be a `DiffID`.

49f72a90 looks good to me.

vbatts:
LGTM

philips:
Added to #333 which is a v1.0.0-rc1 release blocker. Thanks!

PR #303, opened 2016-09-13
Fixies to img: and schema:

runcom:
some fixes for https://github.com/opencontainers/image-spec/pull/297 and another for `schema/manifest_test.go` which was behaving wrongly. In such cases we expected a given error, test passes but the err we got wasn't the one expected (but yet the test passed)

Note: the `img/media-types.png` picture has been generated with the Makefile - I don't know why it has a different size


runcom:
@vbatts @philips @stevvooe PTAL

philips:
LGTM

wking:
3795c5f06 looks fine to me.

I still think we want to make sure the tooling doesn't choke on
nondistributable [1].  However, I'm fine punting that point to after
something like #304 makes the “don't choke” requirement explicit.

[1]: https://github.com/opencontainers/image-spec/pull/297#issuecomment-246439890

vbatts:
LGTM

Issue #286, opened 2016-09-09
cmd/oci-image-tool: validation errors and warnings


runcom:
From https://github.com/opencontainers/image-spec/pull/279#issuecomment-245717705

Action: `oci-image-tool validate` should soften the error given when an unknown media type is encountered when validating a reference under `refs/`.

Discussion: does the above work with media types found in manifests as well? Should we just print a warning and `exit 0` from the tool? Should we a flag which specifies the warning level to suppress? So, should we have more than just one warning level here? (the second level being a manifest, third being media types in descriptors , etc)

/cc @stevvooe @vbatts @philips

wking:
On Fri, Sep 09, 2016 at 01:33:06AM -0700, Antonio Murdaca wrote:
> Action: `oci-image-tool validate` should soften the error given when
> an unknown media type is encountered when validating a reference
> under `refs/`.

I think the softening should go behind a flag like --unpackable [1]
(which makes more sense than my alternative
--ignore-unknown-media-types [2]).  This gives you a way to toggle (b)
in the following cases:

a) The media type is recognized and an expected target of the source
   location (always good).
b) The media type is unrecognized, but the source location does not
   have media-type restrictions (good unless --unpackable)
c) The media type is recognized and the source location does not have
   media-type restrictions, but the media type does not make sense in
   this location (e.g. an image/png in manifest.layers.  Always bad).
d) The media type is not in the set of types allowed by that source
   location (always bad).

The spec does not currently place restrictions on allowed types for
manifest-list.manifests, manifest.layers, or manifest.config, which
makes gradual extensibility easy but leaves the “does not make sense”
call a bit vague (maybe an implementation thinks it does know how to
unpack an image/png layer).  If we do restrict those types, we
wouldn't have any instances of (b) or (c), and there would be no need
of a softening flag.

> So, should we have more than just one warning level here? (the
> second level being a manifest, third being media types in
> descriptors , etc)

I don't think that scales as the number of supported media types grows
(as it presumably will in the future).

[1]: https://github.com/opencontainers/image-spec/pull/262#discussion_r77768227
[2]: https://github.com/opencontainers/image-spec/pull/279#issuecomment-245720837

wking:
On Fri, Sep 09, 2016 at 01:50:04AM -0700, W. Trevor King wrote:
> If we do restrict those types, we wouldn't have any instances of (b)
> or (c), and there would be no need of a softening flag.

[...]

PR #304, opened 2016-09-13
manifest: List media types supported by descriptor references

wking:
Require implementations to support these media types (although an implementation which also supports additional media types is still compliant).  Suggest portable manifests and manifest lists stick to these types so the can rely on implementation support.

A manifest(-list) can safely use a type from outside the list if the author expects the implementation to support the extension type (e.g. application/vnd.docker.image.rootfs.diff.tar.gzip).  But how that out-of-band expectation is setup is beyond the scope of this specification.

Previous discussion in #286 and a few other one-off comments ;).

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -30,7 +30,16 @@ A client will distinguish a manifest list from an image manifest based on the Co
   This REQUIRED property contains a list of manifests for specific platforms.
   While the property MUST be present, the size of the array MAY be zero.
 
-  Each object in the manifest is a [descriptor](descriptor.md) with the following additional properties:
+  Each object in the manifest is a [descriptor](descriptor.md) with the following additional properties and restrictions:
+
+  - **`mediaType`** *object*
+
+    This [descriptor property](descriptor.md#properties) has additional restrictions for `manifests`.
+    Implementations MUST support at least the following media types:
+
+    - [`application/vnd.oci.image.manifest.v1+json`](#image-manifest)
+
+    Manifest lists concerned with portability SHOULD use one of the above media types.
 
   - **`platform`** *object*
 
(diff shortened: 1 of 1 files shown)


vbatts:
This seems to go against the compatibility matrix

philips:
Yes, please add the nondistributable to the compat matrix. Thanks!

wking:
On Tue, Sep 13, 2016 at 12:12:19PM -0700, Brandon Philips wrote:
> Yes, please add the nondistributable to the compat matrix. Thanks!

Should be covered by the just-landed #233.  I'll rebase around that
shortly.

wking:
On Tue, Sep 13, 2016 at 12:00:57PM -0700, Vincent Batts wrote:
> This seems to go against the compatibility matrix

Maybe I'm misunderstanding the compatibility matrix.  I had thought it
was an informative hint to image translators and folks writing
implementations that wanted to support related types.  Are you saying
it is normative, and that complaint *OCI implementations* MUST support
those non-OCI types as well?  Do the other types even have versioned
specifications that we would use to certify compliance with them?

jonboulle:
I now have similar confusion to https://github.com/opencontainers/image-spec/pull/304#issuecomment-246796088 so a clarification would be appreciated.

wking:
On Tue, Sep 13, 2016 at 12:22:54PM -0700, W. Trevor King wrote:
> Tue, Sep 13, 2016 at 12:12:19PM -0700, Brandon Philips:
> > Yes, please add the nondistributable to the compat matrix. Thanks!
> 
> Should be covered by the just-landed #233.  I'll rebase around that
> shortly.

Rebased onto master (no conflicts) with ae138eb → a3b9724.

jonboulle:
This patch LGTM based on my understanding so would love to hear what I'm missing..

philips:
LGTM

vbatts:
LGTM

vbatts:
gd... now it needs a rebase? how did github not show that _right_ before? ...

rebase please

wking:
On Thu, Sep 22, 2016 at 10:49:12AM -0700, Vincent Batts wrote:
> rebase please

Rebased around #305 with a3b9724 → b1f1d9c.

vbatts:
LGTM

jonboulle:
lgtm

PR #305, opened 2016-09-13
*: rename serialization.md -> config.md

vbatts:
Closes: https://github.com/opencontainers/image-spec/issues/283

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

PR #308, opened 2016-09-13
image: manifest.go: remove empty layers check

runcom:
There's a discrepancy between `oci-image-tool`, internal schema tests (https://github.com/opencontainers/image-spec/blob/master/schema/manifest_test.go#L51) and the spec itself (https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions). Schema tests and the spec don't require a _not_ empty layers array while `oci-image-tool` is erroring out on an empty layers array.

### DIFF
--- a/image/manifest.go
+++ b/image/manifest.go
@@ -59,10 +59,6 @@ func findManifest(w walker, d *descriptor) (*manifest, error) {
 			return err
 		}
 
-		if len(m.Layers) == 0 {
-			return fmt.Errorf("%s: no layers found", path)
-		}
-
 		return errEOW
 	}); err {
 	case nil:


wking:
5fd27f222 looks good to me.

wking:
Actually, we probably want to also drop “one or more” [1,2], although
that will have trivial conflicts with the in-flight #305.

[1]: https://github.com/opencontainers/image-spec/blame/90334bd71e3162b36a1a1d43b07d7035d4dcce9d/README.md#L6
[2]: https://github.com/opencontainers/image-spec/blame/90334bd71e3162b36a1a1d43b07d7035d4dcce9d/README.md#L8

runcom:
Closing as per @stevvooe comment here https://github.com/opencontainers/image-spec/issues/313

PR #310, opened 2016-09-14
add unit test for ValidateLayout function

xiekeyang:

### DIFF
--- a/image/image_test.go
+++ b/image/image_test.go
@@ -1,0 +1,259 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package image
+
+import (
+	"archive/tar"
+	"bytes"
+	"crypto/sha256"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"testing"
+)
+
+const (
+	layoutStr = `{"imageLayoutVersion": "1.0.0"}`
+
+	configStr = `{
+    "created": "2015-10-31T22:22:56.015925234Z",
+    "author": "Alyssa P. Hacker <alyspdev@example.com>",
+    "architecture": "amd64",
+    "os": "linux",
+    "config": {
+        "User": "alice",
+        "Memory": 2048,
+        "MemorySwap": 4096,
+        "CpuShares": 8,
+        "ExposedPorts": {
+            "8080/tcp": {}
+        },
+        "Env": [
+            "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+            "FOO=oci_is_a",
+            "BAR=well_written_spec"
+        ],
+        "Entrypoint": [
+            "/bin/my-app-binary"
+        ],
+        "Cmd": [
+            "--foreground",
+            "--config",
+            "/etc/my-app.d/default.cfg"
+        ],
+        "Volumes": {
+            "/var/job-result-data": {},
+            "/var/log/my-app-logs": {}
+        },
+        "WorkingDir": "/home/alice"
+    },
+    "rootfs": {
+      "diff_ids": [
+        "sha256:c6f988f4874bb0add23a778f753c65efe992244e148a1d2ec2a8b664fb66bbd1",
+        "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"
+      ],
+      "type": "layers"
+    },
+    "history": [
+      {
+        "created": "2015-10-31T22:22:54.690851953Z",
+        "created_by": "/bin/sh -c #(nop) ADD file:a3bc1e842b69636f9df5256c49c5374fb4eef1e281fe3f282c65fb853ee171c5 in /"
+      },
+      {
+        "created": "2015-10-31T22:22:55.613815829Z",
+        "created_by": "/bin/sh -c #(nop) CMD [\"sh\"]",
+        "empty_layer": true
+      }
+    ]
+}
+`
+)
+
+var (
+	refStr = `{"digest":"<manifest_digest>","mediaType":"application/vnd.oci.image.manifest.v1+json","size":<manifest_size>}`
+
+	manifestStr = `{
+    "annotations": null,
+    "config": {
+        "digest": "<config_digest>",
+        "mediaType": "application/vnd.oci.image.config.v1+json",
+        "size": <config_size>
+    },
+    "layers": [
+        {
+            "digest": "<layer_digest>",
+            "mediaType": "application/vnd.oci.image.layer.tar+gzip",
+            "size": <layer_size>
+        }
+    ],
+    "mediaType": "application/vnd.oci.image.manifest.v1+json",
+    "schemaVersion": 2
+}
+ `
+)
+
+func TestValidateLayout(t *testing.T) {
+	root, err := ioutil.TempDir("", "oci-test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(root)
+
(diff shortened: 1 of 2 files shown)


xiekeyang:
cc @wking @stevvooe @jonboulle @runcom

image/image_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+const (
+	layoutStr = `{"imageLayoutVersion": "1.0.0"}`
+
+	configStr = `{
+    "architecture": "amd64",
+    "config": {
+        "AttachStderr": false,
+        "AttachStdin": false,
+        "AttachStdout": false,
```
wking: I don't see anything about these in `serialization.md`.  Can we remove them?

image/image_test.go
```diff
@@ -0,0 +1,301 @@
...
+)
+
+func TestValidateLayout(t *testing.T) {
+	root, err := ioutil.TempDir("", "oci-test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(root)
+
+	err = os.MkdirAll(filepath.Join(root, "blobs", "sha256"), 0700)
```
wking: Instead of duplicating this logic from `manifest_test.go`, I'd rather have it [pulled out into a helper][1].

[1]: https://github.com/opencontainers/image-spec/pull/257#discussion_r77942739

image/image_test.go
```diff
@@ -0,0 +1,301 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	desc, err := createLayerFile(root)
+	if err != nil {
+		t.Fatal(err)
+	}
+	manifestStr = strings.Replace(manifestStr, "<layer_digest>", desc.Digest, -1)
+	manifestStr = strings.Replace(manifestStr, "<layer_size>", strconv.FormatInt(desc.Size, 10), -1)
```
wking: No need to use `-1` for these when you know there will only be a single replacement.  And I'm not sure I see the benefit to using templates like this vs. serializing the JSON from the Go types.  In #159 I've floated [`PutJSON`][1] to make adding a CAS blob from a Go type straightforward.  For an example of using it to construct an image, see [here][2].

[1]: https://github.com/opencontainers/image-spec/pull/159/commits/89a36d8af3d17ae72f5f5a8b56fb2562d360bce0#diff-794eda1866432863c39ee59e66210d78R25
[2]: https://github.com/wking/image-spec/commit/2886c655edec1a22d2b246324d5a00c645e57a9f#diff-0e4b5ebdd595bebbfb93671dc3d44ebbR52

image/image_test.go
xiekeyang: > And “create image-layout fodder for testing” is likely to be useful stuff in lots of places. Can we make layer-generator, etc. helpers instead of baking it into TestUnpackLayerRemovePartialyUnpackedFile. I'm missing #159 and 87d6a53 ;).

@wking  Got it. Can we place this code to a function `createTarFile` in `*_test.go`? For now this is only used by unit test, and had better not to be compiled to binary.
For that I will fix `image_test.go` and `manifest_test.go` to reuse the `func createTarFile`

xiekeyang: @wking 
Maybe I didn't fully catch your point. For unit test purpose, I can improve the duplicated code.
But I'm not sure if we must add rename behavior here.

>You can use rename no matter what the source and destination are, and
only have to pay for a copy if they are on different filesystems.  In
this case, tmp1/…, {tmp1}/tmp/… and {tmp1}/blobs/sha256/… are all in
the same filesystem, so rename will be cheap in all cases.

image/image_test.go
xiekeyang: removed.

image/image_test.go
wking: On Wed, Sep 14, 2016 at 12:43:24AM -0700, xiekeyang wrote:
> > +	err = os.MkdirAll(filepath.Join(root, "blobs", "sha256"), 0700)
>
> > And “create image-layout fodder for testing” is likely to be
> > useful stuff in lots of places. Can we make layer-generator,
> > etc. helpers instead of baking it into
> > TestUnpackLayerRemovePartialyUnpackedFile. I'm missing #159 and
> > 87d6a53 ;).
>
> @wking Got it. Can we place this code to a function `createTarFile`
> in `*_test.go`? For now this is only used by unit test, and had
> better not to be compiled to binary.  For that I will fix
> `image_test.go` and `manifest_test.go` to reuse the `func
> createTarFile`

This sounds fine to me.  I'd encourage helpers for:

* create image-layout (directory or tar file) [1]
* create dummy layer
* PutJSON [2]
* PutRef [3]

[1]: https://github.com/opencontainers/image-spec/pull/159/commits/87d6a5351fdcc738b60f5b1612ff528307f8dbe8#diff-745d1b74657ed8e87ac8c4e3f520f73aR223
[2]: https://github.com/opencontainers/image-spec/pull/310#discussion_r78697964
[3]: https://github.com/opencontainers/image-spec/pull/159/commits/2d555835b84d5cf7deaa9b108efac43e683a372e#diff-d8af84c58f6ae3369e0b5cbca06727dfR30

image/image_test.go
xiekeyang: @wking 
fix to replace once.
Here I had thought about serializing the JSON. But I refer to unit tests in `schema` package think here we had better to show plain string which seems more, and we should avoid marshal and un-marshal some complicated structures. So I follow the `schema` unit test files.

xiekeyang:
@wking 
fixed, PTAL.
cc @runcom @jonboulle @stevvooe

vbatts:
LGTM

runcom:
are we merging this here? or does this need to be moved (and possibly merged) in image-tools?

wking:
On Thu, Sep 15, 2016 at 07:07:52AM -0700, Antonio Murdaca wrote:
> are we merging this here? or does this need to be moved (and
> possibly merged) in image-tools?

It can be merged on the command line, just add an image-spec remote.
Or we can move it over if folks want a green button ;).

runcom:
> It can be merged on the command line, just add an image-spec remote.

and that's is awful to me given we have another repo and the code here is just for image-tools, do we want to cherry-pick/merge every outstanding PR from image-spec once merged here - anyway whatever

wking:
On Thu, Sep 15, 2016 at 07:24:40AM -0700, Antonio Murdaca wrote:
> > It can be merged on the command line, just add an image-spec remote.
>
> and that's is awful to me given we have another repo and the code
> here is just for image-tools, do we want to cherry-pick/merge every
> outstanding PR from image-spec once merged here - anyway whatever

You have to migrate them over somehow.  image-tools has preserved the
full image-spec history, so there's no need to cherry-pick or rebase
(unless subsequent image-tools development leads to a conflict you
need to rebase around).  Merging from the command line seems easier
than asking OPs to re-open new PRs, assuming you've already collected
the LGTMs (and the maintainer sets are currently the same, so
PullApprove will flag those appropriately).

vbatts:
Are these sources _just_ for image-tools though?

runcom:
> Are these sources just for image-tools though?

they're seem to be just used in `cmd/` which has moved to image-tools (according to my ability with grep which may be bad - that's basically the reason I left `image/` in image-tools)

wking:
On Thu, Sep 15, 2016 at 07:35:12AM -0700, Antonio Murdaca wrote:
> > Are these sources just for image-tools though?
>
> they're seem to be just used in `cmd/` which has moved to
> image-tools (according to my ability with grep which may be bad -
> that's basically the reason I left `image/` in image-tools)

And I expect image-tools to eventually expose this sort of thing in Go
libraries like runtime-tools has been doing (starting with
opencontainers/runtime-tools#130).

vbatts:
i guess i meant, like runtime-spec has ./specs-go/ that are used by the runtime-tools _and_ runc, I figured there would be a similar relationship here. I could likely be wrong though.

runcom:
> i guess i meant, like runtime-spec has ./specs-go/ that are used by the runtime-tools and runc, I figured there would be a similar relationship here. I could likely be wrong though.

@vbatts I left `./specs-go/` which contains the types here in image-spec indeed. `image/` is just a set of functions used only in `image-tools`

vbatts:
ah. my bad. you're right.

wking:
On Thu, Sep 15, 2016 at 07:47:35AM -0700, Antonio Murdaca wrote:
> > i guess i meant, like runtime-spec has ./specs-go/ that are used
> > by the runtime-tools and runc, I figured there would be a similar
> > relationship here. I could likely be wrong though.
>
> @vbatts I left `./specs-go/` which contains the types here in
> image-spec indeed. `image/` is just a set of functions used only in
> `image-tools`

Yeah.  +1 for leaving dead stuff (JSON Schema, Go types) in image-spec
but putting the moving parts (image/, cmd/) in image-tools.

PR #311, opened 2016-09-14
README: Define "unspecified", "undefined", and "implementation-defined"

wking:
I had been unaware of formal distinctions between these terms until @stephenrwalli [called it out][1] in the context of his [suggestion to use “implementation defined” for uploading `application/vnd.oci.image.layer.nondistributable.tar+gzip`][2].  I couldn't find anything as compact as RFC 2119 around this idea, but linking to a section of the C99 rationale seems reasonable enough.  The PDF I'm linking is “Rationale for International Standard - Programming Languages - C Revision 5.10 April-2003” and the referenced content appears in section 3.

The “rationale for the C99 standard” link text seems pretty informal, but that's [what WG14 uses to refer to the document][3].

[1]: https://github.com/opencontainers/image-spec/pull/233#issuecomment-245416276
[2]: https://github.com/opencontainers/image-spec/pull/233#issuecomment-245354663
[3]: http://www.open-std.org/jtc1/sc22/wg14/

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@ The combination of the image manifest, image configuration, and one or more file
 
 The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
 
+The keywords "unspecified", "undefined", and "implementation-defined" are to be interpreted as described in the [rationale for the C99 standard][c99-unspecified].
+
 ![](img/build-diagram.png)
 
 Once built the OCI Image can then be discovered by name, downloaded, verified by hash, trusted through a signature, and unpacked into an [OCI Runtime Bundle](https://github.com/opencontainers/runtime-spec/blob/master/bundle.md).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -9,6 +9,8 @@ The combination of the image manifest, image configuration, and one or more file
 
 The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" are to be interpreted as described in [RFC 2119](http://tools.ietf.org/html/rfc2119) (Bradner, S., "Key words for use in RFCs to Indicate Requirement Levels", BCP 14, RFC 2119, March 1997).
 
+The keywords "unspecified", "undefined", and "implementation-defined" are to be interpreted as described in the [rationale for the C99 standard][unspecified].
```
jonboulle: "unspecified" is an awkward name for a link (especially a reference one because at first blush it looks like a placeholder) - can you call this c99rationale or similar?

wking: On Wed, Sep 14, 2016 at 02:58:04AM -0700, Jonathan Boulle wrote:
> > +The keywords "unspecified", "undefined", and "implementation-defined" are to be interpreted as described in the [rationale for the C99 standard][unspecified].
>
> "unspecified" is an awkward name for a link (especially a reference
> one because at first blush it looks like a placeholder) - can you
> call this c99rationale or similar?

In case we ever link to another part of the c99 rationale, I've gone
with c99-unspecified in 01687e9 → 2d49333.

jonboulle:
lgtm

PR #313, opened 2016-09-14
An image may have empty layers array

coolljt0725:
from https://github.com/opencontainers/image-spec/pull/305/files/17c1f00097a716ca369d089159dd9569e7a0e0d0#r78707107

make this clear in manifest.md

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -139,7 +139,7 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 
 - **`layers`** *array*
 
-    Each item in the array MUST be a [descriptor](descriptor.md).
+    This property MAY be empty, an image can have an empty layers array. Each item in the array MUST be a [descriptor](descriptor.md).
     The array MUST have the base image at index 0.
     Subsequent layers MUST then follow in the order in which they are to be layered on top of each other.
     The algorithm to create the final unpacked filesystem layout MUST be to first unpack the layer at index 0, then index 1, and so on.


manifest.md
```diff
@@ -139,7 +139,7 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 
 - **`layers`** *array*
 
-    Each item in the array MUST be a [descriptor](descriptor.md).
+    This property MAY be empty, an image can have an empty layers array. Each item in the array MUST be a [descriptor](descriptor.md).
```
wking: I think this should be “This OPTIONAL property”.  And we should stick to the [one sentence per line][1] rule.

To make this change clear, I think we also want the validation fix in #308, `omitempty` in the Go property, and dropping `layers` from `required` in `schema/image-manifest-schema.json`.

[1]: https://github.com/opencontainers/image-spec/tree/v0.4.0#markdown-style

runcom:
Wait, I didn't state that Layers is optional and the other stuff in https://github.com/opencontainers/image-spec/pull/313#discussion_r78716223 - I just opened this discussion in https://github.com/opencontainers/image-spec/pull/308 because there are some discrepancies around the repo.

I would wait for @vbatts @philips @stevvooe and others to chime in here before making any other fix to this PR

manifest.md
```diff
@@ -139,7 +139,7 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
...
     The array MUST have the base image at index 0.
     Subsequent layers MUST then follow in the order in which they are to be layered on top of each other.
```
wking: These two lines should be restructured to avoid requiring a zero-index layer.  Something like:

    The root filesystem MUST match an empty directory onto which the layers have been [applied](layer.md#applying) in the listed order.

With “match” instead of “be” to allow folks to apply via a union filesystem or whatever.  As long as the finished product is right.

wking:
On Wed, Sep 14, 2016 at 02:39:35AM -0700, Antonio Murdaca wrote:
> Wait, I didn't state that Layers is optional and the other stuff in
> https://github.com/opencontainers/image-spec/pull/313#discussion_r78716223
> - I just opened this discussion in
> https://github.com/opencontainers/image-spec/pull/308 because there
> are some discrepancies around the repo.
>
> I would wait for @vbatts @philips @stevvooe and others to chime in
> here before making any other fix to this PR

Waiting for feedback is good.  My motivation for OPTIONAL is that
there's no need to require folks to set an empty array when they could
express the same thing more compactly by leaving the property out of
their JSON.

manifest.md
stevvooe: Let's not mangle this line. The fact that these are descriptors is way more important than whether or not it can be empty.

runcom:
> What is the purpose of a layerless manifest? I don't think makes any sense at all.

Indeed mine was just a question (for which I just removed 3 lines of code and happy to put them back)

runcom:
Damn new UI - @stevvooe this has been started from https://github.com/opencontainers/image-spec/pull/308 - I can re-put the layers check there if this is a non-sense

stevvooe:
> Damn new UI - 

I've already tweeted in anger. 😸 

> @stevvooe this has been started from #308 - I can re-put the layers check there if this is a non-sense

I'll be honest: I am confused here. What is the purpose of an image with no layers? The only thing I can imagine is a common distributed command that would run against a default rootfs (perhaps the host). Very weird and I am not sure if the use case is valid or viable.

runcom:
> Damn new UI -
> I've already tweeted in anger.

I did the same :smile_cat:

wking:
On Wed, Sep 14, 2016 at 01:15:20PM -0700, Stephen Day wrote:
> I'll be honest: I am confused here. What is the purpose of an image
> with no layers?

You might be mounting in the meat from somewhere outside the bundle
directory.  Or you might be using the initially-empty-rootfs container
to hold open a mount namespace for subcontainers [1].

[1]: https://github.com/opencontainers/runtime-spec/pull/395#issuecomment-213605476

stevvooe:
@wking How is that a valid use case for a distributable image? What are you distributing at that point? A command?

wking:
On Wed, Sep 14, 2016 at 01:55:48PM -0700, Stephen Day wrote:
> @wking How is that a valid use case for a distributable image? What
> are you distributing at that point? A command?

You're distributing the config, which has more than the command (but
not all that much more ;).

jonboulle:
While the use case might seem particularly uncommon, I don't think it's necessarily invalid, nor have a strong reason to disallow it (I don't see it as putting any real burden on implementers for example). So I'm okay with this change in principle.

stevvooe:
@jonboulle What does it mean, though?

jonboulle:
@stevooe similar to the previous suggestions - basically a distributable runtime/exec config, where the root filesystem might be supplied from the host, or a volume mount, or ...

stevvooe:
@jonboulle Ok, but I think this needs to be thought through further. Wouldn't it be better for the spec to reserve the layerless image and then we can think it through?

wking:
On Fri, Sep 16, 2016 at 01:15:15PM -0700, Stephen Day wrote:
> @jonboulle What does it mean, though?

With the wording of #318 making it explicit, it means that the
unpacked rootfs will be an empty directory with unspecified
attributes.

stevvooe:
> With the wording of #318 making it explicit, it means that the
unpacked rootfs will be an empty directory with unspecified
attributes.

Ok, I agree. Could we merge the intent of this PR with #318 and close this one?

wking:
On Fri, Sep 16, 2016 at 01:29:27PM -0700, Stephen Day wrote:
> @jonboulle Ok, but I think this needs to be thought through
> further. Wouldn't it be better for the spec to reserve the layerless
> image and then we can think it through?

No need to reserve it, just [1]:

  `layers` MUST contain at least one entry.

for now and drop that line once you're comfortable doing so.

[1]: https://github.com/opencontainers/image-spec/pull/318#issuecomment-247664716

jonboulle:
Let's resolve this via #318.

wking:
On Fri, Sep 16, 2016 at 01:59:25PM -0700, Jonathan Boulle wrote:
> Let's resolve this via #318.

I've closed #318 because there seemed to be a consensus around part of
what it was doing.  Part of what I proposed in #318 has landed via
#349.  Of the remaining bits, #408 has the wording that appears
uncontentious and #407 continues the discussion started by this PR
(which seems to still be unsettled [1]).

[1]: https://github.com/opencontainers/image-spec/pull/318#issuecomment-255177345

PR #317, opened 2016-09-16
layer: clarify wording around applying changesets

jonboulle:
The existing wording was extremely confusing so this is an attempt at
clearing it up. Still not convinced this is quite right but putting it
up for feedback. Alternatively we could scrap the explanation and lean on
something like tar's:
https://www.gnu.org/software/tar/manual/html_node/Dealing-with-Old-Files.html

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -204,7 +204,7 @@ The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
 ./etc/.wh.my-app-config
 ```
 
-where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
+To signify that the resource `./etc/my-app-config` MUST be removed when the changeset is applied, the basename of the entry is prefixed with `.wh.`.
 
 ## Applying Changesets
 
@@ -216,7 +216,7 @@ In the absence of any [whiteout](#whiteouts) files in a layer changeset, the arc
 
 ### Changeset over existing files
 
-This section describes applying an entry from a layer changeset if the file path already exists.
+This section specifies applying an entry from a layer changeset if the target path already exists.
 
 If the entry and the existing path are both directories, then the existing path's attributes MUST be replaced by those of the entry in the changeset.
 In all other cases, the implementation MUST do the semantic equivalent of the following:
--- a/layer.md
+++ b/layer.md
@@ -204,31 +204,29 @@ The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
 ./etc/.wh.my-app-config
 ```
 
-Where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
+where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
 
-## Applying
+## Applying Changesets
 
-Layer Changesets of [mediatype](./media-types.md) `application/vnd.oci.image.layer.v1.tar+gzip` are applied rather than strictly extracted in normal fashion for tar archives.
+Layer Changesets of [mediatype](./media-types.md) `application/vnd.oci.image.layer.v1.tar+gzip` are _applied_, rather than simply extracted as tar archives.
 
-Applying a layer changeset requires consideration for the [whiteout](#whiteouts) files.
+Applying a layer changeset requires special consideration for the [whiteout](#whiteouts) files.
+
 In the absence of any [whiteout](#whiteouts) files in a layer changeset, the archive is extracted like a regular tar archive.
-
 
 ### Changeset over existing files
 
-This section covers applying an entry in a layer changeset, if the file path already exists.
+This section describes applying an entry from a layer changeset if the file path already exists.
 
-If the file path is a directory, then the existing path just has it's attribute set from the layer changeset for that filepath.
-If the file path is any other file type (regular file, FIFO, etc), then the:
-* file path is unlinked (See [`unlink(2)`](http://linux.die.net/man/2/unlink))
-* create the file
-    * If a regular file then content written.
-* set attributes on the filepath
+If the entry and the existing path are both directories, then the existing path's attributes MUST be replaced by those of the entry in the changeset.
+In all other cases, the implementation MUST do the semantic equivalent of the following:
+- removing the file path (e.g. [`unlink(2)`](http://linux.die.net/man/2/unlink) on Linux systems)
+- recreating the file path, based on the contents and attributes of the changeset entry
 
 ## Whiteouts
 
 A whiteout file is an empty file with a special filename that signifies a path should be deleted.
-A whiteout filename consists of the prefix .wh. plus the basename of the path to be deleted.
+A whiteout filename consists of the prefix `.wh.` plus the basename of the path to be deleted.
 As files prefixed with `.wh.` are special whiteout markers, it is not possible to create a filesystem which has a file or directory with a name beginning with `.wh.`.
 
 Once a whiteout is applied, the whiteout itself MUST also be hidden.


jonboulle:
@vbatts 4u

layer.md
```diff
@@ -166,31 +166,29 @@ The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
...
-Applying a layer changeset requires consideration for the [whiteout](#whiteouts) files.
-In the absence of any [whiteout](#whiteouts) files in a layer changeset, the archive is extracted like a regular tar archive.
+Applying a layer changeset requires special consideration for the [whiteout](#whiteouts) files.
 
+In the absence of any [whiteout](#whiteouts) files in a layer changeset, the archive is extracted like a regular tar archive.
 
 ### Changeset over existing files
 
-This section covers applying an entry in a layer changeset, if the file path already exists.
+This section describes applying an entry from a layer changeset if the file path already exists.
```
wking: covers/describes → specifies?

wking: And maybe:

> This section specifies entry application if the target path already exists.

layer.md
```diff
@@ -166,31 +166,29 @@ The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
...
 
-If the file path is a directory, then the existing path just has it's attribute set from the layer changeset for that filepath.
-If the file path is any other file type (regular file, FIFO, etc), then the:
-* file path is unlinked (See [`unlink(2)`](http://linux.die.net/man/2/unlink))
-* create the file
-    * If a regular file then content written.
-* set attributes on the filepath
+If the entry and the existing path are both directories, then the existing path's attributes MUST be replaced by those of the entry in the changeset.
+In all other cases, the implementation MUST do the semantic equivalent of the following:
+- removing the file path (e.g. [`unlink(2)`](http://linux.die.net/man/2/unlink) on Linux systems)
```
wking: “removing the file” → “(recursively) removing the file”?  Most of the discussion in [the GNU docs you linked][1] is for “folks would be mad if we blew away a full tree and replaced it with a broken symlink”.  That makes sense for working filesystems, but we're building the rootfs from scratch here, so losing information is not a concern.  However, you don't want folks blindly using `unlink`, having it fail because the target is a directory and either dying or leaving it in place.  We *do* want them to get that old thing off the filesystem (whatever it takes ;), and the “(recursively)” hint may be enough to get that across.

Or maybe it's sufficient as it stands :p.

[1]: https://www.gnu.org/software/tar/manual/html_node/Dealing-with-Old-Files.html

jonboulle: I'm open to adding "recursively" as long as that is actually what is expected here - @vbatts @stevvooe ?

stevvooe: This cannot act recursively. It must only act on the immediate resource. For example, a link should _not_ be followed.

There are some ordering requirements for this to work correctly. Let's take the following directories from the applying layer:

```
/a/b
/a/b/c
```

`/a/b` should probably come before `/a/b/c` for this to work correctly. If `/a/b/d` is encountered, it needs to be left alone, unless the layer above specifies a removal.

wking:
On Tue, Sep 20, 2016 at 02:11:20PM -0700, Stephen Day wrote:
> -Where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
> +where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
>
> Could we reword this so it is not a partial sentence? It doesn't
> even really complete the clause above.

How about “Where the” → “The”?

> -If the file path is a directory, then the existing path just has it's attribute set from the layer changeset for that filepath.
> -If the file path is any other file type (regular file, FIFO, etc), then the:
> -* file path is unlinked (See [`unlink(2)`](http://linux.die.net/man/2/unlink))
> -* create the file
> -    * If a regular file then content written.
> -* set attributes on the filepath
> +If the entry and the existing path are both directories, then the existing path's attributes MUST be replaced by those of the entry in the changeset.
> +In all other cases, the implementation MUST do the semantic equivalent of the following:
> +- removing the file path (e.g. [`unlink(2)`](http://linux.die.net/man/2/unlink) on Linux systems)
>
> This cannot act recursively. It must only act on the immediate
> resource. For example, a link should _not_ be followed.

Following links (which I think should never be done) and recursively
removing existing conflicts before applying a new entry (which I think
should be done unless both the new enty and existing filesystem object
are both directories) seem like orthogonal considerations.

> There are some ordering requirements for this to work
> correctly. Let's take the following directories from the applying
> layer:
> 
> ```
> /a/b
> /a/b/c
> ```
>
> `/a/b` should probably come before `/a/b/c` for this to work
> correctly. If `/a/b/d` is encountered, it needs to be left alone,
> unless the layer above specifies a removal.

Where /a/b is a directory, I see no constraint on the tar-entry
ordering.  Where /a/b is not a directory (e.g. it's a regular file),
then I expect:

  /a/b
  /a/b/c

to result in a tree with /a/b/c, where /a/b has unspecified
attributes.  And I expect:

  /a/b/c
  /a/b

to result in a tree with /a/b, where /a/b is the file from the
tar-entry.

layer.md
jonboulle: @stevvooe what is supposed to happen if `/a/b/` is a directory and a changeset contains `/a/b` as a file?

layer.md
```diff
@@ -166,31 +166,29 @@ The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
 ./etc/.wh.my-app-config
 ```
 
-Where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
+where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
```
jonboulle: tweaked the wording

vbatts:
yep
LGTM

jonboulle:
Needs an LGTM from @stevvooe since he requested changes

layer.md
```diff
@@ -204,31 +204,29 @@ The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
 ./etc/.wh.my-app-config
 ```
 
-Where the basename name of `./etc/my-app-config` is now prefixed with `.wh.`, and will therefore be removed when the changeset is applied.
+The basename name of `./etc/my-app-config` is prefixed with `.wh.`, which signifies that it MUST be removed when the changeset is applied.
```
stevvooe: This is still fairly awkward. Do you intend to signify or to prefix? I think this sentence can get better.

stevvooe:
@jonboulle The first sentence is still quite awkward. If you're stuck, let me know and I'll suggest a wording. The rest looks fine.

jonboulle:
@stevvooe really not sure how to make this clearer ("signifies" is used in a similar way elsewhere in the document) so suggestions welcome!

layer.md
stevvooe: @jonboulle Maybe this:

> To signify that the resource `./etc/my-app-config` MUST be removed, the basename of the entry is prefixed with `.wh.`.

layer.md
stevvooe: I supposed I missed the distinction here. I think we need to be clear that this is recursive _only_ when the lstat (shallow) type of the resource is dir.

If `/a/b/...` exists, a non-directory entry `/a/b` would cause removal, recursively. Further entries prefixed with an already encountered, non-directory entry, would be ignored or cause an error.

If `/a/b/` exists and a new directory entries `/a/b/...` are encountered, the result will be a union of the two directory trees, unless whiteouts apply.

wking: > I think we need to be clear that this is recursive only when the lstat (shallow) type of the resource is dir.

No, it's only *not* a recursive delete when both the existing entry (lstat) and tar entry are dirs.  See the implementation in opencontainers/image-tools#42 for machine-readable phrasing.

> If `/a/b/…` exists, a non-directory entry `/a/b` would cause removal, recursively.

Yes.

> Further entries prefixed with an already encountered, non-directory entry, would be ignored or cause an error.

Already-encountered-ness has nothing to do with it.  I think we should just unpack as we read through the tarball.  Only the current `.wh.*` handling [jumps the tar order][1], and the remove-before-unpacking behavior is for non-whiteout entries.

> If `/a/b/` exists and a new directory entries `/a/b/…` are encountered, the result will be a union of the two directory trees, unless whiteouts apply.

This is true even without the “directory” limit on `/a/b/…`.  If the filesystem has `/a/b/c` and an `/a/b/not-c` entry is found in the tarball, then the result will have both `/a/b/c` and `/a/b/not-c`.  If an `/a/b` entry is found in the tarball, then it's just a metadata clobber (the result will still have `/a/b/c`).

[1]: https://github.com/opencontainers/image-spec/blob/d4ca161f431b7598b2ca58d58eecea6fcd5dfabb/layer.md#whiteouts

jonboulle:
adopted @stevvooe's wording

stevvooe:
LGTM

vbatts:
LGTM

PR #318, opened 2016-09-16
manifest: Explicitly base manifest.layers on an empty directory

wking:
And point out that layer authors interested in the ownership and other attributes of the unpacked root directory should explicitly set an entry for it.  I expect folks will mostly use `./`, but have made that an example in case they want to use other spellings (which unpackers should respect).

I've used "match"” instead of "be" to allow folks to apply via a union filesystem or whatever. As long as the finished product is right, compliance testers and users should be satisfied.

Spun off from [this comment][1] to answer [this request][2], cc @jonboulle.

[1]: https://github.com/opencontainers/image-spec/pull/313#r78716777
[2]: https://github.com/opencontainers/image-spec/pull/312#issuecomment-247647193

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -45,6 +45,7 @@ Unlike the [Manifest List](manifest-list.md), which contains information about a
     The array MUST have the base image at index 0.
     Subsequent layers MUST then follow in stack order (i.e. from `layers[0]` to `layers[len(layers)-1]`).
     The final filesystem layout MUST match the result of [applying](layer.md#applying) the layers to an empty directory.
+    The [ownership, mode, and other attributes](layer.md#file-attributes) of the initial empty directory are unspecified.
 
     Beyond the [descriptor requirements](descriptor.md#properties), the value has the following additional restrictions:
 
--- a/manifest.md
+++ b/manifest.md
@@ -41,9 +41,9 @@ Unlike the [Manifest List](manifest-list.md), which contains information about a
 
 - **`layers`** *array*
 
+    This OPTIONAL property references the [layers](layer.md) defining the container's root filesystem.
     Each item in the array MUST be a [descriptor](descriptor.md).
-    The array MUST have the base image at index 0.
-    Subsequent layers MUST then follow in stack order (i.e. from `layers[0]` to `layers[len(layers)-1]`).
+    Layers MUST be [applied](layer.md#applying) in stack order (i.e. from `layers[0]` to `layers[len(layers)-1]`).
     The final filesystem layout MUST match the result of [applying](layer.md#applying) the layers to an empty directory.
     The [ownership, mode, and other attributes](layer.md#file-attributes) of the initial empty directory are unspecified.
 
--- a/schema/image-manifest-schema.json
+++ b/schema/image-manifest-schema.json
@@ -30,7 +30,6 @@
   "required": [
     "schemaVersion",
     "mediaType",
-    "config",
-    "layers"
+    "config"
   ]
 }
--- a/specs-go/v1/manifest.go
+++ b/specs-go/v1/manifest.go
@@ -25,7 +25,7 @@ type Manifest struct {
 	Config Descriptor `json:"config"`
 
 	// Layers is an indexed list of layers referenced by the manifest.
-	Layers []Descriptor `json:"layers"`
+	Layers []Descriptor `json:"layers,omitempty"`
 
 	// Annotations contains arbitrary metadata for the manifest list.
 	Annotations map[string]string `json:"annotations"`


wking:
As this stands with 14720125, this PR focuses on the empty-directory base and is largely agnostic on whether `layers[0]` must exist.  I don't see a reason to require it (more discussion on this in #313), but if we want to drop the requirement this PR would also need [these changes][1].  And if we want to keep the requirement, this PR would need a line like:

> `layers` MUST contain at least one entry.

[1]: https://github.com/opencontainers/image-spec/pull/313#r78716223

wking:
Better wording for the unspecified initial directory in 1472012 → 7b92852.

manifest.md
```diff
@@ -140,9 +140,8 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 - **`layers`** *array*
 
     Each item in the array MUST be a [descriptor](descriptor.md).
-    The array MUST have the base image at index 0.
-    Subsequent layers MUST then follow in the order in which they are to be layered on top of each other.
-    The algorithm to create the final unpacked filesystem layout MUST be to first unpack the layer at index 0, then index 1, and so on.
```
stevvooe: Could we preserve the clarity of this wording?

wking:
On Fri, Sep 16, 2016 at 01:31:50PM -0700, Stephen Day wrote:
> > @@ -140,9 +140,8 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
>  - **`layers`** *array*
>  
>      Each item in the array MUST be a [descriptor](descriptor.md).
> -    The array MUST have the base image at index 0.
> -    Subsequent layers MUST then follow in the order in which they are to be layered on top of each other.
> -    The algorithm to create the final unpacked filesystem layout MUST be to first unpack the layer at index 0, then index 1, and so on.
> 
> Could we preserve the clarity of this wording?

One man's clarity is another's redundancy ;).  The “MUST be … in the
listed order” is used in several runtime-spec locations [1,2,3].  Do
you have a particular feature in the lines I'm removing that you feel
it doesn't cover?

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc2/config-linux.md#L203
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc2/config.md#L43
[3]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc2/config.md#L263

manifest.md
```diff
@@ -140,9 +140,8 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 - **`layers`** *array*
 
     Each item in the array MUST be a [descriptor](descriptor.md).
```
jonboulle: must the descriptor point to a layer mediatype?

jonboulle: must the descriptor point to a layer mediatype?

wking:
I've pushed 7b92852 → 809ff0e with a new commit to make layers
optional, as requested here [1] (I think I'm reading that right).
This pulls in my suggestions from [2] except for the change from #308.
#308 addresses content since spun off into image-tools, and that
change should land there if/when this PR lands with the OPTIONAL
semantics.

[1]: https://github.com/opencontainers/image-spec/pull/313#issuecomment-247702236
[2]: https://github.com/opencontainers/image-spec/pull/313#r78716223

manifest.md
stevvooe: @wking Please respond inline on the review.

> Do
you have a particular feature in the lines I'm removing that you feel
it doesn't cover?

You've changed the emphasis and meaning of the clause.

Before, the emphasis was on ordering, where now it is focusing on the nature of the base layer.

wking:
On Fri, Sep 16, 2016 at 01:41:23PM -0700, Jonathan Boulle wrote:
>      Each item in the array MUST be a [descriptor](descriptor.md).
> 
> must the descriptor point to a layer mediatype?

Yes, but I think that's orthogonal (#304).

wking:
On Fri, Sep 16, 2016 at 01:51:32PM -0700, Stephen Day wrote:
> > @@ -140,9 +140,8 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
>  - **`layers`** *array*
>  
>      Each item in the array MUST be a [descriptor](descriptor.md).
> -    The array MUST have the base image at index 0.
> -    Subsequent layers MUST then follow in the order in which they are to be layered on top of each other.
> -    The algorithm to create the final unpacked filesystem layout MUST be to first unpack the layer at index 0, then index 1, and so on.
> 
> @wking Please respond inline on the review.

I'm just replying to GitHub's mail, and it's sending a lot since this
new review feature landed.  I'll experiment with which message I
respond to and see if I can find a pattern to keep things grouped on
the PR.

> > Do you have a particular feature in the lines I'm removing that
> > you feel it doesn't cover?
>
> You've changed the emphasis and meaning of the clause.

I tried to add meaning (for the initial state) without removing
anything from the ordering.  What ordering meaning did I change?

> Before, the emphasis was on ordering, where now it is focusing on
> the nature of the base layer.

We obviously want to emphasize both.  There's a bit more to say about
the initial state here, because layer application is covered
externally in layer.md.  But maybe if I turn the sentence around:

  The root filesystem MUST match the result of applying
  [applied](layer.md#applying) in the listed order to an empty
  directory.

I would rather start with the empty directory in the sentence, since
that's where implementations will start during unpacking, but putting
applications first does increase the changes that readers see “in the
listed order” before they get bored and wander off (or whatever they
were doing with the directory-first order ;).

wking:
I've pushed 809ff0e → c22efd8 changing:

  The root filesystem MUST match an empty directory onto which the
  layers have been [applied](layer.md#applying) in the listed order.

to:

  The root filesystem MUST match the result of
  [applying](layer.md#applying) the entries to an intially empty
  directory in the listed order.

which I like more mostly because it avoid “onto which” and “have been
applied” ;).  Still has “in the listed order” at the end, but the idea
seems clear to me.  If there are clearer rephrasings that avoid
repeating ourselves, I'm happy to roll them in.  I'd just rather not
have “start at 0”, “follow in the order…”, *and* “first unpack the
layer at index 0, then index 1, and so on”.

manifest.md
```diff
@@ -139,10 +139,10 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 
 - **`layers`** *array*
 
+    This OPTIONAL property references the [layers](layer.md) defining the container's root filesystem.
     Each item in the array MUST be a [descriptor](descriptor.md).
-    The array MUST have the base image at index 0.
-    Subsequent layers MUST then follow in the order in which they are to be layered on top of each other.
-    The algorithm to create the final unpacked filesystem layout MUST be to first unpack the layer at index 0, then index 1, and so on.
+    The root filesystem MUST match the result of [applying](layer.md#applying) the entries to an intially empty directory in the listed order.
```
jonboulle: initially (but I'd rather just scrap this word)

jonboulle:
Getting better, though I'm not quite sure we're at the right wording yet...

wking:
I've pushed c22efd8 → 43d2e9a dropping “intially” [1].  It still has
“the initial empty directory” in the second line, and I think that
between that and the context it's clear that the directory only needs
to be empty before you start applying layers ;).

[1]: https://github.com/opencontainers/image-spec/pull/318#pullrequestreview-433790

jonboulle:
LGTM (good enough for now :-)

stevvooe:
@crosbymichael @mrunalp Could you please make sure this makes sense in the context of runc?

In docker, we do have a "scratch" layer but I think we explicitly call that out. I am not sure about the effect of adding an implied empty layer.

crosbymichael:
I don't think having a required empty is needed or a good thing.  If I specify one layer in the array i should only have 1 dir, not two.

wking:
On Tue, Sep 20, 2016 at 01:17:04PM -0700, Michael Crosby wrote:
> I don't think having a required empty is needed or a good thing.  If
> I specify one layer in the array i should only have 1 dir, not two.

How would you have two?  If you specify one layer in the array, you'll
have a rootfs that is just that layer tarball unpacked into an
initially empty directory.  The wording I'm floating here just ensures
that the runtime doesn't unpack the layer tarball into a non-empty
directory.

With:

  $ tar -tf layer.tar 
  ./
  ./abc

This would be the legal result of the unpacking:

  $ mkdir rootfs
  $ tar -C rootfs -xf layer.tar  
  $ tree rootfs/
  rootfs/
  └── abc

  0 directories, 1 file

But this would not be legal:

  $ mkdir rootfs
  $ touch rootfs/def
  $ tar -C rootfs -xf layer.tar
  $ tree rootfs/
  rootfs/
  ├── abc
  └── def

  0 directories, 2 files

crosbymichael:
Then I guess the wording is super confusing and hard to tell what is meant.

wking:
On Tue, Sep 20, 2016 at 01:55:34PM -0700, Michael Crosby wrote:
> Then I guess the wording is super confusing and hard to tell what is meant.  

“MUST match the result of applying the entries to an empty directory
in the listed order” seems fairly clear to me.  Can you explain what
your two-directory interpretation [1] looked like and/or propose
alternative wording to express [2]?

[1]: https://github.com/opencontainers/image-spec/pull/318#issuecomment-248421008
[2]: https://github.com/opencontainers/image-spec/pull/318#issuecomment-248430778

wking:
The bulk of the first commit here (which explicitly required an
initially empty target directory) were picked up in #349.  I've pushed
43d2e9a6 → 28a4a9d5 rebasing around #349, and the only remaining
changes are explicitly setting the attributes of the initial empty
directory unspecified and allowing an empty or unset layers array
(which seemed to be the consensus opinion after #313 [1,2]).

[1]: https://github.com/opencontainers/image-spec/pull/313#issuecomment-247702236
[2]: https://github.com/opencontainers/image-spec/pull/313#issuecomment-247624565

jonboulle:
lgtm

manifest.md
```diff
@@ -157,10 +157,11 @@ Unlike the [Manifest List](#manifest-list), which contains information about a s
 
 - **`layers`** *array*
 
+    This OPTIONAL property references the [layers](layer.md) defining the container's root filesystem.
     Each item in the array MUST be a [descriptor](descriptor.md).
-    The array MUST have the base image at index 0.
-    Subsequent layers MUST then follow in stack order (i.e. from `layers[0]` to `layers[len(layers)]`).
+    Layers MUST must be [applied](layer.md#applying) in stack order (i.e. from `layers[0]` to `layers[len(layers)]`).
```
vbatts: my mistake, but `layers[len(layers)]` -> layers[len(layers)-1]`

wking:
Rebased around #363 with 28a4a9d → ded80c0.

vbatts:
i think this LGTM
But i'm not certain that having layers as optional has a real use case.

(as an aside, it seems almost like the "layers" ought to just be "blobs" or "objects", as the config will reference the diff_ids as well in the expected order)

wking:
On Thu, Oct 06, 2016 at 11:12:45AM -0700, Vincent Batts wrote:
> (as an aside, it seems almost like the "layers" ought to just be
> "blobs" or "objects", as the config will reference the diff_ids as
> well in the expected order)

The manifest links to the config blob too, and that blob is not
referenced from the layers array.  So I prefer the current name to
either ‘blobs’ or ‘objects’.

vbatts:
LGTM

jonboulle:
LGTM but needs approval from steeeeve

stevvooe:
While this is elegant, I am slightly worried there is nothing practical here. From a conceptual point of few, it is interesting that layers are applied to an empty directory.

I did investigate this and ended up filing https://github.com/docker/docker/issues/27568.

I think other language in this PR is sufficient for a merge.

wking:
On Wed, Oct 19, 2016 at 05:48:56PM -0700, Stephen Day wrote:
> From a conceptual point of few, it is interesting that layers are
> applied to an empty directory.

There was previous discussion about an ./ entry in a layer in [1].  If
three image-spec maintainers can have a multi-comment discussion about
whether ./ is useful, I think we need to have a spec entry that
clarifies that it *is* useful.

And the empty initial directory also protects from [2], which I think
is important for sanity ;).

[1]: https://github.com/opencontainers/image-spec/pull/255#r77686727
[2]: https://github.com/opencontainers/image-spec/pull/318#issuecomment-248430778

stevvooe:
@wking 100 image spec maintainers could agree and still be wrong. There are three things at issue:

1. What is the initial state to which layers are applied?
2. Clarification around ordering.
3. The allowance of a container with no layers.

For the most part, the discussion around 1 and 2, which is partially addressed in this PR, is important and relevant. Number 3 is novel, the use cases aren't clear and introduces something that is not already done in practice, the implications of which aren't clear.

wking:
On Thu, Oct 20, 2016 at 10:46:46AM -0700, Stephen Day wrote:
> 1. What is the initial state to which layers are applied?
> 2. Clarification around ordering.
> 3. The allowance of a container with no layers.
> 
> For the most part, the discussion around 1 and 2, which is partially
> addressed in this PR, is important and relevant.

Part of (1) and all of (2) has already been addressed by #349 using
language initially discussed in this PR.  To keep discussion distinct,
I've spun off the two commits in this PR into their own PRs: #407 and
#408.

PR #321, opened 2016-09-18
layer.md: Fix some typo

TrumanLing:

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -90,7 +90,7 @@ Entries for the following files:
 
 ### Populate a Comparison Filesystem
 
-Create a new directory and initialize it with an copy or snapshot of the prior root filesystem.
+Create a new directory and initialize it with a copy or snapshot of the prior root filesystem.
 Example commands that can preserve [file attributes](#file-attributes) to make this copy are:
 * [cp(1)](http://linux.die.net/man/1/cp): `cp -a rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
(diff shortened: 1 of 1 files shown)


layer.md
```diff
@@ -181,7 +181,7 @@ In the absence of any [whiteout](#whiteouts) files in a layer changeset, the arc
 This section covers applying an entry in a layer changeset, if the file path already exists.
 
 If the file path is a directory, then the existing path just has it's attribute set from the layer changeset for that filepath.
-If the file path is any other file type (regular file, FIFO, etc), then the:
+If the file path is any other file type (regular file, FIFO, etc), then:
 * file path is unlinked (See [`unlink(2)`](http://linux.die.net/man/2/unlink))
```
wking: +1 to removing “the”, but we should also rephrase this to an imperative:

> unlink the file path (…)

But this section is getting overhauled in #317, so it may not be worth the trouble of fixing it here ;).

TrumanLing: @wking I have removed modification you mentioned. PTAL.

wking:
ff1291c1 looks good to me, but it would be easier to identify this
change in the logs if the commit message were something like:

  layer: Fix "an copy" -> "a copy" typo

TrumanLing:
@wking thanks for you suggestion. It is OK. PTAL

wking:
Looks great, thanks :)

vbatts:
LGTM

philips:
LGTM

Issue #218, opened 2016-08-29
Some defination in serialization.md may conflict with README.md?


coolljt0725:
According to the description [https://github.com/opencontainers/image-spec/blob/master/README.md#running-an-oci-image] in README.md

>>This entire workflow should support the UX that users have come to expect from container engines like Docker and rkt: primarily, the ability to run an image with no additional arguments:

>>docker run example.com/org/app:v1.0.0
rkt run example.com/org/app,version=v1.0.0
To support this UX the OCI Image Format contains sufficient information to launch the application on the target platform (e.g. command, arguments, environment variables, etc).

So the goal is to run an image with no additional arguments and the OCI Image Format will contains the sufficient information. But during reading the https://github.com/opencontainers/image-spec/blob/master/serialization.md#container-runconfig-field-descriptions , I found the the `Config` MAY be null
>>config struct
The execution parameters which should be used as a base when running a container using the image. This field can be null, in which case any execution parameters should be specified at creation of the container.

and also the `Cmd` and `Entrypoint` MAY be null
>>Entrypoint array of strings
A list of arguments to use as the command to execute when the container starts. This value acts as a default and is replaced by an entrypoint specified when creating a container. This field MAY be "null".
Cmd array of strings
Default arguments to the entrypoint of the container. These values act as defaults and are replaced with any specified when creating a container. If an Entrypoint value is not specified, then the first entry of the Cmd array should be interpreted as the executable to run. This field MAY be "null".

The `config` or  both `Cmd` and `Entrypoit` is null is not sufficient  to run a container with no additional arguments. 

There does some images which `Cmd` and `Entrypoint` are null, we can't launch a container if we don't specified it on creation. So I think we should change the README.md or change the defination in serialization.md or I just misunderstand it? :)

philips:
Just because we should support that UX does not mean that every image MUST support that UX. For example there are Debian base images that don't really make sense to run.

PR #152, opened 2016-06-16
serialization: Entrypoint and Cmd nullable

runcom:
This is built on top of #151 - refer only to the last commit for review

@philips feel free to either cherry-pick my commit from here or wait #151 to be merged first (and I'll rebase)

This patch makes `Entrypoint` and `Cmd` fields in the serialization schema nullable + adds tests

/cc @vbatts

PR #323, opened 2016-09-18
[README] Add a should

coolljt0725:

To support the UX describe above, OCI Image Format `should` contains sufficient information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ This entire workflow supports the UX that users have come to expect from contain
 * docker run example.com/org/app:v1.0.0
 * rkt run example.com/org/app,version=v1.0.0
 
-To support this UX the OCI Image Format contains sufficient information to launch the application on the target platform (e.g. command, arguments, environment variables, etc).
+To support this UX the OCI Image Format should contains sufficient information to launch the application on the target platform (e.g. command, arguments, environment variables, etc).
 
 ## FAQ
 


wking:
I don't think we need this.  The “sufficient information” line guides
the development of this spec, it's not a requirement for either
implementations or image authors.  And see
opencontainers/runtime-spec#536 and some discussion around [1] about
avoiding “should” in favor of either other words or the normative
“SHOULD”.  In this case, I think the text is already appropriately
using the other-words approach.

[1]: https://github.com/opencontainers/image-spec/pull/233#r77873139

coolljt0725:
@wking  I'm not sure if this is need or not
There is a description above this sentence says that:
>This entire workflow supports the UX that users have come to expect from container engines like Docker and rkt: primarily, the ability to run an image with no additional arguments:

>docker run example.com/org/app:v1.0.0
rkt run example.com/org/app,version=v1.0.0

to support this UX, the OCI Image *should* contains sufficient information. If Image doesn't contain some information such as `entrypoint` `cmd`, the UX describe above is impossible.  The image must contains sufficient information to support this UX.

wking:
On Sun, Sep 18, 2016 at 06:29:40PM -0700, Lei Jitang wrote:
> … to support this UX, the OCI Image *should* contains sufficient
> information. If Image doesn't contain some information such as
> `entrypoint` `cmd`, the UX describe above is impossible.  The image
> must contains sufficient information to support this UX.

I agree in prinicple that this is a worthy goal.  However, SHOULD
means it's a normative recommendation, and “contains sufficient
information…” is not specific enough to be testable.  Requirements on
the content should live at the property level, e.g. [1] where we
explicitly allow entrypoint to be null (more on why in #152).

[1]: https://github.com/opencontainers/image-spec/blame/v0.5.0/serialization.md#L259

coolljt0725:
@wking 
> Requirements on
the content should live at the property level, e.g. [1] where we
explicitly allow entrypoint to be null (more on why in #152).

I once open a issue for this #218 , and @philips  answered my question
>Just because we should support that UX does not mean that every image MUST support that UX. For example there are Debian base images that don't really make sense to run.

From what   I understanding, not all the image MUST support that UX, but if the image support that UX, it should contains sufficient information.

wking:
On Sun, Sep 18, 2016 at 11:35:01PM -0700, Lei Jitang wrote:
> From what I understanding, not all the image MUST support that UX,
> but if the image support that UX, it should contains sufficient
> information.

If the image wants to support that UX, it MUST contain entrypoint,
right?  But still, this is something you'll want to point out *where
we define the entrypoint property*.  Making a normative statement here
means you'd have to define “contains sufficient information…”
somewhere, and I don't see that defined anywhere at the moment.
There's just the wishy “e.g. command, …”, and that doesn't sound like
a normative definition.

On the other hand, a bit of handwaving around an informative hint
(“here's a UX to think about when deciding which properties to set”)
is fine.  We just don't want to use the normative “SHOULD” for such a
hint, and I'd rather avoid the informative “should” to avoid confusion
with the normative “SHOULD” (links to previous “avoid ‘should’”
discussion in [1]).

[1]: https://github.com/opencontainers/image-spec/pull/323#issuecomment-247851403

coolljt0725:
@wking  `sufficient information` is hard to define, I think different image has different `sufficient information`, the image provider knows what the `sufficient information` is to support such UX.

>On the other hand, a bit of handwaving around an informative hint
(“here's a UX to think about when deciding which properties to set”)
is fine.  We just don't want to use the normative “SHOULD” for such a
hint, and I'd rather avoid the informative “should” to avoid confusion
with the normative “SHOULD” (links to previous “avoid ‘should’”
discussion in [1]).

If we want drop the `should`, I think we could re-write the sentence.

wking:
On Mon, Sep 19, 2016 at 07:03:24PM -0700, Lei Jitang wrote:
> @wking  `sufficient information` is hard to define…

Agreed, which is why I want to stay away from “SHOULD”.

> If we want drop the `should`, I think we could re-write the
> sentence.

I think the sentance was fine how it was, but am open to other
rephrasings.  Maybe something like:

  The OCI Image Format supports this UX by defining configuration
  properties sufficient to launch the application on the target
  platform (e.g. command, arguments, environment variables, etc.).

which is more clearly a guiding principle for developing the spec and
not a requirement placed on image authors (with those image-author
requirements living where the properties are defined).

vbatts:
@coolljt0725 this "should" seems to be an enforcement on the spec itself, not something enforcing on implementers. Perhaps this is not needed?

coolljt0725:
@vbatts  Yes, it's on the spec itself

PR #325, opened 2016-09-19
HACKING.md:jq should be version 1.5 or higher

TrumanLing:

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -61,7 +61,7 @@ $ make schema-fs
 This target auto-formats all JSON files in the `schema` directory using the `jq` tool.
 
 Prerequisites:
-* [jq](https://stedolan.github.io/jq/)
+* [jq](https://stedolan.github.io/jq/) >=1.5
 
 Invocation:
 ```


wking:
Why do we need ≥ 1.5?  Can you list a specific feature?

TrumanLing:
@wking I tried it with jq version 1.3 and 1.4, and it returned error ``

wking:
On Sun, Sep 18, 2016 at 09:54:03PM -0700, TrumanLing wrote:
> @wking I tried it with jq version 1.3 and 1.4, and it returned error
> ``

``?  Is that a typo?  Or was the error message really an empty string?
It looks like our only jq invocation for the fmt target in the
Makefile.  Which part of that invocation is not supported by 1.3 and
1.4?  Maybe we can find an invocation that also works with the older
versions.

TrumanLing:
@wking it is not a typo, we should show detailed information about jq version so that other people can have good experience of hacking. 

if one type commd `make fmt` to format JSON files under directory schema via `jq --indent 2 -M .`, it will report error `jq: Unknown option --indent` when jq version is below 1.5. I have verified those tree versions 1.3, 1.4 and 1.5, only version 1.5 can be successful.

wking:
On Sun, Sep 18, 2016 at 11:26:08PM -0700, TrumanLing wrote:
> … we should show detailed information about jq version so that other
> people can have good experience of hacking.

And/or adjust our usage so any remotely modern version of jq will
work.

> if one type commd `make fmt` to format JSON files under directory
> schema via `jq --indent 2 -M .`, it will report error `jq: Unknown
> option --indent` when jq version is below 1.5. I have verified those
> tree versions 1.3, 1.4 and 1.5, only version 1.5 can be successful.

And here's the ChangeLog for --indent on 2015-06-03 [1], while 1.4 was
cut 2014-06-09 [2].  So I'm fine merging this requirement with a
commit message that says --indent was added in 1.5, or I'm fine
leaving the requirement unspecified, dropping --indent, and
reformatting our JSON with jq's default indent (two spaces anyway?).

[1]: https://github.com/stedolan/jq/blob/jq-1.5/ChangeLog#L111-L113
[2]: https://github.com/stedolan/jq/tree/jq-1.4

TrumanLing:
@wking I prefer the first way you mentioned, because that will give user more agility than only default indent. And I have updated the commit message. PTAL

wking:
a2b3942b looks good to me.

vbatts:
LGTM

philips:
LGTM

PR #327, opened 2016-09-19
ref file name identical to tag

xiekeyang:
The ref file name should be identical to tag it presents, from implementation side. And this had better to be amended in `Implementor's Note` section.

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -54,6 +54,7 @@ In general the `mediatype` of this [descriptor][descriptors] object will be eith
 A common use case of refs is representing "tags" for a container image.
 For example, an image may have a tag for different versions or builds of the software.
 In the wild you often see "tags" like "v1.0.0-vendor.0", "2.0.0-debug", etc.
+Those tags will often be represented in an image-layout repository with matching refs names like "v1.0.0-vendor.0", "2.0.0-debug", etc.
 
 This illustrates the expected contents of a given ref, the manifest list it points to and the blobs the manifest references.
 


wking:
On Mon, Sep 19, 2016 at 03:50:21AM -0700, xiekeyang wrote:
> The ref file name should be identical to tag it presents, from
> implementation side.

This is not true.  See docker/docker#25779, which currently defaults
to “have the ref match the tag” but lets you override that default to
use any tag you like.  If you require the ref to match the tag,
storing multiple images in a single oci-layout becomes confusing
(always) or impossible (when both images use the same tag).

xiekeyang:
@wking 

> If you require the ref to match the tag,
storing multiple images in a single oci-layout becomes confusing
(always) or impossible (when both images use the same tag).

Here I'm unclear to `both images use the same tag`. Can you present a example?
Normally the layout is like:

```
.
├── blobs
│   └── sha256
│       ├── 2b8fd9751c4c0f5dd266fcae00707e67a2545ef34f9a29354585f93dac906749
│       ├── 8ddc19f16526912237dd8af81971d5e4dd0587907234be2b83e249518d5b673f
│       └── cf51ed98e7bdfd2a0dcd8dba91947e409b015a7582d0ac4703cf6338021e83b9
└── refs
    ├── latest
    └── v2.0
```

What is the layout for same tag case? Do you mean a layout includes images: hello-world:latest and busybox:latest ?

wking:
On Tue, Sep 20, 2016 at 02:09:19AM -0700, xiekeyang wrote:
> What is the layout for same tag case? Do you mean a layout includes
> images: hello-world:latest and busybox:latest ?

Yeah.  The example @stevvooe gives in docker/docker#25779 is to have
refs for both redis:latest and alpine:latest in the same image-layout
tarball.

stevvooe:
@wking I think we can add language to this effect that makes the common ref-tag mapping clear but allows for other mappings.

wking:
On Tue, Sep 20, 2016 at 12:06:28PM -0700, Stephen Day wrote:
> @wking I think we can add language to this effect that makes the
> common ref-tag mapping clear but allows for other mappings.

I think the “In the wild…” line below already makes this use case
pretty clear, and would rather avoid “are identical to” for clarifying
a common use case when tags/refs are not identical in general.  But
I'm open to other rewordings of this line.  How about:

  A common use case is to use “tag” names as refs.

which repaces the current “representing” with “use… as”.

xiekeyang:
@wking @stevvooe 
Fix it. How about new languages?
`with using "tag" as ref file names` is just `A common use case`.
`which is(not MUST) identical to ref file names` is in the case of `you often see`.
Consumers can be clear how to use it, and regard it as common case.

image-layout.md
```diff
@@ -51,9 +51,9 @@ Each object in the `refs` subdirectory MUST be of type `application/vnd.oci.desc
 In general the `mediatype` of this [descriptor][descriptors] object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
 
 **Implementor's Note:**
-A common use case of refs is representing "tags" for a container image.
+A common use case of refs is representing "tags" for a container image, with using "tag" as ref file names.
 For example, an image may have a tag for different versions or builds of the software.
-In the wild you often see "tags" like "v1.0.0-vendor.0", "2.0.0-debug", etc.
+In the wild you often see "tags" like "v1.0.0-vendor.0", "2.0.0-debug", etc which is(not MUST) identical to ref file names.
```
wking: This still seems a bit awkward to me.  This [this suggestion][1] is too brief, how about leaving the current there lines as they stand in master, to convey (by sentence):

1. Refs are often used for tags.
2. Images are often tagged by version or build.
3. Tags look like {examples}.

Then add an additional sentence to cover your additional point:

> Those tags will often be represented in an image-layout repository with matching refs like "v1.0.0-vendor.0”, “2.0.0-debug”, etc.

[1]: https://github.com/opencontainers/image-spec/pull/327#issuecomment-248403718

xiekeyang: Updated.

wking:
14597a36c looks good to me.

xiekeyang:
PTAL.

vbatts:
LGTM

stevvooe:
LGTM

PR #328, opened 2016-09-19
Use layer digest in the description of DiffID

coolljt0725:

`manifest digest` means the digest of manifest, I think this is wrong.

### DIFF
--- a/config.md
+++ b/config.md
@@ -28,7 +28,7 @@ Changing it means creating a new derived image, instead of changing the existing
 A layer DiffID is a SHA256 digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., `sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9`.
 Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using tar-split to save the tar headers.
 
-NOTE: the DiffID is different than the digest in the manifest list because the manifest digest is taken over the gzipped layer for `application/vnd.oci.image.layer.v1.tar+gzip` types.
+NOTE: the DiffID is different than the layer digest because it is taken over the uncompressed gzipped layer for `application/vnd.oci.image.layer.v1.tar+gzip` types.
 
 ### Layer ChainID
 


config.md
```diff
@@ -32,7 +32,7 @@ This specification uses the following terms:
     <dd>
 	A layer DiffID is a SHA256 digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., <code>sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9</code>.
 	Layers must be packed and unpacked reproducibly to avoid changing the layer ID, for example by using tar-split to save the tar headers.
-	NOTE: the DiffID is different than the digest in the manifest list because the manifest digest is taken over the gzipped layer for <code>application/vnd.oci.image.layer.tar+gzip</code> types.
+	NOTE: the DiffID is different than the digest in the manifest list because the digest in the manifest is taken over the gzipped layer for <code>application/vnd.oci.image.layer.tar+gzip</code> types.
```
wking: +1 to avoiding “manifest digest”.  But to make it extra clear that we're comparing with layer digests, how about a separate paragraph:

    The DiffID is different than the layer digest in the <a href="manifest.md#image-manifest-property-descriptions">manifest's <code>layers</code></a> because the layer digest is taken over the blob regardless of compression, while the DiffID is taken after removing any compression.
    For an <code>application/vnd.oci.image.layer.tar+gzip</code> layer.md, the layer digest is taken over the <code>application/vnd.oci.image.layer.tar+gzip</code> content, while the DiffID is take over the <code>application/vnd.oci.image.layer.tar</code> content.

Although that gets into defining a `+gzip` suffix (more on that in #316).

coolljt0725: @wking  You suggestion make it much more clear, but the `application/vnd.oci.image.layer.tar` is not explicit define at the moment.

wking:
On Mon, Sep 19, 2016 at 07:35:57PM -0700, Lei Jitang wrote:
> -	NOTE: the DiffID is different than the digest in the manifest list because the manifest digest is taken over the gzipped layer for <code>application/vnd.oci.image.layer.tar+gzip</code> types.
> +	NOTE: the DiffID is different than the digest in the manifest list because the digest in the manifest is taken over the gzipped layer for <code>application/vnd.oci.image.layer.tar+gzip</code> types.
> 
> @wking You suggestion make it much more clear, but the
> `application/vnd.oci.image.layer.tar` is not explicit define at the
> moment.

So we wait until #316 or similar lands with a definition of the +gzip
suffix?  I'll work up a PR for the suffix.

vbatts:
On 20/09/16 01:37 -0700, W. Trevor King wrote:
>So we wait until #316 or similar lands with a definition of the +gzip
>suffix?  I'll work up a PR for the suffix.

This change does not require waiting on the +gzip wording. That is one
of the ways that removing it here causes confusion.

vbatts:
This LGTM though perhaps it could go a step further to word that "the
digest in the manifest is taken over the entire object" or " ... opaque
blob"? So it is not so specific to just gzipped layers?

wking:
On Wed, Sep 21, 2016 at 12:57:02PM -0700, Vincent Batts wrote:
> On 20/09/16 01:37 -0700, W. Trevor King wrote:
> > So we wait until #316 or similar lands with a definition of the
> > +gzip suffix?  I'll work up a PR for the suffix.
>
> This change does not require waiting on the +gzip wording. That is
> one of the ways that removing it here causes confusion.

With the suffix defined, we could use the wording I've proposed in
[1].  @coolljt0725 liked that wording, but (correctly) pointed out
that it only makes sense once application/vnd.oci.image.layer.tar has
been defined [2].  With the +gzip PR (#332), both
application/vnd.oci.image.layer.tar and
application/vnd.oci.image.layer.tar+gzip will be defined, and we can
use the wording from [1].

[1]: https://github.com/opencontainers/image-spec/pull/328#discussion_r79444393
[2]: https://github.com/opencontainers/image-spec/pull/328#discussion_r79526815

config.md
stevvooe: Mentioning manifests here is the real problem. We should probably just say this is the content of the uncompressed layer tar.

coolljt0725:
updated 
@vbatts  @stevvooe

config.md
```diff
@@ -32,7 +32,7 @@ This specification uses the following terms:
     <dd>
 	A layer DiffID is a SHA256 digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., <code>sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9</code>.
 	Layers must be packed and unpacked reproducibly to avoid changing the layer ID, for example by using tar-split to save the tar headers.
-	NOTE: the DiffID is different than the digest in the manifest list because the manifest digest is taken over the gzipped layer for <code>application/vnd.oci.image.layer.tar+gzip</code> types.
+	NOTE: the DiffID is different than the layer digest because the digest is taken over the content of the uncompressted layer tar for <code>application/vnd.oci.image.layer.tar+gzip</code> types.
```
vbatts: uncompressed

coolljt0725: @vbatts  fixed, thx

coolljt0725:
@vbatts  @stevvooe  rebased after #333

vbatts:
LGTM

vbatts:
the needs another LGTM

config.md
```diff
@@ -28,7 +28,7 @@ Changing it means creating a new derived image, instead of changing the existing
 A layer DiffID is a SHA256 digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., `sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9`.
 Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using tar-split to save the tar headers.
 
-NOTE: the DiffID is different than the digest in the manifest list because the manifest digest is taken over the gzipped layer for `application/vnd.oci.image.layer.v1.tar+gzip` types.
+NOTE: the DiffID is different than the layer digest because the layer digest is taken over the content of the gzipped layer for `application/vnd.oci.image.layer.v1.tar+gzip` types.
```
stevvooe: > NOTE: the DiffID is different than the layer digest because it is taken over the uncompressed content of the gzipped layer for `application/vnd.oci.image.layer.v1.tar+gzip` types.

wking: Your suggested language points out the DiffID being over the uncompressed content.  @coolljt0725's current wording points out the layer digest being over the (raw) gzipped content.  Both of these are important for understanding the difference, and while the uncompressed-ness of DiffID is mentioned in the previous paragraph, I think it's worth talking about both of them in this note.  I have wording for that [in my earlier suggestion][1].  Until something like #332 or #388 lands, we could just use “over the uncompressed content” instead of “over the `application/vnd.oci.image.layer.tar` content”.


[1]: https://github.com/opencontainers/image-spec/pull/328#discussion_r79444393

coolljt0725:
@stevvooe  updated

coolljt0725:
ping @vbatts  needs re-LGTM after rebasing

vbatts:
LGTM

Issue #306, opened 2016-09-13
layer: hardlink verbiage


vbatts:
reference: https://github.com/opencontainers/image-spec/pull/255/files/f5a91ffe001afe2308628b112f3d9534fb94c2da#r78089784

Iron out the wording for a layer changeset preserving hardlinks (where applicable).

philips:
We have three options:

1. Be explicit saying they aren't supported
2. Saying it is implementation specific
3. Trying to write down something about how to explicitly handle it

We can look at what is happening in the current media type handling in Docker Engine. @stevvooe can try and hunt that down.

wking:
On Wed, Sep 14, 2016 at 02:25:55PM -0700, Brandon Philips wrote:
> We have three options:
> 
> 1. Be explicit saying they aren't supported
> 2. Saying it is implementation specific

(1) seems overly restrictive.  I think this is like #304, where you
want to allow implementations to extend the spec (and support
something like hardlinks that it doesn't currently define) but suggest
that portable images avoid these implementation-defined extensions
without and out-of-band reason to expect support.

> 3. Trying to write down something about how to explicitly handle it

This is fine too.  So I think the choices are whether to:

a. Define a hardlink serialization procedure.
b. Require implementations (on a given platform?) to support it.

stevvooe:
[Here](https://github.com/docker/docker/blob/master/pkg/archive/archive.go#L313) is the hard link handling in docker.

PR #336, opened 2016-09-21
layer: add hardlink handling information

vbatts:
Fixes: #306 

/cc @stevvooe @philips 

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -44,9 +44,45 @@ Where supported, MUST include file attributes for Additions and Modifications in
     * Group Name (`gname`) *secondary to `gid`*
 * Mode (`mode`)
 * Extended Attributes (`xattrs`)
-* Symlink reference (`linkname`)
+* Symlink reference (`linkname` + symbolic link type)
+* [Hardlink](#hardlinks) reference (`linkname`)
 
 [Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they lack consistent support across tar implementations.
+
+#### Hardlinks
+
+Hardlinks are a [POSIX concept](http://pubs.opengroup.org/onlinepubs/9699919799/functions/link.html) for having one or more directory entries for the same file on the same device.
+Not all filesystems support hardlinks (e.g. [FAT](https://en.wikipedia.org/wiki/File_Allocation_Table)).
+
+Hardlinks are possible with all [file types](#file-types) except `directories`.
+Non-directory files are considered "hardlinked" when their link count is greater than 1.
+Hardlinked files are on a same device (i.e. comparing Major:Minor pair) and have the same inode.
+The corresponding files that share the link with the > 1 linkcount may be outside the directory that the changeset is being produced from, in which case the `linkname` is not recorded in the changeset.
+
+Hardlinks are stored in a tar archive with type of a `1` char, per the [GNU Basic Tar Format][gnu-tar-standard] and [libarchive tar(5)][libarchive-tar].
+
+While approaches to deriving new or changed hardlinks may vary, a possible approach is:
+
+```
+SET LinkMap to map[< Major:Minor String >]map[< inode integer >]< path string >
+SET LinkNames to map[< src path string >]< dest path string >
+FOR each path in root path
+  IF path type is directory
+    CONTINUE
+  ENDIF
+  SET filestat to stat(path)
+  IF filestat num of links == 1
+    CONTINUE
+  ENDIF
+  IF LinkMap[filestat device][filestat inode] is not empty
+    SET LinkNames[path] to LinkMap[filestat device][filestat inode]
+  ELSE
+    SET LinkMap[filestat device][filestat inode] to path
+  ENDIF
+END FOR
+```
+
+With this approach, the link map and links names of a directory could be compared against that of another directory to derive additions and changes to hardlinks.
 
 ## Creating
 
@@ -279,4 +315,6 @@ Layers that have these restrictions SHOULD be tagged with an alternative mediaty
 [Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
 It is implementation-defined whether or not implementations upload layers tagged with this media type.
 
+[libarchive-tar]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#POSIX_ustar_Archives
+[gnu-tar-standard]: http://www.gnu.org/software/tar/manual/html_node/Standard.html
 [tar-archive]: https://en.wikipedia.org/wiki/Tar_(computing)


vbatts:
View the document: https://github.com/vbatts/oci-image-spec/blob/hardlink-verbiage/layer.md

layer.md
```diff
@@ -44,10 +44,45 @@ Where supported, MUST include file attributes for Additions and Modifications in
...
+#### Hardlinks
+
+Hardlinks are a [POSIX concept](http://pubs.opengroup.org/onlinepubs/9699919799/functions/link.html) for having one or more directory entries for the same file on the same device.
+Not all filesystems support hardlinks (e.g. [FAT](https://en.wikipedia.org/wiki/File_Allocation_Table)).
+
+Hardlinks are possible with all [file types](#file-types) except `directories`.
+Non-directory files with linkcount greater than 1 have hardlinks.
+The corresponding files that share the link with the > 1 linkcount may be outside the directory that the changeset is being produced from, in which case the `linkname` is not recorded in the changeset.
+
+Hardlinks are stored in a tar archive with type of `'1'`, per the [Basic Tar Format][tar-standard].
```
wking: Linking to the GNU docs makes it less clear where this fits in the basket of tar standards ;).  [libarchive's tar(5)][1] attributes this typeflag to ustar (IEEE Std 1003.1-1988).  IEEE Std 1003.1-1988 has been [superseded by IEEE Std 1003.1-2008][2], which seems to have been [superseded by IEEE Std 9945-2009][3].  The IEEE versions are paywalled, so I'm not sure what's inside.  But it's probably better to mention ustar and successors here, and possibly link to the libarchive man page for more historical context, instead of linking to the GNU docs.

[1]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#POSIX_ustar_Archives
[2]: http://standards.ieee.org/findstds/standard/1003.1-1988.html
[3]: http://standards.ieee.org/findstds/standard/1003.1-2008.html
[3]: https://www2.opengroup.org/ogsys/catalog/T101

philips:
LGTM

philips:
@vbatts this upset the pdf build somehow

layer.md
```diff
@@ -44,10 +44,45 @@ Where supported, MUST include file attributes for Additions and Modifications in
...
 
 [Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they lack consistent support across tar implementations.
 
+#### Hardlinks
+
+Hardlinks are a [POSIX concept](http://pubs.opengroup.org/onlinepubs/9699919799/functions/link.html) for having one or more directory entries for the same file on the same device.
+Not all filesystems support hardlinks (e.g. [FAT](https://en.wikipedia.org/wiki/File_Allocation_Table)).
+
+Hardlinks are possible with all [file types](#file-types) except `directories`.
+Non-directory files with linkcount greater than 1 have hardlinks.
```
stevvooe: Might be good to clarify that these must also have the same inode and devid. It looks something like https://github.com/stevvooe/continuity/blob/master/hardlinks_unix.go#L11 in practice.

layer.md
vbatts: You drop so many links and words, but they still resolve to the same thing, that the typeflag is `'1'`

layer.md
vbatts: yeah, I got at that in the pseudocode, but should put it in a sentence here too

layer.md
```diff
@@ -44,10 +44,45 @@ Where supported, MUST include file attributes for Additions and Modifications in
...
 * Extended Attributes (`xattrs`)
-* Symlink reference (`linkname`)
+* Symlink reference (`linkname` + symbolic link type)
+* [Hardlink](#hardlinks) reference (`linkname`)
 
 [Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they lack consistent support across tar implementations.
 
+#### Hardlinks
+
+Hardlinks are a [POSIX concept](http://pubs.opengroup.org/onlinepubs/9699919799/functions/link.html) for having one or more directory entries for the same file on the same device.
```
vbatts: @stevvooe i do say part of that here. So, make it more explicit?

wking:
On Thu, Sep 22, 2016 at 12:32:47PM -0700, Vincent Batts wrote:
> +Hardlinks are stored in a tar archive with type of `'1'`, per the [Basic Tar Format][tar-standard].
>
> You drop so many links and words, but they still resolve to the same
> thing, that the typeflag is `'1'`

*And* the tar flavor must be ustar-compatible.  If we depend on a
versioned spec like IEEE Std 1003.1-1988 (linking to the IEEE page
and/or the libarchive tar(5)), then we have a stable requirement.  If
we link to the GNU docs, and not even a particular version of the GNU
docs, things become too fluid for my taste.  For example, that page
currently has:

  *BEWARE* *BEWARE* *BEWARE* that the following information is still
  boiling, and may change.  Even if the OLDGNU format description
  should be accurate, the so-called GNU format is not yet fully
  decided.

above some newer structures.  And it also has (at the very bottom):

  For references, see ISO/IEC 9945-1:1990 or IEEE Std 1003.1-1990,
  pages 169-173 (section 10.1) for Archive/Interchange File Format;
  and IEEE Std 1003.2-1992, pages 380-388 (section 4.48) and pages
  936-940 (section E.4.48) for pax - Portable archive interchange.

So I think we should take that advice and reference one of those
specs.

vbatts:
updated. PTAL.

layer.md
```diff
@@ -44,10 +44,46 @@ Where supported, MUST include file attributes for Additions and Modifications in
...
+Not all filesystems support hardlinks (e.g. [FAT](https://en.wikipedia.org/wiki/File_Allocation_Table)).
+
+Hardlinks are possible with all [file types](#file-types) except `directories`.
+Non-directory files have hardlinks when the linkcount is greater than 1.
+Hardlinked files are on a same device (i.e. comparing Major:Minor pair), and have the same inode.
```
wking: nit: No need for the comma here; “hardlinked files” is the subject of both portions.

layer.md
```diff
@@ -44,10 +44,46 @@ Where supported, MUST include file attributes for Additions and Modifications in
...
+The corresponding files that share the link with the > 1 linkcount may be outside the directory that the changeset is being produced from, in which case the `linkname` is not recorded in the changeset.
+
+Hardlinks are stored in a tar archive with type of '`1`', per the [GNU Basic Tar Format][gnu-tar-standard] and [libarchive(5)][libarchive].
```
wking: nit: Quoting *and* backticking seems strange.  Can we just backtick or just quote?

wking: Also, the page you are linking is libarchive's tar(5), not libarchive(5).  I think you want a line like:

    Hardlinks are stored in a tar archive with type of `1` as specified by [IEEE Std 1003.1-1988][ieee1003.1-1998] (“ustar”, discussed [here][gnu-tar-standard] and [here][tar.5-ustar]).

    [ieee1003-1998]: http://standards.ieee.org/findstds/standard/1003.1-1988.html
    [gnu-tar-standard]: http://www.gnu.org/software/tar/manual/html_node/Standard.html
    [tar.5-ustar]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#POSIX_ustar_Archives

layer.md
stevvooe: @vbatts I still think that condition needs to be added to the clause above. The issue is that two files with the same inode but different devices ids are two separate hardlinks.

wking: On Thu, Sep 22, 2016 at 02:14:30PM -0700, Stephen Day wrote:
> @vbatts I still think that condition needs to be added to the clause
> above. The issue is that two files with the same inode but different
> devices ids are two separate hardlinks.

He covers that [in Markdown in the next paragraph][1].  Are you just asking for “file” → “inode” here?

[1]: https://github.com/opencontainers/image-spec/pull/336/files#diff-dc2bb3b4eddcffe96e682193926e8ff6R59

layer.md
vbatts: like you said that is paywalled. I'm not going to use that reference for now. The other docs include their own citing.

vbatts:
updated. PTAL

wking:
On Fri, Sep 23, 2016 at 11:33:05AM -0700, Vincent Batts wrote:
> like you said that is paywalled. I'm not going to use that reference
> for now. The other docs include their own citing.

The other docs cite many references, and I'm not sure which one we're
requiring (suggesting?) support for here.  From tar(5), it looks like
the old tar from Version 7 AT&T UNIX already supported linkflag 1
(typeflag in later specs) for hardlinks, and that all subsequent
versions discussed in tar(5) support that too.  IEEE Std 1003.1-2001's
pax requires linkpath to be UTF-8, although this doesn't seem to have
been a requirement before.  The GNU format (with an unversioned spec?)
supports the K typeflag to give linknames > 100 chars for subsequent
entries.  Which of those approaches should image-spec-compatible
libraries use?  Are K typeflags valid?  Are non-UTF-8 linknames valid?
Are we punting to the tar library to pick the tar format based on
magic values?  If so, can we just say that?  If not, can we say which
version of tar we're requiring?

wking:
If we want a freely-accessible standard, I'd suggest pax (IEEE Std
1003.1, 2013 Edition) [1].  It looks like Go has supported it since
v1.1 [2,3].

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13
[2]: https://codereview.appspot.com/6700047
[3]: https://github.com/golang/go/commit/106827990466b9246f9395233882b1e816df398a

vbatts:
On 23/09/16 12:00 -0700, W. Trevor King wrote:
>The other docs cite many references, and I'm not sure which one we're
>requiring (suggesting?) support for here.  From tar(5), it looks like
>the old tar from Version 7 AT&T UNIX already supported linkflag 1
>(typeflag in later specs) for hardlinks, and that all subsequent
>versions discussed in tar(5) support that too.  IEEE Std 1003.1-2001's
>pax requires linkpath to be UTF-8, although this doesn't seem to have
>been a requirement before.  The GNU format (with an unversioned spec?)
>supports the K typeflag to give linknames > 100 chars for subsequent
>entries.  Which of those approaches should image-spec-compatible
>libraries use?  Are K typeflags valid?  Are non-UTF-8 linknames valid?
>Are we punting to the tar library to pick the tar format based on
>magic values?  If so, can we just say that?  If not, can we say which
>version of tar we're requiring?

Really, what are you trying to accomplish with responses like this?
It is not constructive.

wking:
On Fri, Sep 23, 2016 at 12:37:53PM -0700, Vincent Batts wrote:
> On 23/09/16 12:00 -0700, W. Trevor King wrote:
> >The other docs cite many references, and I'm not sure which one we're
> >requiring (suggesting?) support for here.  From tar(5), it looks like
> >the old tar from Version 7 AT&T UNIX already supported linkflag 1
> >(typeflag in later specs) for hardlinks, and that all subsequent
> >versions discussed in tar(5) support that too.  IEEE Std 1003.1-2001's
> >pax requires linkpath to be UTF-8, although this doesn't seem to have
> >been a requirement before.  The GNU format (with an unversioned spec?)
> >supports the K typeflag to give linknames > 100 chars for subsequent
> >entries.  Which of those approaches should image-spec-compatible
> >libraries use?  Are K typeflags valid?  Are non-UTF-8 linknames valid?
> >Are we punting to the tar library to pick the tar format based on
> >magic values?  If so, can we just say that?  If not, can we say which
> >version of tar we're requiring?
> 
> Really, what are you trying to accomplish with responses like this?
> It is not constructive.

The idea with a spec like this is to define behavior so that different
implementations can interoperate reliably.  When there is an interop
problem betweem implementations A and B, it should be clear from the
spec whether implementation A is broken, implementation B is broken,
or the spec is insufficiently clear.  As it stands, if implementation
A produced a spec wit a K typeflag and implementation died unpacking
it, I am not clear whose fault that is.  That's a problem with the
spec.

If we require support for the pax format as defined by IEEE Std
1003.1, 2013 Edition [1] (this would be my preference), then it is
clearly implementation B which is broken.

If we explicitly make the tar version unspecified or
implementation-defined, then we're saying that interop issues like B's
choking on A's layer are not a problem we're trying to solve in this
spec.

But by doing neither of those, layer interop is just confusing.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13

layer.md
```diff
@@ -44,10 +44,46 @@ Where supported, MUST include file attributes for Additions and Modifications in
...
 
 [Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they lack consistent support across tar implementations.
 
+#### Hardlinks
+
+Hardlinks are a [POSIX concept](http://pubs.opengroup.org/onlinepubs/9699919799/functions/link.html) for having one or more directory entries for the same file on the same device.
+Not all filesystems support hardlinks (e.g. [FAT](https://en.wikipedia.org/wiki/File_Allocation_Table)).
+
+Hardlinks are possible with all [file types](#file-types) except `directories`.
+Non-directory files have hardlinks when the linkcount is greater than 1.
```
stevvooe: > Non-directory files are considered "hardlinked" when their link count is greater than 1.

stevvooe:
@vbatts I read it through again it looks good.

vbatts:
updated.

stevvooe:
LGTM

philips:
LGTM

(aside: on mini-vacation so I won't be back into the full swing until Thursday)

Issue #296, opened 2016-09-12
image-tools naming


runcom:
I have a working branch/repo at https://github.com/runcom/image-tools which basically removes from this repo the `cmd/*` bits to be later become `opencontainers/image-tools`.

Basically that big `oci-image-tool` became a set of smaller binaries.
Commands under `cmd/*` looks like this:
```sh
# s/runcom/opencontainers/g to test them out
$ go get github.com/opencontainers/image-tools/cmd/oci-validate
$ go get github.com/opencontainers/image-tools/cmd/oci-unpack
$ go get github.com/opencontainers/image-tools/cmd/oci-create-runtime-bundle
```
The `oci-` prefix is ugly but is making sure commands like `unpack` and `validate` do not conflicts with already installed binaries - should we care? because it's much nicer w/o that prefix.

Thoughts? Preferences?

@philips @vbatts @stevvooe @s-urbaniak

jonboulle:
I don't see a problem with the prefix.

wking:
On Mon, Sep 12, 2016 at 03:18:19AM -0700, Antonio Murdaca wrote:
> Commands under `cmd/*` looks like this:
> ```sh
> # s/runcom/opencontainers/g
> $ go get github.com/opencontainers/image-tools/cmd/oci-validate
> $ go get github.com/opencontainers/image-tools/cmd/oci-unpack
> $ go get github.com/opencontainers/image-tools/cmd/oci-create-runtime-bundle
> ```
> The `oci-` prefix is ugly but is making sure commands like `unpack`
> and `validate` do not conflicts with already installed binaries -
> should we care? because it's much nicer w/o that prefix.
> 
> Thoughts? Preferences?

The prefix should be oci-image-… to avoid conflicting with
oci-runtime-… (which is presumably what ocitools will be installing
after opencontainer/tob#18).

wking:
On Mon, Sep 12, 2016 at 03:18:19AM -0700, Antonio Murdaca wrote:
> $ go get github.com/opencontainers/image-tools/cmd/oci-validate
> $ go get github.com/opencontainers/image-tools/cmd/oci-unpack
> $ go get github.com/opencontainers/image-tools/cmd/oci-create-runtime-bundle
> ```
> The `oci-` prefix is ugly but is making sure commands like `unpack`
> and `validate` do not conflicts with already installed binaries…

Wait, are you switching to install these as separate binaries?
Instead of the current situation where they are sub-commands of a
[...]

Issue #150, opened 2016-06-16
oci-image-tool: Drop http.FileSystem?


wking:
While reading #148, the [`http.FileSystem` section][1] struck me as the wrong approach.  “I'd like to validate this content against this JSON Schema” is easy to support generically, and keeping the schemas themselves outside of that tool allows you to use an existing version of the tool to validate new content against new schemas.  If you compile the schemas in, you have to rebuild the tool to get access to the new schemas.

For an alternative approach, see the in-flight opencontainers/runtime-spec#490.  This approach allows a generic JSON-Schema validator to use whichever schema you point it at.  It would be nice to put the schemas up on www.opencontainers.org/schema/…, but even without that you can [use the raw GitHub URLs][2].  For oci-image-tool, you'd need a method to convert the detected/configured media type / version to a schema URL, but you already do that for `http.FileSystem`.  The URLs would just get a bit longer, and if we start removing schemas you'd have to add logic to pin to the last image-spec version which carried a schema for your media type.

If this sounds like a reasonable approach, I can work up a PR.

[1]: https://github.com/opencontainers/image-spec/pull/148/files#diff-2cb561c5e46512465f54a0f84d421c80R45
[2]: https://github.com/opencontainers/runtime-spec/pull/490/files#diff-751d571a4cba0b0155c3d3bf793e08e8R44

philips:
I would be fine with making this an optional feature. Requiring network traffic by default doesn't seem like the best default however.

wking:
On Thu, Jun 16, 2016 at 03:01:46PM -0700, Brandon Philips wrote:
> I would be fine with making this an optional feature. Requiring
> network traffic by default doesn't seem like the best default
> however.

Does Go have defaults for storing associated data (like Python's [1])?
Baking it into the binary seems like “here's my spell checker with a
compiled in snapshot of today's dictionary” (and I hope nobody does
that ;).

[1]: https://docs.python.org/3/distutils/setupscript.html#installing-package-data

philips:
@wking no, this is the "right" way os doing this for Go. Compiling in the data in a section via a const is the common practice. Not great but that is how it is. 

We have a bunch of tools at CoreOS that use this fake filesystem thingie as default and then let people override in "dev" mode.

s-urbaniak:
Agreeing to all @philips said. I can work on a possibility to look up schema via http/s, or if you @wking would like to do a PR, that'd be fine too.

wking:
[...]

PR #337, opened 2016-09-21
Drop image tools cmd/ and image/

wking:
They are now in opencontainers/image-tools.  Also update the
dependencies.  Details in the commit messages.


philips:
@wking travis build needs to be updated as well.

stevvooe:
I thought the plan was to maintain the validation package here. How will we continue to validate the specification if we accept this PR?

wking:
On Wed, Sep 21, 2016 at 03:05:54PM -0700, Brandon Philips wrote:
> @wking travis build needs to be updated as well.

Oops, thanks :).  Did that, and also updated Makefile, HACKING.md, and
.gitignore with 90bbb29 → 2fcdec9.

wking:
On Wed, Sep 21, 2016 at 03:30:30PM -0700, Stephen Day wrote:
> I thought the plan was to maintain the validation package here. How
> will we continue to validate the specification if we accept this PR?

I don't really have a preference for where validation lives.  The
runtime projects keep the JSON Schema and a generic JSON Schema
validator in runtime-spec [1].  Additional validation beyond what JSON
Schema can cover happens in an runtime-tools, where they will
hopefully soon become their own package
(opencontainers/runtime-tools#188).  The runtime-tools validation will
also hopefully soon include the JSON Schema validation as a portion of
the full validation (opencontainers/runtime-tools#197).  That was the
pattern I expected image-spec to follow, but a lot of that pattern is
still in-flight.

While runtime-spec occasionally has to fix buggy JSON Schema or Go
structures (opencontainers/runtime-spec#491,
opencontainers/runtime-spec#502, opencontainers/runtime-spec#581), I
think that we're far more likely to have churn on the full validator.
I think the version-decoupling which motivated the tool/spec split
(opencontainers/runtime-tools#83) applies to that code.  Plus the
separation makes it easier for the validation tooling to support
multiple spec versions at once (some discussion in
opencontainers/image-tools#9).

However, I agree that having local validation to support ‘make
validate-examples’ is quite useful.  I think that validating the
examples as:

1. legal JSON, and
2. instances of the appropriate JSON Schema

is sufficient to catch most example errors.  Weighing the benefit of
covering errors that cannot be represented in JSON Schema with the
local validate-examples vs. the cost of maintaining a validator
compatible with multiple spec versions inside the spec repository, my
personal preference is to keep that additional testing in runtime-spec
instead of in image-spec.

So having outlined the runtime approach and given some background on
why I think it's a good idea, I'm still happy to go whichever way the
image-spec maintainers want to go on this one ;).  If your target
differs widely from what I have here, it may be faster to just submit
an alternative PR (in which case feel free to cherry-pick anything
from here that seems useful).

[1]: https://github.com/opencontainers/runtime-spec/tree/v1.0.0-rc2/schema

stevvooe:
@wking Could you summarize what you are trying to say?

wking:
On Thu, Sep 22, 2016 at 06:49:58PM -0700, Stephen Day wrote:
> @wking Could you summarize what you are trying to say?

I like the PR as it stands, for the reasons I've listed [1].  It
matches what runtime-spec is doing.  And you'll still have reasonable
validation coverage with just the JSON Schema locally.

[1]: https://github.com/opencontainers/image-spec/pull/337#issuecomment-248816631

stevvooe:
@wking 

> How will we continue to validate the specification if we accept this PR?

wking:
On Fri, Sep 23, 2016 at 03:49:52PM -0700, Stephen Day wrote:
> > How will we continue to validate the specification if we accept
> > this PR?

This PR keeps all of the JSON Schema stuff locally, and I explain in
[1] why I think that is sufficient for validating local examples.

[1]: https://github.com/opencontainers/image-spec/pull/337#issuecomment-248816631

stevvooe:
@wking Explain it in a single paragraph so I don't have to read a book. Again, please answer the question:

> How will we continue to validate the specification if we accept this PR?

Rules for your response:

1. Keep it short.
2. Avoid obtuse references to your own commentary.
3. Keep your language plain and simple to understand.

To be clear, the agreement we had when creating the tools repo was to leave the validation code in the spec repo. This includes multiple versions of validation, since the specification has to consider multiple versions, as well.

The tools repo was just supposed to be a cmd directory.

wking:
On Mon, Sep 26, 2016 at 11:59:06AM -0700, Stephen Day wrote:
> @wking Explain it in a single paragraph so I don't have to read a
> book.

I can give one paragraph responses, but they cut corners.  Linking to
more complete comments gives folks with more time a way to find out
what what those corners were.

> To be clear, the agreement we had when creating the tools repo was
> to leave the validation code in the spec repo. This includes
> multiple versions of validation, since the specification has to
> consider multiple versions, as well.

I think it is better to have validation beyond comparison with the
current JSON Schema in image-tools (which is how the runtime repos
split things up).  Can you explain why a v2.0 specification would have
to be able to validate images compatible with the v1.0 specification?
Or why a v1.4 specification would have to be able to validate a
strictly v1.0 image?  It seems orthongal to me.  The only examples
inlined in a v1.4 spec should be v1.4 examples.

> The tools repo was just supposed to be a cmd directory.

I'm not sure there is a consensus around that among image-spec
maintainers.  For example @philips punted the image/-only issues like
#242 to image-tools.  And even ported some schema/-only issues like
#150 to image-tools.

stevvooe:
> I'm not sure there is a consensus around that among image-spec
maintainers.

We had this consensus in the face to face meeting. I consider the fact that this in the process changed an oversight at minimum.

> Can you explain why a v2.0 specification would have
to be able to validate images compatible with the v1.0 specification?
Or why a v1.4 specification would have to be able to validate a
strictly v1.0 image?  It seems orthongal to me.  The only examples
inlined in a v1.4 spec should be v1.4 examples.

Specifications don't exist in a vacuum. v1.4 needs to be backwards compatible with v1.0 to meet semantic versioning requirements.

This isn't that hard and having the specification and validator in sync is more important than these other concerns.

wking:
On Mon, Sep 26, 2016 at 12:33:33PM -0700, Stephen Day wrote:
> > Can you explain why a v2.0 specification would have to be able to
> > validate images compatible with the v1.0 specification?  Or why a
> > v1.4 specification would have to be able to validate a strictly
> > v1.0 image?  It seems orthongal to me.  The only examples inlined
> > in a v1.4 spec should be v1.4 examples.
>
> Specifications don't exist in a vacuum. v1.4 needs to be backwards
> compatible with v1.0 to meet semantic versioning requirements.

Agreed.  But the backward-compatibility test is “Does v1.4 tooling
interpret v1.0 images with the same semantics as v1.0 tooling on
issues covered by the v1.0 spec?”.  If we're not changing the
semantics of existing fields on minor bumps, I don't know why you'd
need v1.0 schema validation to check that.

> This isn't that hard and having the specification and validator in
> sync is more important than these other concerns.

Not having typos in the spec is important.  But the v1.4 spec isn't
going to *also* include the v1.0 spec (in a subdirectory?).  So the
only thing worth keeping in sync in this repository is a v1.4
validator.  And because of churn concerns [1], I don't think it's
worth keeping more than the JSON Schema validation in sync either.
We're ok with a bit of delay before the tool repos catch up [2] to
cover the cases that JSON Schema cannot handle.

[1]: https://github.com/opencontainers/image-spec/pull/337#issuecomment-248816631
[2]: https://github.com/opencontainers/tob/pull/18/files#diff-07f935bfb8e1871445d8bf2ad3759f9bR52

wking:
On Mon, Sep 26, 2016 at 11:59:06AM -0700, Stephen Day wrote:
> The tools repo was just supposed to be a cmd directory.

I took a local stab at rerolling this branch to only drop cmd/ (since
that seems to be something we agree on), but glide.yml requires
runtime-spec ^1.0.0-rc1, which (for some reason, despite SemVer
pointing out that pre-releases may not satisfy their version's compat
requirements [1]) happily pulls in
opencontainers/runtime-spec@7dab1a24 (tag v1.0.0-rc2).  That brings in
the *Linux change from opencontainers/runtime-spec#502, and I'm not
interested enough in the image→runtime config translation to want to
port it around that change.  So if you want a PR that just drops cmd/,
someone else will have to write it ;).  Or this repo can pin the
runtime-spec dependency at a particular version so it doesn't float
with update-deps.

[1]: http://semver.org/#spec-item-9

vbatts:
@stevvooe is this PR in line with what you described last week on the call?

stevvooe:
@vbatts This looks like it suffices.

@wking Please cleanup the history. This should only be two commits.

wking:
On Wed, Oct 19, 2016 at 04:49:43PM -0700, Stephen Day wrote:
> @wking Please cleanup the history. This should only be two commits.

I think it's easier to review and understand when the more detailed
changes aren't lost in a sea of ‘rm -rf’ noise, but whatever ;).
Rebased onto master and squashed into two commits (the things I did
and the things glide did) with 2fcdec9 → 42b0bea.

stevvooe:
LGTM

vbatts:
LGTM

cmd/oci-image-tool/main.go
```diff
@@ -1,41 +0,0 @@
...
-
-	cmd.AddCommand(newValidateCmd(stdout, stderr))
-	cmd.AddCommand(newUnpackCmd(stdout, stderr))
-	cmd.AddCommand(newBundleCmd(stdout, stderr))
-
-	if err := cmd.Execute(); err != nil {
-		stderr.Println(err)
-		os.Exit(1)
-	}
-}
```
q384566678: You delete this file in this place , but in the image-tools and did not restore the relevant content, why not？ I think it is useful to use.

wking: image-tools has all of the previous subcommands, just promoted to independent binaries.  What were you missing?

q384566678: I mean, like the file, the set of all sub commands into a whole.

wking: I think the image-tools maintainers made an explicit decision to split the former sub-commands out into their own binaries (#296, opencontainers/image-tools#1).  I don't have strong feelings on that one way or the other.  What this PR is about is that none of those commands belong in the spec ;).  If you feel like the wrapping command is important, it's probably best to open an issue arguing your case in image-tools.

PR #339, opened 2016-09-23
canonicalization: Add recommendations for canonicalization

wking:
Spun off from #259, now that we seem to have reached a consensus for “SHOULD canonical JSON” there.  I've set this up so we have space to add canonicalization recommendations for other formats, although the only other basic type discussed in this repository is a gzipped tarball and that's more than I want to bite off at the moment ;).

### DIFF
--- a/canonicalization.md
+++ b/canonicalization.md
@@ -1,0 +1,21 @@
+# Canonicalization
+
+OCI Images [are](descriptor.md) [content-addressable](image-layout.md).
+One benefit of content-addressable storage is easy deduplication.
+Many images might depend on a particular [layer](layer.md), but there will only be one blob in the [store](image-layout.md).
+With a different serialization, that same semantic layer would have a different hash, and if both versions of the layer are referenced there will be two blobs with the same semantic content.
+To allow efficient storage, implementations serializing content for blobs SHOULD use a canonical serialization.
+This increases the chance that different implementations can push the same semantic content to the store without creating redundant blobs.
+
+## JSON
+
+[JSON][] content SHOULD be serialized as [canonical JSON][canonical-json].
+Of the [OCI Image Format Specification media types](media-types.md), all the types ending in `+json` contain JSON content.
+Implementations:
+
+* [Go][]: [github.com/docker/go][], which claims to implement [canonical JSON][canonical-json] except for Unicode normalization.
+
+[canonical-json]: http://wiki.laptop.org/go/Canonical_JSON
+[github.com/docker/go]: https://github.com/docker/go/
+[Go]: https://golang.org/
+[JSON]: http://json.org/
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,8 @@ DOC_FILES := \
 	layer.md \
 	config.md \
 	manifest.md \
-	manifest-list.md
+	manifest-list.md \
+	canonicalization.md
 
 FIGURE_FILES := \
 	img/media-types.png
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ The OCI Image Format project creates and maintains the software shipping contain
 - [Image Configuration](config.md)
 - [Image Manifest](manifest.md)
 - [Image Manifest List](manifest-list.md)
+- [Canonicalization](canonicalization.md)
 
 ## Overview
 


canonicalization.md
```diff
@@ -0,0 +1,19 @@
+# Canonicalization
```
jonboulle: This feels a bit weird lacking context - could we not wrap it into the document talking about the json types?

wking:
On Sat, Sep 24, 2016 at 10:04:53AM -0700, Jonathan Boulle wrote:
> > @@ -0,0 +1,19 @@
> +# Canonicalization
>
> This feels a bit weird lacking context - could we not wrap it into
> the document talking about the json types?

Canonicalization is a way to avoid redundant blobs for a particular
semantic payload.  That idea applies to all media types, it's just
that we currently only have canonicalization advice for JSON.  Once we
get canonicalization for gzipped tarballs (or any other OCI-related
types), we can add that here too.  But as I said in the topic post
here, I'm not clear enough on what gzipped tarball canonicalization
would look like to add it yet ;).

jonboulle:
OK, that makes sense. How about moving this to a more generic "implementers guide"-type document (as a loose example of this see [appc](https://github.com/appc/spec/blob/master/GUIDE.md#implementation-guide))

wking:
On Mon, Sep 26, 2016 at 07:10:31AM -0700, Jonathan Boulle wrote:
> OK, that makes sense. How about moving this to a more generic
> "implementers guide"-type document (as a loose example of this see
> [appc](https://github.com/appc/spec/blob/master/GUIDE.md#implementation-guide))

I'm ok with that, but I prefer a normative SHOULD to informative
“guidelines, tips and suggested best practices”.  I expect the
validation tooling to (optionally) warn about violated SHOULDs but not
about violated informative hints.  Do you see other distinctions
between SHOULD and informative hints?

canonicalization.md
```diff
@@ -0,0 +1,19 @@
...
+
+One benefit of content-addressable storage is easy deduplication.
+Many images might depend on a particular [layer](layer.md), the there will only be one blob in the [store](image-layout.md).
+However, that same semantic layer serialized slightly differently would have a different hash, and if both versions of the layer are referenced there will be two blobs with the same semantic content.
+To allow efficient storage, implementations serializing content for blobs SHOULD use a canonical serialization.
+This increases the chance that different implementations can push the same semantic content to the store without creating redundant blobs.
+
+## JSON
+
+[JSON][] content SHOULD be serialized as [canonical JSON][canonical-json].
```
jonboulle: JSON content (for example, ...)

canonicalization.md
```diff
@@ -0,0 +1,19 @@
+# Canonicalization
+
+One benefit of content-addressable storage is easy deduplication.
+Many images might depend on a particular [layer](layer.md), the there will only be one blob in the [store](image-layout.md).
+However, that same semantic layer serialized slightly differently would have a different hash, and if both versions of the layer are referenced there will be two blobs with the same semantic content.
```
jonboulle: that same layer, if serialized in a different way, would have an entirely different hash, and hence if both versions of the layer, there will be two underlying blobs in the store with the same semantic comment.

wking:
On Tue, Oct 04, 2016 at 06:02:13AM -0700, Jonathan Boulle wrote:
> +One benefit of content-addressable storage is easy deduplication.
>
> Slightly more intro context - something like "One of the goals of
> the OCI Image Specification is to leverage content-addressable
> storage (CAS), which provides benefits like easy deduplication"

I don't think content-addressability is a goal in itself.  It's just a
useful pattern to support for safe, efficient distribution.  In
0b0c8e6e → 3eb6af72, I've rebased around #365 and added another
lead-in sentence pointing out that the current OCI spec does leverage
content-addressability.  Let me know if you'd like further rewording
on this.

> +## JSON
> +
> +[JSON][] content SHOULD be serialized as [canonical JSON][canonical-json].
> 
> JSON content (for example, ...)

This line is a normative SHOULD, and if the PR lands I will file
image-validation code that warns if a JSON blob is not canonical.  I'd
rather not confuse it by adding an open-ended “for example”.  Maybe
what you're looking for is a trailing line in the lead-in paragraph
like:

  The following sections contain canonicalization recommendations for
  OCI-related media types.

> +However, that same semantic layer serialized slightly differently would have a different hash, and if both versions of the layer are referenced there will be two blobs with the same semantic content.
> 
> that same layer, if serialized in a different way, would have an
> entirely different hash, and hence if both versions of the layer,
> there will be two underlying blobs in the store with the same
> semantic comment.

“hence if both versions of the layer” seems to be missing a word or
two.  I've dropped my “However”, shifted the difference condition to
the front, and dropped “slightly” in 3eb6af72.

jonboulle:
@wking if someone does an inline review, please reply inline, otherwise it's impossible to follow

canonicalization.md
```diff
@@ -0,0 +1,20 @@
+# Canonicalization
+
+OCI Images [are](descriptor.md) [content-addressable](image-layout.md).
+One benefit of content-addressable storage is easy deduplication.
```
jonboulle: It's already explicitly called out as a goal @ https://github.com/opencontainers/image-spec/blame/e74c6c703315a8d2baaea691b527194377c33a8d/manifest.md#L4

canonicalization.md
jonboulle: JSON content (for example all of the bla bla media types or bla bla manifests) SHOULD be serialized as

media-types.md
```diff
@@ -47,3 +47,5 @@ The following figure shows how the above media types reference each other:
 
 [Descriptors](descriptor.md) are used for all references.
 The manifest list being a "fat manifest" references one or more image manifests per target platform. An image manifest references exactly one target configuration and possibly many layers.
+
+[JSON]: http://json.org/
```
jonboulle: What's this?

jonboulle:
Looks okay overall after a few more fixes

wking:
On Wed, Oct 05, 2016 at 05:09:38AM -0700, Jonathan Boulle wrote:
> @wking if someone does an inline review, please reply inline,
> otherwise it's impossible to follow

I'll try, but GitHub no longer has a way to do that via email.  When I
do reply by email, I'll be sure to quote enough context so my reply
makes sense.

canonicalization.md
wking: Ah, I [stand corrected][1].  I still think the current wording covers that though.  Would you still prefer [your initial suggestion][2]?

[1]: https://github.com/opencontainers/image-spec/pull/339#issuecomment-251412226
[2]: https://github.com/opencontainers/image-spec/pull/339#discussion_r81730202

wking:
I've pushed 3eb6af7 → 4729086 which:

* Fixed “the there” → “but there” [1]
* Added a new sentence linking to media-types.md for our JSON
  types [2].  This feels pretty obvious to me, and the SHOULD
  canonicalize also applies to all JSON content regardless of
  whether it's an OCI type, so I'm fine dropping the new
  sentence.  Using a new sentence (instead of @jonboulle's
  floated parenthetical) avoids complicating the normative
  SHOULD.

[1]: https://github.com/opencontainers/image-spec/pull/339#discussion_r81958184
[2]: https://github.com/opencontainers/image-spec/pull/339#discussion_r81958004

media-types.md
jonboulle: OK, fine, I asked the wrong question :/. Let me try again: why is it here? (I don't see any JSON references in this document)

canonicalization.md
```diff
@@ -0,0 +1,22 @@
...
+This increases the chance that different implementations can push the same semantic content to the store without creating redundant blobs.
+
+## JSON
+
+[JSON][] content SHOULD be serialized as [canonical JSON][canonical-json].
+Implementations:
+
+* [Go][]: [github.com/docker/go][], which claims to implement [canonical JSON][canonical-json] except for Unicode normalization.
+
+Of the [OCI Image Format Specification media types](media-types.md), all the types ending in `+json` contain JSON content.
```
jonboulle: Move this above the "Implementations" imo.

media-types.md
wking: [First line in the JSON section][1] ;).

[1]: https://github.com/opencontainers/image-spec/pull/339/files#diff-a40e69a354ae90a6066b6786869dc9dbR12

jonboulle: That is a separate file.

wking:
I've pushed 4729086 → b695fb3 shifting the media-types JSON link up
before the list of implementations [1].

[1]: https://github.com/opencontainers/image-spec/pull/339#pullrequestreview-2956854

media-types.md
wking: Heh, so it is.  Thanks for banging my head into that until I understood :p.

wking:
I've pushed b695fb3 → 1685bae dropping the JSON link label from
media-types.md [1].  No idea what I was thinking with that, but thanks
to @jonboulle for pointing it out to me (repeatedly) until I caught
on.

[1]: https://github.com/opencontainers/image-spec/pull/339#discussion_r81958280

jonboulle:
LGTM for now

vbatts:
while i'm not in disagreement here, i fell like this is describing _something_ without describing _anything_.
Like is canonical serialization, the use of symlinks to deduplicate blobs? or something else?

wking:
On Thu, Oct 06, 2016 at 09:44:02AM -0700, Vincent Batts wrote:
> while i'm not in disagreement here, i fell like this is describing
> _something_ without describing _anything_.  Like is canonical
> serialization, the use of symlinks to deduplicate blobs? or
> something else?

How would symlinks deduplicate blobs?  Maybe if you had a cluster of
image-layouts?  If so, that's a CAS-engine optimization (reducing the
disk space used to store a set of blobs by leaning on other
image-layouts).  This PR is about optimizing the *CAS content*
(reducing the number of blobs you need to store).  I expect both are
useful, but advice about CAS-engine optimization should probably be
part of a separate issue/PR.

vbatts:
LGTM

stevvooe:
I know we've already merged this, but how can Canonical JSON be considered a viable specification when the only reference is to a wiki page that could be changed at any time? I brought this up at the time when we started using canonical json in docker. [Docker Distribution](https://github.com/docker/distribution/blob/master/docs/spec/json.md) attempts to take this further, but canonical json is insufficient for this purpose.

I am really confused as to why we consider Canonical JSON a viable specification.

wking:
On Wed, Oct 12, 2016 at 02:21:25PM -0700, Stephen Day wrote:
> I know we've already merged this, but how can Canonical JSON be
> considered a viable specification when the only reference is to a
> wiki page that could be changed at any time?

I agree that a more formal spec would be nice [1].  If you have one,
I'm happy to link to it instead.  But this PR landed a SHOULD.  In the
event that the wiki mutates under us it isn't a disaster, we just
start warning about any incompatibilities.

And the last change to the spec itself was in 2008 [2], so it's fairly
stable.

> I brought this up at the time when we started using canonical json
> in docker. [Docker
> Distribution](https://github.com/docker/distribution/blob/master/docs/spec/json.md)
> attempts to take this further, but canonical json is insufficient
> for this purpose.

The goal is to reduce the number of blobs carrying the same semantic
content.  You don't have to have a complete canonicalization to get
some reduction.  In practice, I think the canonical JSON spec will be
sufficient to avoid most duplication.  And [3] would be fine too (is
that spec versioned independently of docker/distribution as a whole?).
But either of those are better at reducing duplication than not
providing any canonicalization advice at all.

[1]: https://github.com/opencontainers/image-spec/issues/259#issuecomment-245185870
[2]: http://wiki.laptop.org/index.php?title=Canonical_JSON&action=history
[3]: https://github.com/docker/distribution/blob/master/docs/spec/json.md#rules

PR #340, opened 2016-09-23
manifest: DRY annotations and extensibility

wking:
[As](https://github.com/opencontainers/image-spec/pull/164#issuecomment-243952187) [I](https://github.com/opencontainers/image-spec/pull/164#issuecomment-244279771) [suggested](https://github.com/opencontainers/image-spec/pull/164#r77871859) in #164.  I've shifted the extensibility section to stand alone at the end of the file, and pushed the annotations meat into the lower-level manifest specification (linking there from the higher-level manifest-list specification).

The extensibility requirements and annotations field might arguably apply to our other JSON types as well, but I've left them off this commit to focus on making the current requirements more DRY without changing the specified behavior.

cc @jonboulle.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ DOC_FILES := \
 	config.md \
 	manifest.md \
 	manifest-list.md \
-	canonicalization.md
+	considerations.md
 
 FIGURE_FILES := \
 	img/media-types.png
--- a/README.md
+++ b/README.md
@@ -14,7 +14,9 @@ The OCI Image Format project creates and maintains the software shipping contain
 - [Image Configuration](config.md)
 - [Image Manifest](manifest.md)
 - [Image Manifest List](manifest-list.md)
-- [Canonicalization](canonicalization.md)
+- [Considerations](considerations.md)
+    - [Extensibility](considerations.md#extensibility)
+    - [Canonicalization](considerations.md#canonicalization)
 
 ## Overview
 
--- a/considerations.md
+++ b/considerations.md
@@ -1,3 +1,8 @@
+# Extensibility
+
+Implementations that are reading/processing [manifests](manifest.md) or [manifest lists](manifest-list.md) MUST NOT generate an error if they encounter an unknown property.
+Instead they MUST ignore unknown properties.
+
 # Canonicalization
 
 OCI Images [are](descriptor.md) [content-addressable](image-layout.md).
--- a/manifest-list.md
+++ b/manifest-list.md
@@ -66,20 +66,7 @@ This section defines the `application/vnd.oci.image.manifest.list.v1+json` [medi
 
 - **`annotations`** *string-string map*
 
-    This OPTIONAL property contains arbitrary metadata for the manifest list.
-    Annotations MUST be a key-value map where both the key and value MUST be strings.
-    While the value MUST be present, it MAY be an empty string.
-    Keys MUST be unique within this map, and best practice is to namespace the keys.
-    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
-    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications.
-    If there are no annotations then this property MUST either be absent or be an empty map.
-    Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
-
-    See [Pre-Defined Annotation Keys](manifest.md#pre-defined-annotation-keys).
-
-### Extensibility
-Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown property.
-Instead they MUST ignore unknown properties.
+    This OPTIONAL property contains arbitrary metadata for the manifest list with the same semantics as the [manifest's `annotations`](manifest.md#image-manifest-property-descriptions)
 
 ## Example Manifest List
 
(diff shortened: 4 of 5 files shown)


jonboulle:
LGTM, thanks

vbatts:
please rebase

wking:
Rebased around #365 with ff9ead7 → d07ef3f, pulling the extensibility
section out into its own (small) file.

wking:
Fixed `extensibibility.my` → `extensibility.md` typo with d07ef3f → e311e07.

jonboulle:
Ehh, I preferred it where it was - now worried about death by a thousand files (extensibility, canonicalization, etc etc). Can we combine them (into a generic "considerations" document or similar), or just put this back?

wking:
On Mon, Oct 10, 2016 at 09:22:46AM -0700, Jonathan Boulle wrote:
> Ehh, I preferred it where it was - now worried about death by a
> thousand files (extensibility, canonicalization, etc etc). Can we
> combine them (into a generic "considerations" document or similar)…

I doubt we'll have a thousand of these, and I prefer filenames that
are more clear than “considerations”.  But I don't care that much, so
I can combine canonicalization and extensibility into a single
considerations.md later today.

wking:
Combined `extensibility.md` and `canonicalization.md` into a single `considerations.md` with e311e07 → e2a0d6e, although [I liked it better as separate files][1].

[1]: https://github.com/opencontainers/image-spec/pull/340#issuecomment-252680364

jonboulle:
You could probably just come up with a better word than "considerations"
and make us both happy :-)

wking:
On Tue, Oct 11, 2016 at 08:05:48AM -0700, Jonathan Boulle wrote:
> You could probably just come up with a better word than
> "considerations" and make us both happy :-)

These extensibility and canonicalization seem like completely
orthogonal ideas to me.  The only thing tying them together is the
spec behind both is currently short ;).

stevvooe:
"A little duplication is better than the wrong abstraction" 

 - Not sure who to attribute this to

wking:
On Tue, Oct 18, 2016 at 04:13:09PM -0700, Stephen Day wrote:
> "A little duplication is better than the wrong abstraction" 

How is the abstraction I'm proposing wrong?  Maybe we can fix it
without duplication :p.

stevvooe:
@wking You're introducing another document that the reader has to switch to to get the proper context of understanding. It both distract and confuses.

wking:
On Tue, Oct 18, 2016 at 06:29:23PM -0700, Stephen Day wrote:
> @wking You're introducing another document that the reader has to
> switch to to get the proper context of understanding. It both
> distract and confuses.

We already do a lot of linking to handle more basic details of the
spec.  For example, #167 removed local descriptor duplication in favor
of links to descriptor.md.  Many mediaType docs link to
media-types.md#compatibility-matrix.  The platform docs link to Go's
docs for $GOOS and $GOARCH.  I don't see much difference between those
and having the extensibility docs off in their own location.

Would you prefer this PR if I left the “Extensibility” sections alone
for now and limited the change to the manifest-list → manifest
annotations link?  We can revist consolidating the extensibility
sections if/when we add similar extensibility rules to
application/vnd.oci.image.config.v1+json.

stevvooe:
@wking So, `descriptor.md` collected a common idea. It is reflected by a first-class object, even if it is embedded. This PR introduces some arbitrary "considerations" or "extensibility", which is some what meaningless.

It might be better to break out the annotations into its own document, as it describes a common sub-structure, similar to descriptor.

wking:
On Wed, Oct 19, 2016 at 11:19:45AM -0700, Stephen Day wrote:
> It might be better to break out the annotations into its own
> document, as it describes a common sub-structure, similar to
> descriptor.

That makes a nice home for the annotations wording, but I think the
bit of logic without a good home is the extensibility wording.  The
extensibility condition about arbitrary JSON properties beyond those
defined in the spec.  I'd initially put it in an extensibility.md but
moved it into considerations.md to satisfy @jonboulle [1].  Would you
put the extensibility wording in an annotations.md?  It may be a
closer match with annotations (which is already an
extensibility-related location) than it is for canonicalization.  But
it's still distinct from the annotations property itself.

[1]: https://github.com/opencontainers/image-spec/pull/340#issuecomment-252670278

stevvooe:
@wking Frankly, I'd rather have a little repetition here than have to bounce into the middle of other documents without context. Context matters. Though we may store it in our brain in such a manner, most people don't consume information as a graph of references.

wking:
On Wed, Oct 19, 2016 at 11:35:13AM -0700, Stephen Day wrote:
> Frankly, I'd rather have a little repetition here than have to
> bounce into the middle of other documents without context. Context
> matters.

I agree that context is important.  What sort of context are you
looking for for the extensibility constraint?  We have slots on both
sides (e.g. in manifest.md where it links to the extensibility docs
and in the extensibility docs where it discusses the JSON media types
it covers).  I'm happy to provide lots of context at both locations,
but I don't think duplicating the requirement itself for every JSON
media type that it applies to is the scaleable approach.  The
duplicate approach is just redundant reading for anyone reading both
locations and redundant writing (hopefully ;) for anyone writing to
either location.

stevvooe:
@wking Please rebase this or let's close.

wking:
On Wed, Jan 18, 2017 at 04:07:18PM -0800, Stephen Day wrote:
> @wking Please rebase this or let's close.

#501 is also DRYing up the annotations definition and seems to have
more momentum.  Can we leave this PR open until #501 lands, and I'll
rebase or close depending on whether there's still anything here that
isn't addressed by #501?

stevvooe:
@wking Please resubmit after #501 lands.

PR #341, opened 2016-09-23
validate media type of manifest descendants

xiekeyang:
This PR is alternate of https://github.com/opencontainers/image-tools/pull/10.
As we have some suggestion that descendants media type should be checked by JSON schema, I put up this PR.
JSON schema constrains config and layer fields type in manifest file. If this is acceptable, we should then update the vendor to this in #image-tools.

Signed-off-by: xiekeyang xiekeyang@huawei.com

### DIFF
--- a/schema/validator.go
+++ b/schema/validator.go
@@ -16,10 +16,12 @@ package schema
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 
+	"github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"github.com/xeipuuv/gojsonschema"
 )
@@ -27,6 +29,12 @@ import (
 // Validator wraps a media type string identifier
 // and implements validation against a JSON schema.
 type Validator string
+
+type validateDescendantsFunc func(r io.Reader) error
+
+var mapValidateDescendants = map[Validator]validateDescendantsFunc{
+	MediaTypeManifest: validateManifestDescendants,
+}
 
 // ValidationError contains all the errors that happened during validation.
 type ValidationError struct {
@@ -42,6 +50,16 @@ func (v Validator) Validate(src io.Reader) error {
 	buf, err := ioutil.ReadAll(src)
 	if err != nil {
 		return errors.Wrap(err, "unable to read the document file")
+	}
+
+	if f, ok := mapValidateDescendants[v]; ok {
+		if f == nil {
+			return fmt.Errorf("internal error: mapValidateDescendents[%q] is nil", v)
+		}
+		err = f(bytes.NewReader(buf))
+		if err != nil {
+			return err
+		}
 	}
 
 	sl := gojsonschema.NewReferenceLoaderFileSystem("file:///"+specs[v], fs)
@@ -72,4 +90,30 @@ type unimplemented string
 
 func (v unimplemented) Validate(src io.Reader) error {
 	return fmt.Errorf("%s: unimplemented", v)
+}
+
+func validateManifestDescendants(r io.Reader) error {
+	header := v1.Manifest{}
+
+	buf, err := ioutil.ReadAll(r)
+	if err != nil {
+		return errors.Wrapf(err, "error reading the io stream")
+	}
+
+	err = json.Unmarshal(buf, &header)
+	if err != nil {
+		return errors.Wrap(err, "manifest format mismatch")
+	}
+
+	if header.Config.MediaType != string(v1.MediaTypeImageConfig) {
+		fmt.Printf("warning: config %s has an unknown media type: %s\n", header.Config.Digest, header.Config.MediaType)
+	}
+
+	for _, layer := range header.Layers {
+		if layer.MediaType != string(v1.MediaTypeImageLayer) &&
+			layer.MediaType != string(v1.MediaTypeImageLayerNonDistributable) {
+			fmt.Printf("warning: layer %s has an unknown media type: %s\n", layer.Digest, layer.MediaType)
+		}
+	}
+	return nil
 }


wking:
On Fri, Sep 23, 2016 at 01:09:39AM -0700, xiekeyang wrote:
>     A schema/config-descriptor.json (24)

Descriptors can stand on their own (e.g. as ref payloads), so this
should be descriptor-schema.json.

schema/config-descriptor.json
```diff
@@ -0,0 +1,24 @@
+{
+  "description": "OpenContainer Content Descriptor Specification",
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "id": "https://opencontainers.org/schema/descriptor",
+  "type": "object",
+  "properties": {
+    "mediaType": {
+      "type": "string",
+      "pattern": "^application/vnd[.]oci[.]image[.]config[.]v1[+]json$"
```
wking: Ah, descriptors [can point at anything in general][1], but *this* is a descriptor that points at a config file.  That makes sense to me.

[1]: https://github.com/opencontainers/image-spec/pull/341#issuecomment-249217476

wking:
On Fri, Sep 23, 2016 at 01:09:39AM -0700, xiekeyang wrote:
> JSON schema constrains config and layer fields type in manifest
> file. If this is acceptable, we should then update the vendor to
> this in #image-tools.

I don't think we can handle this in JSON Schema.  With 7cb3a3063 this
PR successfully REQUIRES the media types recommended in #304 (still in
flight) for manifest descriptors.  But #304 isn't suggesting that
requirement for manifest instances (which is what JSON Schema tests),
it's MUSTing *implementations* to support those types, and SHOULDing
manifests to use them.  I think we want warnings, not errors, when a
config breaks that SHOULD [1], and I don't know of a way to
distinguish between warnings and errors in JSON Schema.

[1]; https://github.com/opencontainers/image-tools/pull/10#issuecomment-248973124

xiekeyang:
PTAL

wking:
On Fri, Sep 23, 2016 at 08:16:00AM -0700, W. Trevor King wrote:
> I don't think we can handle this in JSON Schema.  With 7cb3a3063
> this PR successfully REQUIRES the media types recommended in #304
> (still in flight) for manifest descriptors.  But #304 isn't
> suggesting that requirement for manifest instances (which is what
> JSON Schema tests), it's MUSTing *implementations* to support those
> types, and SHOULDing manifests to use them.  I think we want
> warnings, not errors, when a config breaks that SHOULD [1], and I
> don't know of a way to distinguish between warnings and errors in
> JSON Schema.

#304 landed a few days ago [1], so the approach you're taking here
(strictly restricting images to those types) is officially too strong.
I think we need to go back to Go and use an option (--strict?
--portable?) to decide whether images that use features beyond the
minimum unpacker requirements are warnings or errors.  I've mocked up
something along those lines (in the context of tar features) in [2]
(linked from #342).

[1]: https://github.com/opencontainers/image-spec/pull/304#event-801307607
[2]: https://github.com/wking/image-tools/blob/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6/validate/layer.go#L60-L65

xiekeyang:
@wking 

> I think we need to go back to Go and use an option (--strict?
--portable?) t

I suppose that you are meaning to modify `schema.MediaTypeManifest.Validate()`, internally, as @stevvooe suggestion https://github.com/opencontainers/image-tools/pull/10#issuecomment-249670609. However, it seem to be a little inelegant to add a code section in it for manifest descendants case branch, like:

```go
func (v Validator) Validate(src io.Reader) error { 

  ...
  if v==MediaTypeManifest{  //or switch-case here
    err:=validateDescendants(src)
	if err!=nil{
	   return err
	}
  }
  ...

}
```

If I am right?

cc @stevvooe

wking:
On Thu, Sep 29, 2016 at 11:25:37PM -0700, xiekeyang wrote:
> @wking 
> 
> > I think we need to go back to Go and use an option (--strict?
> --portable?) t
>
> I suppose that you are meaning to modify
> `schema.MediaTypeManifest.Validate()`, internally, as @stevvooe
> suggestion
> https://github.com/opencontainers/image-tools/pull/10#issuecomment-249670609.

I'm not convinced that the maintainer consensus is to keep more
validation than JSON Schema in this repo (some discussion in #337).
But yeah, the Go I'm suggesting should live wherever the maintainers
want the beyond-the-JSON-Schema Go validation code to live.

> However, it seem to be a little inelegant to add a code section in
> it for manifest descendants case branch…

I've mocked up a callback-based validation approach in that commit.
See [1], where I setup a map of validators for each supported media
type (very stubby at the moment, because my commit is currently based
on opencontainers/image-tools#5 and I don't want to keep rebasing this
callback validation if the ref/CAS interfaces float for another 3
months).  Each media-type validator will check its blob (possibly
using JSON Schema as part of that check) and then recurse into another
Validate() call for any children.  Of course, this is a problem for
validating spec examples (where there is no CAS engine from which to
retrieve the children), so I think the spec should stick to
JSON-Schema-only validation and not worry about any cases that doesn't
cover.

[1]: https://github.com/wking/image-tools/blob/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6/validate/validate.go#L30-L44

xiekeyang:
@wking 
I fix this PR. As previous suggestion, I add OCI and Compatibility Matrix media types to media type constraints of descendants. And these constraints are implemented by Go.
And fix unit test according to latest media type spec.
PTAL.

schema/manifest_backwards_compatibility_test.go
```diff
@@ -137,12 +137,12 @@ func TestBackwardsCompatibilityManifest(t *testing.T) {
...
 		// https://registry-1.docker.io/v2/library/docker/manifests/sha256:888206c77cd2811ec47e752ba291e5b7734e3ef137dfd222daadaca39a9f17bc
 		{
-			digest: "sha256:888206c77cd2811ec47e752ba291e5b7734e3ef137dfd222daadaca39a9f17bc",
+			digest: "sha256:d901d5244c94db8151415bf3a1d3a0b10e7468e4997786b01c287cbcbaeca446",
 			manifest: `{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
    "config": {
-      "mediaType": "application/octet-stream",
+      "mediaType": "application/vnd.docker.container.image.v1+json",
```
wking: Can we do this?  The octet-stream type is from #145, where @philips made it sound like this manifest had been caught in the wild.  If that's the case, we shouldn't be messing with the source.

schema/manifest_backwards_compatibility_test.go
```diff
@@ -137,12 +137,12 @@ func TestBackwardsCompatibilityManifest(t *testing.T) {
 		// "Accept: application/vnd.docker.distribution.manifest.v2+json" \
 		// https://registry-1.docker.io/v2/library/docker/manifests/sha256:888206c77cd2811ec47e752ba291e5b7734e3ef137dfd222daadaca39a9f17bc
 		{
-			digest: "sha256:888206c77cd2811ec47e752ba291e5b7734e3ef137dfd222daadaca39a9f17bc",
+			digest: "sha256:d901d5244c94db8151415bf3a1d3a0b10e7468e4997786b01c287cbcbaeca446",
```
wking: I don't think we want to bump this, but if we do we should probably bump the comment above or explain why the hashes are now different.

schema/schema.go
```diff
@@ -27,6 +27,21 @@ const (
...
+	MediaTypeDockerImageManifest = "application/vnd.docker.distribution.manifest.v2+json"
+
+	// MediaTypeDockerImageLayer specifies the docker image's mediaType used for layers referenced by the manifest.
+	MediaTypeDockerImageLayer = "application/vnd.docker.image.rootfs.diff.tar.gzip"
+
+	// MediaTypeDockerImageConfig specifies the docker image's mediaType for the image configuration.
+	MediaTypeDockerImageConfig = "application/vnd.docker.container.image.v1+json"
+
+	// MediaTypeNonDistributable specifies the non-distributable layer's mediaType.
+	MediaTypeNonDistributable = "application/vnd.oci.image.layer.nondistributable.v1.tar+gzip"
```
wking: I still don't think we want to export these as part of our public API (discussion in #320).

schema/validator.go
```diff
@@ -44,6 +50,13 @@ func (v Validator) Validate(src io.Reader) error {
 		return errors.Wrap(err, "unable to read the document file")
 	}
 
+	if f := mapValidateDescendants[v]; f != nil {
```
wking: I think the idiomatic way to handle this is [with][1]:

    f, ok := mapValidateDescendants[v]
    if ok {
      …validate descendant…
    }

[1]: https://github.com/wking/image-tools/blob/callback-validation/validate/validate.go#L37-L44

schema/validator.go
```diff
@@ -73,3 +86,35 @@ type unimplemented string
...
+			layer.MediaType != MediaTypeNonDistributable {
+			return fmt.Errorf("illegal layer mediaType: %s", layer.MediaType)
+		}
+	}
+	return nil
+}
+
+func init() {
+	mapValidateDescendants[MediaTypeManifest] = validateManifestDescendants
+}
```
wking: No need for `init()`, you can [declare a package-level constant][1].

[1]: https://github.com/wking/image-tools/blob/callback-validation/validate/validate.go#L30-L35

schema/validator.go
```diff
@@ -73,3 +86,35 @@ type unimplemented string
...
+	if header.Config.MediaType != string(v1.MediaTypeImageConfig) &&
+		header.Config.MediaType != MediaTypeDockerImageConfig {
+		return fmt.Errorf("illegal config mediaType: %s", header.Config.MediaType)
+	}
+
+	for _, layer := range header.Layers {
+		if layer.MediaType != string(v1.MediaTypeImageLayer) &&
+			layer.MediaType != MediaTypeDockerImageLayer &&
+			layer.MediaType != MediaTypeNonDistributable {
+			return fmt.Errorf("illegal layer mediaType: %s", layer.MediaType)
```
wking: This is still too strong.  You want to [warn][1] if the type is not an OCI type, but it's not an error.

[1]: https://github.com/opencontainers/image-spec/pull/341#issuecomment-249220321

schema/validator.go
xiekeyang: @wking ,
In the case of setting `nil` to some key, the checking by `ok` will have risk.
`f` is `nil` when `ok` is `true`, the next calling `f` will lead to panic.
So I think it is safe to check `f` is not `nil`.

wking: I'm fine panicking if somebody inserts a `nil` value in `mapValidateDescendants`.  I would be happier if Go's typing was strict enough to error with `cannot convert nil to type validateDescendantsFunc` when they attempted the insertion, but I don't see a way to do that if the [zero value for function types is `nil`][1].  To be extra safe and clear, I'd recommend handling the three cases explicitly:

    f, ok := mapValidateDescendents[v]
    if ok {
        if f == nil {
            …internal error: shame on you for putting a nil value in mapValidateDescendents…
        }
        …validate descendants…
    }
[1]: https://golang.org/ref/spec#Function_types

schema/validator.go
wking: On Wed, Oct 05, 2016 at 07:10:01PM -0700, xiekeyang wrote:
> 1. Do you think we should emit warn for all string not equal to OCI type, even for buzz words? It seems too loose.

We're warning them, and they would be wise to take the warning seriously.  I don't think we can do anything stronger by default; how do we know that the strange type they're using *isn't* handled by their tooling?  But I think image-tools should have a `--strict` flag (for “error on any features not required in compliant unpackers”) and/or an [`--unpackable` flag][1] (for “error on any features *I* don't recognize”).

[1]: https://github.com/opencontainers/image-spec/issues/286#issuecomment-245856987

> 2. Current logger have no `warn` level. Do you agree that we import `logrus` here, to print different  levels log?

I think [logging is orthogonal][1].  These tests are the primary output of validation, so I think a validator should be printing the validation results (including warnings like this) to stdout (or some sort of writer).  In my callback-based mock-up, I'm [using `tap.go`][2], and in that case warnings would [look like][3]:

    ok 7 # skip sha256-a… layers[2] has an unknown media type: image/png

[1]: https://github.com/opencontainers/image-spec/pull/234#issuecomment-245949723
[2]: https://github.com/wking/image-tools/commit/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6#diff-201d416d61af62a246e5b552b6d42c92R31
[3]: https://github.com/opencontainers/image-spec/issues/286#issuecomment-245876396

xiekeyang:
@wking 
Fixed it according to your suggestion. PTAL.

schema/validator.go
```diff
@@ -44,6 +52,13 @@ func (v Validator) Validate(src io.Reader) error {
 		return errors.Wrap(err, "unable to read the document file")
 	}
 
+	if f, ok := mapValidateDescendants[v]; ok && f != nil {
```
wking: There's no reason to check both `ok` and `nil` if you're silently ignoring the `nil` case.  I think you should either [skip the `nil` check and pacnic if someone stuck a `nil` value in `mapValidateDescendants`][1] or [explicitly check for and error out if someone stuck a `nil` value in `mapValidateDescendants`][2]

[1]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82013412
[2]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82113598

schema/validator.go
```diff
@@ -73,3 +88,28 @@ type unimplemented string
...
+		return errors.Wrapf(err, "error reading the io stream")
+	}
+
+	err = json.Unmarshal(buf, &header)
+	if err != nil {
+		return errors.Wrap(err, "manifest format mismatch")
+	}
+
+	if header.Config.MediaType != string(v1.MediaTypeImageConfig) {
+		fmt.Printf("warning: config %s has a non OCI media type: %s\n", header.Config.Digest, header.Config.MediaType)
```
wking: The problem isn't that it has a non-OCI type (`application/vnd.oci.image.manifest.list.v1+json` is an OCI type, and it deserves this warning).  My [suggested wording is][1]:

> {blobDigest} {property} has an unknown media type: {type}

That's a bit complicated with your APIs, because you take a reader and not a descriptor or digest.  In my mock-up, the validator takes [a digest and CAS engine][2], so it's easy to make messages like that.  But it makes sense to want a way to validate a reader without having to inject it into a CAS engine, and for that initial reader-based validation I think the message should just be:

> {blobDigest} {property} has an unknown media type: {type}

For this line here, that's:

    fmt.Println("warning: config has an unknown media type: %s", header.Config.MediaType)

[1]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82114804
[2]: https://github.com/wking/image-tools/commit/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6#diff-8351286f57eb81e245c9c99c07f3b34fR27

schema/validator.go
```diff
@@ -73,3 +88,28 @@ type unimplemented string
...
+	}
+
+	for _, layer := range header.Layers {
+		if layer.MediaType != string(v1.MediaTypeImageLayer) {
```
wking: These can [also be `MediaTypeImageLayerNonDistributable`][1].

[1]: https://github.com/opencontainers/image-spec/pull/304/files#diff-bf8d9e5146f89ef0a8b9bf8cd1026901R173

schema/validator.go
```diff
@@ -28,6 +30,12 @@ import (
 // and implements validation against a JSON schema.
 type Validator string
 
+type validateDescendantsFunc func(r io.Reader) error
+
+var mapValidateDescendants = map[Validator]validateDescendantsFunc{
+	MediaTypeManifest: validateManifestDescendants,
```
wking: This is currently just looking at the types of the descriptors.  It doesn't follow those references to actually validate the descendants (e.g. “the manifest claims sha256:a… is a `application/vnd.oci.image.config.v1+json`.  Does it validate as such?).  I think we eventually want this callback-based validation to handle that, but I don't see an easy way to do it without a CAS-engine API (once you have that it is [pretty easy][1]), so I'm fine punting on that for this PR and only doing these local checks.

[1]: https://github.com/wking/image-tools/commit/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6#diff-8351286f57eb81e245c9c99c07f3b34fR27

At beginning I work this in image-tool project, but seems improper. 
In `schema` package, `validateManifestDescendants` acting as the `validator.Validate` child is what I think the nice way. 
Callback handler must impact on up-layer calling in image-tool project, to which I worry maintainers not fully agree.

wking: I agree that you wouldn't want all of recursive validation, all recursive validation code in image-spec, and CAS-engine code in image-tools.  My personal preference there is to [do non-recursive JSON Schema validation here and to handle recursive validation and the CAS-engine in image-tools][1].  @stevvooe [seems to][2] [prefer][3] keeping more than just JSON Schema validation in image-spec, but @philips has [moved a fair bit of stuff to image-tools][4], so I'm not sure there's a clear maintainer position on this yet.  Once there is a clear position, it should be fairly straightforward to put this logic in the approved location.

[1]: https://github.com/opencontainers/image-spec/pull/337#issuecomment-248816631
[2]: https://github.com/opencontainers/image-spec/pull/337#issuecomment-249673277
[3]: https://github.com/opencontainers/image-tools/pull/10#issuecomment-249670609
[4]: https://github.com/opencontainers/image-spec/pull/337#issuecomment-249670746

schema/validator.go
xiekeyang: fix it.

schema/validator.go
xiekeyang: fix it.

schema/validator.go
```diff
@@ -44,6 +52,16 @@ func (v Validator) Validate(src io.Reader) error {
 		return errors.Wrap(err, "unable to read the document file")
 	}
 
+	if f, ok := mapValidateDescendants[v]; ok {
+		if f == nil {
+			return fmt.Errorf("internal error: shame on you for putting a nil value in mapValidateDescendents")
```
wking: This is taking me too literally ;).  How about:

    return fmt.Errorf("internal error: mapValidateDescendents[%q] is nil", v)

wking:
I have a few minor quibbles with the error messages, but I think
3df641c is close enough to be worth landing (assuming that maintainers
want more than JSON Schema in image-spec [1]).  We can talk about
polishing the error messages [2], TAP [3], recursive validation [4],
etc. in follow-up PRs.

[1]: https://github.com/opencontainers/image-spec/pull/341/files#r82124290
[2]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82125211
[3]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82114804
[4]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82118716

xiekeyang:
Hopefully you can put up some proposal to move this PR forward.

vbatts:
So, the objective here is to descend into a manifest to validate the other objects, right?
What is expected? When a _known_ media-type is specified and an object of the  matching checksum is found, that the object is of the expected media-type. Right?
What is _not_ expected, is if there is an object of *unknown* media-type found, that the test OUGHT to not fail, because it just does not know how to validate unknown objects.

Are we in agreement up to this point?

wking:
On Thu, Oct 06, 2016 at 07:09:05AM -0700, Vincent Batts wrote:
> So, the objective here is to descend into a manifest to validate the
> other objects, right?

This PR doesn't descend into children, because you can't do that
without a CAS engine to fetch them by digest [1].  We'll want that
eventually (but probably not in this repo [2]).  As it stands with
dd4cbf13, this PR just checks #304 on the blob itself (checking the
media types the blob's descriptors claim for the children, but not
actually looking at the child blobs).

> What is expected? When a _known_ media-type is specified and an
> object of the matching checksum is found, that the object is of the
> expected media-type. Right?

That would be nice, but currently this PR just checks to see if the
referencing descriptor has one of the media types that #304 requires
unpackers to support.

> What is _not_ expected, is if there is an object of *unknown*
> media-type found, that the test OUGHT to not fail, because it just
> does not know how to validate unknown objects.

We're not validating the children yet, known media type or not.  But
dd4cbf13 (appropriately, I think) warns on media types that unpackers
are not required to support.  And a future PR could add an
--unpackable or --strict flag to turn those warnings into errors.

[1]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82118716
[2]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82124290
[3]: https://github.com/opencontainers/image-spec/pull/341#discussion_r82114804

xiekeyang:
@vbatts 
As @wking explained, this PR is to fix the insufficient point that children in manifest was not validated. 
It has been token so many challenge as return error or just print error, place it in image tool or image spec, take into Compatibility Matrix or not ...

I've tried to understand maintainers proposal, and improve this PR. Now hopefully it can be accepted.

xiekeyang:
@stevvooe If this is the implementation way what you want?

stevvooe:
@xiekeyang This seems sufficient. It is just making sure the media types are valid. It is too bad that json schema can't do this.

LGTM

vbatts:
LGTM

schema/validator.go
stevvooe: This code is simply validating mediatypes. It doesn't actually validate the descendent objects. This style of validation clearly belongs within the image-spec repo.

The maintainer position is clear and we have discussed on the call multiple times, without dissent.

Please stop fomenting disagreement where there is none.

wking: > Please stop fomenting disagreement where there is none.

The maintainer position has been clear on this [since the 2016-10-12 meeting][1].  I don't think it was particular clear when I made these comments on 2016-10-05.

[1]: http://ircbot.wl.linuxfoundation.org/meetings/opencontainers/2016/opencontainers.2016-10-12-21.01.log.html#l-72

PR #342, opened 2016-09-23
layer: Require ustar (originally from IEEE Std 1003.1-1988 but linking IEEE Std 1003.1-2013)

wking:
The idea with a spec like this is to define behavior so that different implementations can interoperate reliably.  When there is an interop problem betweem implementations A and B, it should be clear from the spec whether implementation A is broken, implementation B is broken, or the spec is insufficiently clear.

For example, pax [defines `g` and `x` typeflags](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_02) that [aren't part of the older ustar (IEEE Std 1003.1-1988)](https://github.com/libarchive/libarchive/wiki/ManPageTar5#POSIX_ustar_Archives).  Before this commit, if implementation A produced a layer with a `g` or `x` typeflag and implementation B died unpacking it, it was unclear whose fault it was.  With this commit, it is clearly B's fault (because it does not understand the pax typeflag).

If implementation A had produces a layer with an `S` typeflag (which [GNU uses for sparse files](https://github.com/libarchive/libarchive/wiki/ManPageTar5#gnu-tar-archives)) and implementation B died unpacking it, it is neither party's fault, because pax explicitly makes those values implementation-defined.  Interop around them is up to out of band communication between the layer author and layer consumer, and is not covered by this spec.

The previous “File Types” section listed sockets, but the pax spec has:

> Attempts to archive a socket using ustar interchange format shall produce a diagnostic message.

And I see no socket entry in [Go's set of typeflag constants](https://golang.org/pkg/archive/tar/#pkg-constants), so I'm not sure how they were supported before.

It looks like Go has supported pax [since](https://codereview.appspot.com/6700047) [v1.1](https://github.com/golang/go/commit/106827990466b9246f9395233882b1e816df398a).

Spun off from discussion in #336.  I don't really care what spec we require here.  Requiring ustar (IEEE Std 1003.1-1988) support would be fine (a lower bar to implement, but does not support path names with more than 100 characters, etc.).  Or requiring support for a number of specifications (e.g. you have to understand both pax _and_ GNU records).  But I think we should be requiring something, otherwise interop relies on unspecified, out-of-band agreements on layer format.

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -6,8 +6,14 @@ This document will use a concrete example to illustrate how to create and consum
 
 ## Distributable Format
 
-Layer Changesets for the [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` MUST be packaged in a [tar archive][tar-archive] compressed with [gzip][gzip].
-Layer Changesets for the [mediatype](./media-types.md) `application/vnd.oci.image.layer.tar+gzip` MUST NOT include duplicate entries for file paths in the resulting [tar archive][tar-archive].
+Layer changesets have the [media type](media-types.md) `application/vnd.oci.image.layer.tar+gzip`.
+Layer changesets SHOULD be packaged in the ustar interchange format, as [specified by IEEE Std 1003.1-2013][ustar], and MUST be compressed using gzip, as [specified by RFC 1952][rfc1952].
+Layer changesets MUST NOT include duplicate entries for target paths (the value computed from `prefix` and `name` in the ustar header).
+
+Implementations consuming layer changesets MUST be able to unpack both gzip and ustar.
+Portable layers SHOULD NOT use features that [ustar][] leaves unspecified, undefined, or implementation-defined.
+For example, pax [extends ustar by specifying `typeflag` values `g` and `x`][pax-header], so support for unpacking such entries may be mixed.
+[Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they [are a GNU extension][tar.5-gnu].
 
 ## Change Types
 
@@ -20,69 +26,6 @@ Types of changes that can occur in a changeset are:
 Additions and Modifications are represented the same in the changeset tar archive.
 
 Removals are represented using "[whiteout](#whiteouts)" file entries (See [Representing Changes](#representing-changes)).
-
-### File Types
-
-Throughout this document section, the use of word "files" or "entries" includes:
-
-* regular files
-* directories
-* sockets
-* symbolic links
-* block devices
-* character devices
-* FIFOs
-
-### File Attributes
-
-Where supported, MUST include file attributes for Additions and Modifications include:
-
-* Modification Time (`mtime`)
-* User ID (`uid`)
-    * User Name (`uname`) *secondary to `uid`*
-* Group ID (`gid `)
-    * Group Name (`gname`) *secondary to `gid`*
-* Mode (`mode`)
-* Extended Attributes (`xattrs`)
-* Symlink reference (`linkname` + symbolic link type)
-* [Hardlink](#hardlinks) reference (`linkname`)
-
-[Sparse files](https://en.wikipedia.org/wiki/Sparse_file) SHOULD NOT be used because they lack consistent support across tar implementations.
-
-#### Hardlinks
-
-Hardlinks are a [POSIX concept](http://pubs.opengroup.org/onlinepubs/9699919799/functions/link.html) for having one or more directory entries for the same file on the same device.
-Not all filesystems support hardlinks (e.g. [FAT](https://en.wikipedia.org/wiki/File_Allocation_Table)).
-
-Hardlinks are possible with all [file types](#file-types) except `directories`.
-Non-directory files are considered "hardlinked" when their link count is greater than 1.
-Hardlinked files are on a same device (i.e. comparing Major:Minor pair) and have the same inode.
-The corresponding files that share the link with the > 1 linkcount may be outside the directory that the changeset is being produced from, in which case the `linkname` is not recorded in the changeset.
-
-Hardlinks are stored in a tar archive with type of a `1` char, per the [GNU Basic Tar Format][gnu-tar-standard] and [libarchive tar(5)][libarchive-tar].
-
-While approaches to deriving new or changed hardlinks may vary, a possible approach is:
-
-```
-SET LinkMap to map[< Major:Minor String >]map[< inode integer >]< path string >
-SET LinkNames to map[< src path string >]< dest path string >
-FOR each path in root path
-  IF path type is directory
-    CONTINUE
-  ENDIF
-  SET filestat to stat(path)
-  IF filestat num of links == 1
-    CONTINUE
-  ENDIF
-  IF LinkMap[filestat device][filestat inode] is not empty
-    SET LinkNames[path] to LinkMap[filestat device][filestat inode]
-  ELSE
-    SET LinkMap[filestat device][filestat inode] to path
-  ENDIF
-END FOR
-```
-
-With this approach, the link map and links names of a directory could be compared against that of another directory to derive additions and changes to hardlinks.
 
 ## Creating
 
@@ -112,7 +55,7 @@ rootfs-c9d-v1/
         my-app-tools
 ```
 
-The `rootfs-c9d-v1` directory is then created as a plain [tar archive][tar-archive] with relative path to `rootfs-c9d-v1`.
+The `rootfs-c9d-v1` directory is then created as a plain [tar archive][ustar] with paths relative to `rootfs-c9d-v1`.
 Entries for the following files:
 
 ```
@@ -127,7 +70,7 @@ Entries for the following files:
 ### Populate a Comparison Filesystem
 
 Create a new directory and initialize it with a copy or snapshot of the prior root filesystem.
-Example commands that can preserve [file attributes](#file-attributes) to make this copy are:
+Example commands that can preserve [file attributes][ustar] to make this copy are:
 * [cp(1)](http://linux.die.net/man/1/cp): `cp -a rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
 * [rsync(1)](http://linux.die.net/man/1/rsync):  `rsync -aHAX rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
 * [tar(1)](http://linux.die.net/man/1/tar): `mkdir rootfs-c9d-v1.s1 && tar --acls --xattrs -C rootfs-c9d-v1/ -c . | tar -C rootfs-c9d-v1.s1/ --acls --xattrs -x` (including `--selinux` where supported)
@@ -188,7 +131,7 @@ This reflects the removal of `/etc/my-app-config` and creation of a file and dir
 
 ### Representing Changes
 
-A [tar archive][tar-archive] is then created which contains *only* this changeset:
+A [tar archive][ustar] is then created which contains *only* this changeset:
 
 - Added and modified files and directories in their entirety
 - Deleted files or directories marked with a [whiteout file](#whiteouts)
@@ -315,7 +258,7 @@ Layers that have these restrictions SHOULD be tagged with an alternative mediaty
 [Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
 It is implementation-defined whether or not implementations upload layers tagged with this media type.
 
-[libarchive-tar]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#POSIX_ustar_Archives
-[gnu-tar-standard]: http://www.gnu.org/software/tar/manual/html_node/Standard.html
-[tar-archive]: https://en.wikipedia.org/wiki/Tar_(computing)
-[gzip]: http://www.zlib.org/rfc-gzip.html
+[ustar]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_06
+[pax-header]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_02
+[rfc1952]: https://tools.ietf.org/html/rfc1952
+[tar.5-gnu]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#gnu-tar-archives


vbatts:
Trevor, this is a bit excessive. Honestly i'm inclined to just close this.
@jonboulle @stevvooe thoughts?

wking:
On Sat, Sep 24, 2016 at 05:44:38AM -0700, Vincent Batts wrote:
> Trevor, this is a bit excessive.

Would it be less excessive if I moved the bar from pax (IEEE Std
1003.1-2013) to ustar (IEEE Std 1003.1-1988)?  Or is it the idea of
requiring compatibility with a tar-ish standard of any kind that you
find excessive?

jonboulle:
I do like the spirit of this because going down the path of defining this all in-spec (e.g. #317, #336) is a rabbit hole and I'd rather defer to external standards wherever possible. Unfortunately pax is not super well known or popular so I fear we risk a lot of confusion by changing the references everywhere like this. Can we still generally talk about tar but then just define more specifically what we mean by it? (most modern tar implementations support pax right?)

wking:
On Mon, Sep 26, 2016 at 02:09:44AM -0700, Jonathan Boulle wrote:
> Unfortunately pax is not super well known or popular so I fear we
> risk a lot of confusion by changing the references everywhere like
> this.

If we want to require pax support, I can put in links from anywhere
1833fd1 said “pax archive”.  Or just drop the archive information from
most places outside of layer.md, because the fact that layers are
using tar or pax or whatever does not matter outside of generating and
unpacking layers.

> Can we still generally talk about tar but then just define more
> specifically what we mean by it?

If we're requiring pax support, I'd rather call it pax.  It seems like
it would be more confusing to say “tar” when we meant “at least pax”.
But I agree that there is less name-recognition for pax.  I've pushed
1833fd1 → 93b5819 shifting the bar from pax to ustar.  I also
tightened the wording to make it clear that ustar is the minimum
required bar for unpackers, and that layer generators are free to use
other formats if they are comfortable with the posibility of an
OCI-compliant unpacker choking on them.

> (most modern tar implementations support pax right?)

I don't know how many tar implementations there are, but libarchive
(obviously?) supports pax [1].  Go supports pax (links in my topic
post here).  Python supports pax [2].  Perl's Archive::Tar supports
pax [3] (and defines a socket typeflag as 8, which I haven't been able
to turn up a spec for).  GNU tar has supported pax since v1.14 [4].
If there's anyone you're curious about, I'm happy to look into their
pax support.

[1]: https://github.com/libarchive/libarchive/blob/v3.2.1/libarchive/archive_write_set_format_pax.c
[2]: https://docs.python.org/3/library/tarfile.html
[3]: http://perldoc.perl.org/Archive/Tar.html
[4]: https://www.gnu.org/software/tar/manual/html_node/posix.html#SEC145

stevvooe:
This is way out in left field. No one know what any of this language means and it makes the specification impossible to read.

What this is called is _affectation_:

> a behavior, speech, or writing that is artificial and designed to impress.

wking:
On Mon, Sep 26, 2016 at 11:41:45AM -0700, Stephen Day wrote:
> This is way out in left field. No one know what any of this language
> means and it makes the specification impossible to read.

Can you point out a particular bit of the language I'm adding which
needs clarification?

I agree with @jonboulle that when external specs defining ustar (or
whatever tar flavor we want to require in unpackers) and gzip exist,
we want to punt to them instead of defining our own tar flavor [1].
We already to some punting to other standards, e.g. to RFC 2119 [2]
and RFC 4634 [3].  I don't see anything especially confusing about
linking to an external spec for our chosen tar flavor and gzip.  But
that doesn't mean that the wording I'm using to do so couldn't be
improved, and I welcome suggestions to improve that wording.

[1]: https://github.com/opencontainers/image-spec/pull/342#issuecomment-249518142
[2]: https://github.com/opencontainers/image-spec/blame/v0.5.0/README.md#L7
[3]: https://github.com/opencontainers/image-spec/blob/v0.5.0/descriptor.md#sha-256

runcom:
I never heard about ustar in my entire life

stevvooe:
> Can you point out a particular bit of the language I'm adding which
needs clarification?

The entire thing is an example of affectation, even the comment that you responded with. The links you keep adding everywhere don't help to provide context or understanding. They are just a way to make your points look intelligent "'cause references".

The fact is, no one knows what pax or ustar is (nor are these very well written specs). They know what tar is and there are several implementations. The stuff you are linking to are attempts at standards but most people defer to the implementations in GNU and BSD.

Writing a specification like this doesn't add clarity, it just adds volume. Help your reader, don't drown them.

wking:
On Mon, Sep 26, 2016 at 12:05:41PM -0700, Stephen Day wrote:
> The fact is, no one knows what pax or ustar is (nor are these very
> well written specs). They know what tar is and there are several
> implementations.
>
> The stuff you are linking to are attempts at standards but most
> people defer to the implementations in GNU and BSD.

I think “the current GNU (or BSD) implementation” or even “the v1.14
GNU tar implementation” are much worse *specs* from a validation
standpoint than the ustar spec in [1].  The point of a spec is to
define these things to promote interop.  If someone hands you a
tarball using an S typeflag, does “everyone know” how to unpack that?
@vbatts seems to think not [2] (and I agree that folks outside of GNU
tar are less likely to have implemented a GNU-tar extension like
sparse files).  If we set a minimum bar for unpackers, then layer
authors will know how far they can push things before they risk
portability issues.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/utilities/pax.html#tag_20_92_13_06
[2]: https://github.com/opencontainers/image-spec/pull/255#r77822800

stevvooe:
@wking 

> Have you ever written a tar specification, validator, or unpacker?

I know @runcom's credentials. Have you?

wking:
On Mon, Sep 26, 2016 at 12:35:11PM -0700, Stephen Day wrote:
> > Have you ever written a tar specification, validator, or unpacker?
> 
> I know @runcom's credentials. Have you?

Nope.  Which is why I want to stay out of that business ;).  But given
that GNU's tar(1), libarchive's tar(5), the POSIX pax spec, etc. all
reference ustar, it's hard to imagine someone who has done any of
those things *not* having stumbled across “ustar” before.  I'd never
heard of it before image-spec either, but I haven't been involved in
writing a spec that needs tar before.  I think this sort of prior-art
research is an important part of writing such a spec, and where
someone else has already done all the work writing a tar standard, and
that standard has been widely supported for almost 30 years, I think
it is wise to lean on that standard instead of waving the “everybody
already knows this” wand around it.

vbatts:
On 26/09/16 12:55 -0700, W. Trevor King wrote:
>On Mon, Sep 26, 2016 at 12:35:11PM -0700, Stephen Day wrote:
>> > Have you ever written a tar specification, validator, or unpacker?
>>
>> I know @runcom's credentials. Have you?
>
>Nope.  Which is why I want to stay out of that business ;).  But given
>that GNU's tar(1), libarchive's tar(5), the POSIX pax spec, etc. all
>reference ustar, it's hard to imagine someone who has done any of
>those things *not* having stumbled across “ustar” before.  I'd never
>heard of it before image-spec either, but I haven't been involved in
>writing a spec that needs tar before.  I think this sort of prior-art
>research is an important part of writing such a spec, and where
>someone else has already done all the work writing a tar standard, and
>that standard has been widely supported for almost 30 years, I think
>it is wise to lean on that standard instead of waving the “everybody
>already knows this” wand around it.

I've heard of it. And used it. As well know that many features that
folks rely on for distributing container images is not possible with
ustar and pax.
This whole web of references is not accomplishing the task at hand.

wking:
On Mon, Sep 26, 2016 at 01:08:19PM -0700, Vincent Batts wrote:
> I've heard of it. And used it. As well know that many features that
> folks rely on for distributing container images is not possible with
> ustar and pax.

This is useful information to put in the spec.  Which features are you
looking for that aren't covered by ustar?  Link names over 100 chars?
Something else?  Which features aren't covered by pax?  Sparse files?
Something else?  Are these features used commonly enough that we
unpackers MUST handle them?  Or are they rare enough that we want to
leave a low bar like ustar and leave interop for these advanced
features up to out-of-band communication between the layer author and
unpacker?

stevvooe:
@wking Linking out to a large document can often add more confusion than it provides.

We need to be clear, in context, which portion of those specifications matter. For example, handling of sockets or pipes is probably not in scope and that needs to be clear this specification.

wking:
On Mon, Sep 26, 2016 at 02:43:13PM -0700, Stephen Day wrote:
> We need to be clear, in context, which portion of those
> specifications matter.

Do we?  I expect all consumers will be using an off-the-shelf tar
library which supports all of ustar (or all of pax, or whatever).
Requiring part of ustar (or part of pax, or whatever) seems like more
complication than it's worth.

> For example, handling of sockets or pipes is probably not in scope
> and that needs to be clear this specification.

I think handling FIFOs (aka “named pipes”) *is* important, and ustar
supports them.

Our current spec also claims (ish) support for sockets [1], but I
don't see the point to that (whoever is listening should be able to
create their own socket when they start up), have not found any spec
covering it, and have not found support for storing them outside of
Perl's Archive::Tar [2].  So ustar, pax, and I agree with you that
sockets are not in scope.

[1]: https://github.com/opencontainers/image-spec/blame/8f4f36be0563cc7c1234046ea00afe2e767d68ca/layer.md#L30
[2]: https://github.com/opencontainers/image-spec/pull/342#issuecomment-249619230

wking:
Rebased around #316 and #336 with 93b5819 → 37ec118.  There are no
changes in the lines I'm adding, but with #316 in place I'm no longer
adding the gzip requirement.

I'm dropping the hardlink discussion from #336, because the ustar spec
covers hardlinks without us having to explain them locally.  The
discussion in #336 is more detailed than the ustar equivalent, but
folks have been able to write interoperable tar implementations
sharing hardlinks since 1988 based on the ustar spec, so I doubt we
need to add the additional background and implementation hints here.
On the other hand, I'm not strongly against keeping some of these if
folks do want to keep some or all of the hardlink tutorial, I'm fine
restoring it.

stevvooe:
> I'm dropping the hardlink discussion from #336, because the ustar spec
covers hardlinks without us having to explain them locally.  The
discussion in #336 is more detailed than the ustar equivalent, but
folks have been able to write interoperable tar implementations
sharing hardlinks since 1988 based on the ustar spec, so I doubt we
need to add the additional background and implementation hints here.

From experience, this is hardly the case. This spec needs to be clear about hardlinks _in the context of layer files,_ which aren't covered in the 1988 specification.

Remember, while these are indeed tar files, from the point of view of the specification, they represent the semantic concept of a filesystem changeset. One cannot simply unpack these on top of each other and get a useable result. The tar entries need to be interpreted and this specification instructs that interpretation.

wking:
On Tue, Sep 27, 2016 at 01:32:32PM -0700, Stephen Day wrote:
> > I'm dropping the hardlink discussion from #336, because the ustar
> > spec covers hardlinks without us having to explain them locally.
> > The discussion in #336 is more detailed than the ustar equivalent,
> > but folks have been able to write interoperable tar
> > implementations sharing hardlinks since 1988 based on the ustar
> > spec, so I doubt we need to add the additional background and
> > implementation hints here.
>
> From experience, this is hardly the case. This spec needs to be
> clear about hardlinks _in the context of layer files,_ which aren't
> covered in the 1988 specification.

I don't see anything that is layer-specific in the hardlink section
except the psuedocode and following line.  I'm fine keeping that if
you like.

> Remember, while these are indeed tar files, from the point of view
> of the specification, they represent the semantic concept of a
> filesystem changeset. One cannot simply unpack these on top of each
> other and get a useable result.

I think that's only because of whiteouts, which are orthogonal to
hardlinks.  In the absence of whiteouts, you *can* unpack them on top
of each other and get the correct result, regardless of hardlink
presence.

runcom:
> On Mon, Sep 26, 2016 at 12:04:41PM -0700, Antonio Murdaca wrote:
>  I never heard about ustar in my entire life
> Have you ever written a tar specification, validator, or unpacker?

@wking I'm going to document myself if I don't know something. But, my comment was referring to the fact that _in my opinion_ you're trying to put stuff into this spec which aren't _so common overall_. The beauty of this spec (as well as the Docker spec) is its **simplicity**. By requiring all this other stuff you're trying to add everywhere you make that simplicity simply go away. This is just my opinion of course.

wking:
On Wed, Sep 28, 2016 at 08:07:08AM -0700, Antonio Murdaca wrote:
> … my comment was referring to the fact that _in my opinion_ you're
> trying to put stuff into this spec which aren't _so common overall_.

I think ustar *is* common.  Are you aware of a tar tool or library
that has seen any development since 2000 which does not support it?

> The beauty of this spec (as well as the Docker spec) is its
> **simplicity**. By requiring all this other stuff you're trying to
> add everywhere you make that simplicity simply go away.

The Wikipedia entry currently linked from the spec has ~3k words.  The
ustar section I'm linking instead has ~2k words, which have been
carefully honed since 1988 as a normative spec.  If we're looking for
a normative link, I think the latter is simpler, and it makes it very
clear what the OCI requires unpackers to handle.

runcom:
> On Wed, Sep 28, 2016 at 08:07:08AM -0700, Antonio Murdaca wrote:
>  … my comment was referring to the fact that _in my opinion_ you're
>  trying to put stuff into this spec which aren't _so common overall_.
> I think ustar *is* common.  Are you aware of a tar tool or library
> that has seen any development since 2000 which does not support it?
>  The beauty of this spec (as well as the Docker spec) is its
>  **simplicity**. By requiring all this other stuff you're trying to
>  add everywhere you make that simplicity simply go away.
> The Wikipedia entry currently linked from the spec has ~3k words.  The
> ustar section I'm linking instead has ~2k words, which have been
> carefully honed since 1988 as a normative spec.  If we're looking for
> a normative link, I think the latter is simpler, and it makes it very
> clear what the OCI requires unpackers to handle.

you probably didn't get what I said if you are still quoting stuff which _I understand_ - so, fair enough, I'm out of this conversation :)

wking:
On Wed, Sep 28, 2016 at 08:22:26AM -0700, Antonio Murdaca wrote:
> you probably didn't get what I said if you are still quoting stuff
> which _I understand_…

I was confused, but @runcom clarified on IRC that he considers the
ustar requirement to be 2k words of new complexity [1].  I, on the
other hand, think we were already suggesting ustar support, because
the Wikipedia article we previously linked had:

  The file structure to store this information was later standardized
  in POSIX.1-1988 and later POSIX.1-2001.

And I think this PR simplifies things by cutting straight to that spec
(currently the ustar stuff originally from the 1988 version) and
bypassing the rest of the stuff in the Wikipedia article.

It also makes it clear that ustar comatibility is only a requirement
for unpackers, and that layer authors are free to generate non-ustar
entries if they're comfortable with the portability risk, while
previously we made no clear requirements of either the layer author or
the unpacker.

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2016-09-28.log.html#t2016-09-28T15:34:38
[2]: https://en.wikipedia.org/wiki/Tar_%28computing%29#UStar_format

wking:
As an example of the sort of thing you can do after this change, I've
stubbed in a callback-based validator with a stict mode [1].  In
strict mode it errors out for typeflags not defined by ustar, but in
permissive mode it does not.

However, without this change there are no grounds in the spec for
making that distinction around ustar typeflags.  There's only the
warning about ‘S’ (GNU's sparse), but no warnings for the other GNU
extensions like ‘K’ or ‘L’ or pax extensions like ‘g’ and ‘x’.

[1]: https://github.com/wking/image-tools/blob/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6/validate/layer.go#L60-L65

runcom:
> However, without this change there are no grounds in the spec for
> making that distinction around ustar typeflags.  There's only the
> warning about ‘S’ (GNU's sparse), but no warnings for the other GNU
> extensions like ‘K’ or ‘L’ or pax extensions like ‘g’ and ‘x’.

and maybe it's not the right time to do this? or it's not necessary altogether? you can see it yourself, that change isn't _required_ anywhere, the spec is still working even w/o it. it's _you_ adding it and then making that change.

wking:
On Thu, Sep 29, 2016 at 03:33:34AM -0700, Antonio Murdaca wrote:
> > However, without this change there are no grounds in the spec for
> > making that distinction around ustar typeflags.  There's only the
> > warning about ‘S’ (GNU's sparse), but no warnings for the other
> > GNU extensions like ‘K’ or ‘L’ or pax extensions like ‘g’ and ‘x’.
> 
> and maybe it's not the right time to do this?

It's hard for me to imagine a modern tar implementation not even
supporting ustar (but this isn't my field, maybe they're out there).
If we don't land a requirement like this before 1.0, the risk is that
layer authors publish layers that use more advanced tar features and
then have their assumed compatibility rug pulled out from under them
when we later land something like this to say “actually, when we said
‘tar’ before, we're only actually requiring ustar support.  Sorry”.
I'd rather avoid that by setting our low bar before 1.0.

> or it's not necessary altogether? you can see it yourself, that
> change isn't _required_ anywhere, the spec is still working even w/o
> it. it's _you_ adding it and then making that change.

The current layer-publisher ecosystem is (as I understand it, again,
not my field) a monoculture around Go's archive/tar library.  The idea
behind the OCI is to pin down standards to facilitate interop between
multiple implementations.  If I create a tar file using
SCHILY.acl.default, will all OCI implementations be able to unpack it?
Libarchive [1], star [2], GNU tar [3], and tar-split [4] understand
it.  But I can't find code in Docker's unpacker to support it [5].
With this PR, it's clear that OCI-compliant unpackers are *not*
required to support SCHILY.acl.default (although they can support it
if they want), and layer authors can push layers using
SCHILY.acl.default if they're comfortable figuring out that
compatibility on their own.  And layer authors can push layers that
only use ustar *without* having to worry about compatibility with
OCI-compliant unpackers.

[1]: https://github.com/libarchive/libarchive/blob/v3.2.1/libarchive/archive_read_support_format_tar.c#L1803-L1852
[2]: http://cdrtools.sourceforge.net/private/man/star/star.4.html
[3]: http://git.savannah.gnu.org/cgit/tar.git/tree/src/xheader.c?h=release_1_29#n1785
[4]: https://github.com/vbatts/tar-split/blob/v0.10.1/archive/tar/common.go#L194
[5]: https://github.com/docker/docker/blob/v1.12.1/pkg/archive/archive.go#L699-L839

vbatts:
I'm still calling this ridiculous and excessive. At best linking to some IEEE pay-walled doc. But for many, that is a loose link that they will not research the validity of. I'm likely going to close this issue.

wking:
On Thu, Oct 06, 2016 at 10:48:24AM -0700, Vincent Batts wrote:
> I'm still calling this ridiculous and excessive. At best linking to
> some IEEE pay-walled doc.

It currently links to a non-paywalled 2013 version of the ustar spec.

aecolley:
This discussion is exactly what the Unix world used to be like before Posix: we just wanted our programs to run without having to make conditionals for every different distribution's way of doing every single trivial thing. Posix didn't standardize the `tar` command, so it's still a mess. Now it's an anachronistic mess.

Why not just say [GNU tar](http://www.gnu.org/software/tar/)? Meaning that conforming images will be extractable by GNU tar without unusual options. If you really wanted to tie it down, you could specify a tar version (1.29 is current). GNU is already central to Linux and tolerated on Windows, so it's a good-enough choice.

wking:
On Mon, Oct 17, 2016 at 04:08:12PM -0700, Adrian Colley wrote:
> Why not just say [GNU tar](http://www.gnu.org/software/tar/)?
> Meaning that conforming images will be extractable by GNU tar
> without unusual options. If you really wanted to tie it down, you
> could specify a tar version (1.29 is current). GNU is already
> central to Linux and tolerated on Windows, so it's a good-enough
> choice.

I'm ok with requiring GNU tar, although I'd definitely want to pin to
a version, and it would be nice to have a clear definition of what
that spec is.  Linking to [1] (or docs built from it [2], like we
currently do [3]) seems fairly loose.  For example:

  In addition to entries describing archive members, an archive may
  contain entries which tar itself uses to store information. See
  section Including a Label in the Archive, for an example of such an
  archive entry.

is pretty open-ended.  The questions are “are the GNU docs
sufficiently spec-like?” and “do we actually want to require
implementations to support everything in the GNU spec?”.  I'm not sure
about either, but would be happier linking to a spec of any stripe
than leaving the required tar support unspecified.

[1]: http://git.savannah.gnu.org/cgit/tar.git/tree/doc/intern.texi?h=release_1_29
[2]: http://www.gnu.org/software/tar/manual/html_node/Standard.html
[3]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc1/layer.md#L319

vbatts:
I'm closing this as docs currently reference GNU tar "standard" format, as well as having requirements for xattrs. This PR goes far enough that would be immediately incompatible with what folks are using and expecting.

PR #351, opened 2016-09-30
OPTIONAL field MAY be null

coolljt0725:
This field MAY be null is duplicate with the `OPTIONAL`

### DIFF
--- a/config.md
+++ b/config.md
@@ -48,6 +48,8 @@ Since the [configuration JSON](#image-json) that gets hashed references hashes o
 
 ## Properties
 
+Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being absent.
+
 - **created** *string*, OPTIONAL
 
   An ISO-8601 formatted combined date and time at which the image was created.
@@ -113,7 +115,7 @@ Since the [configuration JSON](#image-json) that gets hashed references hashes o
    - **Entrypoint** *array of strings*, OPTIONAL
 
      A list of arguments to use as the command to execute when the container starts.
-     This value acts as a default and is replaced by an entrypoint specified when creating a container. This field MAY be null.
+     This value acts as a default and is replaced by an entrypoint specified when creating a container.
 
    - **Cmd** *array of strings*, OPTIONAL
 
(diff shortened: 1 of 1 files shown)


jonboulle:
I think this was talked about in a previous issue somewhere - it's not quite redundant - the field being optional means it might not be present, but that's different to it being present but having an empty/null value.

jonboulle:
Could probably do with a once-over of the spec to make sure we're consistent about this for all fields.

coolljt0725:
>Could probably do with a once-over of the spec to make sure we're consistent about this for all fields.

+1 for this to make it more clear.

coolljt0725:
@jonboulle  add a sentence to clarify this

jonboulle:
@coolljt0725 how is this clearer than what was originally there? it's no longer in context

coolljt0725:
@jonboulle  maybe I just misunderstood your point. I think all the `OPTIONAL` field could be null

jonboulle:
That makes sense, but if so, I'd rather move it to the top of the
PROPERTIES section, and tweak the wording a bit -

Any OPTIONAL field MAY also be set to null, which is equivalent to being
unspecified.

duglin:
To be be really anal:

`Any OPTIONAL field MAY also be set to null, which is equivalent to being absent.`

wking:
On Fri, Sep 30, 2016 at 07:39:34AM -0700, Jonathan Boulle wrote:
> That makes sense, but if so, I'd rather move it to the top of the
> PROPERTIES section, and tweak the wording a bit -

If we can make a general rule of this (and that sounds good to me), I
think that rule belongs in the README or some other file to cover all
of our JSON, instead of making a rule that applies only to
application/vnd.oci.image.config.v1+json.

> Any OPTIONAL field MAY also be set to null, which is equivalent to
> being unspecified.

I like this wording^.  I've floated using ‘null’ (vs. leaving the
property off) as a way to explicitly set a property unlimited [1], but
that could be tedious to handle in Go and the idea hasn't caught on
yet.  @crosbymichael has also pushed for a distinction between ‘null’
and Go's default value [2]; I'm not clear on why, but it may impact
this as well.

[1]: https://github.com/opencontainers/runtime-spec/pull/499#issuecomment-226572995
[2]: https://github.com/opencontainers/runtime-spec/pull/576#issuecomment-248759913

vbatts:
LGTM

philips:
LGTM

PR #358, opened 2016-10-01
schema/defs-config.json: Require rootfs properties and enumerate type

wking:
The property had been (erroneously) "layers"; this commit fixes that by making the property "type" and using an enum to require the value to be "layers".  It also adds "required" entries for both rootfs.diff_ids and rootfs.type.


vbatts:
LGTM

philips:
LGTM

PR #365, opened 2016-10-04
*: split manifest and manifest list documents

jonboulle:
The combined manifest/manifest-list document is long and unwieldy.
Furthermore, the Image Manifest is arguably more important (since the
Manifest List is optional), and so it should stand alone as a
first-class document.


vbatts:
yes
LGTM

philips:
LGTM

PR #145, opened 2016-06-16
schema: add a docker v2.2 backwards compat test

philips:
This adds a test of the schema compatibility by taking the manifest of
library/docker from the Docker Hub, with sha256 digest verification, and
simply finding/replacing the four constant strings documented in
media-types.md and ensuring that it passes OCI validation.

https://github.com/opencontainers/image-spec/blob/master/media-types.md

This test is functionaly equivalent to downloading a Docker v2.2
manifest into the file `docker-manifest`:

```
$ curl -L -H "Authorization: Bearer ..." \
	-H "Accept: application/vnd.docker.distribution.manifest.v2+json" \
	https://registry-1.docker.io/v2/library/docker/manifests/sha256fg:888206c77cd2811ec47e752ba291e5b7734e3ef137dfd222daadaca39a9f17bc > docker-manifest
```

And then running a regex to change `docker.distribution` to `oci.image`
and changing `v2` to `v1`

```
$ cat docker-manifest | sed -e "s%docker\.distribution%oci\.image%g" -e "s%v2%v1%g" > oci-manifest
```

And finally testing the manifest with oci-image-tool:

```
$ oci-image-tool validate oci-manifest
oci-manifest: OK
```

Signed-off-by: Brandon Philips <brandon.philips@coreos.com>

PR #368, opened 2016-10-04
remove unit test for compatability matrix

xiekeyang:
From 
#304 
and
https://github.com/opencontainers/image-spec/pull/320#issuecomment-250998880 

We seems agree not to support Docker media types.
If so, this unit test for Docker media support should be removed.
And we should add constraint only for OCI types.
cc @wking 

Signed-off-by: xiekeyang <keyang.xie@gmail.com>


wking:
fb63aee rolls back #145.  I'm fine leaving it in to compare the
current spec against Docker JSON that has been caught in the wild.

I don't see Docker-compat as a requirement for external
implementations (it is explicitly not a requirement with #304), and I
don't think we need to facilitate Docker-compat in external
implementations (which is why I don't think we need #320).  But I
don't have a problem with a *spec* constraint that image-spec v1
accept any Docker v2 content after a media-type change (which is what
these tests are about).  On the other hand, I wouldn't mind making the
change more involved either (#224).  So I don't mind either way on
this PR, but think #320 should stay closed.

xiekeyang:
@wking 
Agree. I'd close #320.
Hopefully, this PR can be accepted, because this is required by #341.
#341 resolve the descendants validation in manifest. But UT can not pass for beyond OCI cases. So this PR should be stand firstly.

wking:
On Tue, Oct 04, 2016 at 09:33:13PM -0700, xiekeyang wrote:
> Hopefully, this PR can be accepted, because this is required by
> #341. #341 resolve the descendants validation in manifest. But UT
> can not pass for beyond OCI cases. So this PR should be stand
> firstly.

I'm not entirely sure I understand you here, but while
application/vnd.docker.distribution.manifest.v2+json is not an OCI
type, a compliant application/vnd.oci.image.manifest.list.v1+json MAY
reference one.  #304 only requires OCI unpackers to understand the OCI
types, and an OCI manifest-list pointing at a Docker manifest should
get a “this may not be portable” warning from the image-validator but
should not raise a validation error.  The same applies to other Docker
types.  So I don't see a need to remove these compatibility tests (but
I wouldn't mind if we did).

On the other hand, #341 would error instead of printing a “may not be
portable” warning, so I agree that we don't want to handle this
portion of the validation in JSON Schema [1].

[1]: https://github.com/opencontainers/image-spec/pull/341#issuecomment-250671146

PR #371, opened 2016-10-06
config: adding labels

vbatts:
the "Labels" field is not specified in the docker image spec, but is
present and used.
https://github.com/docker/docker/blob/master/image/spec/v1.2.md

This carries over that functionality

Signed-off-by: Vincent Batts vbatts@hashbangbash.com

### DIFF
--- a/config.md
+++ b/config.md
@@ -133,7 +133,7 @@ Since the [configuration JSON](#image-json) that gets hashed references hashes o
      Sets the current working directory of the entrypoint process in the container.
      This value acts as a default and is replaced by a working directory specified when creating a container.
 
-   - **Labels ** *object*, OPTIONAL
+   - **Labels** *object*, OPTIONAL
 
      The field contains arbitrary metadata for the container.
      Labels MUST be a key-value map where both the key and value MUST be strings.
--- a/config.md
+++ b/config.md
@@ -133,6 +133,16 @@ Since the [configuration JSON](#image-json) that gets hashed references hashes o
      Sets the current working directory of the entrypoint process in the container.
      This value acts as a default and is replaced by a working directory specified when creating a container.
 
+   - **Labels ** *object*, OPTIONAL
+
+     The field contains arbitrary metadata for the container.
+     Labels MUST be a key-value map where both the key and value MUST be strings.
+     Keys MUST be unique within this map, and best practice is to namespace the keys.
+     Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+     Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by subsequent specifications.
+     If there are no annotations then this property MAY either be absent or an empty map.
+     Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown annotation key.
+
 - **rootfs** *object*, REQUIRED
 
    The rootfs key references the layer content addresses used by the image.
@@ -213,7 +223,11 @@ Here is an example image configuration JSON document:
             "/var/job-result-data": {},
             "/var/log/my-app-logs": {}
         },
-        "WorkingDir": "/home/alice"
+        "WorkingDir": "/home/alice",
+        "Labels": {
+            "com.example.project.git.url": "https://example.com/project.git",
+            "com.example.project.git.commit": "45a939b2999782a3f005621a8d0f29aa387e1d6b"
+	}
     },
     "rootfs": {
       "diff_ids": [


config.md
```diff
@@ -213,7 +223,11 @@ Here is an example image configuration JSON document:
             "/var/job-result-data": {},
             "/var/log/my-app-logs": {}
         },
-        "WorkingDir": "/home/alice"
+        "WorkingDir": "/home/alice",
+	"Labels": {
```
jonboulle: indenting is off

jonboulle:
Where is it specified in Docker? Only in code?

vbatts:
yep

config.md
vbatts: fixed

philips:
LGTM

jonboulle:
@vbatts how can we be confident we aren't missing other fields that are present, used, but only defined in code?

coolljt0725:
@vbatts  @jonboulle  @philips   
I think there are several other fields that are present and used in docker images, but not defined in the spec. The entire `config` of a docker image is https://github.com/docker/docker/blob/master/api/types/container/config.go#L36 , but `Memory MemorySwap CpuShares` are actually not used in docker images any more . Here is a config of an image pull from docker hub use `skopeo`. 
````

  "config": {
    "Labels": {},
    "OnBuild": null,
    "OpenStdin": false,
    "Tty": false,
    "AttachStderr": false,
    "AttachStdout": false,
    "AttachStdin": false,
    "User": "",
    "Domainname": "",
    "Hostname": "b18b9cafe0e0",
    "StdinOnce": false,
    "Env": [
      "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
    ],
    "Cmd": [
      "/bin/bash"
    ],
    "ArgsEscaped": true,
    "Image": "sha256:16ca893df3275f0d891c5d004788609363adce7b6e6c36d67a8d5bf89c12d708",
    "Volumes": null,
    "WorkingDir": "",
    "Entrypoint": null
  },

````

From what I know, there are concepts of `config`(https://github.com/docker/docker/blob/master/api/types/container/config.go#L36) and `hostConfig`(https://github.com/docker/docker/blob/master/api/types/container/host_config.go#L278) in docker. `config` are supposed to be portable and will be commit to images, `hostConfig` are supposed to be non-portable and are not to be commit to images.
`Memory` `MemorySwap` `CpuShares` does exist in `config` of image before docker 1.7.0, but after https://github.com/docker/docker/pull/10298, resource config will not commit to image any more. 

Not related to this, I'm always curious about why we only have `Memory` `MemorySwap` `CpuShares` defined in the spec since there are many other cgroup resource config for container, after viewing the git history of `config.go`, I think it's because docker only have these three filed defined at the very beginning(https://github.com/docker/docker/pull/4085/files#diff-2660508eef7a8eeadb6d27b8bbf6d718R15),so the spec (https://github.com/docker/docker/pull/9560/files?short_path=f10cb3a#diff-f10cb3a0965ab8eb7051f8e367802e69) at that moment only define these three filed.  and this spec doesn't updated after adding more fields and removing some fields.

So I'm just wondering should we add are all the missing but used and present filed of docker image to the config of spec and also removing `Memory` `MemorySwap` `CpuShares` so we can keep consistent with docker images?
My thoughts is that the config should only contains portable container runtime config like `Env` `Cmd`... and we define a annotations just like #372  does, so user can add implementation specified filed.

wking:
On Sat, Oct 08, 2016 at 01:28:59AM -0700, Lei Jitang wrote:
> So I'm just wondering should we add are all the missing but used and
> present filed of docker image to the config of spec and also
> removing `Memory` `MemorySwap` `CpuShares` so we can keep consistent
> with docker images?  My thoughts is that the config should only
> contains portable container runtime config like `Env` `Cmd`... and
> we define a annotations just like #372 does, so user can add
> implementation specified filed.

The problem here is that “portable container runtime config” is a
pretty fuzzy boundary (opencontainers/runtime-spec#284).  I think we
want to drop the image-spec runtime configuration, distribute the
runtime-spec runtime configuration (with whatever the image author
wants to put in it), and sanitize the config after unpacking it [1].
Sanitization code in opencontainers/runtime-tools#219.

If that's too big a departure from the current Docker approach to fit
into 1.0, I think we should blindly do whatever Docker is doing now
until we are ready to drop the image-spec runtime configuration ;).
But if we're intending to unpack a runtime-spec config, we'll have to
figure out the application/vnd.docker.container.image.v1+json →
runtime-spec config translation somewhere.  I'd rather have that be an
image-build-time problem.  And santization/localization would be a
post-unpack, just-before-runtime problem.

[1]: https://github.com/opencontainers/image-spec/pull/372#issuecomment-252084933

vbatts:
@jonboulle we can be reasonably certain that there are other fields that are not defined, but only in code.

vbatts:
@coolljt0725 on that note, I would be fine seeing Memory*  and CPU* removed from the OCI image spec. Further, having an annotation that provides a *hint* for memory and CPU demands of the contained application.

vbatts:
I'm tagging this for the v1.0.0-rc3 milestone, because it is needed for compatibility and it is something that the ecosystem has grown to depend on.

vbatts:

stevvooe:
@vbatts I'm not quite sure what exactly is going on here with the extra fields. Relevant code is in https://github.com/docker/docker/blob/2a2f183627af25fb6f1c93c5fc2c4d04ab7e75bb/daemon/commit.go#L196. The interesting one to me is `container_config` and `config`. __Further analysis must be done to see if these are applied at runtime.__

Regarding labels, we need to be careful about the semantics of label inheritance. It is not well-defined in docker, at all, and there are some tricky situations to address, such as removing labels. It may be sufficient to say to inheritance model isn't specified, but we should at least study the problem.

For the most part, these missing fields are of little consequence, as docker maintains the original byte content, rather than relying on serialization round trips.

vbatts:
@stevvooe i really do not follow you regarding "if these are applied at runtime". What are you expecting? Because it seems to me you are describing something more than just arbitrary metadata, which is the perception labels have to the majority of folks I have seen using them. It is not like an environment variable or something that is exposed into the container.

stevvooe:
@vbatts I would not expect that labels on image be applied to a running container. With the way labels are used, it is a massive security hole. Unfortunately, looks like that cat is out of the bag: https://github.com/docker/docker/blob/2a2f183627af25fb6f1c93c5fc2c4d04ab7e75bb/daemon/commit.go#L67.

vbatts:
I am not seeing this massive security hole. :-\

Is this cat the one regarding inheritance of labels? If so, it is as I
figured. They just stack or clobber.

stevvooe:
LGTM

> I am not seeing this massive security hole. :-\

Let's say I have a label-based load balancer or ACL system. One can then craft an image such that the label on the image will cause the created container to join the load balancing group or gain privileges that it doesn't already have.

This cat has outgrown the bag, so I'm not sure there is much to do from OCI's perspective.

stevvooe:
@philips PTAL

philips:
LGTM

PR #373, opened 2016-10-08
set paragraph to massive description

xiekeyang:
this big section looks a little hard to read, so I set it to paragraphs.

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -44,11 +44,14 @@ afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256/
 ```
 
 Object names in the `refs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
+
 Object names in the `blobs` subdirectories are composed of a directory for each hash algorithm, the children of which will contain the actual content.
 A blob, referenced with digest `<alg>:<hex>` (per [descriptor](descriptor.md#digests-and-verification)), MUST have its content stored in a file under `blobs/<alg>/<hex>`.
 The character set of the entry name for `<hex>` and `<alg>` MUST match the respective grammar elements described in [descriptor](descriptor.md#digests-and-verification).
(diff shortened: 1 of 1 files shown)


image-layout.md
```diff
@@ -44,11 +44,16 @@ afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 ./blobs/sha256/
 ```
 
 Object names in the `refs` subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
+
 Object names in the `blobs` subdirectories are composed of a directory for each hash algorithm, the children of which will contain the actual content.
+
 A blob, referenced with digest `<alg>:<hex>` (per [descriptor](descriptor.md#digests-and-verification)), MUST have its content stored in a file under `blobs/<alg>/<hex>`.
+
 The character set of the entry name for `<hex>` and `<alg>` MUST match the respective grammar elements described in [descriptor](descriptor.md#digests-and-verification).
 For example `sha256:5b` will map to the layout `blobs/sha256/5b`.
```
wking: I think you want to drop the two previous blank lines.  All of this is about mapping blob digests into the `blobs` tree.  With that change you'd have one paragraph about the `refs` tree, one about the `blobs` tree, one about dangling blobs, and one about missing blobs.  That seems reasonable to me.

xiekeyang: @wking 
Yes, I'm separating one big paragraph to three, each of them express an individual concern.

wking: I think you *should* be separating one big paragraph [into four][1].  With 833917c3 you're separating it into six.  The [`blob`-tree paragraph][1] I'd prefer would contain several sentences: two for the `blobs/<algo>/<hash>` structure, one for the algo and hash charsets, and one example.  I think those are all closely-related enough to belong in the same paragraph.

[1]: https://github.com/opencontainers/image-spec/pull/373#discussion_r82899107

wking:
0c624cb looks good to me.

xiekeyang:
@philips PTAL.

vbatts:
LGTM

philips:
LGTM

PR #375, opened 2016-10-08
Clarify User and MemorySwap are platform specific

### LINKED ISSUE #329: container runconfig in Image Configuration is linux centric
The container runconfig(https://github.com/opencontainers/image-spec/blob/master/config.md#container-runconfig-field-descriptions) in Image Configuration is very linux centric.

I think this field should be platform specific, I think we should have a clarification about this.

coolljt0725:
based on the runtime spec, the user are platform-specific and `MemorySwap`  are for Linux based systems.

Fixed #329 

Actually, I'd like to remove `Memory` `MemorySwap` `CpuShares`

### DIFF
--- a/config.md
+++ b/config.md
@@ -86,7 +86,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
 
    - **MemorySwap** *integer*, OPTIONAL
 
-     Total memory usage (memory + swap); set to `-1` to disable swap.
+     MemorySwap is a platform-specific field to set total memory usage (memory + swap) for Linux based systems; set to `-1` to disable swap.
      This acts as a default value to use when the value is not specified when creating a container.
 
    - **CpuShares** *integer*, OPTIONAL
--- a/config.md
+++ b/config.md
@@ -74,9 +74,9 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
 
    - **User** *string*, OPTIONAL
 
-     The username or UID which the process in the container should run as.
+     The username or UID which is a platform-specific structure that allows specific control over which user the process run as.
      This acts as a default value to use when the value is not specified when creating a container.
-     All of the following are valid: `user`, `uid`, `user:group`, `uid:gid`, `uid:group`, `user:gid`.
+     For Linux based systems, all of the following are valid: `user`, `uid`, `user:group`, `uid:gid`, `uid:group`, `user:gid`.
      If `group`/`gid` is not specified, the default group and supplementary groups of the given `user`/`uid` in `/etc/passwd` from the container are applied.
 
    - **Memory** *integer*, OPTIONAL


wking:
On Sat, Oct 08, 2016 at 02:33:13AM -0700, Lei Jitang wrote:
> based on the runtime spec, the user are platform-specific and
> `MemorySwap ` are for Linux based systems.

Solaris also supports the user fields [1].  Both Solaris [2] and
Windows [3] have similar-ish memory settings, but they aren't quite
MemorySwap.

> Actually, I'd like to remove `Memory` `MemorySwap` `CpuShares`

I'd rather use the runtime-spec config.json [4,5,6].  It saves us from
having to copy over settings from runtime-spec, and leaves the “what
settings do I think are safe?” descision up to the image-consumer (who
can use a stock sanitization script if they want).

[1]: https://github.com/opencontainers/runtime-spec/blob/a39be468c9cfd7e873889687ce64376ee41e25d3/config.md#linux-and-solaris-user
[2]: https://github.com/opencontainers/runtime-spec/blob/a39be468c9cfd7e873889687ce64376ee41e25d3/config-solaris.md#cappedmemory
[3]: https://github.com/opencontainers/runtime-spec/blob/a39be468c9cfd7e873889687ce64376ee41e25d3/config-windows.md#memory
[4]: https://github.com/opencontainers/image-spec/issues/329#issuecomment-248238359
[5]: https://github.com/opencontainers/image-spec/pull/372#issuecomment-252084933
[6]: https://github.com/opencontainers/image-spec/pull/371#issuecomment-252706689

coolljt0725:
@wking 
> Solaris also supports the user fields [1].  Both Solaris [2] and
Windows [3] have similar-ish memory settings, but they aren't quite
MemorySwap.

`Solaris ` does have user, I didn't say that Solaris doesn't user field. but we doesn't clarify that the image spec is support for Solaris at the moment  :)
Windows and Solaris do have memory limit, but it's not `MemorySwap`

wking:
On Tue, Oct 11, 2016 at 01:41:09AM -0700, Lei Jitang wrote:
> > Solaris also supports the user fields [1].  Both Solaris [2] and
> > Windows [3] have similar-ish memory settings, but they aren't
> > quite MemorySwap.
>
> `Solaris` does have user, I didn't say that Solaris doesn't user
> field. but we doesn't clarify that the image spec is support for
> Solaris at the moment :)

As long as we have an image-spec-specific config with a user field,
I'd rather clarify it by requiring config translators to populate the
runtime-spec config's process.user for all runtime-spec OSes.  Solaris
and Linux have the same runtime-spec config, so I'd expect they'd have
the same image-spec-config support.

> Windows and Solaris do have memory limit, but it's not `MemorySwap`

As long as we continue to support an image-spec-specific MemorySwap
setting, I think it makes sense to add support for the related Solaris
and Windows fields.  On the other hand, if we decide to drop
MemorySwap, then we can skip the related Solaris and Windows fields.
But it seems inconsistent to only distribute memory/swap limits if the
target OS happens to be Linux.

vbatts:
Largely, LGTM
I am in favor of just removing Memory MemorySwap CpuShares too

vbatts:
needs a rebase

coolljt0725:
@vbatts  rebased

vbatts:
LGTM

jonboulle:
lgtm for now

PR #380, opened 2016-10-11
README.md: Add high-level link to optional image-layout

wking:
To make it clear that supporting image-layout is optional, because there's no way to test support without requiring a particular backend (tar, HTTP, etc.).

Spun off from [here][1].

[1]: https://groups.google.com/a/opencontainers.org/d/msg/dev/s_SuN47XGC4/ycyxzzXnCAAJ

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ The [OCI Image Media Types](media-types.md) document is a starting point to unde
 The high-level components of the spec include:
 
 * An archival format for container images, consisting of an [image manifest](manifest.md), an [image layout](image-layout.md), a set of [filesystem layers](layer.md), and [image configuration](config.md) (base OCI layer)
-* A process of hashing the image format for integrity and content-addressing (base OCI layer)
+* A [process of referencing container images by a cryptographic hash of their content](descriptor.md) (base OCI layer)
+* A format for [storing CAS blobs and references to them](image-layout.md) (optional OCI layer)
 * Signatures that are based on signing image content address (optional OCI layer)
 * Naming that is federated based on DNS and can be delegated (optional OCI layer)
 


philips:
LGTM

jieyu:
LGTM!

README.md
```diff
@@ -40,7 +40,8 @@ The [OCI Image Media Types](media-types.md) document is a starting point to unde
 The high-level components of the spec include:
 
 * An archival format for container images, consisting of an [image manifest](manifest.md), an [image layout](image-layout.md), a set of [filesystem layers](layer.md), and [image configuration](config.md) (base OCI layer)
-* A process of hashing the image format for integrity and content-addressing (base OCI layer)
+* A [process of hashing the image format](descriptor.md) for integrity and content-addressing (base OCI layer)
+* A format for [storing CAS blobs and references to them](image-layout.md) (optional OCI layer)
```
jonboulle: Where is this in the OCI Scope Table?

README.md
```diff
@@ -40,7 +40,8 @@ The [OCI Image Media Types](media-types.md) document is a starting point to unde
 The high-level components of the spec include:
 
 * An archival format for container images, consisting of an [image manifest](manifest.md), an [image layout](image-layout.md), a set of [filesystem layers](layer.md), and [image configuration](config.md) (base OCI layer)
-* A process of hashing the image format for integrity and content-addressing (base OCI layer)
+* A [process of hashing the image format](descriptor.md) for integrity and content-addressing (base OCI layer)
```
jonboulle: It's confusing to link to the top of that document because the first half of it says nothing about the "process of hashing the image format". Probably better to either link to [this section](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests-and-verification) or elsewhere.

wking: If the current text does not justify a link to the top of the file, I think we should change the link text.  If there is no link text that justifies a link to the top of that file, we should probably remove content from the file until there is ;).

jonboulle: How about "a process of referencing container images by a cryptographic hash of their content", which accords better with the relevant table layers ("Use of Hash as Content Addressable name for immutable containers")

README.md
wking: Somewhere around "Use of Hash as Content Addressable name for immutable containers" and "Archival Format".  But if we can't find it in the scope table, we should either update the scope table (my preference) or remove image-layout from the spec.

README.md
wking: Sounds good.  Will update when I get back to my laptop.

wking: Updated with 1a6c4f9 → f5f98cf.

README.md
jonboulle: As thrilled as I would be for the scope table to be updated I can assure you that will be considerably more involved than just tweaking this PR. IMO image layout is basically covered in the first bullet point in this list - it's essentially working towards the "Archival Format" (see ["longer term"](https://github.com/opencontainers/image-spec/pull/94#issue-157118241).

wking: If you're satisfied with the way it is, I'm happy.  If you want me to tweak wording here, just tell me what to put in ;).

vbatts:
LGTM

jonboulle:
lgtm

PR #381, opened 2016-10-11
[ReadMe] Add Compliance Language

RobDolinMS:
Add language parallel to what is in the OCI Runtime.

Signed-off-by: Rob Dolin robdolin@microsoft.com

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,9 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 
 The keywords "unspecified", "undefined", and "implementation-defined" are to be interpreted as described in the [rationale for the C99 standard][c99-unspecified].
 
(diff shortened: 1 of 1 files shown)


philips:
Agreed, we will have entire sections of the spec that are optional vs parts of sections.

vbatts:
This SGTM. It has a whitespace that needs to be fixed also.

RobDolinMS:
@philips Since @wking has a few items in his comment, can you elaborate on what you agree on?

RobDolinMS:
Per discussion on today's OCI Dev ConCall, recommendation is to have more detailed language as Trevor suggests.  

There may be a separate issue related to manifest lists.

philips:
LGTM

We need to add clarifying language manifest list top matter but I have no further concerns for other objects.

vbatts:
Moved to #432

PR #384, opened 2016-10-12
README: add correct dial in

philips:
The 888 numbers don't seem to work; I called and it was an empty room.


jonboulle:
needs DCO :roll_eyes:

philips:
@jonboulle see https://github.com/opencontainers/image-spec/pull/394

PR #385, opened 2016-10-13
Add arm64 to supported architecture

coolljt0725:

`arm64` image is already supported for a long time under `aarch64/` namespace on dockerhub and we 
have run docker containers for a long time on arm64.

### DIFF
--- a/config.md
+++ b/config.md
@@ -60,7 +60,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
 - **architecture** *string*, REQUIRED
 
   The CPU architecture which the binaries in this image are built to run on.
-  Possible values include: `386`, `amd64`, `arm`.
+  Possible values include: `386`, `amd64`, `arm`, `arm64`.
   More values may be supported in the future and any of these may or may not be supported by a given container runtime implementation.
   New entries SHOULD be submitted to this specification for standardization and be inspired by the [Go language documentation for $GOOS and $GOARCH](https://golang.org/doc/install/source#environment).
 


wking:
On Thu, Oct 13, 2016 at 01:27:47AM -0700, Lei Jitang wrote:
> `arm64` image is already supported for a long time under `aarch64/`
> namespace on dockerhub and we have run docker containers for a long
> time on arm64.

For what it's worth, the runtime-spec architecture setting is $GOARCH
plus local additions [1] (although there are currently no local
additions).  Is there a point to *not* supporting an existing $GOARCH
value?  This repo's current “Possible values include” wording is
open-ended, but it seems like it would be less work to either:

a. Delegate the enumeration of OSes and architectures to runtime-spec,
   or
b. Mirror runtime-spec in using $GOARCH plus local additions.

I'm personally in favor of (a), because I don't see a point in having
OSes or architectures that are only supported by the runtime-spec but
not the image-spec (or vice versa).  But if image-spec maintainers
don't trust runtime-spec maintainers to respond quickly enough when
new entries are needed, then (b) would be a reasonable choice.

[1]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#platform

coolljt0725:
@wking  I prefer (b), because the platform defined in https://github.com/opencontainers/image-spec/blob/master/manifest-list.md is

>This REQUIRED property describes the platform which the image in the manifest runs on. A full list of valid operating system and architecture values are listed in the Go language documentation for $GOOS and $GOARCH

ping @vbatts  @stevvooe  @philips  WDYT?

I'll update this if we agree to do so.

stevvooe:
LGTM

philips:
LGTM

jonboulle:
lgtm

PR #386, opened 2016-10-14
Use specs-go/v1 to reference media type instead of a contant string

coolljt0725:

### DIFF
--- a/schema/manifest_backwards_compatibility_test.go
+++ b/schema/manifest_backwards_compatibility_test.go
@@ -22,13 +22,14 @@ import (
 	"testing"
 
 	"github.com/opencontainers/image-spec/schema"
+	"github.com/opencontainers/image-spec/specs-go/v1"
 )
 
 var compatMap = map[string]string{
-	"application/vnd.docker.distribution.manifest.list.v2+json": "application/vnd.oci.image.manifest.list.v1+json",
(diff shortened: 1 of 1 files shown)


stevvooe:
LGTM

glestaris:
I could be misunderstanding this, but there may be a reason behind explicitely using literals instead of constants. Otherwise a regression in a constant may go unoticed. I didn't look deep tho so I could be wrong about this.

wking:
On Tue, Oct 18, 2016 at 03:18:04AM -0700, George Lestaris wrote:
> I could be misunderstanding this, but there may be a reason behind
> explicitely using literals instead of constants. Otherwise a
> regression in a constant may go unoticed. I didn't look deep tho so
> I could be wrong about this.

Can you describe a potential constant-regression workflow?  I'm
personally fine with using string literals instead of variable names,
but the only small difference I see in robustness is that the variable
names will be compared at compile time while the strings would only be
compared at runtime.  For a test suite, that's not a big deal (compile
time and runtime being so close together), but it certainly doesn't
seem like using variables would be a drop in robustness.

stevvooe:
@glestaris This is using constants on one side of the map and moves to use literals on the other. We can add further constants in the future, but this analysis is really outside the scope of this PR.

philips:
LGTM

PR #392, opened 2016-10-17
manifest-list: Drop HTTP Content-Type sentence

### LINKED ISSUE #390: Manifest list should not be identified by HTTP headers
The manifest-list spec says:

> A client will distinguish a manifest list from an image manifest based on the Content-Type returned in the HTTP response.

This is incorrect. A client will always discover a manifest list by following a descriptor in the `refs/` directory, and the descriptor will state the media type. The media type will unambiguously distinguish manifests from manifest lists. No part of the spec requires obtaining a manifest/manifest list over HTTP.

In the event that a manifest list is obtained over HTTP, any mismatch between the `Content-Type` in the HTTP response header and the `mediaType` in the corresponding descriptor should be resolved in the spec rather than being implementation-defined. I suggest that the descriptor's mediaType should win.

wking:
As Adrian Colley points out in #390, clients may not be aquiring the manifest over HTTP.  In most cases, folks will know (from a Content-Type header, a descriptor media type, or other means) what type of blob they're dealing with before they look at the blob.  I expect client behavior [like][2]:

> If you can verify the digest, ignore Content-Type. If you can't verify the digest,   respect the Content-Type and require it to match your expected media type (if any).

But in the absence of an HTTP-API spec I don't think we need to talk about this at all.

Fixes #390.

[1]: https://github.com/opencontainers/image-spec/issues/390
[2]: https://github.com/opencontainers/image-spec/issues/390#issuecomment-254029311

### DIFF
--- a/manifest-list.md
+++ b/manifest-list.md
@@ -2,7 +2,6 @@
 
 The manifest list is a higher-level manifest which points to specific [image manifests](manifest.md) for one or more platforms.
 While the use of a manifest list is OPTIONAL for image providers, image consumers SHOULD be prepared to process them.
-A client will distinguish a manifest list from an image manifest based on the Content-Type returned in the HTTP response.
 
 This section defines the `application/vnd.oci.image.manifest.list.v1+json` [media type](media-types.md).
 


jonboulle:
LGTM (though I'd also be happy to change this to a suggestion, e.g. "In the case of transferring images over HTTP clients COULD distinguish ..." or similar)

wking:
On Tue, Oct 18, 2016 at 07:46:35AM -0700, Jonathan Boulle wrote:
> (though I'd also be happy to change this to a suggestion, e.g. "In
> the case of transferring images over HTTP clients COULD distinguish
> ..." or similar)

I'm fine with that or the guidelines I quote in the topic post.
Either way, this sort of advice is not manifest-list specific.  We can
push it into media-types.md or the considerations.md that is in flight
with #340.  Would you like me to move it to media-types.md in this PR?

philips:
LGTM

I like @jonboulle's suggestion; lets do this in media-types.md

wking:
On Tue, Oct 18, 2016 at 09:07:21PM -0700, Brandon Philips wrote:
> I like @jonboulle's suggestion…

“COULD distinguish” isn't much guidance ;).  Do you have specific
concerns with the “If you can verify…” language I quote in [1]?  That
quote tries to set up the following behavior tree:

1. You know the media type ahead of time.
   a. The Content-Type matches your expected media type.
      Use your expected media type to select the appropriate blob
      parser / validator.
   b. The Content-Type does not match the expected media type.
      i. You know the digest ahead of time.
         Use your expected media type to select the appropriate blob
         parser / validator.  Possibly warn about the mismatch.
      ii. You don't know the digest ahead of time.
          Error out citing the mismatch.
2. You don't know the media type ahead of time.
   Use Content-Type to select the appropriate blob parser / validator.

Would you change any of those?  Recommend a different tree?  Just
rephrase the wording for that tree?

[1]: https://github.com/opencontainers/image-spec/pull/392#issue-183555915

jonboulle:
> Do you have specific
> concerns with the “If you can verify…” language I quote in [1]? 

This seems fine. Please file a followup PR.

PR #397, opened 2016-10-19
config|manifest-list: Punt to runtime-spec for platform OS/arch

wking:
The new wording makes it clear that implementations are free to support only a subset of OSes or architectures, and both implementations and blobs are free to support/use OSes and architectures not listed in runtime-spec.  Leaning on runtime-spec here lets us consolidate any OCI `GOOS`/`GOARCH` extension in a single location (runtime-spec).

Spun off from [this comment][1] in #385.

[1]: https://github.com/opencontainers/image-spec/pull/385#issuecomment-253640211


philips:
LGTM

vbatts:
LGTM

PR #404, opened 2016-10-20
schema: Limit schemaVersion to 2 and enumerate mediaType

wking:
For manifests and manifest lists, to match the requirements from their Markdown definitions.

### DIFF
--- a/schema/fs.go
+++ b/schema/fs.go
@@ -205,7 +205,7 @@ var _escData = map[string]*_escFile{
 	"/config-schema.json": {
 		local:   "config-schema.json",
 		size:    774,
-		modtime: 1478057674,
+		modtime: 1478292182,
 		compressed: `
 H4sIAAAJbogA/5SRvW7rMAyFdz+F4WS8ju7QKWsfoEPHooMqUzEDWFRJZgiKvHv1EzcxEBTuEsSH/M4R
 ya+mbbsBxDFGRQrdvu1eIoRnCmoxALfpn8dD+xrBoUdnS9e/jG3FjTDZjIyqcW/MUSj0Vd0RH8zA1mv/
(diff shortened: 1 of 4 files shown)


xiekeyang:
@wking
1. Do you forget to refresh the generated virtual file system, to see https://github.com/opencontainers/image-spec/blob/master/HACKING.md#virtual-schema-httpfilesystem?
2. It seems that you want to reject all media types in [compatibility-matrix](https://github.com/opencontainers/image-spec/blob/master/media-types.md#compatibility-matrix), that may be conflict to our previous conclusion. I worry that we fall into the confusion again that media type should be loose or strict. After read #406 , I still think this PR is too strict.

wking:
On Thu, Oct 20, 2016 at 09:48:41PM -0700, xiekeyang wrote:
> 1. Do you forget to refresh the generated virtual file system…

Oops, yeah.  Fixed with the just-pushed 00f177b.

> 2. It seems that you want to reject all media types in compatibility-matrix…

I think there are two things going on here: what you are and what you reference.  #304 is about who you reference, and the pattern is that we require implementations to support appropriate OCI types but allow images and implementations to use other types as well.  *This* PR, on the other hand, is about declaring what you are, and there's only one good answer for that.

An `application/vnd.oci.image.manifest.v1+json` with `application/vnd.docker.container.image.v1+json` in `config.mediaType` is legal (although there may be compliant runtimes that don't support it).  An `application/vnd.oci.image.manifest.v1+json` with `application/vnd.docker.distribution.manifest.v2+json` in `mediaType` is *not* legal.

wking:
I've just pushed an additional commit (801699e) fixing the `convertFormats` direction to [make Travis happy][1], so this stricter checking has already been useful :).

[1]: https://travis-ci.org/opencontainers/image-spec/builds/169440828#L335

vbatts:
needs a rebase

wking:
On Tue, Nov 01, 2016 at 07:51:41AM -0700, Vincent Batts wrote:
> needs a rebase

Rebased around #413's schema/fs.go bump with 801699e → c23165c.

philips:
LGTM

Conflict though.

wking:
On Wed, Nov 16, 2016 at 02:38:58PM -0800, Brandon Philips wrote:
> Conflict though.

Rebased onto master and rebuilt schema/fs.go with c23165c → dad01a1.

It's probably best if you order up all the JSON-Schema-touching PRs
(maybe tag them all?), because *all* of them are going to conflict on
fs.go.

An alternative would be to ask contributors to *not* update fs.go and
have the merging maintainer rebuild fs.go after merging the change.
In this case you'd also want to add a Travis line rebuilding fs.go
before running the tests.

philips:
LGTM

ptal @opencontainers/image-spec-maintainers

stevvooe:
LGTM

PR #408, opened 2016-10-20
manifest: Explicitly unspecified attributes for the initial layer directory

wking:
Point out that the layer's initial empty directory has unspecified attributes.  Layer authors interested in the ownership and other attributes of the unpacked root directory should explicitly set an entry for it.

Spun off from #318 to focus on a single point.


stevvooe:
LGTM

vbatts:
LGTM

PR #410, opened 2016-10-21
Some omitempty usage modifications

q384566678:
Optional fields should be added omitempty 

### DIFF
--- a/specs-go/v1/config.go
+++ b/specs-go/v1/config.go
@@ -17,34 +17,34 @@ package v1
 // ImageConfig defines the execution parameters which should be used as a base when running a container using an image.
 type ImageConfig struct {
 	// User defines the username or UID which the process in the container should run as.
-	User string `json:"User"`
+	User string `json:"User,omitempty"`
 
 	// Memory defines the memory limit.
-	Memory int64 `json:"Memory"`
+	Memory int64 `json:"Memory,omitempty"`
 
 	// MemorySwap defines the total memory usage limit (memory + swap).
-	MemorySwap int64 `json:"MemorySwap"`
+	MemorySwap int64 `json:"MemorySwap,omitempty"`
 
 	// CPUShares is the CPU shares (relative weight vs. other containers).
-	CPUShares int64 `json:"CpuShares"`
+	CPUShares int64 `json:"CpuShares,omitempty"`
 
 	// ExposedPorts a set of ports to expose from a container running this image.
-	ExposedPorts map[string]struct{} `json:"ExposedPorts"`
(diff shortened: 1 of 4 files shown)


specs-go/v1/config.go
```diff
@@ -17,34 +17,34 @@ package v1
 // ImageConfig defines the execution parameters which should be used as a base when running a container using an image.
 type ImageConfig struct {
 	// User defines the username or UID which the process in the container should run as.
-	User string `json:"User"`
+	User string `json:"User,omitepty"`
```
wking: “omitepty” → “omitempty”.

wking:
Besides the typos, 5a318bf7 looks good to me.

philips:
LGTM

vbatts:
LGTM

stevvooe:
`omitempty` has nothing to do with REQUIRED vs OPTIONAL.

Please revert this change.

q384566678:
@stevvooe I have already deleted this branch, Can I open another PR to repair?

q384566678:
@stevvooe In addition, after reading your discussion, I am not very clear when `omitempty` should be used, that is, why do you think this change is wrong, I did not see the specific provisions.

stevvooe:
`omitempty` influences when Go `json` package emits values for zero-valued fields. Please read https://golang.org/pkg/encoding/json/. It should not be used to try to comply with the required and optional provisions of the specification, as simply doesn't do that.

Typically, `omitempty` should be used on every field unless you want it to emit a zero-value. An example might be if you want to have an empty slice be emitted, rather than omitting the field. I don't think there are any cases in OCI where it make sense to not have `omitempty`.

q384566678:
@stevvooe So, would I need to reopen a PR to revert it? Because the branch has been deleted.

wking:
On Mon, Feb 06, 2017 at 07:43:01PM -0800, Stephen Day wrote:
> Typically, `omitempty` should be used on every field unless you want
> it to emit a zero-value. An example might be if you want to have an
> empty slice be emitted, rather than omitting the field. I don't
> think there are any cases in OCI where it make sense to not have
> `omitempty`.

The manifest-list ‘manifests’ entry is currently explicitly REQUIRED
and it explicitly allows a zero-length array [1].  I'm not sure why we
don't just make it OPTIONAL (less work for the same amount of required
information), but as the spec stands it would not be a good idea to
make ManifestList.Manifests omitempty (as I pointed out in [2]).

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc4/manifest-list.md#L23-L24
[2]: https://github.com/opencontainers/image-spec/issues/557#issuecomment-277839297

Issue #24, opened 2016-04-14
Any chance of changing the whiteout file approach


timthelion:
It seems unfortunate that a new standard should use a method which unecesarilly limits the standard. With the .wh file approach, base images can no longer contain arbitrary data. This means, for example, that you cannot have an image with example OCI image-spec data stored in it. Is there any possibility of changing this to use, for example, a white out list instead. So that the files that made a layer would be:

````
VERSION
layer.tar
whiteouts
json
````

That would mean that truely arbitrary data could be stored in the images, which would be really nice :)

vbatts:
To be clear, you are saying that the .wh approach is limited? And the
whiteout file is preferred?

If so, I agree.

timthelion:
Creating a list of files to be whited out, and keeping that outside of `layer.tar` is better than putting `.wh` files in `layer.tar`. We wouldn't want a situation like this: http://git-annex.branchable.com/forum/Storing_git_repos_in_git-annex/

vbatts:
I wholly agree and intend to see it done as a whiteout file list.

philips:
I think this is something we should think about for post v1.0.0. I whole heartedly agree but in the ideal case the initial v1 serialization spec is binary compatible with the existing docker serialization to save the sanity of all of the existing registry folks like acr, gcr.io, Quay, Hub, etc.

timthelion:
Would it be possible to use the whiteout list file IFF a whiteout list file exists, and otherwise use the `.wh` approach?

philips:
@timthelion Yes, that would be the right way of approaching it. It would be a schema bump which would be a version break. Happy to consider adding this feature to fix the issue but I do want to hold off until we get post v1.0.0 (in a couple of months).

timthelion:
@philips so basically, you want to have the 1.0 release be supported by Docker/CoreOS/whatever, without actually haven't to change Docker/CoreOS/whatever's code? Basically, there will be no actual technical changes to the spec before 1.0?

cyphar:
@timthelion No *breaking* technical changes.

wking:
On Thu, Apr 14, 2016 at 02:26:16PM -0700, Timothy Hobbs wrote:
> Creating a list of files to be whited out, and keeping that outside
> of `layer.tar` is better than putting `.wh` files in `layer.tar`.

If we don't mind picking up a non-standard tar entry, star
(Schilling's tar) uses pax extension headers and defines
[...]

PR #411, opened 2016-10-21
specs-go: clarify mediatypes

vbatts:
Due to the conflicting use of the `mediatType` field across                                   
    documents, and after discussion on                                                          
    https://github.com/opencontainers/image-spec/pull/411,                                                
    this changeset removes the use of `mediaType` where it is used to refers                                                     
    to a document's own type. Leaving only the use of `mediaType` for                                      
    descriptors, where it is used to describe the type of a referenced object.

### DIFF
--- a/manifest-list.md
+++ b/manifest-list.md
@@ -4,6 +4,7 @@ The manifest list is a higher-level manifest which points to specific [image man
 While the use of a manifest list is OPTIONAL for image providers, image consumers SHOULD be prepared to process them.
 
 This section defines the `application/vnd.oci.image.manifest.list.v1+json` [media type](media-types.md).
+For the media type(s) that this document is compatible with, see the [matrix][matrix].
 
 ## *Manifest List* Property Descriptions
 
@@ -14,9 +15,8 @@ This section defines the `application/vnd.oci.image.manifest.list.v1+json` [medi
 
 - **`mediaType`** *string*
 
-  This REQUIRED property contains the media type of the manifest list.
-  For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.list.v1+json`.
-  For the media type(s) that this is compatible with, see the [matrix](media-types.md#compatibility-matrix).
+  This property is *reserved* for use, to [maintain compatibility][matrix].
+  When used, this field contains the media type of this document, which differs from the [descriptor](descriptor.md#properties) use of `mediaType`.
 
 - **`manifests`** *array of objects*
 
@@ -119,3 +119,4 @@ Instead they MUST ignore unknown properties.
 ```
 
 [runtime-platform2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#platform
+[matrix]: media-types.md#compatibility-matrix
--- a/specs-go/v1/config.go
+++ b/specs-go/v1/config.go
@@ -78,6 +78,7 @@ type History struct {
 }
 
 // Image is the JSON structure which describes some basic information about the image.
+// This provides the `application/vnd.oci.image.config.v1+json` mediatype when marshalled to JSON.
 type Image struct {
 	// Created defines an ISO-8601 formatted combined date and time at which the image was created.
 	Created string `json:"created,omitempty"`
--- a/specs-go/v1/descriptor.go
+++ b/specs-go/v1/descriptor.go
@@ -15,9 +15,10 @@
 package v1
 
 // Descriptor describes the disposition of targeted content.
+// This structure provides `application/vnd.oci.descriptor.v1+json` mediatype when marshalled to JSON
 type Descriptor struct {
-	// MediaType contains the MIME type of the referenced object.
-	MediaType string `json:"mediaType"`
+	// MediaType is the media type of the object this schema refers to.
+	MediaType string `json:"mediaType,omitempty"`
 
 	// Digest is the digest of the targeted content.
 	Digest string `json:"digest"`
--- a/specs-go/v1/manifest_list.go
+++ b/specs-go/v1/manifest_list.go
@@ -50,7 +50,8 @@ type ManifestDescriptor struct {
 	Platform Platform `json:"platform"`
 }
 
-// ManifestList  references manifests for various platforms.
+// ManifestList references manifests for various platforms.
+// This structure provides `application/vnd.oci.image.manifest.list.v1+json` mediatype when marshalled to JSON.
 type ManifestList struct {
 	specs.Versioned
 
--- a/image-layout.md
+++ b/image-layout.md
@@ -62,7 +62,7 @@ The blobs directory MAY be missing referenced blobs, in which case the missing b
 
 No semantic restriction is given for object names in the `refs` subdirectory.
 Each object in the `refs` subdirectory MUST be of type `application/vnd.oci.descriptor.v1+json`.
-In general the `mediatype` of this [descriptor][descriptors] object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
+In general the `mediaType` of this [descriptor][descriptors] object will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json` although future versions of the spec MAY use a different mediatype.
 
 **Implementor's Note:**
 A common use case of refs is representing "tags" for a container image.
@@ -85,7 +85,6 @@ $ cat ./refs/v1.0 | jq
 $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f | jq
 {
   "schemaVersion": 2,
-  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
   "manifests": [
     {
       "mediaType": "application/vnd.oci.image.manifest.v1+json",
@@ -103,7 +102,6 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 $ cat ./blobs/sha256/afff3924849e458c5ef237db5f89539274d5e609db5db935ed3959c90f1f2d51 | jq
 {
   "schemaVersion": 2,
-  "mediaType": "application/vnd.oci.image.manifest.v1+json",
   "config": [
     "mediaType": "application/vnd.oci.image.config.v1+json",
     "size": 7023,
--- a/manifest.md
+++ b/manifest.md
@@ -7,6 +7,7 @@ In OCI, this is codified in a [Manifest List](manifest-list.md).
 The third goal is to be translatable to the [OCI Runtime Specification](https://github.com/opencontainers/runtime-spec).
 
 This section defines the `application/vnd.oci.image.manifest.v1+json` [media type](media-types.md).
+For the media type(s) that this is compatible with see the [matrix](media-types.md#compatibility-matrix).
 
 # Image Manifest
 
@@ -21,9 +22,8 @@ Unlike the [Manifest List](manifest-list.md), which contains information about a
 
 - **`mediaType`** *string*
 
-    This REQUIRED property contains the media type of the image manifest.
-    For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.v1+json`.
-    For the media type(s) that this is compatible with see the [matrix](media-types.md#compatibility-matrix).
+  This property is *reserved* for use, to [maintain compatibility][matrix].
+  When used, this field contains the media type of this document, which differs from the [descriptor](descriptor.md#properties) use of `mediaType`.
 
 - **`config`** *[descriptor](descriptor.md)*
 
--- a/schema/image-manifest-schema.json
+++ b/schema/image-manifest-schema.json
@@ -11,13 +11,6 @@
       "minimum": 2,
       "maximum": 2
     },
-    "mediaType": {
-      "id": "https://opencontainers.org/schema/image/manifest/mediaType",
-      "type": "string",
-      "enum": [
-        "application/vnd.oci.image.manifest.v1+json"
-      ]
-    },
     "config": {
       "$ref": "content-descriptor.json"
     },
@@ -35,7 +28,6 @@
   },
   "required": [
     "schemaVersion",
-    "mediaType",
     "config",
     "layers"
   ]
--- a/schema/manifest-list-schema.json
+++ b/schema/manifest-list-schema.json
@@ -11,13 +11,6 @@
       "minimum": 2,
       "maximum": 2
     },
-    "mediaType": {
-      "id": "https://opencontainers.org/schema/image/manifest-list/mediaType",
-      "type": "string",
-      "enum": [
-        "application/vnd.oci.image.manifest.list.v1+json"
-      ]
-    },
     "manifests": {
       "type": "array",
       "items": {
@@ -31,7 +24,6 @@
   },
   "required": [
     "schemaVersion",
-    "mediaType",
     "manifests"
   ]
 }
--- a/specs-go/v1/manifest.go
+++ b/specs-go/v1/manifest.go
@@ -16,7 +16,7 @@ package v1
 
 import "github.com/opencontainers/image-spec/specs-go"
 
-// Manifest defines a schema2 manifest
+// Manifest provides `application/vnd.oci.image.manifest.list.v1+json` mediatype structure when marshalled to JSON.
 type Manifest struct {
 	specs.Versioned
 
--- a/specs-go/versioned.go
+++ b/specs-go/versioned.go
@@ -20,7 +20,4 @@ package specs
 type Versioned struct {
 	// SchemaVersion is the image manifest schema that this image follows
 	SchemaVersion int `json:"schemaVersion"`
-
-	// MediaType is the media type of this schema.
-	MediaType string `json:"mediaType"`
 }


specs-go/v1/config.go
```diff
@@ -75,6 +75,7 @@ type History struct {
 }
 
 // Image is the JSON structure which describes some basic information about the image.
+// This provides the `application/vnd.oci.image.config.v1+json` mediatype when marshalled to JSON.
```
vbatts: i was thinking the same. Especially since the field is omitEmpty for JSON. Perhaps make it *optional* with a SHOULD, so it will have compat with older docker? Thoughts @stevvooe ?

vbatts: (to that end, also the `oci-layout` file, and it doesn't even have a mediatype assigned for it)

specs-go/v1/manifest_list.go
```diff
@@ -50,8 +50,10 @@ type ManifestDescriptor struct {
 	Platform Platform `json:"platform"`
 }
 
-// ManifestList  references manifests for various platforms.
+// ManifestList references manifests for various platforms.
+// This structure provides `application/vnd.oci.image.manifest.list.v1+json` mediatype when marshalled to JSON.
 type ManifestList struct {
+	specs.MediaTyped
 	specs.Versioned
```
wking: If `Versioned` contains `MediaTyped` (which is what you currently have), isn't it redundant to list `MediaTyped` here?

vbatts: yes, technically. Does not change anything and is very apparent that it is media typed

jonboulle:
Why did we ever do this?

specs-go/v1/descriptor.go
```diff
@@ -14,10 +14,12 @@
...
 package v1
 
+import specs "github.com/opencontainers/image-spec/specs-go"
+
 // Descriptor describes the disposition of targeted content.
+// This structure provides `application/vnd.oci.descriptor.v1+json` mediatype when marshalled to JSON
 type Descriptor struct {
-	// MediaType contains the MIME type of the referenced object.
-	MediaType string `json:"mediaType"`
+	specs.MediaTyped
```
jonboulle: oh, so this is actually wrong because the mediatype here is the REFERENCED object not itself.

jonboulle:
@vbatts can we rather remove media type fields?

vbatts:
@jonboulle like completely? or just this PR?

jonboulle:
@vbatts like completely

vbatts:
well, honestly I like the self describing nature of it. I wish the descriptor field names could be fixed, such that it could self describe and have a separate field that describes the object pointing to.
Self-describing, while this approach is slightly more expensive, allows for an algorithm like
```golang 
mt := v1.MediaTyped{}
if err := json.Marshal(buf, &mt); err != nil {
  // handle yer biz
}
switch mt.MediaType {
case v1.MediaTypeImageManifest:
  // something specific
default:
  // handle yer biz
}
```
So we can have discovery without having a `map[string]interface{}{}`.

wking:
On Fri, Oct 28, 2016 at 07:07:03AM -0700, Vincent Batts wrote:
> well, honestly I like the self describing nature of it.

There was a bit more discussion of this in
opencontainers/image-tools#5 starting with [1].  @cyphar was arguing
for this peek-inside auto-detection in the proposed oci-refs because
tracking media types is tedious.  I was arguing against peek-inside
auto-detection because it doesn't always work (it requires a JSON blob
whose mediaType is self-describing, so not the current descriptors or
any non-JSON type, e.g. a layer tarball).

We SHOULD descriptors for all blob references [2], so if the initial
publisher references the blob they published (and presumably they
would or future garbage collection would remove it), that descriptor
reference will have media type information for the blob.  Anyone else
who stumbles across the same blob either:

a. Created it independently (in which case they should know what it is),
b. Found it by following a descriptor (in which case that descriptor
   tells them what it is), or
c. Found it by walking the CAS (e.g. during an interal GC pass).

(c) is the only tricky case, and the solution is to have the GC engine
walk the CAS from pinned descriptors (so it's like (c)) instead of
walking the CAS via readdir(3) on the blob filesystem.

[1]: https://github.com/opencontainers/image-tools/pull/5#issuecomment-256353980
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc1/descriptor.md#L10

cyphar:
My main concern is that without `mediaType` things are properly self-describing -- you have no clue what each object represents unless you've walked from a `ref` to the object itself. For me, this is just tedious to deal with (as a consumer I shouldn't be caring about the internals of all of the referencing if there's `oci-image-tools` that should handle this for me and just output the JSON blob that I am trying to look at).

@wking gave the example that `mediaType` won't on the layer tar+gzip archives. But that isn't helping the case against `mediaType`, because both **tar** and **gzip** have **unique headers** that identify what they are! There's no ambiguity with regards to whether a layer object is a layer object or not. The only thing that's ambiguous is the JSON blobs we're using which aren't adequately expressing what they represent.

In particular, the fact that descriptors don't say that they're descriptors is quite silly (`mediaType` is broken in that instance).

wking:
The tar+gzip headers do not distinguish `application/vnd.oci.image.layer.v1.tar+gzip` from `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip` or from a [future `application/vnd.oci.image.layer.v1.1.tar`][1].

Walking the DAG sounds like an integral part of using CAS.  Can you provide more details about what you mean by "just output the JSON blob that I am trying to look at"?

[1]: https://github.com/opencontainers/image-spec/issues/24#issuecomment-254359960

cyphar:
> Walking the DAG sounds like an integral part of using CAS. Can you provide more details about what you mean by "just output the JSON blob that I am trying to look at"?

Lets say I want to modify the config for an image. Currently the way I need to do this is:

1. Get the reference descriptor using `oci-refs get`. I then have to parse this JSON to get the right blob for the manifest.
2. Get the manifest using `oci-cas get`. I then have to parse this JSON to get the right blob for the config.
3. I now have the config JSON and I can modify it using `jq` or whatever.

In order to update the image I now need to:

1. Push the new modified config and get the new digest, which I do with `oci-cas put`.
2. Update the manifest to use the new config, then push it to get a new digest with `oci-cas put`.
3. Create a new reference descriptor and push it to get the new reference with `oci-refs put`.
4. Clean everything up with the (currently non-existant) `oci-gc`.

Now, my main point is that **the OCI tooling hasn't added anything**. The process takes precisely as long as it would using `sha256sum`, `jq`, `ls`, `cat` and `rm`! The fact that **so many** of the things need to be done manually even with having tooling just makes this ridiculous. Imagine if `oci-create-runtime-bundle` or `oci-unpack` would only let you extract a single layer at a time -- that would also be useless.

I'm going to end up writing a wrapper for all of this, but I think we should be considering who the **target user** of these utilities is. Looking at `oci-image-init`, `oci-unpack`, `oci-validate` and `oci-create-runtime-bundle` I would assume that the target is people who just want to use images without having to understand the intricacies of the specification and how things need to have 15 layers of referencing and double-checking of sizes and so on. But maybe I was mistaken and the intended users are just us? In which case I'll be happy to just go and write the tooling I need and maintain it outside of this project -- it's just a bit of a shame that we're not concrete on **who** will be using this tooling and what they would want from such tooling.

wking:
> The process takes precisely as long as it would using sha256sum, jq, ls, cat and rm!

But ls, cat, and rm don't abstract away the CAS registry implementation.  And the presence of low-level tools like oci-cas doesn't preclude additional tooling like oci-jq-manifest (or whatever your high-level tweaker looks like).  In fact, oci-cas makes it much easier to write a CAS-engine agnostic oci-jq-manifest.

I'm missing the connection between "I'll wrap DAG walking because it's tedious" and "I need to discover media types by inspecting the blob content".  In your example above you start with `oci-refs get`, so you'll be following descriptors (which tell you the media type before you have the blob content) all the way down.

cyphar:
@wking My point will be more clear if I just make my wrapping tool. I'm working on it now, but in order to handle it properly I need to merge this PR into my branch -- so we can just continue with merging this (and we can discuss higher-level handlers later).

wking:
My point is that I see no need for peek-inside type detection.  But the `mediaType` properties (excepting Descriptor's) will be the same amount of work to remove regardless of this PR, so I don't have a problem with it landing.

cyphar:
@wking ... You can't have it both ways. In opencontainers/image-tools#5 you want to make the tooling around images as flexible as possible (so that you can support extensions that don't match the spec). That's fine, but you can't then also say that all of the objects should **no longer be self-describing**. The result is that you cannot use **any** generic OCI tooling for images, because all of the objects are defined relative to one another and thus you couldn't effectively modify a single object (you'd have to parse the image each time to know what object you're modifying -- and even then you'd need to track the object types implicitly).

Explicit is better than implicit.

wking:
It's a Merkle DAG, you can *never* modify a single blob.  Peek-inside type detection doesn't change that.

wking:
Even if you're modifying the root blob, you'd still need to push a new ref.

stevvooe:
@cyphar The model for CAS-based engine distribution typically requires referencing an object through type-qualified references, which are called descriptors. I disagree that using CAS in this manner prevents extension, as I don't think OCI could be introduced at all if the concepts defined in manifests and manifest lists didn't work. We have been doing this in compilers and runtime since forever. 

The difference is the propagation of the modification. Typically, the process of modifying something requires a walk don't the qualified reference path, then anything that was touched or references something that was touched gets update. This is what happens with this technology (I also suspect this is part of why CAS is not adopted more often). When you touch something, it modifies the "address" of that thing and those changes and their references need to be updated.

I don't know what is going on in opencontainers/image-tools#5, but if CAS access to a tar file is being exposed over CLI, something is going way wrong.

In general, I'd like to see a proposal for the image-tools that are useful. I am not familiar with your work, but it would be helpful if you could identify the problems that have led to a "wrapper" or even how we can incorporate that work into OCI.

wking:
On Mon, Oct 31, 2016 at 02:22:54PM -0700, Stephen Day wrote:
> I don't know what is going on in opencontainers/image-tools#5, but
> if CAS access to a tar file is being exposed over CLI, something is
> going way wrong.

I disagree, but I think we should probably be discussing this point in
opencontainers/image-tools#5.  Can you file your concerns in more
detail over there?

> In general, I'd like to see a proposal for the image-tools that are
> useful. I am not familiar with your work, but it would be helpful if
> you could identify the problems that have led to a "wrapper" or even
> how we can incorporate that work into OCI.

+1.

stevvooe:
> Can you file your concerns in more
detail over there?

I did, then you just opened three more PRs. I have no more time to waste.

cyphar:
@stevvooe Okay, so here's what I'm working on at the moment (the only reason I haven't published it yet is just because right now I'm doing a bunch of go-mtree related work rather than actually getting `imagectl` to work at the moment):

The main issue I have with the current incarnation of @wking's `oci-cas` and `oci-refs` is that it doesn't appear to me as though the tooling was written with the image-spec in mind. In particular, it looks more like "generic CAS tooling" that just happens to handle the CAS layout of the OCI image spec.

The tooling I would like (and am planning on writing) actually is written around the spec. In particular, it lets you modify **images** rather than blobs. So you can say "modify this field in the config of this image ref" and that will translate to replacing the config, the manifest blob that referenced it and the ref blob that referenced that. Same thing applies to changing the set of layers and so on -- the tooling is actually written around the end-user goal of *actually modifying images as a first-class concept*.

And in my case, `imagectl` already knows the types of everything implicitly (because I had to walk through the CAS to get there) -- though that precludes a descriptor telling me what type it points to (or every blob being self-describing). However, the main issue I have with @wking's PR (aside from not being sure why we're exposing `tar` handling code) is that it exposes a generic CAS interface that isn't actually helpful to anyone unless we're also planning on writing some more tooling that can take our JSON blobs and do useful things with them -- and to do that you need to have self-describing blobs.

wking:
On Mon, Oct 31, 2016 at 08:22:31PM -0700, Aleksa Sarai wrote:
> The main issue I have with the current incarnation of @wking's
> `oci-cas` and `oci-refs` is that it doesn't appear to me as though
> the tooling was written with the image-spec in mind. In particular,
> it looks more like "generic CAS tooling" that just happens to handle
> the CAS layout of the OCI image spec.

oci-cas is generic CAS tooling.  oci-refs is generic mutable-refs
tooling.  They're written with image-spec in mind because image-layout
is a generic CAS/refs format (although it's obviously not the only
possible CAS or refs format).  The idea in putting a layer of
abstraction between the image-layout tooling and the rest of the
image-spec support is that it lets you adjust image-layout or swap in
some completely different CAS/refs engine without rewriting the
domain-specific logic.

> The tooling I would like (and am planning on writing) actually is
> written around the spec. In particular, it lets you modify
> **images** rather than blobs. So you can say "modify this field in
> the config of this image ref" and that will translate to replacing
> the config, the manifest blob that referenced it and the ref blob
> that referenced that. Same thing applies to changing the set of
> layers and so on -- the tooling is actually written around the
> end-user goal of *actually modifying images as a first-class
> concept*.

This sounds useful to me too.  I don't see why having this
higher-level code and command-line interface makes image-cas or
image-refs a bad idea.

> And in my case, `imagectl` already knows the types of everything
> implicitly (because I had to walk through the CAS to get there) --
> though that precludes a descriptor telling me what type it points to
> (or every blob being self-describing).

So your imagectl doesn't need to self-describing blobs or a type-aware
CAS engine?  I'm glad, although not particularly surprised, since I
expect most CAS consumers will be publishing content with known types
or walking down from a ref.

> However, the main issue I have with @wking's PR (aside from not
> being sure why we're exposing `tar` handling code)…

It's not just exposing tar-handling code, it's exposing a generic
CAS/refs handling API.  One backend for those CAS and ref stores is
tar, but it also supports directories, and can be extended in the
future to support zip, HTTP, FTP, rsync, etc. [1].  It's also possible
that it could be extended to support the Docker registry and other
existing CAS/ref stores.

> … is that it exposes a generic CAS interface that isn't actually
> helpful to anyone unless we're also planning on writing some more
> tooling that can take our JSON blobs and do useful things with them…

I completely agree that we don't want oci-cas and oci-refs to be our
highest-level commands.  oci-image-validate, oci-unpack, and
oci-create-runtime-bundle are already higher-level commands.  Your
imagectl will be another higher-level command.  But the CAS and refs
implementations decouple cleanly from the higher-level,
domain-specific logic.  I don't think the presence of the high-level
commands is a reason to *not* expose the low-level CAS/refs API on the
command-line.

> … and to do that you need to have self-describing blobs.

I'm still not understanding this leap.  Can you present a workflow
that requires self-describing blobs?

[1]: https://github.com/opencontainers/image-spec/blame/5dbe4a342b6de21149f5c9f23c13b62ec1503082/image-layout.md#L4

wking:
On Mon, Oct 31, 2016 at 09:45:02PM -0700, W. Trevor King wrote:
> > … and to do that you need to have self-describing blobs.
> 
> I'm still not understanding this leap.  Can you present a workflow
> that requires self-describing blobs?

The only situation I've seen where self-describing blobs is useful is
outside-of-CAS validation, where we currently attempt to auto-detect
the type when the user doesn't pass it through to us [1].  That is
useful, but it doesn't involve the CAS engine.  And I don't think it
is *so* useful that we need to preserve it going forwards.  Do we
expect folks to frequently validate content whose type they don't
know?  “Is this something that OCI can autodetect and validate?”
seems like a much less useful question to ask than “Is this a valid
application/vnd.oci.image.manifest.v1+json?”.  But still,
autodetection for out-of-CAS validation was useful enough for folks to
add validation support for it.  Is that a sufficient use-case for
extending self-describing media types to
application/vnd.oci.image.config.v1+json?  How about
application/vnd.oci.descriptor.v1+json or
application/vnd.oci.image.layer.v1.tar+gzip?  I don't think so, but I
consider it to be a policy question and not a technical question
(folks *do* know what they're validating, they just don't want to type
out the media type), so I'm comfortable with image-spec maintainers
ruling either way.

Are there other use-cases for self-describing blobs that do make this
a technical requirement?

[1]: https://github.com/opencontainers/image-tools/blob/0a1e5dc0c9b67619d46bd8cbcc948bda1b143b27/cmd/oci-image-validate/main.go#L148-L152

stevvooe:
@cyphar I completely agree. CAS is too-level to be useful. Thank you for the break down!

https://github.com/docker/docker/pull/27455 is a proposal to docker to add a manifest tool which has the same UX problem. Basically, you need to "check out" an image, modify it, then check it back in.

In the spirt of OCI being made up of proven technologies, I think it would be worthwhile to let `imagectl` inform the work in image tools, if possible.

specs-go/v1/config.go
vbatts: @stevvooe woah. The `mediatype` is as necessary as a document providing the version or schema that it represents. Why have any of them, if this is the case?
Regardless, this leaves a mess for implementations to require a big case statement for reflect and discovery what kind of document and version is even being dealt with.

vbatts:
lots of philosophy and bike sheds, it seems to me. Is the motivation to remove `mediatype` from all but the descriptor, @stevvooe @jonboulle ?

specs-go/v1/config.go
wking: > The mediatype is as necessary as a document providing the version or schema that it represents.

Agreed.  But if you know the media type *from the descriptor*, why repeat it in the blob itself?  The only case I've seen is for outside-of-CAS reference like [verification][1], where the caller presumably knows the media type they're trying to verify but doesn't feel like typing it out.  I'm ok with providing a self-describing `mediaType` to save them that effort, but don't feel like we have to provide it.

[1]: https://github.com/opencontainers/image-spec/pull/411#issuecomment-257489701

wking: > `cat blob | some-tool --add-option` won't work properly because you cannot deal with multiple versions and also error handling

But:

```
$ DESCRIPTOR="$(oci-refs get some-image v1.0)"
$ DIGEST="$(echo "${DESCRIPTOR}" | jq .digest)"
$ MEDIA_TYPE="$(echo "${DESCRIPTOR}" | jq .mediaType)"
$ oci-cas get some-image "${DIGEST}" | some-tool --add-option --media-type "${MEDIA_TYPE}"
```

works fine.  And the only case I can think of where you *aren't* getting to the digest by walking down from a ref is still validating a blob before you push it into CAS.  And there you presumably know the type, and just want to be saved from typing it out on the command line (and as I've said, I'm ok keeping a self-describing `mediaType` if/where we want to support this).

> And since oci-refs and oci-cas appear to be on the path to "images are just a CAS" you need to have that.

I think “images can be stored in CAS as a Merkle tree” is true, but I don't think they're *just* a set of Merkle blobs.  Otherwise we could have written `descriptor.md` and `image-layout.md` and skipped `manifest.md`, `manifest-list.md`, `layer.md`, etc.  CAS (and refs) are one level in the implementation, but there are additional, higher levels with more domain-specific spec.  I'd rather have image-tools provide APIs and tools for handling images at all of these levels.

cyphar: > `oci-cas get some-image "${DIGEST}" | some-tool --add-option --media-type "${MEDIA_TYPE}"`

IMO that's ridiculous. You now need to carry around the media type of an object separate to the object. Why? What is the benefit? Why not just follow along with literally every other format that exists and make it self-describing? Why has this discussion been going on for so long?

cyphar: @wking You're arguing that there's dichotomy between "free for all, no need to have `mediaType`" and "hueristic so that we can recognise every blob type". I don't think there is one.

The benefit of being able to know *what a JSON blob is meant to represent* is entirely separate from "I can tell what every blob in the image is without references". If you're not happy with detecting `tar` files (like `file` and `libmagic` do) that's fine. But please let's not make all of our JSON objects meaningless blobs that require jumping through references in order to even understand what we're looking at (or keeping the type information out-of-band).

wking: > But please let's not make all of our JSON objects meaningless blobs that require jumping through references in order to even understand what we're looking at…

You're saying “assuming (for some out-of-band reason) that the blob is a JSON object which contains a self-describing `mediaType` field, we can use that `mediaType` field to unambiguously identify the content”.  That initial assumption is what I'm worried about.  If you see cases where you are comfortable making that assumption (for whatever out-of-band reasons), then great, use peek-inside type detection based on the `mediaType` value.  But I'd strongly recommend consumers use the referencing descriptor's `mediaType` to avoid having to rely on that assumption.

cyphar: > But I'd strongly recommend consumers use the referencing descriptor's mediaType to avoid having to rely on that assumption.

Your consistent implication that all consumers will have access to the entire image is getting annoying. If I have a tool like `oci-do-something` which I pipe a JSON object to, I don't expect that it will be reading the repository. In fact, I might have a service that modifies the JSON objects (and therefore actually cannot access the original repo). So you can't "use the referencing descriptor" because there **isn't** one.

Now, you might argue that we should send the out-of-band media type with it. But why should that be a requirement? What are you gaining by removing `mediaType`?

wking: > Now, you might argue that we should send the out-of-band media type with it. But why should that be a requirement?

That's exactly what I'll argue ;).  And unless you implement [completely generic peek-inside detection][1] (which I don't think anyone's arguing for), you're going to have to transmit *some* amount of media-type guidance along with your blob content.  I'm suggesting that guidance be the media type.

You seem to be suggesting that that guidance be “this blob is a JSON object which contains a self-describing `mediaType` field”.  Maybe you transmit that information because the tool-caller knows the tool can only handle such media types and therefore only feeds matching blobs into the tool.  That's how `oci-image-validate` works, and [I'm comfortable with that][2] from a keystroke-saving perspective.  However, I don't think we should pretend that this approach is completely free of out-of-band type guidance.

> What are you gaining by removing `mediaType`?

I'm not suggesting we remove `mediaType`, because some users (e.g. you with `oci-do-something`, or a number of people with `oci-image-validate`'s autodection) can't be bothered to pass media types around.  And I'm fine with that (typing out a long media type is not something I'd like to do repeatedly).

I'm just suggesting image-handling tools follow the spec's [SHOULD][3] and use descriptors to reference blob content, with peek-inside type detection being reserved for [signed-assertions][4].  And having acquired the media type from the referencing descriptor (or because we authored the blob ourselves), I see no need for image-handling tools to use peek-inside type detection.

Perhaps our difference here is that I see (almost) all tooling as being descriptor-based, while you see the tooling as being isolated-blob based.  Since I'm fine leaving existing `mediaType` entries in place, maybe we can just wait a year to see how that plays out and revisit this discussion then?

[1]: https://github.com/opencontainers/image-spec/pull/411#discussion_r86466669
[2]: https://github.com/opencontainers/image-spec/pull/411#issuecomment-257489701
[3]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc1/descriptor.md#L10
[4]: https://github.com/opencontainers/image-spec/pull/411#discussion_r86466669

jonboulle:
@vbatts I think I need to convert this PR to an epub and put it on my kindle to have any hope of catching up

jonboulle:
@vbatts as I see it, two things going on:
i) inconsistency of whether types have a `mediaType` field or not
ii) the descriptor type has a `mediaType` field which has a different meaning than the other types that have it (in other cases it's self-describing, in the descriptor case it's describing the entity the descriptor references)

IMHO ii MUST be fixed. i I don't feel _incredibly_ strongly about whether we address that by removing it everywhere or adding it to the places it's missing (see: ad infinitum arguments either way), but would still prefer consistency.

cyphar:
@jonboulle I'm :+1: on just doing `ii` in this PR and no longer discussing `i` (I'm starting to get frustrated). @wking can have his discussion elsewhere.

wking:
On Fri, Nov 04, 2016 at 04:13:06AM -0700, Jonathan Boulle wrote:
> ii) the descriptor type has a `mediaType` field which has a
> different meaning than the other types that have it (in other cases
> it's self-describing, in the descriptor case it's describing the
> entity the descriptor references)
>
> IMHO ii MUST be fixed. i I don't feel _incredibly_ strongly about
> whether we address that by removing it everywhere or adding it to
> the places it's missing (see: ad infinitum arguments either way),
> but would still prefer consistency.

If you feel (ii) must be fixed, it's because you want to support
peek-inside typing and not confuse the peekers with the different
descriptor semantics.  But if you address that by removing mediaType
from non-descriptor types, you're still not supporting the peekers.

If you don't care about peek-inside typing (this is where I'm at), it
doesn't matter if different types have different semantics for
mediaTyped.  You already know the blob type from the referencing
descriptor or because you just created thte blob yourself.  With this
position, you can leave the existing schemas alone (and land this PR
as it stands or not depending on whether you like having MediaTyped
independent of Versioned).

If you do care about peek-inside typing but don't want descriptor
confusing things, you'd need to rename either descriptor's mediaType
or the non-descriptors' mediaType.  I don't think either of those are
possible unless we drop our current “we can parse Docker blobs
directly into our types without semantic changes” requirement.  Even
if (ii) is bugging you, I'd be surprised if it is bugging you enough
to push through the current Docker-blob-semantics requirement ;).

stevvooe:
@jonboulle 

We should remove or deprecate the `mediaType` field from types that aren't descriptors.

> i) inconsistency of whether types have a mediaType field or not
> ii) the descriptor type has a mediaType field which has a different meaning than the other types that have it (in other cases it's self-describing, in the descriptor case it's describing the entity the descriptor references)

`i` was never supposed to exist, which created the inconsistency described in `ii`.

Types are supposed to be resolved through a qualified references. This both ensures secure access of resources, as well as easily supportable type-equivalence and versioning.

jonboulle:
OK can we just remove it everywhere else then.

vbatts:
Is the consensus here to remove `mediaType` from everything except the descriptor?
Will this not break compatibility, or are we saying we're okay with that, @stevvooe ?

stevvooe:
> Will this not break compatibility, or are we saying we're okay with that, @stevvooe ?

This should be okay.

I think we should reserve the field, so it is not used.

stevvooe:
@vbatts Do you need help in removing/reserving the `mediaType` field from other types?

vbatts:
Updated. Also updated the description on this PR to reflect the commit message.
PTAL.

wking:
e517a60 still needs Markdown and JSON Schema updates.

vbatts:
@stevvooe by removing `mediaType` from places like https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions it will absolutely make the OCI manifest and manifest-list not expressly compatible with the docker equivalents. Thoughts?

wking:
> ... it will absolutely make the OCI manifest and manifest-list not expressly compatible with the docker equivalents.

Removing `mediaType` from the OCI schema makes things *more* compatible in the Docker -> OCI direction (because you can just leave the Docker `mediaType` alone).  In the OCI -> Docker direction we'll need `mediaType` injection (where we previously needed `mediaType` translation), but that doesn't seem like a major compatible drop.

cyphar:
@vbatts We discussed this in the call. Basically @philips said that the `mediaType` was a mistake for Docker to have and that we don't need it. Conversion isn't really that significant of a reason, if you look at the conversion code I wrote for skopeo containers/image#172 the conversion is just a matter of changing mediaTypes. If we remove the `mediaType` from non-`Descriptor`s then it's just a matter of removing one line from the manifest configuration.

Since the config doesn't have a `mediaType` we don't need to worry about the sha256sum of objects changing through conversion.

wking:
> If we remove the mediaType from non-Descriptors then it's just a matter of removing one line from the manifest configuration.

Both manifests and manifest lists explicitly allow arbitrary extension properties, so unless we reserve the field (and I don't see a point to reserving it), you wouldn't need to remove the self-decribing `mediaType`.  You probably will still need to translate descriptor `mediaType`s, but you have to do that already.

> Since the config doesn't have a mediaType we don't need to worry about the sha256sum of objects changing through conversion.

Manifest digests are going to change regardless of this PR, since you're translating `layers[].mediaType`, etc.  Config digests are not going to change regardless of this PR, since config contains no `mediaType` properties.

cyphar:
> Manifest digests are going to change regardless of this PR, since you're translating layers[].mediaType, etc. Config digests are not going to change regardless of this PR, since config contains no mediaType properties.

That was my point. In Docker, manifests aren't "objects" in the same sense as in OCI. But configuration is.

stevvooe:
LGTM

As far as I can tell, this looks to remove embedded mediaTypes.

manifest-list.md
```diff
@@ -14,9 +15,8 @@ This section defines the `application/vnd.oci.image.manifest.list.v1+json` [medi
 
 - **`mediaType`** *string*
 
-  This REQUIRED property contains the media type of the manifest list.
-  For this version of the specification, this MUST be set to `application/vnd.oci.image.manifest.list.v1+json`.
-  For the media type(s) that this is compatible with, see the [matrix](media-types.md#compatibility-matrix).
+  This property is *reserved* for use, to [maintain compatibility][matrix].
+  When used, this field contains the media type of this document, which differs from the [descriptor](descriptor.md#properties) use of `mediaType`.
```
jonboulle: I kinda agree, this confused me on my read-through. Can we just say something like: 
```
This OPTIONAL property is RESERVED by the specification for backwards compatibility with older versions of Docker. The value of this field will not change. This field MAY be removed in a future version of the specification.
```
and leave it at that.

philips:
LGTM

cyphar:
Just want to note that this has broken several things in the past few weeks. While it's overall _fine_ (this is an `-rc` level spec after all) I am getting **increasingly distressed** about the **BLATANT lack of testing** people are doing within the spec writing community...

jonboulle:
@cyphar to be clear are you talking about writing additional tests to cover these cases, or not actually running existing tests?

cyphar:
Both. In the case of `umoci`, the current test suite is sufficient enough that if *anyone* had run it **before** merging breaking code into `skopeo`, `image-tools` and several other projects the breakages [in the other tools due to spec changes that weren't properly handled] would've been caught. But I'm also helping out with writing more tests for tools like `skopeo` which appear to be having a hard time keeping OCI image translation in working order.

While I'm doing my best to keep packages in openSUSE up to date enough so that `umoci`'s test suite is run against the latest tools, there's only one of me.

stevvooe:
@cyphar Remember that dependencies in _your_ project are _your_ responsibility.

To be honest, this spec was thrust into RC mode long before it was even close to being ready. I've seen more than one vendor declare "OCI Image support" in registries and countless projects claiming all sorts of OCI things related to images that are specious, at best.

If you want to be part of the OCI land grab, I commend you and we'll support you. But you'll have to deal with these kinds of problems along the way until we mark a 1.0 release.

What testing would you like us to pursue that we are not?

cyphar:
> Remember that dependencies in your project are your responsibility.

`skopeo` and `image-tools` are not dependencies of `umoci`. They are used as part of the test suite to verify that `umoci` is doing the right thing. So imagine my surprise when spec changes mean that `umoci` starts finding issues in these other projects.

> If you want to be part of the OCI land grab, I commend you and we'll support you. But you'll have to deal with these kinds of problems along the way until we mark a 1.0 release.

... so nobody who wants working code should implement any tooling around OCI images until 1.0? In other words, nobody should test that the spec is actually _sane_ before we declare it as "ready"? I'd like to remind you that the reason why we haven't hit 1.0 yet is because there weren't enough implementations for maintainers to agree that the spec was ready.

> What testing would you like us to pursue that we are not?

Making sure that `image-tools` is well-maintained would be a good first start (I've seen several issues with it, where it would). Also, I'm 95% sure that neither `image-tools` nor `runtime-tools` would be sufficient to actually claim that a project is significantly in compliance with OCI specs -- I understand that's something that is being worked on by the CertWG but we need to put more effort into maintaining and extending validation.

It would also be nice if spec changes involved checking with existing implementations, which is how other specifications handle things too. I'm not saying "make a PR against every project" but at the very least when we're updating our tools can we please check that we didn't break them (when interacting with a correct third-party implementation) first.

I'm going to be honest, my current hesitation about putting parts of `umoci` into OCI (something that I've planned to do from the beginning and still fully intend to do!) is because I'm just not sure that it will be maintained at its current standard once it becomes part of `image-tools`. And recent experiences are exacerbating this concern of mine.

cyphar:
> > If you want to be part of the OCI land grab, I commend you and we'll support you. But you'll have to deal with these kinds of problems along the way until we mark a 1.0 release.
>
> ... so nobody who wants working code should implement any tooling around OCI images until 1.0? In other words, nobody should test that the spec is actually sane before we declare it as "ready"? I'd like to remind you that the reason why we haven't hit 1.0 yet is because there weren't enough implementations for maintainers to agree that the spec was ready.

Sorry, to be clear I have no problem dealing with breakages because the spec has changed drastically -- that's obviously to be expected. It's just a little bit frustrating that people aren't testing their implementations (and the `image-tools` repo is an example of this) so it ends up with me having to start fixing fairly trivial issues all over the place.

Just because the spec is `-rc` doesn't mean we should stop caring about making code implementing the spec maintainable.

cyphar:
> I've seen more than one vendor declare "OCI Image support" in registries and countless projects claiming all sorts of OCI things related to images that are specious, at best.

`umoci`'s integration test suite runs `image-tools validate` after every single operation. If there's a more complete way of checking that OCI images generated are valid, please tell me about it because I will gladly add it to our tests. It's a bit hard to say that "support" is specious if our tools for validating whether something is supported aren't sufficiently checking things.

stevvooe:
@cyphar I'm sorry. I was harsh in my initial response. I agree with most of what you said.

The goal is to make a maintainable specification. My goal is that a 1.0 won't go live with these silly problems. We need early adopters to make sure things are sane. Nobody is saying that you shouldn't use it until 1.0. I am just saying that you are going to run into these kinds of issues until a 1.0 is released. It is frustrating and it is a pain, but that is the current state.

> I'm going to be honest, my current hesitation about putting parts of umoci into OCI (something that I've planned to do from the beginning and still fully intend to do!) is because I'm just not sure that it will be maintained at its current standard once it becomes part of image-tools. And recent experiences are exacerbating this concern of mine.

This is completely fair. I had this same concern with `go-digest`. The last time I checked out and worked with `image-tools`, it accepted `docker save/load` files as valid oci images, which is ridiculous. The commands are ill-conceived and I don't see their use case. It is _way_ too focused on image layout, which is a minor part of the image format. I'd like to see something closer to the [`dist` tool POC](https://github.com/docker/containerd/blob/master/reports/2017-01-27.md#distribution-tool) in 
`containerd`, specifically, at least `dist apply`. The goal of `image-tools` should be to get the _hard_ stuff right.

Continuing with the honesty here, `image-tools` would be in a lot better place if we had the work and vision that you've put into `umoci` in `image-tools`. I think we could get a lot more done their with someone with a strong idea of what the actual usability problems are, as well help to inform what is needed for the validating images.

To put in another way, I'd much prefer that we put your frustration into action in getting the `image-tools` project into shape. If that means you becoming a maintainer, or leading it, let's make that happen. The current situation is unacceptable and I think you can make it better.

caniszczyk:
@cyphar if you're interested in becoming a maintainer in OCI image-tools, please feel free to open up an issue and we can have the other maintainers vote if you plan to work more heavily on image-tools (we'd love to have you)

cyphar:
I've made a proposal in opencontainers/image-tools#118. But there are some further discussions that need to happen wrt. the history of `umoci`.

stevvooe:
> I've made a proposal in opencontainers/image-tools#118. But there are some further discussions that need to happen wrt. the plan of umoci and how it'll interact with image-tools.

We can discuss this on the call next week.

cyphar:
@stevvooe I'm going to send a proposal on the ML later today, so we can also discuss it there.

PR #412, opened 2016-10-24
Adjust the doc order of the content table

coolljt0725:

To make it more readable, make `manifest.md` and `manifest-list.md` above `layer.md` and `config.md`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,10 +27,10 @@ DOC_FILES := \
 	media-types.md \
 	descriptor.md \
 	image-layout.md \
-	layer.md \
-	config.md \
 	manifest.md \
 	manifest-list.md \
+	layer.md \
+	config.md \
 	canonicalization.md
 
 FIGURE_FILES := \
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ The goal of this specification is to enable the creation of interoperable tools 
     - [Media Types](media-types.md)
 - [Content Descriptors](descriptor.md)
 - [Image Layout](image-layout.md)
-- [Filesystem Layers](layer.md)
-- [Image Configuration](config.md)
 - [Image Manifest](manifest.md)
 - [Image Manifest List](manifest-list.md)
+- [Filesystem Layers](layer.md)
+- [Image Configuration](config.md)
 - [Canonicalization](canonicalization.md)
 
 ## Overview


vbatts:
LGTM

jonboulle:
lgtm

wking:
On Sun, Oct 23, 2016 at 08:30:47PM -0700, Lei Jitang wrote:
> To make it more readable, make `manifest.md` and `manifest-list.md` above `layer.md` and `config.md` You can view, comment on, or merge this pull request online at:

The previous order was [intended to be bottom-up][1], so folks reading in order would never be reading about undefined dependencies.  The new order is mixed.  If the intention is to make it top-down, manifest-list should come first.

[1]: https://github.com/opencontainers/image-spec/pull/275#r77881759

jonboulle:
Missed that. I guess it makes a bit of sense but I'd probably rather go for the top-down

wking:
On Mon, Oct 24, 2016 at 09:05:07AM -0700, Jonathan Boulle wrote:
> I guess it makes a bit of sense but I'd probably rather go for the top-down

Would that be:

```make
DOC_FILES := \
  README.md \
  code-of-conduct.md \
  project.md \
  media-types.md \
  manifest-list.md \
  manifest.md \
  layer.md \
  config.md \
  descriptor.md \
  image-layout.md \
  canonicalization.md
```

?

coolljt0725:
@wking  I though make `manifest-list.md` before `manifest.md`, but after reading the `manifest.md`, I think we could make `manifest.md` at first because at the beginning of `manifest.md`, it defines the [manifest specification](https://github.com/opencontainers/image-spec/blob/master/manifest.md#oci-image-manifest-specification).

wking:
On Mon, Oct 24, 2016 at 06:18:22PM -0700, Lei Jitang wrote:
> … I think we could make `manifest.md` at first because at the beginning of `manifest.md`, it defines the manifest specification.                                                                                                             

There is similar context-setting at the start of manifest-list.md.  I think manifest.md's current “three main goals” language is a bit stale though, since it seems to be introducing goals that lead to manifest lists (“fat manifest”) and configurations (“translatable to the OCI Runtime Specification”).  Regardless of whether we use top-down or bottom-up ordering, I think manifest.md's lead-in should be trimmed to focus on the role of the manifest in the OCI image modeling.

PR #416, opened 2016-10-26
 Descriptor should be materialized

q384566678:
When I modify the program in the image-tools, I found that if the Descriptors do not concrete, you can not call it, so I think it should be specific
Signed-off-by: zhouhao zhouhao@cn.fujitsu.com

### DIFF
--- a/schema/defs-image.json
+++ b/schema/defs-image.json
@@ -15,22 +15,12 @@
       "id": "https://opencontainers.org/schema/image/manifestDescriptor",
       "type": "object",
       "required": [
-        "mediaType",
-        "size",
-        "digest",
+        "config",
         "platform"
       ],
       "properties": {
-        "mediaType": {
-          "description": "the mediatype of the referenced object",
-          "$ref": "#definitions/mediaType"
-        },
-        "size": {
-          "description": "the size in bytes of the referenced object",
-          "$ref": "defs.json#/definitions/int64"
-        },
-        "digest": {
-          "$ref": "#definitions/digest"
+        "config": {
+          "$ref": "#content-descriptor.json"
         },
         "platform": {
           "id": "https://opencontainers.org/schema/image/platform",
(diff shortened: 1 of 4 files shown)


schema/defs-image.json
```diff
@@ -15,22 +15,12 @@
...
-          "$ref": "#definitions/mediaType"
-        },
-        "size": {
-          "description": "the size in bytes of the referenced object",
-          "$ref": "defs.json#/definitions/int64"
-        },
-        "digest": {
-          "$ref": "#definitions/digest"
+        "config": {
+          "$ref": "#content-descriptor.json"
```
wking: I'm fine shifting the descriptor properties out into their own property, but manifest-lists currently [don't do that][1].  I think you might be confusing `manifestDescriptor` ([used in the manifest-list][2]) with the [manifest schema][3].

[1]: https://github.com/opencontainers/image-spec/blob/7a6820ed62065744e9142ec93da4b37cc86fed5a/manifest-list.md#manifest-list-property-descriptions
[2]: https://github.com/opencontainers/image-spec/blob/7a6820ed62065744e9142ec93da4b37cc86fed5a/schema/manifest-list-schema.json#L16-L19
[3]: https://github.com/opencontainers/image-spec/blob/7a6820ed62065744e9142ec93da4b37cc86fed5a/schema/image-manifest-schema.json

q384566678: Perhaps I naMaybe I'm a little problem on the naming or understand， but I think in the manifest_list.go, Descriptor is the need for specific, or else when the call will be given the wrong, like“Descriptor has no field or method validate”

wking: It sounds like you're talking about [this Travis error][1] from opencontainers/image-tools#51.  But that's a Go compilation error; I don't see how it relates to the JSON Schema change you're proposing here.

[1]: https://travis-ci.org/opencontainers/image-tools/jobs/170981561#L248

philips:
I don't understand the reason for this change.

PR #419, opened 2016-10-27
layer.md: fix spelling error

TrumanLing:

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -63,7 +63,7 @@ The corresponding files that share the link with the > 1 linkcount may be outsid
 
 Hardlinks are stored in a tar archive with type of a `1` char, per the [GNU Basic Tar Format][gnu-tar-standard] and [libarchive tar(5)][libarchive-tar].
 
-While approaches to deriving new or changed hardlinks may vary, a possible approach is:
+While approaches to derive new or changed hardlinks may vary, a possible approach is:
 
 ```
 SET LinkMap to map[< Major:Minor String >]map[< inode integer >]< path string >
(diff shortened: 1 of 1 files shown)


layer.md
```diff
@@ -63,7 +63,7 @@ The corresponding files that share the link with the > 1 linkcount may be outsid
 
 Hardlinks are stored in a tar archive with type of a `1` char, per the [GNU Basic Tar Format][gnu-tar-standard] and [libarchive tar(5)][libarchive-tar].
 
-While approaches to deriving new or changed hardlinks may vary, a possible approach is:
+While approaches to derive new or changed hardlinks may vary, a possible approach is:
```
jonboulle: The original was correct

layer.md
```diff
@@ -204,7 +204,7 @@ The resulting tar archive for `rootfs-c9d-v1.s1` has the following entries:
 ./etc/.wh.my-app-config
 ```
 
-To signify that the resource `./etc/my-app-config` MUST be removed when the changeset is applied, the basename of the entry is prefixed with `.wh.`.
+To signify that, the resource `./etc/my-app-config` MUST be removed when the changeset is applied, the basename of the entry is prefixed with `.wh.`.
```
jonboulle: This changes the meaning. "that" is used as a conjuction here, not a pronoun.

jonboulle:
Sorry, neither of these are errors.

layer.md
TrumanLing: I think it should be ```to derive``` or ```for deriving```.

layer.md
TrumanLing: As for here, I agree with you.

TrumanLing: As for here, I agree with you.

PR #421, opened 2016-10-28
Add regular expression for object names in 'refs'

TrumanLing:
it will be used in 'image-tools' or other project for checking regular
expression of 'refs', which writes as "Object names in the refs
subdirectories MUST NOT include characters outside of the set of "A" to
"Z", "a" to "z", "0" to "9", the hyphen -, the dot ., and the underscore
_." at https://github.com/opencontainers/image-spec/blob/master/image-layout.md

Signed-off-by: Ling FaKe lingfake@huawei.com

### DIFF
--- a/specs-go/v1/layout.go
+++ b/specs-go/v1/layout.go
@@ -14,8 +14,15 @@
 
 package v1
 
+import "regexp"
+
 // ImageLayout is the structure in the "oci-layout" file, found in the root
 // of an OCI Image-layout directory.
 type ImageLayout struct {
 	Version string `json:"imageLayoutVersion"`
 }
+
+var (
+	// RefsRegexp matches requirement of image-layout 'refs' charset.
+	RefsRegexp = regexp.MustCompile(`^[a-zA-Z0-9-._]+$`)
+)


specs-go/v1/image_layout.go
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+package v1
+
+import "regexp"
+
+var (
+	// refRegexp for string ref should be complied with description "Object names in the refs subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen -, the dot ., and the underscore _." at https://github.com/opencontainers/image-spec/blob/master/image-layout.md.
+	RefsRegexp = regexp.MustCompile(`^[a-zA-Z0-9\-\.\_]+?$`)
+)
```
wking: Shift to `specs-go/v1/layout.go`, which already has some image-layout stuff (and may get more soon, #414)?

TrumanLing: Got it!

TrumanLing:
@wking it is done, PTAL.

specs-go/v1/layout.go
```diff
@@ -14,8 +14,15 @@
...
+import "regexp"
+
 // ImageLayout is the structure in the "oci-layout" file, found in the root
 // of an OCI Image-layout directory.
 type ImageLayout struct {
 	Version string `json:"imageLayoutVersion"`
 }
+
+var (
+	// RefsRegexp for string ref should be complied with description "Object names in the refs subdirectories MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen -, the dot ., and the underscore _." at https://github.com/opencontainers/image-spec/blob/master/image-layout.md.
```
wking: No need for the URL anymore now that we're in the same repository.  How about:

    // RefsRegexp matches refs that meet the image-layout charset requirements.

or something?

specs-go/v1/layout.go
```diff
@@ -14,8 +14,15 @@
...
+	RefsRegexp = regexp.MustCompile(`^[a-zA-Z0-9\-\.\_]+?$`)
```
wking: You don't have to backslash-escape hyphen, period, or underscore inside your brackets.  And I'm not sure why you have the `?`.  So I think this should be something like:

```go
RefsRegexp = regexp.MustCompile(`^[a-zA-Z0-9-._]+$`)
```

wking:
Cross-referencing opencontainers/image-tools#61.

specs-go/v1/layout.go
TrumanLing: OK, I will simplify the comment phrase.

specs-go/v1/layout.go
TrumanLing: I agree with you.

TrumanLing:
@wking updated it, PTAL.

wking:
b3e0e1a3 looks good to me.

philips:
LGTM

jonboulle:
lgtm

PR #422, opened 2016-10-28
image-layout.md: fix refs mediatype

Ladicle:
Object type in the `refs` subdirectory is `application/vnd.oci.descriptor.v1+json`, right?
It's type is `image.manifest.list` in the example code. I fix it.


jonboulle:
The object _itself_ is a descriptor, but that descriptor is _pointing_ to (referencing) an image manifest list. The descriptor doesn't have its own media type embedded inside it (see also #411 for more discussion of this)

PR #434, opened 2016-11-01
image-layout: clarification of oci-layout

vbatts:
The section for this file was too crammed together. Breaking it out into
individual items allows for more straightforward validation.

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -18,9 +18,12 @@ The image layout MUST contain two top level directories:
 
 Both `blobs` and `refs` MAY be empty.
 
-The image layout MUST also contain a file that is used to identify the layout version:
+The image layout MUST also contain an `oci-layout` file:
 
-- `oci-layout` MUST contain a JSON object with a version field `{"imageLayoutVersion": "1.0.0"}` and MAY include additional fields.
+- It MUST be a JSON object
+- It MUST contain an `imageLayoutVersion` field
+    - The `imageLayoutVersion` value will align with the OCI Image Specification version at the time changes to the layout are made, and will pin a given version until changes to the layout are required
+- It MAY include additional fields
 
 This is an example image layout:
 


image-layout.md
```diff
@@ -15,9 +15,12 @@ The image layout has two top level directories:
 - "refs" contains [descriptors][descriptors]. Commonly pointing to an [image manifest](manifest.md#image-manifest) or an [image manifest list](manifest-list.md#oci-image-manifest-list-specification).
 
 
-It also contains a file that is used to identify the layout version:
+The image layout MUST contain a file "oci-layout":
 
-- "oci-layout" MUST contain a JSON object with a version field `{"imageLayoutVersion": "1.0.0"}` and MAY include additional fields.
+- It MUST be a JSON object
+- It MUST contain a `imageLayoutVersion` field for determing the version of the layout used
```
wking: “a `imageLayoutVersion`” → “an `imageLayoutVersion`” and “determing” → “determining”.

I'm also fine with the shorter:

> - It MUST contain an `imageLayoutVersion` field.

since the purpose of that field seems fairly obvious.

image-layout.md
```diff
@@ -15,9 +15,12 @@ The image layout has two top level directories:
...
+    - The `imageLayoutVersion` value will align with the `specs.Version` at the time changes to the layout are made, and will pin a given version until changes to the layout are required
```
wking: It may be worth pointing out that this means `imageLayoutVersion` is not semantically versioned.  I'm ok with that, since I expect most image-layout bumps will break backwards compat anyway, but my personal preference would be to have a semantically versioned `imageLayoutVersion` that is more independent of the spec version.  You'd bump the spec version if *any* constituent component added a new feature (minor bump) or a backwards-incompatible feature (major bump).  So:

* A major `imageLayoutVersion` bump would require a major spec version bump.
* A minor `imageLayoutVersion` bump would require at least a minor spec version bump.
* A new layer media type (e.g. for #24) would require a minor spec version bump but no `imageLayoutVersion` bump.
* Dropping an old media type (e.g. after types based on `.wh.*` fall out of use) would require a major spec version bump but no `imageLayoutVersion` bump.

wking: Also, I'd rather change “`specs.Version`” → “OCI Image Format Specification version” to avoid referencing the informative Go from Markdown spec.

image-layout.md
```diff
@@ -15,9 +15,12 @@ The image layout has two top level directories:
 - "refs" contains [descriptors][descriptors]. Commonly pointing to an [image manifest](manifest.md#image-manifest) or an [image manifest list](manifest-list.md#oci-image-manifest-list-specification).
 
 
-It also contains a file that is used to identify the layout version:
+The image layout MUST contain an `oci-layout` file:
 
-- "oci-layout" MUST contain a JSON object with a version field `{"imageLayoutVersion": "1.0.0"}` and MAY include additional fields.
+- It MUST be a JSON object
+- It MUST contain an `imageLayoutVersion` field
+    - The `imageLayoutVersion` value will align with the OCI Image Specification version at the time changes to the layout are made, and will pin a given version until changes to the layout are required
```
wking: You've gone with “OCI Image Specification”, but the title for this spec is “[Open Container Initiative Image Format Specification][1]”.  I'm fine changing “Open Container Initiative” → “OCI”, but don't think we should be dropping “Format” here unless we also drop it from the spec title.  I don't mind which way we go, but I think we should stay consistent.

I also [still][2] think we should either semantically version `imageLayoutVersion` or point out that this is not semantically versioned.

[1]: https://github.com/opencontainers/image-spec/blob/a061b53a054decf479b006866bbe9f2642b50129/README.md#open-container-initiative-image-format-specification
[2]: https://github.com/opencontainers/image-spec/pull/434#discussion_r85972877

vbatts: I went with this because the image-layout.md doc is titled "OCI Image Layout Specification".

wking: This file is about the OCI Image Layout Specification, and if `imageLayoutVersion` semantically versions that, I'm happy.  However, it seems like the “pin” language was adding a relationship to the project-wide OCI Image Format Specification version.

To distinguish the two cases, pretend we had already cut 1.0.0, and say there is a backwards-compatible change to the image layout (e.g. you add an optional index, #438) and a backwards-incompatible change to something else (e.g. we drop support for `.wh.*`, #24).  The backwards-incompatible change means the project as a whole goes to 2.0.0.  With the independently-SemVer'ed approach, the backwards-compatible image-layout change moves `imageLayoutVersion` to 1.1.0.  With the pinned-to-the-project-version approach, `imageLayoutVersion` moves to 2.0.0.  Which of those is your intended result?

vbatts: Dare I say that some kind of specific-feature based compatibility would most closely be like the bittorrent BEP process, where specific features or enhancements are assigned an indicator (i.e. BEP0001). Rather than rev'ing a version that requires compliance of all features, an implementation can specify the specific features they accommodate.
Your pretending may certainly find benefits and drawbacks to both approaches. Neither is perfect. What is your intended result?

wking: I'm fine with tagged features (opencontainers/runtime-spec#15), but the OCI specs seemed to be leaning towards linearized versioning.  With linearized versioning, my preferred result is 1.1.0 (because it avoids confusing implementers and layout authors who are interested in image-layout but not with the other spec features).  I don't expect much feature fragmentation in image-layout, so shifting to tagged features would be a small gain.  However, just because I don't foresee fragmentation doesn't mean we won't have it ;).  So I'm also fine switching to tagged features proactively.

philips:
LGTM

vbatts:
updated. PTAL.

jonboulle:
lgtm

stevvooe:
LGTM

Do we have any language about how implementations must error if the Major is different, but must be able to process future minors/patch?

wking:
On Wed, Dec 07, 2016 at 02:26:50PM -0800, Stephen Day wrote:
> Do we have any language about how implementations must error if the
> Major is different, but must be able to process future minors/patch?

Are we intending to major-bump the OCI Image Format Specification
version when there is a backwards incompatible change to image-layout?
Or are we reserving major bumps for backwards-incompatible changes in
the blob schemas?  As I read it (still with 9c333a), the current
wording does not imply semantic versioning for image-layout (and I
think we should either require semantic versioning for image-layout or
point out that we're not requiring it [1,2]).

[1]: https://github.com/opencontainers/image-spec/pull/434#discussion_r85972877
[2]: https://github.com/opencontainers/image-spec/pull/434#discussion_r86073619

PR #438, opened 2016-11-03
image-layout: describe an OPTIONAL index

### LINKED ISSUE #374: reference management in image layout
Current spec doesn't care the layout management for `ref` folder. Actually consumer can set up many child folder and put the reference file to any where. It seems to be a little chaos. I think it had better to build rules for this. like:

`<domain>/<repository>/<reference>`

`domain` identifies a directory.
`repository` identifies the image name, which may be multi-level directory, like `library/ubuntu`
`reference` identifies the image tag.

the tree looks like:

```
.
└── refs
    └── docker.registry.io
        └── library
            └── ubuntu
                └── latest
                └── 14.04
            └── busybox
                └── latest
```

Any `ref` layout inconsistent to above will be regarded as illegal.

Previous [comment](https://github.com/opencontainers/image-spec/pull/327#issuecomment-248256133) shows that the image layout actually play as an image repository, which contains many images(as it contains manifest list). So I think the directory should be managed with multi-levels.

vbatts:
Depending on where and how an OCI image-layout is provided (tar, zip,
HTTP, etc) the mechanism to "walk" the refs can vary in complexity and
expense. For HTTP in example, getting a directory listing is often
disabled, but even would require multiple fetches to build the map of refs.

Closes #374

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -24,6 +24,7 @@ The image layout MUST also contain an `oci-layout` file:
 - It MUST contain an `imageLayoutVersion` field
     - The `imageLayoutVersion` value will align with the OCI Image Specification version at the time changes to the layout are made, and will pin a given version until changes to the layout are required
 - It MAY include additional fields
+- It MAY contain an `index` field ([described below](#index))
 
 This is an example image layout:
 
@@ -127,6 +128,42 @@ $ cat ./blobs/sha256/5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a43335
 ```
 $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f
 [tar stream]
+```
+
+## Index
+
+Index is intended to be generated from the content in an image-layout after any updates have been made.
+Allows for mapping objects in the image-layout, and needing only to read from the `oci-layout` once to collect this index.
+
+**index** *array of objects*, OPTIONAL
+
+  Provides a single access to top-level information about the layout.
+
+  - Each object has the base properties of [descriptor][descriptors]
+  - OPTIONAL **annotations** field with the same constraints as [manifest `annotations`](manifest.md#image-manifest-property-descriptions)
+
+### Index Example
+
+```json
+{
+  "imageLayoutVersion": "1.0.0",
+  "index": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "digest": "sha256:0169d60159b9e6cf7aad602377ad6e3f551a7b0d01c51c8620cea5358f71546e",
+      "size": 426
+    },
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "digest": "sha256:42f20fd615e5ab67f1b43c99f5f4a262d3e952ee6aa70b88f61207c24f2df2fb",
+      "size": 426,
+      "annotations": {
+        "org.waffles.with": "syrup,raspberries",
+        "io.flatpak.app.name": "org.gnome.gedit"
+      }
+    }
+  ]
+}
 ```
 
 [descriptors]: ./descriptor.md


image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 [tar stream]
 ```
 
+## Index
```
wking: [Trailing whitespace][1].

[1]: https://travis-ci.org/opencontainers/image-spec/builds/172964573#L318

image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+Index is intended to be generated from the content in an image-layout after any updates have been made.
+Allows for mapping random access to the references in the image-layout, and needing only to read from the `oci-layout` once to build this map of references.
+
+**index** *array of objects*, OPTIONAL
+
+  Provides a single access to top-level information about the layout.
+  Each object has the following properties:
+
+  - **ref** *string*, REQUIRED
+
```
wking: [Trailing whitespace][1].

[1]: https://travis-ci.org/opencontainers/image-spec/builds/172964573#L318

image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 [tar stream]
 ```
 
+## Index 
+
+Index is intended to be generated from the content in an image-layout after any updates have been made.
```
wking: “any updates” → “any ref additions or removals”.  You shouldn't need to update it after other changes (e.g. blob addition or removal, or changing a ref's descriptor payload).

image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+Allows for mapping random access to the references in the image-layout, and needing only to read from the `oci-layout` once to build this map of references.
```
wking: I'd rather replace this line with your HTTP use-case.  Maybe just:

> For example, HTTP servers often disable directory listing.

image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+
+**index** *array of objects*, OPTIONAL
+
+  Provides a single access to top-level information about the layout.
```
wking: Four-space indents nside list entries where we'll be using sub-lists (opencontainers/runtime-spec#495)?

And for the wording, maybe just “Provides a list of available refs.”

image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 [tar stream]
 ```
 
+## Index 
+
+Index is intended to be generated from the content in an image-layout after any updates have been made.
+Allows for mapping random access to the references in the image-layout, and needing only to read from the `oci-layout` once to build this map of references.
+
+**index** *array of objects*, OPTIONAL
```
wking: The ref list could get large, and I'd rather not overload the “do I understand how to parse this?“ `oci-layout`.  Can we put this index in `refs.json` instead?  You could just dive in with the array:

```
$ cat refs.json
["latest", "v1"]
```

Since the filename + the version from `oci-layout` sufficiently namespace it.

image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+  Provides a single access to top-level information about the layout.
+  Each object has the following properties:
+
+  - **ref** *string*, REQUIRED
+  
+    The name of the ref (correspond JSON descriptor found in `./refs/`)
+
+  - **annotations** *string-string map*, OPTIONAL
+
+    This field is under the same constaints as [manifest `annotations`](manifest.md#image-manifest-property-descriptions)
```
wking: I'd rather not add this field to the ref index, since it wouldn't be supported by `ref/`-walking listers.  Can you motivate a use-case for it?  Folks who wanted to annotate a ref can already do so by adding a new ref (e.g. `v1.0.0.annotations`) pointing at a CAS blob with their desired payload.

wking:
On Thu, Nov 03, 2016 at 07:41:35AM -0700, Vincent Batts wrote:
> Depending on where and how an OCI image-layout is provided (tar,
> zip, HTTP, etc) the mechanism to "walk" the refs can vary in
> complexity and expense. For HTTP in example, getting a directory
> listing is often disabled, but even would require multiple fetches
> to build the map of refs.

Ref payloads are just descriptors, so they usually won't be very big.
For cases where a JSON listing of all refs is feasible, we can
probably inline the descriptor value in the ref-listing JSON file.
For example:

  $ cat refs.json
  {
    "v1.0" {
      "size": 4096,
      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
      "mediaType": "application/vnd.oci.image.manifest.list.v1+json"}
    },
    …
  }

Once you have that, maybe you no longer need /refs?  There are pros
and cons to either approach:

* A `refs.json` works well over HTTP, where listing refs/ may not be
  enabled.  It may provide slow access to a known ref, since you have
  to unmarshal a potentially large JSON document before you can lookup
  the ref you need.

* A `refs/` directory avoids racy conflicts around reading/writing a
  single file.  Multiple consumers can read from and atomically update
  entries in the directory without problems.  It also provides fast
  access to a known ref.

My feeling is that in situations where you're using a ref-listing JSON
(e.g. over HTTP):

* You don't really care about speed, because it's dwarfed by network
  latency.
* You don't really care about update conflicts, because if you are
  running a service supporting concurrent modifications, you probably
  have something more sophisticated than direct access to refs.json.

So I'd argue for refs.json as a complete replacement for refs/ which
should live under a 2.x imageLayoutVersion series.  Refs and CAS are
completely orthogonal, so we might want to replace imageLayoutVersion
with separate version numbers for refs and CAS.  Then you could say:

  $ cat oci-layout
  {
    "refsLayoutVersion": "1.0.0",
    "casLayoutVersion": "1.0.0"
  }

for refs/ and:

  $ cat oci-layout
  {
    "refsLayoutVersion": "2.0.0",
    "casLayoutVersion": "1.0.0"
  }

for refs.json.

image-layout.md
vbatts: k

image-layout.md
vbatts: Those are both are updates of one kind or another.

image-layout.md
vbatts: it has the potential for being long, no matter where it lives. And as the oci-layout is required file, it will already be stat'ed or fetched. Deferring to another file means _another_ read, that could be a ENOENT or a 404. Plus this truncated list you mention does not allow for having tagged metadata.

image-layout.md
vbatts: This is the same indentation used in the config.md and manifest.md for its field descriptions.

image-layout.md
wking: But there are updates that are not included in “ref additions or removals” (e.g. blob addition, or changing a ref's descriptor payload) which would *not* require index regeneration.  I'm suggesting we make our regeneration suggestions as narrowly-scoped as possible, because conflicts around regenerating this index are [one of the major pain-points with this approach][1].

[1]: https://github.com/opencontainers/image-spec/pull/438#issuecomment-258199793

image-layout.md
wking: As it stands, the line is not a sentence (that would be “The index allows…” or some such).  And you already have random access with `refs/`.  The think the index adds is single-file list support when you can't walk `refs/` with `readdir(3)` or similar.  And the HTTP sentence makes that clear with a concrete example.

image-layout.md
wking: With a separate file, a long index only impacts ref-listing.  By putting it in `oci-layout`, a long index impacts *any* image-layout activity (because everyone should be checking to see “is this an image-layout version I understand” before trying to use the bits they need).

> Plus this truncated list you mention does not allow for having tagged metadata.

If annotating refs is important, I'd rather have `annotations` added to the descriptor model.  Do you have an example use case that shows that `annotations` are useful for refs but *not* for descriptors in general?

image-layout.md
wking: If you want to punt for now, that's fine with me (I'm not a consumer of the compiled HTML or PDF output).  But runtime-spec has had both opencontainers/runtime-spec#495 and opencontainers/runtime-spec#608 to avoid issues with two-space indents for sub-lists.  It seems like we should be sticking to that pattern here for new additions, and eventually someone will be bothered enough [by][1]:

```html
<ul>
<li><strong><code>mediaType</code></strong> <em>string</em></li>
</ul>
<p>This REQUIRED property contains the MIME type of the referenced content.</p>
<p>The OCI image specification defines <a href="media-types.md">several of its \
own MIME types</a> for resources defined in the specification.</p>
```

That they'll go through and fix the existing Markdown ;).

[1]: https://github.com/opencontainers/image-spec/releases/download/v1.0.0-rc2/oci-image-spec-v1.0.0-rc2.html

image-layout.md
wking: I've spun this off into #440.  Can we drop it from this PR and focus on the ref-listing use case?  We can come back and add it to the index if the consensus after #440 is “yes we need these, and no we can't add `annotations` to the descriptor model” (that's not the consensus I expect, but I'm not a maintainer ;).

image-layout.md
vbatts: That still means that a read per ref.
The whole notion of an index will be for those sharing it in their own use-case being able to bubble up the important pieces of metadata per ref. Adding it to the descriptor might be fine as well, but as RPM, debian, and other repos have shown, there is a vast amount of metadata provided as early as possible. It is not my job to limit to these unknown users what information they're allowed to bubble up. Again, I am not just thinking through my own personal use-case of what is needed in my home directory of my personal machine.

image-layout.md
vbatts: really spreading out the topics. there is hardly a consensus reachable when the topic is so diluted.

wking: Folks have suggested that I keep too many points in a single PR and end up distracting from that PR.  If you feel like the annotations discussion is fundamental to this PR, I'm happy to keep it here (and you should probably close #440 with “Discussion in #438” or some such).

vbatts: oh, it's just here, there and everywhere. The further dilution of topics never seems to have the intention to arrive at a conclusion, only to further add items to come to agreement on.

image-layout.md
wking: Adding `annotations` to descriptors is not directly coupled to removing it from an index.  It just means that folks who find the ref payload via `refs/` will still have access to it (and can skip loading the potentially large index JSON).  Once you have descriptor annotations, you can certainly cache them in the index as a separate performance optimization.  But *only* supporting ref annotations via the index (and not via `refs/`) seems like a strange position.

image-layout.md
wking: I expect we will more quickly arrive at a consensus for a ref index to support our existing ref listing over HTTP (when the HTTP server disables directory listing), since we both agree that's important.  I'm suggesting we decouple the orthogonal ref-annotation issue so that it doesn't bog down the ref-indexing discussion, which should help us reach a consensus on ref-indexing more quickly.

wking:
On Thu, Nov 03, 2016 at 09:41:26AM -0700, W. Trevor King wrote:
> Refs and CAS are completely orthogonal, so we might want to replace
> imageLayoutVersion with separate version numbers for refs and CAS.

Spun off into #441 to avoid distracting here.

image-layout.md
```diff
@@ -115,4 +116,42 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+  Provides a single access to top-level information about the layout.
+  Each object has the following properties:
+
+  - **ref** *string*, REQUIRED
+
+    The name of the ref (correspond JSON descriptor found in `./refs/`)
+
+  - **annotations** *string-string map*, OPTIONAL
+
+    This field is under the same constaints as [manifest `annotations`](manifest.md#image-manifest-property-descriptions)
```
coolljt0725: `constaints `  a typo here :)

vbatts: fixed

vbatts:
updated. PTAL

image-layout.md
```diff
@@ -118,4 +119,40 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 [tar stream]
 ```
 
+## Index
+
+Index is intended to be generated from the content in an image-layout after any udates have been made.
```
philips: nit: udates

stevvooe:
This is a great change, but the structure is nearly identical to manifest-list. Let's do the following:

1. Remove the `refs` concept. It overlaps too much and isn't very useful.
1. Add an `annotations` field to descriptor.
2. Modify manifest-list type to clarify that it can be used in this role.
3. Have the image layout specify and `index.json` as a manifest-list.

In general, this addresses concerns I've had with `refs`. Thanks for taking this problem on!

image-layout.md
vbatts: fxied

philips:
@stevvooe I think I largely agree but what replaces the concept of a "ref name" in the [manifest list](https://github.com/opencontainers/image-spec/blob/master/manifest-list.md)? Any arbitrary annotation?

vbatts:
Perhaps a specific annotation. "org.opencontainer.ref.name" or similar?

stevvooe:
@philips @vbatts I was leaving that as arbitrary for now. We could have tools on import pick up various fields then work on conventions from there.

Perhaps we work through this change and address this in a follow up.

I can already imagine a tool that dumps the index and annotations:

```
$ oci-image inspect foo.tar.gz
sha256:abcdef
  org.opencontainer.ref.name=foo
  org.opencontainer.morejunk=bar
```

Maybe integrate such that the consumer doesn't need to have a convention:

```
$ docker load -t $(oci-image inspect -f '{.Annotations["org.opencontainer.ref.name"}' foo.tar.gz) foo.tar.gz
```

stevvooe:
@vbatts Any movement on updating this approach to use the manifest list type?

vbatts:
i'm still mulling on this @stevvooe . I realize this is very overlapped with the manifest-list, though the manifest list is meant for pointing to only manifests with an election based on platform, though the act of indexing could allow for pointing to any mediatypes (even arbitrary types). I feel like this is different enough.

It would take some restructuring of the manifest-list that would make it generic.

wking:
On Tue, Dec 13, 2016 at 08:45:39AM -0800, Vincent Batts wrote:
> It would take some restructuring of the manifest-list that would
> make it generic.

With a new Descriptor.annotations, I think manifest-list is *already*
generic enough (although it would be good to provide more context
around these dual use-cases, e.g. near [1]).  What else are you
missing?

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/manifest-list.md#L35

vbatts:
@wking besides the `manifests` field name with the sentence `This REQUIRED property contains a list of manifests for specific platforms.`

wking:
On Tue, Dec 13, 2016 at 09:07:32AM -0800, Vincent Batts wrote:
> @wking besides the `manifests` field name with the sentence `This
> REQUIRED property contains a list of manifests for specific
> platforms.`

The ‘manifests’ name could be made more generic.  Are you suggesting
lifting the requirement to allow for indexes that contain no entries?
I'm ok with that.  The “list of manifests” context could be made more
generic too, but it's currently not blocking you putting non-manifests
in because the descriptor types are flexible [1].

We'd probably also want to move the currently-REQUIRED platform stuff
[2] into Descriptor.annotations.

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/manifest-list.md#L35
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/manifest-list.md#L39

vbatts:
i do agree all the platform values would be better as annotations. Especially given their inconsistent naming convention (to speak of, we use `under_score`, `camelCase`, and `field.name` :grimacing: )

stevvooe:
@vbatts The manifest-list is already generic enough. I'd prefer not to see a secondary data type that is the same but not.

> though the manifest list is meant for pointing to only manifests with an election based on platform

The manifest-list was never meant to be limited to this particular use case. It got de-featured during the design process, since no one could understand its usefulness in other cases (like this one), as I had argued back at that time. Subsequently, this was the use case that got it in, but the limitation is arbitrary.

Adding annotations is all it would take. No other movements are necessary.

alexlarsson:
Dropping `platform` from the manifest list and having a generic annotations with some common standard fields seems like a great improvement to me. The current set of fields has a bunch of essentially free-form fields with zero guidance to how you would use them anyway.

For the manifest list digest it would be quite nice if you could extend it with other types. For instance, when using OCI registries for flatpak I would like to put in an index in the registry that points to an appstream[1] xml file which could be used by a graphical app-store frontent. I.e. it would get me metadata like title, description, icons, screenshots, etc for all the apps in the registry. But I agree with @wking in that it seems the spec already allows this.

[1] https://www.freedesktop.org/wiki/Distributions/AppStream/

vbatts:
i've just rebased this, but haven't had a chance to merge it with @stevvooe thoughts on an `index.json` that is a manifest-list. Nor updates to the `manifest-list.md` that would make it more generic to be a descriptor-list. :grimacing:

vbatts:
@stevvooe if we were to move forward with changes to the manifest-list to make it more of a descriptor-list, would that not allow the current docker use of this media type in the registry et al to just begin use as is without requiring the backwards compatibility? seems like @mikebrow (via https://github.com/docker/distribution/pull/2076) would have an input here

stevvooe:
@vbatts For the most part, we just ignore extra fields, so this shouldn't be a problem.

philips:
LGTM

What is left to do on this? Should we make it part of the -rc4 milestone?

RobDolinMS:
@vbatts has some hesitation on this potentially being too generic.
@stevvooe Can you comment on this?

stevvooe:
@vbatts @RobDolinMS I thought we had agreed that these were structurally identical to manifest lists.

That would entail closing this PR and fixing it to take that into account.

xiekeyang:
@vbatts 
If we should restrain that objects in `Index` field MUST be `manifest` (maybe or `layer`) type? These might be all possible types for `Index`.

And if we should NOTE the different of Index and manifest-list in this section? Currently manifest-list is mainly for pointing to one manifest under specific platform. While `Index` is for including many manifests, with some sequential dependency.

vbatts:
Next steps here to move forward and satisfy the requirements would be:
* keeping manifest-list as "the entry point" ( it was a lot work to get it there for schema 2)
* loosen up required fields
* optional `/index.json`
* add to manifest-list a field like "channels" or "descriptors"

Last thing for me to do is clarify the scope/context of annotations and expectations around conveying them.

vbatts:
#533 merged

PR #439, opened 2016-11-03
image-layout: Fix 'mediatype' -> 'mediaType' in ref payload

wking:
To match the descriptor property, fixing a typo from #94, although that was developed in parallel with the descriptor spec.


vbatts:
LGTM

jonboulle:
lgtm

PR #431, opened 2016-11-01
specs-go: add missing urls field of descriptor

coolljt0725:
Signed-off-by: Lei Jitang <leijitang@huawei.com>

PR #442, opened 2016-11-03
descriptor: MUST URIs and SHOULD http(s) for descriptor's 'urls'

### LINKED ISSUE #436: Define URLs' scheme
Descriptor's props at https://github.com/opencontainers/image-spec/blob/master/descriptor.md#properties do not talk about which scheme a URL should have.
Should we define that URLs SHOULD have either http or https scheme? leaving other scheme (ftp, file) to be used as well, but enforcing http[s]? or should we leave it as is and maybe link https://tools.ietf.org/html/rfc1738?

/cc @vbatts @philips @jonboulle

wking:
The in-flight #431 is using JSON Schema's [`uri` format][1], which requires RFC 3986.  This commit makes that an official spec requirement by adding a MUST to the Markdown.  It also SHOULDs the http(s) schemes, because they're widely supported and are deployed in existing Docker images.  And @vbatts and @runcom both like SHOULDing them ;).

Fixes #436.

[1]: https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-7.3.6

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -36,7 +36,9 @@ The following fields contain the primary properties that constitute a Descriptor
 
 - **`urls`** *array*
 
-  This OPTIONAL property specifies a list of URLs from which this object MAY be downloaded.
+  This OPTIONAL property specifies a list of URIs from which this object MAY be downloaded.
+  Each entry MUST conform to [RFC 3986][rfc3986].
+  Entries SHOULD use the `http` and `https` schemes, as defined in [RFC 7230][rfc7230-s2.7].
 
 ### Reserved
 
@@ -117,3 +119,6 @@ The following example describes a [_Manifest_](manifest.md#image-manifest) with 
   "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"
 }
 ```
+
+[rfc3986]: https://tools.ietf.org/html/rfc3986
+[rfc7230-s2.7]: https://tools.ietf.org/html/rfc7230#section-2.7


runcom:
Seems fine to me

wking:
I can also add a link to [RFC 7540 §3][1] if folks want to see HTTP/2 punting to HTTP/1.1 on this (although [HTTP/2's `:scheme` pseudo-header is more open-ended][2]).

[1]: https://tools.ietf.org/html/rfc7540#section-3
[2]: https://tools.ietf.org/html/rfc7540#section-8.1.2.3

descriptor.md
```diff
@@ -34,9 +34,11 @@ The following fields contain the primary properties that constitute a Descriptor
   This property exists so that a client will have an expected size for the content before processing.
   If the length of the retrieved content does not match the specified length, the content SHOULD NOT be trusted.
 
-- **`urls`** *array*
+- **`urls`** *array of strings*
```
wking: Done in 8729ba4 → 876489a, and I've filed #447 with a handful of these “array of {type}” fixes.

stevvooe:
LGTM

philips:
LGTM

descriptor.md
```diff
@@ -36,7 +36,9 @@ The following fields contain the primary properties that constitute a Descriptor
 
 - **`urls`** *array*
 
-  This OPTIONAL property specifies a list of URLs from which this object MAY be downloaded.
+  This OPTIONAL property specifies a list of URIs from which this object MAY be downloaded.
+  Each entry MUST conform to [RFC 3986][rfc3986].
+  Entries SHOULD use the `http` and `https` schemes, as defined in [RFC 7230][rfc7230-s2.7].
```
wking: There's no danger of using a maliciously substituted blob, because you'll hash it when you get it.  Using HTTP just means detecting that sort of denial of service is more expensive.  And it means packet sniffers can see what you're asking for.  I expect there are situations where both of those are acceptable risks.

PR #444, opened 2016-11-04
Bug Fix: error syntax in schema definitions

xiekeyang:
Validate manifest list JSON object via schema.Validator.
* To validate full fields;
* To validate required fields only;
* To validate customized reference manifest media type;
* To validate invalid media type case;
* To validate invalid field type case;

### DIFF
--- a/schema/content-descriptor.json
+++ b/schema/content-descriptor.json
@@ -6,14 +6,15 @@
   "properties": {
     "mediaType": {
       "description": "the mediatype of the referenced object",
-      "$ref": "defs-image.json#definitions/mediaType"
+      "$ref": "defs-image.json#/definitions/mediaType"
     },
     "size": {
       "description": "the size in bytes of the referenced object",
       "$ref": "defs.json#/definitions/int64"
     },
     "digest": {
-      "$ref": "defs-image.json#definitions/digest"
+      "description": "the cryptographic checksum digest of the object, in the pattern '<hash>:<hexadecimal digest>'",
+      "$ref": "defs-image.json#/definitions/digest"
     },
     "urls": {
       "description": "a list of urls from which this object may be downloaded",
--- a/schema/defs-image.json
+++ b/schema/defs-image.json
@@ -4,7 +4,7 @@
     "mediaType": {
       "id": "https://opencontainers.org/schema/image/mediaType",
       "type": "string",
-      "pattern": "^[a-z]+/[0-9a-zA-Z.+]+$"
+      "pattern": "^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}/[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}$"
     },
     "digest": {
       "description": "the cryptographic checksum digest of the object, in the pattern '<hash>:<hexadecimal digest>'",
--- a/schema/defs-image.json
+++ b/schema/defs-image.json
@@ -23,14 +23,15 @@
       "properties": {
         "mediaType": {
           "description": "the mediatype of the referenced object",
-          "$ref": "#definitions/mediaType"
+          "$ref": "#/definitions/mediaType"
         },
         "size": {
           "description": "the size in bytes of the referenced object",
           "$ref": "defs.json#/definitions/int64"
         },
         "digest": {
-          "$ref": "#definitions/digest"
+          "description": "the cryptographic checksum digest of the object, in the pattern '<hash>:<hexadecimal digest>'",
+          "$ref": "#/definitions/digest"
         },
         "urls": {
           "description": "a list of urls from which this object may be downloaded",
--- a/schema/manifest_test.go
+++ b/schema/manifest_test.go
@@ -95,6 +95,23 @@ func TestManifest(t *testing.T) {
 			fail: true,
 		},
 
+		// expected failure: config.digest is not hex hashed format
+		{
+			manifest: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
+  "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
+    "size": 1470,
+    "digest": "sha256:c86f7763873b6c0aaehhhhhhhhhhhhhhhhmmmmmmmmmmmmmmmm22d9h3bab59b4f"
+  },
+  "layers": []
+}
+`,
+			fail: true,
+		},
+
 		// valid manifest with optional fields
 		{
 			manifest: `
--- a/schema/manifestlist_test.go
+++ b/schema/manifestlist_test.go
@@ -1,0 +1,257 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/opencontainers/image-spec/schema"
+)
+
+func TestManifestList(t *testing.T) {
+	for i, tt := range []struct {
+		manifestList string
+		fail         bool
+	}{
+		// expected failure: mediaType does not match pattern
+		{
+			manifestList: `
+{
+  "schemaVersion": 2,
+  "mediaType": "invalid",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": 7143,
+      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
+      "platform": {
+        "architecture": "ppc64le",
+        "os": "linux"
+      }
+    }
+  ]
+}
+`,
+			fail: true,
+		},
+
+		// expected failure: manifest.size is string, expected integer
+		{
+			manifestList: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": "7682",
+      "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270",
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux",
+        "features": [
+          "sse4"
+        ]
+      }
+    }
+  ]
+}
+`,
+			fail: true,
+		},
+
+		// expected failure: manifest.digest is missing, expected required
+		{
+			manifestList: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": 7682,
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux",
+        "features": [
+          "sse4"
+        ]
+      }
+    }
+  ]
+}
+`,
+			fail: true,
+		},
+
+		// expected failure: manifest.platform is missing, expected required
+		{
+			manifestList: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": 7682,
+      "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"
+    }
+  ]
+}
+`,
+			fail: true,
+		},
+
+		// expected failure: invalid referenced manifest media type
+		{
+			manifestList: `
+{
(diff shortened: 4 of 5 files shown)


schema/manifestlist_test.go
```diff
@@ -0,0 +1,195 @@
...
+	for i, tt := range []struct {
+		manifestList string
+		fail         bool
+	}{
+		// expected failure: mediaType does not match pattern
+		{
+			manifestList: `
+{
+  "schemaVersion": 2,
+  "mediaType": "invalid"
```
wking: Probably also want a test where `schemaVersion` is 0, although support for checking that is still in flight with #404.

schema/manifestlist_test.go
```diff
@@ -0,0 +1,195 @@
...
+      "size": 7143,
+      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
+      "platform": {
+        "architecture": "ppc64le",
+        "os": "linux"
+      }
+    },
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": "7682",
```
wking: No need to include multiple manifests to exercise this bit of the JSON Schema.  Can we drop the `sha256:e69…` entry?

schema/manifestlist_test.go
```diff
@@ -0,0 +1,195 @@
...
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": 7143,
+      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
+      "platform": {
+        "architecture": "ppc64le",
+        "os": "linux"
+      }
+    },
+    {
+      "mediaType": "foo",
```
wking: This is not a valid media type (you need [both a type and a subtype][1]).  I recommend using `text/plain` or some other off-the-shelf, non OCI media type.

[1]: https://tools.ietf.org/html/rfc4288#section-4.2

schema/manifestlist_test.go
```diff
@@ -0,0 +1,195 @@
...
+        "features": [
+          "sse4"
+        ]
+      }
+    }
+  ],
+  "annotations": {
+    "com.example.key1": "value1",
+    "com.example.key2": "value2"
+  }
```
wking: You've already exercised `annotations`, so I'd leave it off of tests that focused on other properties.

schema/manifestlist_test.go
xiekeyang: It is reasonable and I take it in mind. This case should be added after #404 is landed.

schema/manifestlist_test.go
xiekeyang: Although it is done in `manifest_test.go`, but I think we should still cover it here, against that schema definition for manifest and manifest list split. After all, they are different stories.
And I agree to test no required entry case.

schema/manifestlist_test.go
wking: #404 also adds testing for the self-deceiving media type you're attempting to verify here ;).  I expect this case is failing for a different reason.

schema/manifestlist_test.go
xiekeyang: We have talked about how to look and handle the media type of reference fields. I remember we agree to only omit warning for not OCI canonical media type and later, we need to add strict mode to defer it.
In unit test, I propose to use free string to show the current compatibility and reality. Using restricted string  maybe conceal the defect on design.
IMO.

xiekeyang:
It is undated in https://github.com/opencontainers/image-spec/pull/444/commits/a0e029099f2320418241951af824e293b538c624.
Drop some duplicated lines, and add some case missing required fields.

schema/manifestlist_test.go
xiekeyang: It is now on line https://github.com/opencontainers/image-spec/pull/444/files#diff-b1e07ac69261c70ec3feda53adffb906R34.

I read #404. It strict the manifest and list media type, but it seems not to set up unit test for manifest list media type, it just modify Backwards Compatibility, to let manifest list be compatible to Docker's. It doesn't test invalid string.

this pr looks not duplicated with #404, right, or I miss something?

schema/manifestlist_test.go
wking: The descriptor spec (which [manifest-list leans on][1]) [requires this to be a media type][2] (but doesn't require it to be a specific media type).  So there are a few cases for validation:

a. A media type that implementations MUST support (currently just `application/vnd.oci.image.manifest.v1+json`).
b. Another media type (e.g. `text/plain`).
c. A non-media-type string (e.g. `foo`), which doesn't match [RFC 4288's name requirements][3].
d. A non-string.

We may not be able to recognize the media types in (b), but we *can* distinguish them from the non-media-type strings in (c).  I'd expect validation to fail for (c), but am happy to PR more explicit language to the spec before you make changes here.

[1]: https://github.com/opencontainers/image-spec/blame/28b83df0bcc065285be064710d09971256a4e08e/descriptor.md#L22
[2]: https://github.com/opencontainers/image-spec/blame/28b83df0bcc065285be064710d09971256a4e08e/manifest-list.md#L30
[3]: https://tools.ietf.org/html/rfc4288#section-4.2

schema/manifestlist_test.go
wking: Yeah, we want both this PR and #404.  Which one lands first depends on whether we're doing Test-Driven Development ;).

xiekeyang: @wking, maybe I express inaccurately. I mean if my test case is not redundant(as my explain above), and this PR looks OK?

https://github.com/opencontainers/image-spec/pull/444/files#diff-b1e07ac69261c70ec3feda53adffb906R29,

```json
{
  "schemaVersion": 2,
  "mediaType": "invalid",
  "manifests": [
    {
      "size": 7143,
      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
      "platform": {
        "architecture": "ppc64le",
        "os": "linux"
      }
    }
  ]
}
```

xiekeyang: @wking WIP

schema/manifestlist_test.go
xiekeyang: @wking 

You are right. Actually current schema exist some bugs, to lead illegal cases passed. 
I fix them in https://github.com/opencontainers/image-spec/pull/444/commits/7f792fa5bf925b01bc280c8b6ba6aec6655097a2 and https://github.com/opencontainers/image-spec/pull/444/commits/2da8885c2242fb08e550d83e4ee68ab9024fb12b.
With those supporting, we can test and get the excepted result. After fixing, I add the necessary test cases, including what you suggested.

https://github.com/opencontainers/image-spec/pull/444/commits/2da8885c2242fb08e550d83e4ee68ab9024fb12b MUST be done, otherwise some customized but legal media types like `application/octet-stream`(Docker are using) will be rejected, and `TestBackwardsCompatibilityManifest` be failed. 

I think this PR should be correct, PTAL, thanks.

schema/defs-image.json
```diff
@@ -4,7 +4,7 @@
     "mediaType": {
       "id": "https://opencontainers.org/schema/image/mediaType",
       "type": "string",
-      "pattern": "^[a-z]+/[0-9a-zA-Z.+]+$"
+      "pattern": "^[a-z]+/[0-9a-zA-Z-_+.]+$"
```
wking: This is still not quite right.  I've spun off a PR focused on this issue and descriptor tests (pulling in one of your commits from this series) in #448.  Although it would be good if we could get some of these validation PRs landed :p.  Having so many in the air at once is making it hard to keep them orthogonal ;).

schema/content-descriptor.json
```diff
@@ -6,14 +6,15 @@
...
-      "$ref": "defs-image.json#definitions/mediaType"
+      "$ref": "defs-image.json#/definitions/mediaType"
     },
     "size": {
       "description": "the size in bytes of the referenced object",
       "$ref": "defs.json#/definitions/int64"
     },
     "digest": {
-      "$ref": "defs-image.json#definitions/digest"
+      "description": "the cryptographic checksum digest of the object, in the pattern '<hash>:<hexadecimal digest>'",
```
xiekeyang: I carry on the description from https://github.com/opencontainers/image-spec/blob/master/schema/defs-image.json#L10 directly. I think it had better to keep same words for one entry.

xiekeyang:
@wking
I drop my commit about media type, which is not really quite right.
 I patch your commit https://github.com/opencontainers/image-spec/pull/444/commits/d05baa5f802c895669443191d6b973bf2e3d55f0 directly, for go pass the unit test(maybe we need more work about https://github.com/opencontainers/image-spec/pull/444/commits/d05baa5f802c895669443191d6b973bf2e3d55f0).
Our PRs seems conflict, which should be resolved after one landed.

wking:
On Sat, Nov 05, 2016 at 09:52:12PM -0700, xiekeyang wrote:
> I patch your commit
> https://github.com/opencontainers/image-spec/pull/444/commits/d05baa5f802c895669443191d6b973bf2e3d55f0
> directly, for go pass the unit test…

To unpack that comment a bit:

You've cherry-picked that in in place of your previous 2da8885.  And
you need one or the other to cover application/octet-stream (from [1])
not fitting the current master's media type pattern [2].

> Our PRs seems conflict, which should be resolved after one landed.

Sounds good.

[1]: https://github.com/opencontainers/image-spec/blob/7cab1f8c0dcccbfd23201b87c6591390d0ecd706/schema/manifest_backwards_compatibility_test.go#L146
[2]: https://github.com/opencontainers/image-spec/blob/7cab1f8c0dcccbfd23201b87c6591390d0ecd706/schema/defs-image.json#L7

xiekeyang:
@stevvooe @vbatts This PR is fixing definitions bug in schema. I think it to be ready, and PTAL.

xiekeyang:
Rebased, PTAL.

schema/defs-image.json
```diff
@@ -4,7 +4,7 @@
     "mediaType": {
       "id": "https://opencontainers.org/schema/image/mediaType",
       "type": "string",
-      "pattern": "^[a-z]+/[0-9a-zA-Z.+]+$"
+      "pattern": "^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}/[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}$"
```
stevvooe: Is `^ and `#` really allowed? What is the source of these additions?

stevvooe: I wonder if we should just do an enum here. Is there a way we can define an enum, then have callers of the validator add values to it as needed?

wking: That change is cherry-picked from #448, which focuses on updating the `mediaType` pattern to match [RFC 6838][1].  And yeah, for some reason they allow both `#` and `^` :p.

I don't think an enum would work, because descriptors are generic.  On the other hand, particular *instances* of descriptors may not be generic, and #404 is an example of using enums in those cases.

[1]: https://tools.ietf.org/html/rfc6838#section-4.2

xiekeyang:
@stevvooe We've discussed in some issues before, and finally think it is best to let json schema pattern just cover spec declaration, and check future more via strict controllable function.
Customized media type is self-defined. We should allow and can't use enum to involve it.
I pick @wking commit to use media type pattern. I review it, and sure it match [rfc6838](https://tools.ietf.org/html/rfc6838#section-4.2), which we had better to follow up now.

xiekeyang:
@stevvooe If the above comment https://github.com/opencontainers/image-spec/pull/444#issuecomment-262144283 and this patch is acceptable? Thanks a lot!

vbatts:
LGTM

xiekeyang:
Rebased, PTAL, thanks a lot!

vbatts:
LGTM

xiekeyang:
@opencontainers/image-spec-maintainers I think this is ready to me merged. PTAL, thanks.

vbatts:
LGTM

xiekeyang:
@opencontainers/image-spec-maintainers  rebased, PTAL.

schema/defs-image.json
xiekeyang: @stevvooe enum seems too strict, which will lead customized types failed during verifying. I checked `RFC 6838`, the pattern really follows it. I think `RFC 6838`  rule should be OK here.

xiekeyang:
I worried that schema bugs will exist in spec without this PR. That let schema work incorrectly.

schema/manifest_test.go
```diff
@@ -89,6 +89,23 @@ func TestManifest(t *testing.T) {
...
+		// expected failure: config.digest is not hex hashed format
+		{
+			manifest: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
+  "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
+    "size": 1470,
+    "digest": "sha256:c86f7763873b6c0aae22d9h3bab59b4f5debbed6685761b5951584f6efb0633b"
```
stevvooe: Could you make this less subtle?

There is an `h` in the middle for anyone not keeping count.

stevvooe:
LGTM

I do have a nit but you can PR that later.

schema/manifest_test.go
xiekeyang: fixed in https://github.com/opencontainers/image-spec/pull/444/commits/003ac5f168dbc83f3793352ce4820975c8173c08#diff-6493def706b32045f8d30e7a4b860c46R101

xiekeyang: need re-LGTM thanks a lot!

stevvooe:
LGTM

xiekeyang:
@stevvooe @vbatts Rebase again, please re-LGTM, thanks a lot!

stevvooe:
LGTM

xiekeyang:
@vbatts Please re-LGTM, thanks.

xiekeyang:
@vbatts should be ready to merge.

vbatts:
LGTM

PR #446, opened 2016-11-05
manifest test case should involve required fields

xiekeyang:
This test case want to return failure only when `invalid` media type.
However, it lack the required fields, and will return failure whatever `mediaType` is.
So, it should be completed the accurate required fields, to make sure it to be our expectation.

### DIFF
--- a/schema/manifest_test.go
+++ b/schema/manifest_test.go
@@ -31,7 +31,19 @@ func TestManifest(t *testing.T) {
 			manifest: `
 {
   "schemaVersion": 2,
-  "mediaType": "invalid"
+  "mediaType": "invalid",
+  "config": {
+    "mediaType": "application/vnd.oci.image.config.v1+json",
+    "size": 1470,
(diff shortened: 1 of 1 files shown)


wking:
On Sat, Nov 05, 2016 at 04:40:37AM -0700, xiekeyang wrote:
> This test case want to return failure only when `invalid` media type.
> However, it lack the required fields, and will return failure whatever `mediaType` is.
> So, it should be completed the accurate required fields, to make sure the test aim.

fb409fd1 looks good to me.

Now it fails with:

  mediaType: Does not match pattern '^[a-z]+/[0-9a-zA-Z.+]+$'

So a similar test with mediaType set to text/png would pass, although
the manifest would obviously still be invalid.  The fix for *that* is
in #404.

xiekeyang:
@stevvooe @vbatts This PR is what I think to be ready, and PTAL, thanks.

philips:
LGTM

jonboulle:
LGTM

PR #333, opened 2016-09-21
config: reformat and add optional/required

philips:
Reformat this file to look more like manifest.md and add
OPTIONAL/REQUIRED tags on all fields.

Fixes https://github.com/opencontainers/image-spec/issues/251

PR #447, opened 2016-11-05
*: Replace '*array*' with '*array of {type}*'

wking:
Covering all cases turned up by:

    $ git grep '\*array'

runtime-spec does this pretty consistently, and the first such usage appeared in image-spec with f3d0f78 (config: reformat and add optional/required, 2016-09-20, #333).  And before that, c22ca799 (serialization: docker v1 image format media type, 2016-04-04, #6) had landed some:

    <code>array of strings</code>


philips:
LGTM

jonboulle:
lgtm, thank you

PR #448, opened 2016-11-05
schema: Add descriptor tests and fix mediaType pattern

wking:
Cherry-picking the $ref fix from #444, since there was a lot going on
in that PR around the manifest list.  Fixing those things is good, but
I thought it would help review to have something that just focused on
descriptor bugs.

CC @xiekeyang.

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -19,7 +19,8 @@ The following fields contain the primary properties that constitute a Descriptor
 
 - **`mediaType`** *string*
 
-  This REQUIRED property contains the MIME type of the referenced content.
+  This REQUIRED property contains the media type of the referenced content.
+  Values MUST comply with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2].
 
   The OCI image specification defines [several of its own MIME types](media-types.md) for resources defined in the specification.
 
@@ -139,4 +140,6 @@ In the following example, the descriptor indicates that the referenced manifest 
 ```
 
 [rfc3986]: https://tools.ietf.org/html/rfc3986
+[rfc6838]: https://tools.ietf.org/html/rfc6838
+[rfc6838-s4.2]: https://tools.ietf.org/html/rfc6838#section-4.2
 [rfc7230-s2.7]: https://tools.ietf.org/html/rfc7230#section-2.7
--- a/schema/descriptor_test.go
+++ b/schema/descriptor_test.go
@@ -1,0 +1,213 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package schema_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/opencontainers/image-spec/schema"
+)
+
+func TestDescriptor(t *testing.T) {
+	for i, tt := range []struct {
+		descriptor string
+		fail     bool
+	}{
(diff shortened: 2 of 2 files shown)


schema/defs-image.json
```diff
@@ -4,7 +4,7 @@
     "mediaType": {
       "id": "https://opencontainers.org/schema/image/mediaType",
       "type": "string",
-      "pattern": "^[a-z]+/[0-9a-zA-Z.+]+$"
+      "pattern": "^[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}/[A-Za-z0-9][A-Za-z0-9!#$&-^_.+]{0,126}$"
```
wking: We previously didn't limit the length at all ;).  I'd rather go with the most generous interpretation of the current RFC (which I think matches what I have here :p) and leave it to consumers to add additional checks if they like.

wking:
On Sat, Nov 05, 2016 at 10:03:13PM -0700, xiekeyang wrote:
> meanwhile, we should also keep test cases in `manifest` and
> `manifest list` files cover important definitions for descriptor,
> because we have been aware that they use different schema branch
> sometime…

+1 for exactly this reason.

descriptor.md
```diff
@@ -19,7 +19,8 @@ The following fields contain the primary properties that constitute a Descriptor
 
 - **`mediaType`** *string*
 
-  This REQUIRED property contains the MIME type of the referenced content.
+  This REQUIRED property contains the media type of the referenced content.
+  Values MUST comply with [RFC 6838][rfc6838].
```
stevvooe: Restrict this `MUST` to the format in the rfc.

wking: Was there a particular normative portion of RFC 6838 that you feel does not apply to `mediaType` values?  For example, the [vendor tree][1] semantics are outside of the [naming requirements section][2], and I think both of those apply to `mediaType` values.

[1]: https://tools.ietf.org/html/rfc6838#section-3.2
[2]: https://tools.ietf.org/html/rfc6838#section-4.2

wking:
Rebased onto master now that #444 has landed.

descriptor.md
stevvooe: Just reference 4.2 directly. Most of the rest of the specification is related to registration.

wking: > Most of the rest of the specification is related to registration

Doesn't all the registration stuff still apply?  For example [§3.2][1] has:

> While public exposure and review of media types to be registered in the vendor tree are not required, using the media-types@iana.org mailing list for review is encouraged, to improve the quality of those specifications.

so I think our unregistered vendor types are fine.  And I don't think our spec should allow unregistered standards-tree type (violating [§3.1]) or other RFC-violating media types, even if they happen to match the §4.2 syntax.

If you'd like to draw attention to §4.2 in particular, I'd be fine with:

> Values MUST comply with [RFC 6838][2], including the naming requirements in it's [§4.2][3].

[1]: https://tools.ietf.org/html/rfc6838#section-3.2
[2]: https://tools.ietf.org/html/rfc6838
[3]: https://tools.ietf.org/html/rfc6838#section-4.2
[4]: https://tools.ietf.org/html/rfc6838#section-3.1

stevvooe: @wking That's fine.

wking: > @wking That's fine.

Done with 119ef52 → 15dbdf5.

stevvooe:
LGTM

wking:
This PR currently fails Travis [1], but the problem is unrelated
(#530).

[1]: https://travis-ci.org/opencontainers/image-spec/builds/195291380#L361

wking:
On Wed, Jan 25, 2017 at 04:25:28PM -0800, W. Trevor King wrote:
> This PR currently fails Travis [1], but the problem is unrelated
> (#530).

Actually, since #444 picked up my JSON Schema change (15f9f74), I can
drop the fs.go rebuild from this PR (it's covered by #530 and #531).
I've done that with 15dbdf5 → 070cd2a, and now Travis is happy.

stevvooe:

wking:
I just pushed 070cd2a → be553fb fixing ‘SHA256’ → ‘sha256’ for the
“invalid hash characters” test.  Now SHA256 only occurs in the
“invalid aglorithm characters” test.

vbatts:
LGTM

stevvooe:
LGTM

PR #432, opened 2016-11-01
[ReadMe] Add Compliance Language

vbatts:
Add language parallel to what is in the OCI Runtime.

This is a carry of #381, with fixed whitespace.

Signed-off-by: Rob Dolin <robdolin@microsoft.com>
Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

PR #452, opened 2016-11-07
Support different levels validation for JSON reader stream

xiekeyang:
As previous discussion among developers, to use selectable strict mode, for validating media type of referenced objects.  After enabling strict media types unmatched to OCI will be expected failure.
* each object media type in manifest and manifest list should be validated.
* add unit test cases for un/strict mode, with expected result.
* image-tool repo should introduce the new flag if this PR be landed.
* rename previous functions from xxDescendants to xxMediaType, because functions validate all objects media type, not only descendants.

### DIFF
--- a/schema/manifest_test.go
+++ b/schema/manifest_test.go
@@ -193,7 +193,7 @@ func TestManifest(t *testing.T) {
 		},
 	} {
 		r := strings.NewReader(tt.manifest)
-		err := schema.ValidatorMediaTypeManifest.Validate(r)
+		err := schema.ValidatorMediaTypeManifest.Validate(r, []schema.ValidateFunc{schema.ValidateSchema})
 
 		if got := err != nil; tt.fail != got {
 			t.Errorf("test %d: expected validation failure %t but got %t, err %v", i, tt.fail, got, err)
--- a/schema/spec_test.go
+++ b/schema/spec_test.go
@@ -77,7 +77,8 @@ func validate(t *testing.T, name string) {
 			continue
 		}
 
-		err = schema.Validator(example.Mediatype).Validate(strings.NewReader(example.Body))
+		err = schema.Validator(example.Mediatype).Validate(strings.NewReader(example.Body),
+			[]schema.ValidateFunc{schema.ValidateSchema})
 		if err == nil {
 			printFields(t, "ok", example.Mediatype, example.Title)
 			t.Log(example.Body, "---")
--- a/schema/validator.go
+++ b/schema/validator.go
@@ -26,14 +26,42 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
+type validateRefMediaFunc func(r io.Reader) error
+
+var validateRefMediaMap = map[Validator]validateRefMediaFunc{
+	ValidatorMediaTypeImageIndex: validateImageIndexRefObject,
+	ValidatorMediaTypeManifest:   validateManifestRefObject,
+}
+
+// ValidateFunc provides the type of function to validate specific object.
+type ValidateFunc func(r io.Reader, v Validator) error
+
 // Validator wraps a media type string identifier
 // and implements validation against a JSON schema.
 type Validator string
 
-type validateDescendantsFunc func(r io.Reader) error
+// Validate validates the given read r to memory and call funcList on the content.
+// r: the given reader.
+// funcList: the optional functions to validate the wrapped media type.
+func (v Validator) Validate(r io.Reader, funcList []ValidateFunc) error {
+	buf, err := ioutil.ReadAll(r)
+	if err != nil {
+		return errors.Wrap(err, "unable to read the document file")
+	}
 
-var mapValidateDescendants = map[Validator]validateDescendantsFunc{
-	ValidatorMediaTypeManifest: validateManifestDescendants,
+	for _, f := range funcList {
+		err := f(bytes.NewReader(buf), v)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+type unimplemented string
+
+func (v unimplemented) Validate(src io.Reader) error {
+	return fmt.Errorf("%s: unimplemented", v)
 }
 
 // ValidationError contains all the errors that happened during validation.
@@ -45,21 +73,13 @@ func (e ValidationError) Error() string {
 	return fmt.Sprintf("%v", e.Errs)
 }
 
-// Validate validates the given reader against the schema of the wrapped media type.
-func (v Validator) Validate(src io.Reader) error {
+// ValidateSchema validates the given reader against the schema of the wrapped media type.
+// src: the given reader.
+// v: the expected media type.
+func ValidateSchema(src io.Reader, v Validator) error {
 	buf, err := ioutil.ReadAll(src)
 	if err != nil {
 		return errors.Wrap(err, "unable to read the document file")
-	}
-
-	if f, ok := mapValidateDescendants[v]; ok {
-		if f == nil {
-			return fmt.Errorf("internal error: mapValidateDescendents[%q] is nil", v)
-		}
-		err = f(bytes.NewReader(buf))
-		if err != nil {
-			return err
-		}
 	}
 
 	sl := gojsonschema.NewReferenceLoaderFileSystem("file:///"+specs[v], fs)
@@ -86,16 +106,26 @@ func (v Validator) Validate(src io.Reader) error {
 	}
 }
 
-type unimplemented string
+// ValidateRefMedia validates the referenced objects against OCI media type defined in spec.
+// src: the given reader.
+// v: the expected media type.
+func ValidateRefMedia(src io.Reader, v Validator) error {
+	f, ok := validateRefMediaMap[v]
+	if ok {
+		if f == nil {
+			return fmt.Errorf("referenced media type validation %q unimplemented", v)
+		}
 
-func (v unimplemented) Validate(src io.Reader) error {
-	return fmt.Errorf("%s: unimplemented", v)
+		return f(src)
+	}
+
+	return nil
 }
 
-func validateManifestDescendants(r io.Reader) error {
+func validateManifestRefObject(src io.Reader) error {
 	header := v1.Manifest{}
 
-	buf, err := ioutil.ReadAll(r)
+	buf, err := ioutil.ReadAll(src)
 	if err != nil {
 		return errors.Wrapf(err, "error reading the io stream")
 	}
@@ -115,6 +145,27 @@ func validateManifestDescendants(r io.Reader) error {
 			layer.MediaType != string(v1.MediaTypeImageLayerNonDistributable) &&
 			layer.MediaType != string(v1.MediaTypeImageLayerNonDistributableGzip) {
 			fmt.Printf("warning: layer %s has an unknown media type: %s\n", layer.Digest, layer.MediaType)
+		}
+	}
+	return nil
+}
+
+func validateImageIndexRefObject(src io.Reader) error {
+	header := v1.ImageIndex{}
+
+	buf, err := ioutil.ReadAll(src)
+	if err != nil {
+		return errors.Wrapf(err, "error reading the io stream")
+	}
+
+	err = json.Unmarshal(buf, &header)
+	if err != nil {
+		return errors.Wrap(err, "manifest list format mismatch")
+	}
+
+	for _, manifest := range header.Manifests {
+		if manifest.MediaType != string(v1.MediaTypeImageManifest) {
+			return fmt.Errorf("manifest %s has an unknown media type: %s", manifest.Digest, manifest.MediaType)
 		}
 	}
 	return nil
--- a/schema/backwards_compatibility_test.go
+++ b/schema/backwards_compatibility_test.go
@@ -116,7 +116,7 @@ func TestBackwardsCompatibilityImageIndex(t *testing.T) {
 
 		imageIndex := convertFormats(tt.imageIndex)
 		r := strings.NewReader(imageIndex)
-		err := schema.ValidatorMediaTypeImageIndex.Validate(r)
+		err := schema.ValidatorMediaTypeImageIndex.Validate(r, []schema.ValidateFunc{schema.ValidateSchema})
 
 		if got := err != nil; tt.fail != got {
 			t.Errorf("test %d: expected validation failure %t but got %t, err %v", i, tt.fail, got, err)
@@ -178,7 +178,7 @@ func TestBackwardsCompatibilityManifest(t *testing.T) {
 
 		manifest := convertFormats(tt.manifest)
 		r := strings.NewReader(manifest)
-		err := schema.ValidatorMediaTypeManifest.Validate(r)
+		err := schema.ValidatorMediaTypeManifest.Validate(r, []schema.ValidateFunc{schema.ValidateSchema})
 
 		if got := err != nil; tt.fail != got {
 			t.Errorf("test %d: expected validation failure %t but got %t, err %v", i, tt.fail, got, err)
@@ -217,7 +217,7 @@ func TestBackwardsCompatibilityConfig(t *testing.T) {
 
 		config := convertFormats(tt.config)
 		r := strings.NewReader(config)
-		err := schema.ValidatorMediaTypeImageConfig.Validate(r)
+		err := schema.ValidatorMediaTypeImageConfig.Validate(r, []schema.ValidateFunc{schema.ValidateSchema})
 
 		if got := err != nil; tt.fail != got {
 			t.Errorf("test %d: expected validation failure %t but got %t, err %v", i, tt.fail, got, err)
--- a/schema/config_test.go
+++ b/schema/config_test.go
@@ -212,7 +212,7 @@ func TestConfig(t *testing.T) {
 		},
 	} {
 		r := strings.NewReader(tt.config)
-		err := schema.ValidatorMediaTypeImageConfig.Validate(r)
+		err := schema.ValidatorMediaTypeImageConfig.Validate(r, []schema.ValidateFunc{schema.ValidateSchema})
 
 		if got := err != nil; tt.fail != got {
 			t.Errorf("test %d: expected validation failure %t but got %t, err %v", i, tt.fail, got, err)
--- a/schema/descriptor_test.go
+++ b/schema/descriptor_test.go
@@ -204,7 +204,7 @@ func TestDescriptor(t *testing.T) {
 		},
 	} {
 		r := strings.NewReader(tt.descriptor)
-		err := schema.ValidatorMediaTypeDescriptor.Validate(r)
+		err := schema.ValidatorMediaTypeDescriptor.Validate(r, []schema.ValidateFunc{schema.ValidateSchema})
 
 		if got := err != nil; tt.fail != got {
 			t.Errorf("test %d: expected validation failure %t but got %t, err %v", i, tt.fail, got, err)
--- a/schema/imageindex_test.go
+++ b/schema/imageindex_test.go
@@ -239,7 +239,7 @@ func TestImageIndex(t *testing.T) {
 		},
 	} {
 		r := strings.NewReader(tt.imageIndex)
-		err := schema.ValidatorMediaTypeImageIndex.Validate(r)
+		err := schema.ValidatorMediaTypeImageIndex.Validate(r, []schema.ValidateFunc{schema.ValidateSchema})
 
 		if got := err != nil; tt.fail != got {
 			t.Errorf("test %d: expected validation failure %t but got %t, err %v", i, tt.fail, got, err)


wking:
There's a reasonable amount of overlap between this PR and #403, but
I'm happy to rebase #403 if this one lands first.

schema/validator.go
```diff
@@ -105,13 +137,26 @@ func validateManifestDescendants(r io.Reader) error {
 		return errors.Wrap(err, "manifest format mismatch")
 	}
 
+	if header.MediaType != string(v1.MediaTypeImageManifest) {
+		if strict {
+			return fmt.Errorf("error: manifest has an unknown media type: %s\n", header.MediaType)
+		}
+		fmt.Printf("warning: manifest has an unknown media type: %s\n", header.MediaType)
```
xiekeyang: Updated in 
https://github.com/opencontainers/image-spec/pull/452/commits/6934d41165d1441fe3bfd67dbd87e827395ce8d1
https://github.com/opencontainers/image-spec/pull/452/commits/3da0cd8feb09f373d0a8e99a5c430fc49f23667a
https://github.com/opencontainers/image-spec/pull/452/commits/a2d0b365daafe02246c57221f181277e988ce799

I code it as,
```go
  msg := fmt.Sprintf(...)
  if strict {
          return fmt.Errorf("error: %s", msg)
  }   
  fmt.Printf("warning: %s\n", msg)
```

I honestly like `fmt.Sprintf` to generate message, because it may be error or warning type.
`fmt.Errorf` and `fmt.Printf` can distinguish returning or omitting behavior.

xiekeyang:
> There's a reasonable amount of overlap between this PR and #403, but
I'm happy to rebase #403 if this one lands first.

Ah they have really some overlap. We should resolve the conflict if one landed.

schema/validator.go
```diff
@@ -88,11 +89,43 @@ func (v Validator) Validate(src io.Reader) error {
...
+
+	err = json.Unmarshal(buf, &header)
+	if err != nil {
+		return errors.Wrap(err, "manifest list format mismatch")
+	}
+
+	if header.MediaType != string(v1.MediaTypeImageManifestList) {
+		msg := fmt.Sprintf("manifest list has an unknown media type: %s", header.MediaType)
+		if strict {
+			return fmt.Errorf("error: %s", msg)
```
wking: I'm fine with the “warning: ” prefix you have below, but I don't think you need an “error: ” prefix for the error string.  You know it's an error because its Go type is `error`, having “error” in the string payload is redundant.  If you want to display these to users with an error prefix (which is fine with me), then prefix the error with “error: ” when you print it to stderr (or wherever).  But don't add the prefix here.

schema/validator.go
```diff
@@ -88,11 +89,43 @@ func (v Validator) Validate(src io.Reader) error {
...
+		}
+		fmt.Printf("warning: %s\n", msg)
+	}
+
+	for _, manifest := range header.Manifests {
+		if manifest.MediaType != string(v1.MediaTypeImageManifest) {
+			if strict {
+				return fmt.Errorf("error: manifest %s has an unknown media type: %s\n", manifest.Digest, manifest.MediaType)
+			}
+			fmt.Printf("warning: manifest %s has an unknown media type: %s\n", manifest.Digest, manifest.MediaType)
```
wking: You can do the same `Sprintf` (or whatever) unification here.

schema/validator.go
xiekeyang: Agree with you and fix in https://github.com/opencontainers/image-spec/pull/452/commits/25825f3505911d69915bcfd9b2df083c23ed4e02 https://github.com/opencontainers/image-spec/pull/452/commits/7b138962e1a4b5b83da3f45c20c3f9babdc61c67 and https://github.com/opencontainers/image-spec/pull/452/commits/4daf4f71a42487d551ca41a6598d9f057d497348

I have hesitated to this, and then comply that we are usual to wrap error message. Actually I like error messages to be concise.

schema/validator.go
xiekeyang: Yeah, it is a nit, fix in https://github.com/opencontainers/image-spec/pull/452/commits/25825f3505911d69915bcfd9b2df083c23ed4e02#diff-4a672c41642ea30ebf88bb87685cf72aR119

schema/manifestlist_test.go
```diff
@@ -0,0 +1,106 @@
+// Copyright 2016 The Linux Foundation
```
wking: This looks like it will conflict with your in-flight #444 (also adding a `schema/manifestlist_test.go` from scratch).

xiekeyang:
@stevvooe @vbatts 
This PR is adding strict mode for schema media type, as maintainers suggestion. I think it is ready, PTAL, thanks.

schema/validator.go
```diff
@@ -46,17 +47,17 @@ func (e ValidationError) Error() string {
 }
 
 // Validate validates the given reader against the schema of the wrapped media type.
-func (v Validator) Validate(src io.Reader) error {
+func (v Validator) Validate(src io.Reader, strict bool) error {
```
philips: can you please describe what strict does?

schema/validator.go
```diff
@@ -88,11 +89,44 @@ func (v Validator) Validate(src io.Reader) error {
...
+		fmt.Printf("warning: %s\n", msg)
+	}
+
+	for _, manifest := range header.Manifests {
+		if manifest.MediaType != string(v1.MediaTypeImageManifest) {
+			msg := fmt.Sprintf("manifest %s has an unknown media type: %s", manifest.Digest, manifest.MediaType)
+			if strict {
+				return fmt.Errorf("%s", msg)
+			}
+			fmt.Printf("warning: %s\n", msg)
```
philips: erm, I don't think this package should just randomly printf stuff

xiekeyang: @philips , here I keep previous non-strict behavior(warning) unchanged. And I think it seems had better to warn customer his media-type is not OCI format, even be accepted.
Or do you have some proposal for this?

schema/validator.go
xiekeyang: I add doc in https://github.com/opencontainers/image-spec/pull/452/commits/d9b5499553734cfb3e6fcc1588a230ce304560cf#diff-4a672c41642ea30ebf88bb87685cf72aR49. PTAL.

schema/validator.go
wking: I think we should stick to printing warnings for this PR.

In follow-up work we can add a validation argument which can hold a test-harness handle (e.g. a more generic form of the [`harness *tap.T`][1] I mocked up in opencontainers/image-tools@4f0b3c1b).  Then these test functions can register errors and warnings without worrying about how the results are being collected or displayed.  But we don't need to bite all of that off in one chunk ;).

[1]: https://github.com/opencontainers/image-tools/commit/4f0b3c1bae97d212680dc73ee8b35b6e47b8a8c6#diff-8351286f57eb81e245c9c99c07f3b34fR28

schema/manifestlist_test.go
```diff
@@ -0,0 +1,106 @@
...
+		manifestList string
+		fail         bool
+	}{
+		// valid manifest list accepting customized media types
+		{
+			strict: false,
+			manifestList: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/customized+json",
```
wking: This [fails now][1] (since #404), because your example is not a manifest list.  I think you want to update this `mediaType` to `application/vnd.oci.image.manifest.list.v1+json`, but you should be able to leave the rest of this particular JSON blob alone.

[1]: https://travis-ci.org/opencontainers/image-spec/builds/176600776#L350

schema/manifestlist_test.go
```diff
@@ -0,0 +1,106 @@
...
+			fail: false,
+		},
+
+		// expected failure strictly: manifestList.mediaType is unknown
+		{
+			strict: true,
+			manifestList: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/customized+json",
```
wking: This will fail regardless of strict (since #404).  I'm fine leaving a test for that in place (with `fail: true`), but the test should use `strict: false` to show that the failure is due to you not passing a manifest-list, and not due to particular validation strictness.

schema/manifestlist_test.go
wking: I haven't looked into #444 again, but I think `manifestlist_test.go` needs a success test with OCI-defined `mediaType`s for both the root `mediaType` and the `manifests[0].mediaType`.

schema/validator.go
```diff
@@ -46,17 +47,18 @@ func (e ValidationError) Error() string {
 }
 
 // Validate validates the given reader against the schema of the wrapped media type.
-func (v Validator) Validate(src io.Reader) error {
+// strict: accept the customized media type when value is false, or reject when true.
```
xiekeyang: @wking I think your words may be good as help message for customer, but I really want to present the design purpose here, as my words.

xiekeyang:
@opencontainers/image-spec-maintainers @wking ,
Now validation rule for JSON object is that the root object media-type MUST be OCI format, or will be reject. 
But we allow the referenced objects, like manifest in manifest-list, config in manifest, to be not OCI format(non-strict mode).
If we really want that? I remember we had discussed this, but I still not fully understand why. WDYT?
I would fix this patch after agreement achieved.

xiekeyang:
@opencontainers/image-spec-maintainers @wking 
I fix the patch, in accordance with #404. 
* Only valid referenced object;
* Unit test only for referenced object;

PTAL

schema/manifest_test.go
```diff
@@ -113,9 +118,75 @@ func TestManifest(t *testing.T) {
...
+			manifest: `
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
+  "config": {
+    "mediaType": "application/vnd.customized.config+json",
+    "size": 1470,
+    "digest": "sha256:c86f7763873b6c0aae22d963bab59b4f5debbed6685761b5951584f6efb0633b"
+  },
+  "layers": []
```
wking: This is currently illegal (although possibly not validated).  See #407, which is trying to make it legal.  For the purpose of this test, you should use the usual strictly-valid dummy layer from the other tests.

schema/spec_test.go
```diff
@@ -73,7 +73,7 @@ func validate(t *testing.T, name string) {
 			continue
 		}
 
-		err = schema.Validator(example.Mediatype).Validate(strings.NewReader(example.Body))
+		err = schema.Validator(example.Mediatype).Validate(strings.NewReader(example.Body), false)
```
wking: It seems like we should be able to validate spec examples strictly.

schema/manifest_test.go
xiekeyang: @wking , but current schema really allows empty layer slice(I've debug it). I think spec also allow empty. I honest dislike to avoid unclear of undetermined stuff in specific patch.
To add more test case for optional layers is another story(IMO).

schema/validator.go
xiekeyang: fixed in https://github.com/opencontainers/image-spec/pull/452/commits/b68d69226243c79b275a3c5caeac04e3dad84231#diff-4a672c41642ea30ebf88bb87685cf72aR50

schema/spec_test.go
xiekeyang: Yes, you are right. This test is for spec validation, which MUST be strict OCI format coincidence. `false` will lead this test case loss effectiveness. fix in https://github.com/opencontainers/image-spec/pull/452/commits/b68d69226243c79b275a3c5caeac04e3dad84231#diff-5b73117945d63fe6253dccb9c210d74fR76.

schema/manifest_test.go
wking: On Thu, Nov 17, 2016 at 06:00:35PM -0800, xiekeyang wrote:
> … but current schema really allows empty layer slice (I've debug it). I think spec also allow empty…

I don't think the spec allows empty layers, based [on][1]:

> The array MUST have the base image at index 0.

#407, and #318 before it, are attempting to lift that restriction.  But I agree that you don't want to get involved in that discussion in this PR.  By using the usual single-entry `layers` array here, you would avoid getting involved.

[1]: https://github.com/opencontainers/image-spec/blame/de440dbfe41f35eb03117204bc99eeb0e6a186b8/manifest.md#L45

wking:
The diff from both b68d692 and a183979 looks fine to me, although I
think we might want to squash the commits together.  As it stands
running the test suite on just b68d692 raises:

  $ make test
  go test -race -cover github.com/opencontainers/image-spec/schema github.com/opencontainers/image-spec/specs-go github.com/opencontainers/image-spec/specs-go/v1
  # github.com/opencontainers/image-spec/schema_test
  schema/manifest_test.go:118: not enough arguments in call to schema.MediaTypeManifest.Validate
  FAIL    github.com/opencontainers/image-spec/schema [build failed]
  …

xiekeyang:
squashed in https://github.com/opencontainers/image-spec/pull/452/commits/833391c49b3f170e4e78836edb3d17331812ecb8

wking:
833391c looks good to me :).

stevvooe:
How can we define a validation framework that can support different levels of validation? This PR sub-divides the validation space into media types and not media types. This creates the problem that we now have to define what belongs in strict and what doesn't.

Wouldn't it make sense if we have options to control the suites of validation? For example, to get the "strict" behavior, one might configure this as:

```go
Validate(r, ValidateFields, ValidateMediaTypes)
```

We could also have explicit sets of validation that meet certain support-levels:

```go
var ValidateV1Strict []Validator{ValidateFields, ValidateMediaTypes}

Validate(r, ValidateV2Strict...)
```

Without a definition of what "strict" means, we'll be in the position that this will be under definition of the implementation, where if we are clear what is being validated, we can assure users they are in line with the specification.

wking:
On Mon, Nov 21, 2016 at 01:24:03PM -0800, Stephen Day wrote:
> This creates the problem that we now have to define what belongs in
> strict and what doesn't.

There's a definition here [1].  I think we can tighten that defintion
up (maybe on the heels of #432?), but the wording in [1] seems like a
reasonable start (and we can update it if we think of better wording
before #432 gives us somewhere to put formal definitions).

> Wouldn't it make sense if we have options to control the suites of validation?…
> 
> ```go
> Validate(r, ValidateFields, ValidateMediaTypes)
> ```

I'm fine with this sort of breakdown, but if we want to support it I
think we're want to supply functions:

* Validate (everything)
* ValidateFields (only validate the fields)
* ValidateMediaTypes (only validate the referenced media types)

And it seems orthogonal to how strictly we do each of those
validations.  E.g. you can have “uses a field that's not defined in
the spec” passing a relaxed validation (it's legal [2]) but failing
strict validation (the extension field doesn't have OCI-defined
semantics).  Or you can have “uses a layer media type that's not
defined in the spec” passing relaxed validation (it's legal [3]) but
failing strict validation (an application/zip layer doesn't have
OCI-defined semantics).

[1]: https://github.com/opencontainers/image-spec/pull/452/files#diff-4a672c41642ea30ebf88bb87685cf72aR50
[2]: https://github.com/opencontainers/image-spec/blob/d65dbee74ecbb3fa6a07e2af6d091245da19fb1b/manifest.md#extensibility
[3]: https://github.com/opencontainers/image-spec/blame/d65dbee74ecbb3fa6a07e2af6d091245da19fb1b/manifest.md#L60

stevvooe:
@wking Please let the submitter respond to code review comments.

Again, you linked me to something without any context and I have no clue what you mean.

xiekeyang:
@stevvooe I mostly understand and agree your expectation. Let me refactor the PR and re-submit sooner.

xiekeyang:
@stevvooe 
With current PR, image tool usage will be like,
```
$ oci-valid-tool --strict=false customized-manifest.json
Warn: referenced config has an unknown media type: application/vnd.oci.customized.config.v1+json
OK!

$ oci-valid-tool --strict=true customized-manifest.json
Error: referenced config has an unknown media type: application/vnd.oci.customized.config.v1+json
Failed!
```
Customized json is failed under strict mode, and prompt warning non-strict mode. 

I think your proposal want tool to be more flexible, like,
```
$ oci-valid-tool --level [fields] customized-manifest.json
OK!

$ oci-valid-tool --level [fields,ref-media] customized-manifest.json
Failed!
```
And you want to involve all possible objects(in future) to `level` array, right?

xiekeyang:
I refactor patch according to @stevvooe proposal of level usage and function interface. It is done in https://github.com/xiekeyang/image-spec/blob/6b47a182a3b6499df2bead20016d6e4eec7fdb4e/schema/validator.go.
* To provide validating function array for different level objects. 
* To support schema and referenced media type validation.
* To allow image tools or consumer to select different validation levels for specific require.
* To add unit test cases for new interface.
cc @stevvooe @philips @vbatts PTAL.

xiekeyang:
@stevvooe @philips If the latest commit of https://github.com/opencontainers/image-spec/pull/452/commits/21ef4e86646f400f32186b1f3c5527d3d7cab89e and https://github.com/opencontainers/image-spec/pull/452/commits/d898abe12403701769557132def5981df82d2eda OK? PTAL, thanks a lot!

schema/validator.go
```diff
@@ -92,7 +99,56 @@ func (v unimplemented) Validate(src io.Reader) error {
...
+	}
+
+	err = json.Unmarshal(buf, &header)
+	if err != nil {
+		return errors.Wrap(err, "manifest list format mismatch")
+	}
+
+	for _, manifest := range header.Manifests {
+		if manifest.MediaType != string(v1.MediaTypeImageManifest) {
+			return fmt.Errorf("manifest %s has an unknown media type: %s", manifest.Digest, manifest.MediaType)
```
After former proposal from @stevvooe, I think we have made agreement. Current commit follows up the proposal thought, to use function slice. Consumer can select objects(fields, media type...) at will.
> I don't see a way to invoke this to error out when a media type is truly invalid (i.e. not compliant with RFC 6838, #448) vs. not necessarily supported by an OCI-compliant implementation (e.g. a non-OCI type). 

Under truly invalid case, JSON schema prompts error.
Under non-OCI type, validator will prompt error only if ref-type(here is media type) function is selected.
Honest I'm not very willing roll back the former strict framework, which has been rejected. And we had better to wait @stevvooe's feedback.
cc @stevvooe WDYT?

xiekeyang:
@stevvooe If current patch looks as what you want?

schema/validator.go
wking: > I don't think it belongs to this patch.

I agree that we don't want to cover the reserved `data` as part of this PR.  This PR is about two things:

* Broadly, setting a precedent for handling verification for behaviour which images MAY use but which implementations MUST support.
* Narrowly, adding optional validation for per-location media types, which are a special case of the broader issue.

So while I have no desire to expand the narrow target, I want to try and solve it in a way that will cover the broader issue.  I think the `strict` option sets a precedent which will scale well as we extend it to other spec statements.

I'm less clear on how the “pass in an array of validator functions” will scale.  It seems like it might end up with a separate validator for each spec statement?  Or maybe we'll end up with a `ValidateOnlyContainsOCIRequiredSettings` validator?  Or…?  It's not clear to me how we're expected to apply this approach to future spec statements.

vbatts:
This SGTM. Though @stevvooe recommends this needs a technical proposal on how this kind of feature ties to the requirements of compliance and certification

stevvooe:
@vbatts We can go forward with this approach until we have better requirements.

xiekeyang:
@vbatts  @stevvooe got it. I'd go on finishing this pr.

xiekeyang:
@stevvooe I think you agree the current implementation of pr, to present selectable media validating function, which is controlled by function-map. In future we can add more val function for another objects if necessary. Is that right?

xiekeyang:
@stevvooe I'm considering to use interface for `Validator` functions, but not figure out yet.

vbatts:
Is this needing to block a vote for RC5?

xiekeyang:
According to milestone, it is blocking RC5. If @stevvooe @wking @philips accept its map framework for validation, I can finish update in 2 or 3 days. Or for any other proposal, please put up freely, and I would take more analysis. Thanks a lot!

xiekeyang:
Or we can consider to move this milestone to next release.

xiekeyang:
@opencontainers/image-spec-maintainers I update this, to keep functional map, and add a bug fix commit. PTAL.

image-layout.md
```diff
@@ -134,7 +134,7 @@ The `imageLayoutVersion` value will align with the OCI Image Specification versi
 
 ### oci-layout Example
 
-```json
+```json,title=Image%20Layout&valid=enabling&mediatype=oci%2Dlayout
```
wking: I'm in favor of adding a title, but “Image Layout” is the whole structure, not just the `oci-layout` file.  Can we follow the section headers and call this “oci-layout file” or some such?

What is `valid=enabling` about?

I don't think we want to invent a one-off media type here.  If we want a media type for this file format (e.g. for example validation), I think we should declare a media type for the schema like we do with our other schemas (e.g. [here][1] and [here][2] for manifests).

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc4/manifest.md#L9
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc4/media-types.md#L7

image-layout.md
```diff
@@ -162,7 +162,7 @@ Those tags will often be represented in an image-layout repository with matching
 
 ### Index Example
 
-```json,title=Image%20Index&mediatype=application/vnd.oci.image.index.v1%2Bjson
+```json
```
wking: I think we still want to validate this example as an index even if it doesn't occur in `image-index.md`.  And titles seem like they'd always be a good idea.  Why are you removing this?

schema/validator.go
```diff
@@ -26,14 +26,42 @@ import (
 	"github.com/xeipuuv/gojsonschema"
 )
 
+type validateRefMediaFunc func(r io.Reader) error
+
+var vlidateRefMediaMap = map[Validator]validateRefMediaFunc{
```
wking: `vlidateRefMediaMap` → `validateRefMediaMap`.

schema/validator.go
```diff
@@ -26,14 +26,42 @@ import (
...
+	ValidatorMediaTypeImageIndex: validateImageIndexRefObject,
+	ValidatorMediaTypeManifest:   validateManifestRefObject,
+}
+
+// ValidateFunc provides the type of function to validate specific object.
+type ValidateFunc func(r io.Reader, v Validator) error
```
wking: If we expect to be applying a series of these to validate a single blob, using an `io.Reader` seems wasteful.  Is there some Go type magic we can use to pass a parsed JSON object through to the validator to let us only deserialize the JSON once?

schema/validator.go
```diff
@@ -26,14 +26,42 @@ import (
...
+
 // Validator wraps a media type string identifier
 // and implements validation against a JSON schema.
 type Validator string
 
-type validateDescendantsFunc func(r io.Reader) error
+// Validate validates the given reader against the schema and OCI format defined in spec.
```
wking: I don't see anything about “the schema” or “OCI format defined in the spec” in your current implementation (3ad5d1b).  This function is just “read `r` into memory and call all of `funcList` on the content, dying on the first error”.

schema/validator.go
```diff
@@ -86,13 +106,28 @@ func (v Validator) Validate(src io.Reader) error {
...
+		if f == nil {
+			return fmt.Errorf("referenced media type validation %q unimplemented", v)
+		}
 
-func (v unimplemented) Validate(src io.Reader) error {
-	return fmt.Errorf("%s: unimplemented", v)
+		buf, err := ioutil.ReadAll(r)
+		if err != nil {
+			return errors.Wrap(err, "unable to read the document file")
+		}
```
wking: No need to bother with this, since you're only calling a single `f`.  Just pass your reader along:

```
if ok {
  if f == nil {
    …
  }
  return f(r)
}

File: schema/validator.go [hunk 19]
```diff
@@ -86,13 +106,28 @@ func (v Validator) Validate(src io.Reader) error {
...
+
+		return f(bytes.NewReader(buf))
+	}
+
+	return nil
```
wking: I think we should be raising the same unimplemented error here as you currently have for `f == nil`.  How about:

```
f, ok := vlidateRefMediaMap[v]
if ok && f != nil {
    return f(r)
}
return fmt.Errorf("referenced media type validation %q unimplemented", v)
```

image-layout.md
xiekeyang: Sorry, `valid=enabling` is a redundant debugging item, should be removed.

image-layout.md
xiekeyang: I remove this title as not to valid it. I understand that you want to valid it. 
But I think `spec_test.go` should only valid those who are announced only in itself *.md. So I remove it, for it is just a reference. It is IMO, to be not sure.

image-layout.md
xiekeyang: I extract this and put it on #579.

schema/validator.go
xiekeyang: Yes, this should be fixed.

schema/validator.go
xiekeyang: I'm hard to agree for this. Some validations such as descriptor, layout doesn't exist in function map, which make no sense.  They should be ignore, but not prompt error.

schema/validator.go
xiekeyang: I didn't find out the better way yet. At the same time, I still think it had better to input io reader stream to. ValidateRefFuncs are public functions. Consumer should be allowed called them directly. Here we provide io reader to customer, who can use them freely, and it allows expend their job in future.

schema/validator.go
wking: > Some validations such as descriptor, layout doesn't exist in function map, which make no sense.

We can validate descriptors against the JSON Schema.  And if you want JSON Schema validation handled elsewhere, we can define a no-op validator for types where we know there are no children.  But I think silently ignoring validating for all types that have no validator registered is setting ourselves up for “oops, that wasn't valid, but we forgot to add a validator for $TYPE”.

schema/validator.go
wking: > ValidateRefFuncs are public functions.

That doesn't really matter.  We could always provide `io.Reader` and parsed-JSON flavored versions if we figure out how to write a function that takes an already-parsed-from-JSON object.

xiekeyang:
@stevvooe The stuff under discussion is that the argument variable for ValidRefFuncs should be io.Reader type or json type. I prefer to reader. WDYT?

xiekeyang:
I'm so regretful that I've make big effort on this, but it can't attract maintainers attention. Although I still think it really bring benefit according to pre discussion, I've to close it.

PR #457, opened 2016-11-15
manifest-list.md: fix example

coolljt0725:
The json tag define in
https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest_list.go#L34
for os.features is `os.features`, we should use
`os.features` in the example of manifest list.

### DIFF
--- a/manifest-list.md
+++ b/manifest-list.md
@@ -105,7 +105,7 @@ Instead they MUST ignore unknown properties.
       "platform": {
         "architecture": "amd64",
         "os": "linux",
-        "features": [
+        "os.features": [
           "sse4"
         ]
       }


jonboulle:
LGTM

wking:
On Tue, Nov 15, 2016 at 02:53:54AM -0800, Lei Jitang wrote:
> The json tag define in
> https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest_list.go#L34
> for os.features is `os.features`…

And the JSON Schema agrees [1].  So while these keys don't seem
consistent with our usual camelCase, bfec1c00 looks good to me.

[1]: https://github.com/opencontainers/image-spec/blob/7cab1f8c0dcccbfd23201b87c6591390d0ecd706/schema/defs-image.json#L63

stevvooe:
LGTM

Issue #287, opened 2016-09-09
cmd/oci-image-tool: is having empty refs/ a valid image?


runcom:
$SUBJECT and basically, is it valid to have stuff under `blobs/` not referenced? Can I still package it up and validation is ok given there are no refs/? Should we print a warning (which can be silenced)?

wking:
On Fri, Sep 09, 2016 at 01:40:15AM -0700, Antonio Murdaca wrote:
> $SUBJECT

I think so.  At least I don't see any “and there MUST be at least one
ref” language.  And if you cared about running images from the layour
that would be “and there MUST be at least one unpackable ref”
language.  And then you have to match the oci-layout version with the
media types that were unpackable when it was current (or is validity
in the eye of the validator?)… This is a bowl that I do not want to
fathom ;).

Maybe folks are just using an image-layout to ship some missing blobs.

I don't see any incentive to image-authors to publish ref-less blobs
and then pretend they are runnable, so I don't see a need to get into
the business of restricting refs.

> … and basically, is it valid to have stuff under `blobs/` not
> referenced?

Yes [1].

> Should we print a warning?

Warnings for “there are no refs” and “dangling blob sha256-abc…” sound
useful to me.  Maybe put the first behind ‘--log-level warn’ and the
second behind ‘--log-level info’ based on how likely I think they are
to be a mistake and how many lines will get printed if they exist ;).

[1]: https://github.com/opencontainers/image-spec/blame/e6a431f1584468f112a52c01a02d326199b99021/image-layout.md#L46

vbatts:
Both seem valid to me

runcom:
Alright, can we keep this open so we won't forget about adding some sort of warning/log when this happens

philips:
This issue was moved to opencontainers/image-tools#15

wking:
On Tue, Sep 20, 2016 at 07:46:54PM -0700, Brandon Philips wrote:
> This issue was moved to opencontainers/image-tools#15

It seems like whether a refs-less image-layout is valid or not is
primarily a question for the image-layout spec.  Or have we already
reached a consensus on that in favor of the spec as it stands (with no
requirement for having refs or restrictions on media types for those
refs).  The spec's current position makes sense to me.

If we've already reached a consensus on the spec position, the
remaining issue is just “should we warn users when validating layouts
without refs (with recognized types)?”, and that seems like it can be
handled in image-tools without spec clarity around the warning
condition.

PR #459, opened 2016-11-16
image-layout: Require blobs/refs to exist but allow them to be empty

wking:
The previous wording (“has” and “contains”) was not clear enough to [avoid confusion][1].  I consider this PR to be a spec clarification, and not a spec change, but others will [probably disagree][2] (which is why I think we need the clarification).

If you cared about running images from the layout, you'd need “and there MUST be at least one unpackable ref” language.  And then you have to match the oci-layout version with the media types that were unpackable when it was current (or is validity in the eye of the validator?)…  This is a bowl that I do not want to fathom ;).

Maybe folks are just using an image-layout to ship some missing blobs (and have refs empty).  I don't see any incentive to image-authors to publish ref-less blobs and then pretend they are runnable, so I don't see a need to get into the business of restricting refs.

Or maybe they're shipping some missing refs (and have blobs empty).  Maybe they expect all blobs to be fetched via the descriptor's `urls`.  Those sound fine to me too, so I don't think we should be in the business of restricting blobs (and we already have “[The blobs directory MAY be missing referenced blobs…][3]”).

I am fine with validation code *warning* users about either case (e.g. “this image-layout has no refs” or “refs a, b, and c require blobs which are not stored in this image-layout”), but I don't think the spec should block either of those.

Ping @q384566678 and @xiekeyang (for opencontainers/image-tools#83) and @runcom (for opencontainers/image-spec#287).

[1]: https://github.com/opencontainers/image-spec/issues/287
[2]: https://github.com/opencontainers/image-tools/pull/83#issuecomment-260841249
[3]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc2/image-layout.md#L52

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -9,15 +9,18 @@ Given an image layout and a ref, a tool can create an [OCI Runtime Specification
 * [Applying the filesystem layers](layer.md#applying) in the specified order
 * Converting the [image configuration](config.md) into an [OCI Runtime Specification `config.json`](https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md)
 
-The image layout has two top level directories:
+The image layout MUST contain two top level directories:
 
-- "blobs" contains content-addressable blobs. A blob has no schema and should be considered opaque.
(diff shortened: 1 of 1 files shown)


philips:
LGTM

stevvooe:
LGTM

PR #461, opened 2016-11-17
image-layout.md: clarify the layout file name MUST be `oci-layout`

coolljt0725:

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -18,9 +18,10 @@ The image layout MUST contain two top level directories:
 
 Both `blobs` and `refs` MAY be empty.
 
-The image layout MUST also contain a file that is used to identify the layout version:
+The image layout MUST also contain an `oci-layout` file that is used to identify the layout version:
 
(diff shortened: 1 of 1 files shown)


image-layout.md
```diff
@@ -18,7 +18,7 @@ The image layout MUST contain two top level directories:
 
 Both `blobs` and `refs` MAY be empty.
 
-The image layout MUST also contain a file that is used to identify the layout version:
+The image layout MUST also contain a file that is used to identify the layout version, the file name MUST be `oci-layout`:
 
 - `oci-layout` MUST contain a JSON object with a version field `{"imageLayoutVersion": "1.0.0"}` and MAY include additional fields.
```
wking: How about rephrasing both of these lines to something like:

> The image layout MUST also contain an `oci-layout` file that is used to identify the layout version.
> `oci-layout` MUST contain a JSON object with `imageLayoutVersion` set to `1.0.0`.
> `oci-layout` MAY include additional fields.

coolljt0725: I'm ok with this rephrasing.

coolljt0725:
updated per @wking 's suggestion

coolljt0725:
This is a dup of #434  closing

PR #462, opened 2016-11-18
config.md: remove "the ID of its parent image"

coolljt0725:
we doesn't define parent image in the configuration, we
should remove this sentence.


wking:
0fa275ca looks good to me.

vbatts:
LGTM

philips:
LGTM

PR #463, opened 2016-11-18
manifest.md: use `base layer` instead of `base image`

coolljt0725:
We doesn't define a `base image`, use `base layer` is
correct. This issue first being addressed in pr #312,
and then being addressed in pr #318, and then in pr #407,
but landing #407 has a long way to go. We could address
this first to avoid confusing.

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -42,7 +42,7 @@ Unlike the [Manifest List](manifest-list.md), which contains information about a
 - **`layers`** *array*
 
     Each item in the array MUST be a [descriptor](descriptor.md).
-    The array MUST have the base image at index 0.
+    The array MUST have the base layer at index 0.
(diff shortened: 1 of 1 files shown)


wking:
6bc08bc looks good to me.

wking:
And note that this is *exactly* the same patch as #312.  That PR was
closed in the naïve expectation that landing optional layers would be
fairly straightforward [1].  Since that has proven to not be the case,
I think it makes sense to come back and handle this smaller cleanup
independently (as this PR is doing).

[1]: https://github.com/opencontainers/image-spec/pull/312#issuecomment-247707098

jonboulle:
LGTM

vbatts:
LGTM

PR #466, opened 2016-11-18
limit min layer length in manifest to one

xiekeyang:
Spec declares `The array MUST have the base image at index 0. `
So that we should limit it in JSON shcema, and add test case.

### DIFF
--- a/schema/fs.go
+++ b/schema/fs.go
@@ -205,7 +205,7 @@ var _escData = map[string]*_escFile{
 	"/config-schema.json": {
 		local:   "config-schema.json",
 		size:    774,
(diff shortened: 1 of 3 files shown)


wking:
I hope we can lift that restriction in the future (#407), but 419b76f8
looks good to me as an encoding of the current spec.

xiekeyang:
@opencontainers/image-spec-maintainers I think this PR is ready to be merged, PTAL.
cc @stevvooe @vbatts

stevvooe:
LGTM

jonboulle:
lgtm

PR #468, opened 2016-11-19
fix bug: config test cases MUST include all required fields

xiekeyang:
Test cases should contain required fields, otherwise result will be always failure whether the validated field be legal or not. That is not our expection.
And add a new test case for only required fields, to check the `optional` and `required` fields in spec.

### DIFF
--- a/schema/config_test.go
+++ b/schema/config_test.go
@@ -31,7 +31,13 @@ func TestConfig(t *testing.T) {
 			config: `
 {
     "architecture": "amd64",
-    "os": 123
+    "os": 123,
+    "rootfs": {
+      "diff_ids": [
+        "sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef"
+      ],
+      "type": "layers"
+    }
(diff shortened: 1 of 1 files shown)


schema/config_test.go
```diff
@@ -31,7 +31,13 @@ func TestConfig(t *testing.T) {
 			config: `
 {
     "architecture": "amd64",
-    "os": 123
+    "os": 123,
+    "rootfs": {
```
xiekeyang: @wking 
> This one (and a few others) is still missing architecture and os.

Excuse me I might not understand your comment. `architecture` and `os` seems have been added to all cases.

wking: Ah, sorry, I'm just blind.  ceb8cf8e0 looks good to me.

xiekeyang:
@opencontainers/image-spec-maintainers I think this PR is ready to be merged, PTAL.
cc @stevvooe @vbatts

vbatts:
LGTM

stevvooe:
LGTM

PR #469, opened 2016-11-21
schema: add checkPlatform

q384566678:
Fixes https://github.com/opencontainers/image-tools/pull/39

### DIFF
--- a/schema/validator.go
+++ b/schema/validator.go
@@ -33,7 +33,9 @@ type Validator string
 type validateDescendantsFunc func(r io.Reader) error
 
 var mapValidateDescendants = map[Validator]validateDescendantsFunc{
-	MediaTypeManifest: validateManifestDescendants,
+	MediaTypeImageConfig:  validateConfigDescendants,
+	MediaTypeManifest:     validateManifestDescendants,
+	MediaTypeManifestList: validateManifestListDescendants,
 }
 
 // ValidationError contains all the errors that happened during validation.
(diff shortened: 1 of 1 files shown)


stevvooe:
LGTM

wking:
acf4adc37 looks good to me, although this is not really about
validating descendants (especially for the config platform check).  It
would be nice to have something like #403 land soonish so we had a
generic name for this sort of media-type-map-based verification.

q384566678:
@philips @vbatts @jonboulle @stevvooe  PTAL

jonboulle:
These are SHOULDs, not MUSTs. Why should we validate them?
https://github.com/opencontainers/image-spec/blame/master/config.md#L63

vbatts:
i'm okay with them being _checked_, but not to FAIL a validation.

wking:
On Wed, Nov 30, 2016 at 01:42:16PM -0800, Vincent Batts wrote:
> i'm okay with them being _checked_, but not to FAIL a validation.

+1.  We should probably table this until we settle on strict handling
(see the discussion in opencontainers/image-spec#452 and me trying to
outline the issue in opencontainers/image-tools#66).

q384566678:
> These are SHOULDs, not MUSTs. Why should we validate them?

Because these two are required attributes, and as the specification says: **The name of the operating system which the image is built to run on**.  If it is the verification of the system, only by JSON to restrict the type is not rigorous, it should be the value of its specific judgment.

q384566678:
Due to my mistakes, resulting in this PR can not be updated, I will close this, and then open a new PR.

PR #470, opened 2016-11-21
test.go:Modify the unsupported architecture

q384566678:
According to [runtime-spec's platform.arch](https://golang.org/doc/install/source#environment),when os is 'linux', does not support 's390x'


q384566678:
In the test, is it necessary to test all the corresponding Linux？

philips:
@q384566678 

From the docs you link to:

"s390x (IBM System z)
Supports Linux binaries. New in 1.7 and not as well exercised as other ports."

PR #474, opened 2016-11-29
platform object should not reject additional properties

xiekeyang:
According to [extensibility](https://github.com/opencontainers/image-spec/blob/master/manifest-list.md#extensibility), platform object in manifest list should accept the additional properties. So the limitation of `"additionalProperties": false` should be removed.
BTW, here was a syntax error, and this limitation didn't work. Even that, it still should be removed.

### DIFF
--- a/schema/defs-image.json
+++ b/schema/defs-image.json
@@ -74,8 +74,7 @@
               "type": "array",
               "items": {
                 "type": "string"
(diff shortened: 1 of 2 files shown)


philips:
LGTM

vbatts:
LGTM

PR #476, opened 2016-11-30
validator.go: add the layers of the judgment

q384566678:
Since the following use of ` range header.Layers`, so in the previous layer should increase the judge.


xiekeyang:
@q384566678 This seems to have been done in schema JSON.

q384566678:
@xiekeyang Yes, you are right

PR #483, opened 2016-12-07
layer: wordsmith non-distributable section

### LINKED ISSUE #475: Clarify when non-distributable layer policies should be enforced
This came up in this discussion: https://github.com/docker/docker/pull/28903#issuecomment-263696321.

For docker, the mediaType for a nondistributable layer is only enforced on push, leaving the mediaType ignored on pull. On pull, the `urls` field is used instead. While this specification doesn't define push and pull, we should clarify that it is important that these operations be only enforced on the correct data flow direction.

This is very subtle, but preserving this behavior ensures flexibility.

jonboulle:
Attempt to clarify when non-distributable policies should be enforced
(only on upload, never on download); as discussed in + fixes #475

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -314,7 +314,7 @@ Such "non-distributable" layers are typically downloaded directly from a distrib
 Non-distributable layers SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`.
 Implementations SHOULD NOT upload layers tagged with this media type; however, such a media type SHOULD NOT affect whether an implementation downloads the layer.
 
-[Descriptors](descriptor.md) referencing non-distributable layers MAY include `urls` for downloading these layers directly.
+[Descriptors](descriptor.md) referencing non-distributable layers MAY include `urls` for downloading these layers directly; however, the presence of the `urls` field SHOULD NOT be used to determine whether or not a layer is non-distributable.
 
 [libarchive-tar]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#POSIX_ustar_Archives
 [gnu-tar-standard]: http://www.gnu.org/software/tar/manual/html_node/Standard.html
--- a/layer.md
+++ b/layer.md
@@ -308,12 +308,13 @@ Any given image is likely to be composed of several of these Image Filesystem Ch
 
 # Non-Distributable Layers
 
-Certain layers, due to legal requirements, may not be regularly distributable.
-Typically, such layers are downloaded directly from a distributor but are never uploaded.
+Due to legal requirements, certain layers may not be regularly distributable.
+Such "non-distributable" layers are typically downloaded directly from a distributor but never uploaded.
 
-Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`.
-[Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
-It is implementation-defined whether or not implementations upload layers tagged with this media type.
+Non-distributable layers SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`.
+Implementations SHOULD NOT upload layers tagged with this media type; however, such a media type SHOULD NOT affect whether an implementation downloads the layer.
+
+[Descriptors](descriptor.md) referencing non-distributable layers MAY include `urls` for downloading these layers directly.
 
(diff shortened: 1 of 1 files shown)


wking:
I like (and don't see anything controversial in) everything about
a2dda98 except for the new SHOULD NOT's reliance on fuzzy “upload” and
“download” wording [1].  Is there any chance of splitting that off
into it's own PR?  Then we can land the cleaner wording quickly while
we hash out whether we need to define upload/download.  Unless I'm the
only person who sees uncertainty around what upload/download mean in
the cache/proxy context ;).

[1]: https://github.com/opencontainers/image-spec/pull/483/files#r91417509

jonboulle:
I'm not sure what you're suggesting - just strike the line here for now?

wking:
Yeah, drop your new "Implementations SHOULD NOT upload..." line and keep the old "It is implementation-defined..." line.  Then we land this one quick and easy, and you open a new PR with just that "implementation-defined" -> "SHOULD NOT" line swap.

layer.md
```diff
@@ -308,12 +308,13 @@ Any given image is likely to be composed of several of these Image Filesystem Ch
...
+Due to legal requirements, certain layers may not be regularly distributable.
+Such "non-distributable" layers are typically downloaded directly from a distributor but never uploaded.
 
-Layers that have these restrictions SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`.
-[Descriptors](descriptor.md) referencing these layers MAY include `urls` for downloading these layers.
-It is implementation-defined whether or not implementations upload layers tagged with this media type.
+Non-distributable layers SHOULD be tagged with an alternative mediatype of `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`.
+Implementations SHOULD NOT upload layers tagged with this media type; however, such a media type SHOULD NOT affect whether an implementation downloads the layer.
+
+[Descriptors](descriptor.md) referencing non-distributable layers MAY include `urls` for downloading these layers directly.
```
stevvooe: > If the `urls` field is present, they SHOULD be favored for retrieval over resolving through the `digest` identifiers.
> The presence of the `urls` field MUST NOT be used to determine whether or not a layer is non-distributable.

stevvooe:
Added a few lines, otherwise, LGTM.

You must have a fine forge to be smithing at such a level.

layer.md
stevvooe: > I don't think we want this, since it sounds like “don't cache these locally”.

Only if you read into it too much and skip words like "retrieval" and "resolving" when trying to understand the meaning. It is also a SHOULD.

wking: Does “resolving through `digest` identifiers” somehow not include fetching from a local cache?  If so, I think it's worth spelling that out explicitly (but defining “a local cache” might be tricky).

And I don't think the SHOULD-ness means we can be sloppy about defining the relevant terms.  SHOULD just means compliant implementations can skip the requirement; it shouldn't mean that the requirement itself is ambiguous.

stevvooe: @wking More "precision" here is going to infer a hierarchy, which actually sacrifices accuracy. The term "retrieval" implies relative displaced locality. In this context, since we are talking about `urls`, the term is both accurate and precise, since the relative locality is `urls`, provided in the first clause, which implies network access. Local cache clearly does not fall into the locality level of `urls`.

Adding more here will just obscure the intent of the statement. It is neither sloppy nor imprecise and ultimately accurate in context.

But let's see what the contributor has to say, as he might have an opinion.

stevvooe:
@jonboulle Ping.

layer.md
jonboulle: I don't understand why we need to call out `urls` as being preferential - what's wrong with the digest resolving?

jonboulle: > The presence of the urls field MUST NOT be used to determine whether or not a layer is non-distributable.
 
Mixing SHOULDs and MUSTs here is awkward. With the current wording I really don't think there's any reason to think a reader or implementer would infer the opposite of this; in the descriptor definition, `urls` is mentioned as a common property, so clearly its presence doesn't imply anything about distributability.

stevvooe: @jonboulle How so? There has already been confusion that `urls` == nondistributable, on both docker PRs and OCI PRs. Making this distinction clear will ensure that `urls` will be useful without being tied to `nondistributable`.

jonboulle: I still don't think it's necessary but in the interests of moving forward I added that line, ptal

stevvooe:
LGTM

stevvooe:
@jonboulle Looks like the DCO check failed.

jonboulle:
le sigh. now with fixed DCO

stevvooe:
LGTM

philips:
LGTM

PR #484, opened 2016-12-08
Fix "manifest list" typos for manifest annotations

q384566678:

### DIFF
--- a/specs-go/v1/manifest.go
+++ b/specs-go/v1/manifest.go
@@ -27,6 +27,6 @@ type Manifest struct {
 	// Layers is an indexed list of layers referenced by the manifest.
 	Layers []Descriptor `json:"layers"`
 
-	// Annotations contains arbitrary metadata for the manifest.
+	// Annotations contains arbitrary metadata for the image manifest.
 	Annotations map[string]string `json:"annotations,omitempty"`
 }


q384566678:
@stevvooe @vbatts @jonboulle PTAL

stevvooe:
@q384566678 I have no clue what this PR is doing. Please expand on "Modify the error message". What is this fixing?

q384566678:
@stevvooe Should be a writing error, manifest the annotations field id should be `manifest/annotations ` instead of `manifestlist-list/annotations`.

stevvooe:
@q384566678 What should be writing an error? Do you have an example of what happens with and without this patch?

q384566678:
@stevvooe Uh, if not revised, then there should not be any mistakes. I think here mainly belongs to writing errors.

q384566678:
@stevvooe  @vbatts @philips I have updated this PR, I think these two should be modified. It can be seen from the specification that [manifest list](https://github.com/opencontainers/image-spec/blame/master/manifest-list.md#L74) and [image manifest](https://github.com/opencontainers/image-spec/blame/master/manifest.md#L68) are different in `Annotations`.PTAL

stevvooe:
@q384566678 Are you saying that this is missing? Could you provide an example of the output before and after this change?

q384566678:
@stevvooe Uh......I mean that in my two changes, the manifest and manifest list confused, if not changed, then it will not be wrong, because it is only the role of the description, not the specific data call.

wking:
65c70950e looks good to me.  image-manifest-schema.json should not
have manifest-list IDs and manifest.go should not mention manifest
lists in its comments.  I expect both were copy/paste errors from when
annotations were added to manifests, but I haven't gone back into the
Git history to confirm that.

wking:
On Tue, Feb 14, 2017 at 11:56:44AM -0800, W. Trevor King wrote:
> 65c70950e looks good to me…

The PR subject and commit message summary are pretty bad though ;).
Maybe “Fix "manifest list" typos for manifest annotations”?

q384566678:
> The PR subject and commit message summary are pretty bad though ;).

My title does have a problem.updated.thanks.

wking:
On Tue, Feb 14, 2017 at 05:37:36PM -0800, Zhou Hao wrote:
> My title does have a problem.updated.thanks.

Want to update the commit message as well?

vbatts:
LGTM

jonboulle:
lgtm

Issue #482, opened 2016-12-07
Clarify role of ChainID


stevvooe:
The OCI Image config document covers the calculation of the `ChainID` but it doesn't go into why this is useful or how to best leverage.

The best way to view it is a hash of ordering of applied layers.

Let's say we have layers A, B, C, ordered from bottom to top, where A is the base and C is the top. Defining `|` as a binary application operator, the root filesystem may be `A|B|C`. While it is implied that `C` is only useful when applied to `A|B`, the identifier `C` is insufficient to identify this result, as we'd have the equality `C = A|B|C`, which isn't true.

The main issue is when we have two definitions of `C`, `C = C` and `C = A|B|C`. If this is true (with some handwaving), `C = x|C` where `x = any application` must be true. This means that if an attacker can define `x`, relying on `C` provides no guarantee that the layers were applied in any order.

The `ChainID` addresses this problem by being defined as a compound hash. __We differentiate the changeset `C`, from the order dependent application `A|B|C` by saying that the resulting rootfs is identified by ChainID(A|B|C), which can be calculated by `ImageConfig.rootfs`.__

The definition from the spec is something like this (also, see the [base implementation](https://github.com/docker/docker/blob/85bc735b4a56223c84971839d819ff8dc494c181/layer/layer.go#L254)):

```
ChainID(layer[N]) = SHA256hex(ChainID(layer[N-1]) + " " + DiffID(layer[N])).
```

(Note that this definition is slightly insufficient, because it implies that layer[N] is `layer[0]|...|layer[N-1]|layer[N]`, which we indicate doesn't quite add up above)

With our expanded example, the we can have a symbolic definition of `ChainID(C)`, which is a variation on some function `Hchain(A|B|C)`, with some notation hand-waving. 

```
ChainID(A) = DiffID(A)
ChainID(A|B) = SHA256(ChainID(A) + " " + DiffID(B))
ChainID(A|B|C) = SHA256(ChainID(A|B) + " " + DiffID(C))
```

(Note that we may be missing the base case, `ChainID(A) = DiffID(A)`, as well)

Let's expand this, for fun:

```
ChainID(A|B|C) = SHA256(SHA256(DiffID(A) + " " + DiffID(B)) + " " + DiffID(C))
```

Hopefully, the above is illustrative of the _actual_ contents of the `ChainID`.

[...]

PR #148, opened 2016-06-16
Vendor dependencies, add hacking guide

s-urbaniak:
This PR vendors all OCI image spec Go dependencies using glide and glide-vc.

Additionally it adds a hacking guide since we accumulated quite some tools/prerequisites for various make targets. The hacking guide documents them.

Fixes #76, #162

PR #486, opened 2016-12-08
identity: add implementation of ChainID

stevvooe:
The specification defines an algorithm to calculate a `ChainID`, which
can be used to identify the result of subsequent applications of layers.
Because this algorithm is subtle and only needs to implemented in a
single place, we provide a reference implementation.

For convenience, we provide functions that calculate all the chain ids
and just the top-level one. It is is integrated with the
distribution/digest type for safety and convenience.

As part of this, the `identity` package has been introduced. For
consuming code, a few helpers have been provide to ease transition as
the name of the upstream package has not yet been finalized. Users of
this package should employ `FromBytes`, `FromString` and `FromReader`
where appropriate, which should ease the transition if these packages
change.

Tests are formulated based on pre-calculation of chain identifiers to
ensure correctness.

Relates to #482.
Requires #487.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,10 +1,11 @@
 package: github.com/opencontainers/image-spec
 import:
 - package: github.com/pkg/errors
-  version: ">=0.7.0"
+  version: '>=0.7.0'
 - package: github.com/xeipuuv/gojsonschema
   version: d5336c75940ef31c9ceeb0ae64cf92944bccb4ee
 - package: github.com/russross/blackfriday
   version: ~v1.4
 - package: github.com/shurcooL/sanitized_anchor_name
   version: 10ef21a441db47d8b13ebcc5fd2310f636973c77
+- package: github.com/opencontainers/go-digest
--- a/identity/helpers.go
+++ b/identity/helpers.go
@@ -1,0 +1,40 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import (
+	_ "crypto/sha256" // side-effect to install impls, sha256
+	_ "crypto/sha512" // side-effect to install impls, sha384/sh512
+
+	"io"
+
+	digest "github.com/opencontainers/go-digest"
+)
+
+// FromReader consumes the content of rd until io.EOF, returning canonical
+// digest.
+func FromReader(rd io.Reader) (digest.Digest, error) {
+	return digest.Canonical.FromReader(rd)
+}
+
+// FromBytes digests the input and returns a Digest.
+func FromBytes(p []byte) digest.Digest {
+	return digest.Canonical.FromBytes(p)
+}
+
+// FromString digests the input and returns a Digest.
+func FromString(s string) digest.Digest {
+	return digest.Canonical.FromString(s)
+}
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,10 @@
-hash: 61c12670baff1b8a58d601c5afeae07ac831ef85cb67de3d4823ccdbb0085f56
-updated: 2016-09-21T13:59:32.039338003-07:00
+hash: 731b4c55f527573960d8a2e09881f77a431ecf2cb434192768a9e5be84ff1800
+updated: 2017-01-11T12:19:22.946780728-08:00
 imports:
+- name: github.com/opencontainers/go-digest
+  version: 21dfd564fd89c944783d00d069f33e3e7123c448
 - name: github.com/pkg/errors
-  version: 17b591df37844cde689f4d5813e5cea0927d8dd2
+  version: 645ef00459ed84a119197bfb8d8205042c6df63d
 - name: github.com/russross/blackfriday
   version: 0b647d0506a698cca42caca173e55559b12a69f2
 - name: github.com/shurcooL/sanitized_anchor_name
@@ -14,7 +16,7 @@ imports:
 - name: github.com/xeipuuv/gojsonschema
   version: d5336c75940ef31c9ceeb0ae64cf92944bccb4ee
 - name: go4.org
-  version: 27882e489dcd7ade28c033b82c6d498661c4dc42
+  version: 3b6fde061c9c94928618caa51aff24646db549df
   subpackages:
   - errorutil
 testImports: []
--- a/identity/chainid.go
+++ b/identity/chainid.go
@@ -1,0 +1,67 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package identity provides implementations of subtle calculations pertaining
+// to image and layer identity.  The primary item present here is the ChainID
+// calculation used in identifying the result of subsequent layer applications.
+//
+// Helpers are also provided here to ease transition to the
+// github.com/opencontainers/go-digest package, but that package may be used
+// directly.
+package identity
+
+import "github.com/opencontainers/go-digest"
+
+// ChainID takes a slice of digests and returns the ChainID corresponding to
+// the last entry. Typically, these are a list of layer DiffIDs, with the
+// result providing the ChainID identifying the result of sequential
+// application of the preceding layers.
+func ChainID(dgsts []digest.Digest) digest.Digest {
+	chainIDs := make([]digest.Digest, len(dgsts))
+	copy(chainIDs, dgsts)
+	ChainIDs(chainIDs)
+
+	if len(chainIDs) == 0 {
+		return ""
+	}
+	return chainIDs[len(chainIDs)-1]
+}
+
+// ChainIDs calculates the recursively applied chain id for each identifier in
+// the slice. The result is written direcly back into the slice such that the
+// ChainID for each item will be in the respective position.
+//
+// By definition of ChainID, the zeroth element will always be the same before
+// and after the call.
+//
+// As an example, given the chain of ids `[A, B, C]`, the result `[A,
+// ChainID(A|B), ChainID(A|B|C)]` will be written back to the slice.
+//
+// The input is provided as a return value for convenience.
+//
+// Typically, these are a list of layer DiffIDs, with the
+// result providing the ChainID for each the result of each layer application
+// sequentially.
+func ChainIDs(dgsts []digest.Digest) []digest.Digest {
+	if len(dgsts) < 2 {
+		return dgsts
+	}
+
+	parent := digest.FromBytes([]byte(dgsts[0] + " " + dgsts[1]))
+	next := dgsts[1:]
+	next[0] = parent
+	ChainIDs(next)
+
+	return dgsts
+}
--- a/identity/chainid_test.go
+++ b/identity/chainid_test.go
@@ -1,0 +1,95 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package identity
+
+import (
+	_ "crypto/sha256" // required to install sha256 digest support
+	"reflect"
+	"testing"
+
+	"github.com/opencontainers/go-digest"
+)
+
+func TestChainID(t *testing.T) {
+	// To provide a good testing base, we define the individual links in a
+	// chain recursively, illustrating the calculations for each chain.
+	//
+	// Note that we use invalid digests for the unmodified identifiers here to
+	// make the computation more readable.
+	chainDigestAB := digest.FromString("sha256:a" + " " + "sha256:b")              // chain for A|B
+	chainDigestABC := digest.FromString(chainDigestAB.String() + " " + "sha256:c") // chain for A|B|C
+
+	for _, testcase := range []struct {
+		Name     string
+		Digests  []digest.Digest
+		Expected []digest.Digest
+	}{
+		{
+			Name: "nil",
+		},
+		{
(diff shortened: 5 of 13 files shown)


stevvooe:
PTAL @opencontainers/image-spec-maintainers

wking:
Travis points out some missing license headers [1].

[1]: https://travis-ci.org/opencontainers/image-spec/builds/182417754#L327

wking:
And it has some problems with *testing.T.Run [1].

[1]: https://travis-ci.org/opencontainers/image-spec/builds/182417754#L336

specs-go/chain/chainid.go
```diff
@@ -0,0 +1,46 @@
+package chain
+
+import "github.com/docker/distribution/digest"
+
+// ChainID takes a slice of digests and returns the ChainID corresponding to
+// the last entry. Typically, these are a list of layer DiffIDs, with the
```
wking: I'd rather drop “to the last entry”.  The ChainID is a property of [the slice as a whole][1], not a property of one entry in particular.

[1]: https://github.com/opencontainers/image-spec/issues/482#issuecomment-265565694

stevvooe:
See #487.

Also, note that the dependency hasn't been added. Considering adding digest package here, as well.

specs-go/chain/chainid.go
```diff
@@ -0,0 +1,46 @@
...
+// result providing the ChainID identifying the result of sequential
+// application of the preceeding layers.
```
wking: “preceeding” → “preceding”

specs-go/chain/chainid.go
```diff
@@ -0,0 +1,46 @@
...
+func ChainID(dgsts []digest.Digest) digest.Digest {
+	chainIDs := make([]digest.Digest, len(dgsts))
+	copy(chainIDs, dgsts)
+	ChainIDs(chainIDs)
+
+	if len(chainIDs) == 0 {
+		return ""
```
wking: Neither the [current definition][1] nor any of the alternative phrasings proposed in #482 define chainIDs for empty arrays.  I think we want:

    func ChainID(digests []digest.Digest) (digest.Digest, error)

with an error being returned when `len(digests) == 0`.

[1]: https://github.com/opencontainers/image-spec/blob/656fb2f82121e8bd611b5aae4a6a3471e1bfe078/config.md#layer-chainid

specs-go/chain/chainid.go
```diff
@@ -0,0 +1,46 @@
...
+
+// ChainIDs calculates the recursively applied chain id for each identifier in
+// the slice. The result is written direcly back into the slice such that the
+// ChainID for each item will be in the respective position.
+//
+// By definition of ChainID, the zeroth element will always be the same before
+// and after the call.
+//
+// As an exmaple, given the chain of ids `[A, B, C]`, the result `[A,
+// ChainID(A|B), ChainID(A|B|C)]` will be written back to the slice.
```
wking: “exmaple” → “example”

And the pipe syntax is not particularly intuitive for me.  Can we use:

> As an example, given the digests `[A, B, C]`, the result `[A, ChainID([A, B]), ChainID([A, B, C])]` will be written back to the slice.

specs-go/chain/chainid.go
```diff
@@ -0,0 +1,46 @@
...
+//
+// The input is provided as a return value for convenience.
```
wking: I'd rather drop this, since it might mask the mutation you're doing.

specs-go/chain/chainid.go
stevvooe: There is no need to create an error for this condition.

specs-go/chain/chainid.go
stevvooe: My did you quote yourself as the canonical definition of ChainID?

Actually, a list of digests has a ChainID for each entry in the array. This function returns the last entry.

wking:
On Thu, Dec 08, 2016 at 02:03:15PM -0800, Stephen Day wrote:
> Also, note that the dependency hasn't been added. Considering adding
> digest package here, as well.

If we are including a reference ChainID implementation in this spec, I
think we *have* to include the digest package here too (or enough of
it to get ChainIDs working, which is not a big piece).  I don't like
the spec consumes Docker, which consumes the spec, etc. reference
cycle.

specs-go/chain/chainid.go
wking: This is a reference implementation for the [definiton][1], so I think we do need an “undefined behavior” error for the undefined case of `ChainID([])`.  Alternatively, we could define the ChainID for that case to be the empty string, but I don't see the point since we [always require at least one layer][2] (with #407 still in flight).

It would have been nice if the chainID algorithm was:

    ChainID([L₀, …, Lₙ₋₁]) = { SHA256hex(''), when n == 0
                             { SHA256hex(ChainID([L₀, …, Lₙ₋₂]) + " " + DiffID(Lₙ₋₁)), when n > 0

to cleanly cover the `n == 0` case, but I think changing to that now would create more confusion than the increased cleanliness is worth.

[1]: https://github.com/opencontainers/image-spec/blob/656fb2f82121e8bd611b5aae4a6a3471e1bfe078/config.md#layer-chainid
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/manifest.md#L45

stevvooe:
>  I don't like
the spec consumes Docker, which consumes the spec, etc. reference
cycle.

The spec does not "consume" docker. The digest package is just a Go package from docker. Is the problem that the dependency has the name "docker" in it, like it is some big-bad word?

specs-go/chain/chainid.go
stevvooe: @wking And, as I said in #482, that definition needs to be adjusted. As the specification is written currently, it is undefined.

specs-go/chain/chainid.go
wking: > My did you quote yourself as the canonical definition of ChainID?

I quoted myself as the clearest (to me ;) description of ChainID being a property of an array (which you point out is not blindingly obvious in the current spec wording, #482.

> Actually, a list of digests has a ChainID for each entry in the array. This function returns the last entry.

The [spec definition][1] is for ChainID, singular, not a ChainIDs array.  And the ChainID being returned here corresponds to the whole slice, not just the last entry.  Yes, the returned ChainID is the same digest as the last entry in the mutated slice from your `ChainIDs` function, but I think that's an implementation detail that's too peripheral for the opening sentence of this comment.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc3/config.md#layer-chainid

specs-go/chain/chainid.go
wking: And with all of the proposals in #482, the chainID for an empty array is *still* undefined.  In the absence of a spec definition (which is where things stand with PR), I think we should be raising an “undefined” error.

specs-go/chain/chainid.go
stevvooe: @wking Seriously, if you continue to split hairs on topics you have literally _no clue about_, I am going to block you. In fact, I remember explaining chain ids to you _months_ ago and having you walk away with no understanding in their use or calculation.

Now, you show up, quoting yourself, rather than the proposed explanation, as canonical?

> And the ChainID being returned here corresponds to the whole slice, not just the last entry.

These are the same thing.

wking:
On Thu, Dec 08, 2016 at 02:22:48PM -0800, Stephen Day wrote:
> The spec does not "consume" docker. The digest package is just a Go
> package from docker. Is the problem that the dependency has the name
> "docker" in it, like it is some big-bad word?

Yes ;).  To a spec reader, it is unclear how tightly
github.com/docker/distribution/digest is tied into the rest of
github.com/docker/distribution or what the process is for managing
github.com/docker/distribution/digest releases (are they independent
of github.com/docker/distribution as a whole?).

I'd be completely fine with a github.com/docker/digest that had no
non-Go-stdlib deps.

specs-go/chain/chainid.go
wking: > > And the ChainID being returned here corresponds to the whole slice, not just the last entry.
>
> These are the same thing.

No they aren't, and you pointed this out in #482.  If the chain ID was a property of an *entry*, the function signature here would be:

    func ChainID(digest digest.Digest) digest.Digest

and you'd call it with `ChainID(config.RootFS.DiffIDs[3])`, etc.  But the chain ID is really a property of the *slice* (defined for slices with length > 0), so the signature is:

    func ChainID(digests []digest.Digest) (digest.Digest, error)

And you call it with `ChainID(config.RootFS.DiffIDs[:4])`, etc.

specs-go/chain/chainid.go
wking: I agree that reading the docs is an important step for the dev who first uses the function.  Maybe they know (for whatever reason) that the current code base no longer needs the original diffIDs, and they save a copy by using:

    middleChainIDs := chain.ChainIDs(config.RootFS.DiffIDs)[2:4]

All very nice.

Some time later, another dev is going to come along and read the consuming code, not realize that `config.RootFS.DiffIDs` has been mutated, and use the mutated values.  Are they going to go back and read the docs (“I wonder if `ChainIDs` is mutating its argument…”)?  Probably not.

You *could* guard against that with the initial dev using:

    // warning: mutates config.RootFS.DiffIDs
    middleChainIDs := chain.ChainIDs(config.RootFS.DiffIDs)[2:4]  

On the other hand, *not* returning the mutated slice makes it calls that look more suspicious for mutation:

    chain.ChainIDs(config.RootFS.DiffIDs)

^ that does not sound like a logging function, so it's probably mutating the argument.

And I doubt the diffID slice will ever be so long that allocating a new slice for the chainIDs is a big performance drag.  I'd rather allocate and return a new slice.  But there is *some* performance impact to that, so if you have to keep the mutation, I'm ok with that, but I'd rather make the mutation as obvious as possible in the function signature.

stevvooe:
@vbatts @philips Could I get some feedback on this before I take another pass?

specs-go/chain/chainid.go
```diff
@@ -0,0 +1,62 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package chain provides implementations of the ChainID calculation used in
+// identifying the result of subsequent layer applications.
+package chain
+
+import "github.com/docker/distribution/digest"
```
philips: We need to vendor this, right?

Or, we move that package into this repo. Given the strong dependence on that library and that these string digests are [part of the spec](https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests-and-verification) seems we should add that package here as well. We already have one open-coded impl here: https://github.com/opencontainers/image-spec/blob/ebee8d2c7a1f3ea37a4ca0fed459db487456a16b/schema/manifest_backwards_compatibility_test.go#L113

stevvooe: Given we have a solid package and mature implementation, I'm thinking that we can bring the `distribution/digest` into OCI. I'm seeing some disconcerting patterns being propagated that this package solves.

jonboulle: Yes, I'd be strongly in favour of moving it in here so that any spec implementers can/will easily leverage it.

stevvooe: @jonboulle @philips After some consideration, I've broken this out into a separate package, https://github.com/docker/go-digest. The reason being that this has wider use that just OCI images.

If possible, it might be good to move this over to `opencontainers/go-digest` but I think this meets the current concern.

jonboulle: to be clear my concerns were around importing an unversioned package from a large out-of-org repository (nothing to with the particular provenance or import path of said package). Moving to importing an unversioned package from a small out-of-org repository is a minor improvement but ideally I'd rather see this in-tree, with a secondary preference for another opencontainers/* repo and vendored from there.

stevvooe: @jonboulle @philips I'm not sure if your implications apply here, seeing the history of the digest repository. In fact, moving it out will make it _less_ stable, but I want to make everyone happy.

If you can get me an `opencontainers/go-digest` repository by the 1pm PST, I'll be more than happy to move it over and support the transition in Docker itself.

stevvooe:
@philips Updated to vendor `docker/go-digest`. Moved to package `identity` and added some helpers to ease transition if package naming is changes.

identity/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	_ "crypto/sha256" // side-effect to install impls, sha256
+	_ "crypto/sha512" // side-effect to install impls, sha384/sh512
+
+	"io"
+
+	digest "github.com/docker/go-digest"
+)
+
+// FromReader returns the most valid digest for the underlying content using
```
jonboulle: What does "most valid digest" mean?

jonboulle:
Thanks, generally this looks good - unfortunately it looks like the git validation checker is choking on the vendor commits again :-( - @vbatts what did we end up doing for that?

stevvooe:
@jonboulle I looked for an easy patch to `git-validation` but it doesn't look straightforward.

jonboulle:
@stevvooe I forgot about https://github.com/opencontainers/image-spec/pull/148#issuecomment-227968453 - did you use `make update-deps`?

stevvooe:
@philips @jonboulle Updated to use `go-digest` package.

There may be some follow up after we work out some administrative details on go-digest (such as a release), but let's get this in.

vendor/github.com/russross/blackfriday/LICENSE.txt
```diff
@@ -1,29 +0,0 @@
-Blackfriday is distributed under the Simplified BSD License:
```
wking: Accidental removal?

vendor/github.com/shurcooL/sanitized_anchor_name/LICENSE
```diff
@@ -1,19 +0,0 @@
-Copyright (c) 2015 Dmitri Shuralyov
```
wking: Another accidental removal?

vendor/github.com/russross/blackfriday/LICENSE.txt
stevvooe: @wking I ran the `make update-deps`.

stevvooe: The problem is that `glide-vc` is not taking test dependencies into account.

We also need to stop committing broken garbage into the Makefile. None of this is making this easier.

stevvooe: Ok, this PR should be correct now.

I'm sending a separate PR with the fixes to the `Makefile`.

stevvooe:

glide.yaml
```diff
@@ -1,10 +1,11 @@
...
 - package: github.com/pkg/errors
-  version: ">=0.7.0"
+  version: '>=0.7.0'
 - package: github.com/xeipuuv/gojsonschema
   version: d5336c75940ef31c9ceeb0ae64cf92944bccb4ee
 - package: github.com/russross/blackfriday
   version: ~v1.4
 - package: github.com/shurcooL/sanitized_anchor_name
   version: 10ef21a441db47d8b13ebcc5fd2310f636973c77
+- package: github.com/opencontainers/go-digest
```
jonboulle: Can you add a version, please?

wking: > Can you add a version, please?

We still need to cut one (opencontainers/go-digest#23), as @stevvooe mentions at [in his earlier comment][1].

[1]: https://github.com/opencontainers/image-spec/pull/486#issuecomment-271982906

jonboulle: Look at the immediately prior dependency. Version can be a hash.

stevvooe: That is in the glide.lock file. Only specify a version here if it matters when updating.

jonboulle: No, it's here in glide.yaml - every other package has an associated version. Can we just keep it consistent for now? IMO glide-update should ideally be idempotent unless someone intentionally requests otherwise.

jonboulle:
LGTM
I still prefer https://github.com/opencontainers/image-spec/pull/486/files#r96237294 but am not going to block right now.

philips:
LGTM

PR #487, opened 2016-12-08
build: upgrade to Go 1.7

stevvooe:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - 1.6
+  - 1.7
 
 sudo: required
 


stevvooe:
PTAL @opencontainers/image-spec-maintainers

wking:
This is presumably to address an expansion of the test API [1].  But
if we are attempting to provide a ChainID package for external
consumption, I think it might make more sense to have an
implementation/test-suite which worked on a range of recent Go
versions.

[1]: https://github.com/opencontainers/image-spec/pull/486#issuecomment-265868884

stevvooe:
@wking Go 1.7 has been out since August. By the time the specification is released, Go 1.8 will be out. Since these are in-project tests and we don't provide any backwards compatibility for Go versions, we should target the latest possible. Note that these packages will be consumable by pre-Go 1.7, anyways, since they don't consume tests.

Anyways, I would kindly ask that you stop wasting my time.

wking:
On Thu, Dec 08, 2016 at 02:34:59PM -0800, Stephen Day wrote:
> Since these are in-project tests and we don't provide any backwards
> compatibility for Go versions…

If no Go backwards compatibility is the project policy, then that's
fine.  Do we document that anywhere?

> Note that these packages will be consumable by pre-Go 1.7, anyways,
> since they don't consume tests.

I'm hesistant about “don't worry, this works on Go <1.7, we just don't
test it there”.  Travis makes a build matrix pretty easy, so we can
test this in as many flavors of Go as the project decides to support.

stevvooe:
@vbatts @philips Could we please get this merged?

vbatts:
LGTM

philips:
LGTM

PR #488, opened 2016-12-10
README.md: move a line to spec.md

coolljt0725:

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,6 @@ Additional documentation about how this group operates:
 - [Releases](RELEASES.md)
 - [Project Documentation](project.md)
 
-The _optional_ and _base_ layers of all OCI projects are tracked in the [OCI Scope Table](https://www.opencontainers.org/governance/oci-scope-table).
-
 ## Running an OCI Image
 
 The OCI Image Format partner project is the [OCI Runtime Spec project](https://github.com/opencontainers/runtime-spec).
--- a/spec.md
+++ b/spec.md
@@ -48,4 +48,6 @@ The high-level components of the spec include:
 * Signatures that are based on signing image content address (optional OCI layer)
 * Naming that is federated based on DNS and can be delegated (optional OCI layer)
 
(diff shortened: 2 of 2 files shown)


wking:
I think that line should stay in the README.  The scope table is really guidance from the TOB to project maintainers about what should go in the spec, it's not part of the spec itself.

coolljt0725:
I think the line 
`The _optional_ and _base_ layers of all OCI projects are tracked in the [OCI Scope Table](https://www.opencontainers.org/governance/oci-scope-table).`
should next to https://github.com/opencontainers/image-spec/pull/488/files#diff-958e7270f96f5407d7d980f500805b1bL49

wking:
On Fri, Dec 09, 2016 at 09:20:27PM -0800, Lei Jitang wrote:
> I think the line `The _optional_ and _base_ layers of all OCI
> projects are tracked in the [OCI Scope
> Table](https://www.opencontainers.org/governance/oci-scope-table).`
> should next to
> https://github.com/opencontainers/image-spec/pull/488/files#diff-958e7270f96f5407d7d980f500805b1bL49

Moving the Scope Table link next to the spec's mention of
optional/base is reasonable, but I'd still rather keep this guiding
doc in the non-spec, project-structure section of the repo.  If the
consensus is to move it into the spec, we'll want to be very clear
that the Scope Table is not normative (because it's not maintained by
the image-spec maintainers, it's not versioned, and we can't link to a
pinned version of it).

And I doubt the Scope Table will be particularly interesting once the
spec's compliance language grows the “Protocols defined by this
specification are” section that we currently have in runtime-spec
[1,2].  The scope table will just be “if you want to read these
protocols again in language written by people less closely tied to the
spec…” ;).

[1]: https://github.com/opencontainers/image-spec/pull/381#discussion_r85947950
[2]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc2/README.md#L30

stevvooe:
This is meant for the readme, not the spec.

PR #458, opened 2016-11-16
Clean-up spec to remove non-spec-ish items

duglin:
Per the oci f2f at kubecon.  Similar changes what is in: https://github.com/opencontainers/runtime-spec/pull/626

Signed-off-by: Doug Davis <dug@us.ibm.com>

PR #490, opened 2016-12-10
spec: Restore compliance language

wking:
This landed in #432 but was removed in #458, likely accidentally during a rebase.


wking:
And I [still][1] think we have more work to do on this section, but we can work that out in later PRs.  This PR just restores the language we'd agreed on before and then accidentally removed.

[1]: https://github.com/opencontainers/image-spec/pull/432#discussion_r85947950

vbatts:
LGTM

stevvooe:
LGTM

Issue #479, opened 2016-11-30
config: document how to translate image config into runtime config


philips:
There are several fields in the OCI Image Config that directly translate to a fields on the Runtime Config. We should document how those fields translate.

The obvious ones:

- User
- Env
- Entrypoint
- Cmd
- WorkingDir

UX ones (used to help users but not automatically mapped):

- ExposedPorts
- Volumes

Optional as they depend on the user and largely shouldn't be dictated by image author:

- Memory
- MemorySwap
- CPUShares

philips:
As discussed @wking @stevvooe

philips:
@cyphar could you take a swing at adding this language?

wking:
See also #454, which would set the required runtime-spec config version.

cyphar:
@philips Yeah, I'll give it a shot.

cyphar:
I've done this in #492.

RobDolinMS:
@philips Can this Issue be closed with active discussion now happening in PR #492 ?

stevvooe:
@RobDolinMS We should really wait for the issue to get resolved before closing it.

PR #492, opened 2016-12-12
conversion: add document about image -> runtime configuration

cyphar:
Add documentation about how to convert from an OCI image configuration
to an OCI runtime configuration. In particular, describe precisely what
fields need to be filled given a particular configuration.

The fields have been grouped into several categories, because some of
the image fields are not as well-thought-out as others (such as the
resource limitation fields, which should really be decided by the user
not by the image creator). In addition, some fields (such as Volumes)
cannot be understood by a generic configuration converter.

Fixes: #479

### DIFF
--- a/conversion.md
+++ b/conversion.md
@@ -1,0 +1,119 @@
+# Conversion to OCI Runtime Configuration
+
+When extracting an OCI Image into an [OCI Runtime bundle][oci-runtime-bundle], two orthogonal components of the extraction are relevant:
+
+1. Extraction of the root filesystem from the set of [filesystem layers](layers.md).
+2. Conversion of the [image configuration blob](config.md) to an [OCI Runtime configuration blob][oci-runtime-config].
+
+This section defines how to convert an `application/vnd.oci.image.config.v1+json` blob to an [OCI runtime configuration blob][oci-runtime-config] (the latter component of extraction).
+The former component of extraction is defined [elsewhere](layers.md) and is orthogonal to configuration of a runtime bundle.
+The values of runtime configuration properties not specified by this document are implementation-defined.
+
+A converter MUST rely on the OCI image configuration to build the OCI runtime configuration as described by this document; this will create the "default generated runtime configuration".
+
+The "default generated runtime configuration" MAY be overridden or combined with externally provided inputs from the caller.
+In addition, a converter MAY have its own implementation-defined defaults and extensions which MAY be combined with the "default generated runtime configuration".
+The restrictions in this document refer only to combining implementation-defined defaults with the "default generated runtime configuration".
+Externally provided inputs are considered to be a modification of the `application/vnd.oci.image.config.v1+json` used as a source, and such modifications have no restrictions.
+
+For example, externally provided inputs MAY cause an environment variable to be added, removed or changed.
+However an implementation-defined default SHOULD NOT result in an environment variable being removed or changed.
+
+[oci-runtime-bundle]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/bundle.md
+[oci-runtime-config]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc5/config.md
+
+## Verbatim Fields
+
+Certain image configuration fields have an identical counterpart in the runtime configuration.
+Some of these are purely annotation-based fields, and have been extracted into a [separate subsection](#annotation-fields).
+A compliant configuration converter MUST extract the following fields verbatim to the corresponding field in the generated runtime configuration:
+
+| Image Field         | Runtime Field   | Notes |
+| ------------------- | --------------- | ----- |
+| `architecture`      | `platform.arch` |       |
+| `os`                | `platform.os`   |       |
+| `Config.WorkingDir` | `process.cwd`   |       |
+| `Config.Env`        | `process.env`   | 1     |
+| `Config.Entrypoint` | `process.args`  | 2     |
+| `Config.Cmd`        | `process.args`  | 2     |
+
+1. The converter MAY add additional entries to `process.env` but it SHOULD NOT add entries that have variable names present in `Config.Env`.
+2. If both `Config.Entrypoint` and `Config.Cmd` are specified, the converter MUST append the value of `Config.Cmd` to the value of `Config.Entrypoint` and set `process.args` to that combined value.
+
+### Annotation Fields
+
+These fields all affect the `annotations` of the runtime configuration, and are thus subject to [precedence](#annotations).
+
+| Image Field         | Runtime Field   | Notes |
+| ------------------- | --------------- | ----- |
+| `author`            | `annotations`   | 1,2   |
+| `created`           | `annotations`   | 1,3   |
+| `Config.Labels`     | `annotations`   |       |
+| `Config.StopSignal` | `annotations`   | 1,4   |
+
+1. If a user has explicitly specified this annotation with `Config.Labels`, then the value specified in this field takes lower [precedence](#annotations) and the converter MUST instead use the value from `Config.Labels`.
+2. The value of this field MUST be set as the value of `org.opencontainers.image.author` in `annotations`.
+3. The value of this field MUST be set as the value of `org.opencontainers.image.created` in `annotations`.
+4. The value of this field MUST be set as the value of `org.opencontainers.image.stopSignal` in `annotations`.
+
+## Parsed Fields
+
+Certain image configuration fields have a counterpart that must first be translated.
+A compliant configuration converter SHOULD parse all of these fields and set the corresponding fields in the generated runtime configuration:
+
+| Image Field         | Runtime Field    |
+| ------------------- | ---------------  |
+| `Config.User`       | `process.user.*` |
+
+The method of parsing the above image fields are described in the following sections.
+
+### `Config.User`
+
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are numeric (`uid` or `gid`) then the values MUST be copied verbatim to `process.user.uid` and `process.user.gid` respectively.
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are not numeric (`user` or `group`) then a converter SHOULD resolve the user information using a method appropriate for the container's context.
+For Unix-like systems, this MAY involve resolution through NSS or parsing `/etc/passwd` from the extracted container's root filesystem to determine the values of `process.user.uid` and `process.user.gid`.
+
+In addition, a converter SHOULD set the value of `process.user.additionalGids` to a value corresponding to the user in the container's context described by `Config.User`.
+For Unix-like systems, this MAY involve resolution through NSS or parsing `/etc/group` and determining the group memberships of the user specified in `process.user.uid`.
+If the value of [`user`](config.md#properties) in `Config.User` is numeric, the converter SHOULD NOT modify `process.user.additionalGids`.
+
+If `Config.User` is not defined, the converted `process.user` value is implementation-defined.
+If `Config.User` does not correspond to a user in the container's context, the converter MUST return an error.
+
+## Optional Fields
+
+Certain image configuration fields are not applicable to all conversion use cases, and thus are optional for configuration converters to implement.
+A compliant configuration converter SHOULD provide a way for users to extract these fields into the generated runtime configuration:
+
+| Image Field           | Runtime Field      | Notes |
+| --------------------- | ------------------ | ----- |
+| `Config.ExposedPorts` | `annotations`      | 1     |
+| `Config.Volumes`      | `mounts`           | 2     |
+
+1. The runtime configuration does not have a corresponding field for this image field.
+   However, converters SHOULD set the [`org.opencontainers.image.exposedPorts` annotation](#config.exposedports).
+2. If a converter implements conversion for this field using mountpoints, it SHOULD set the `destination` of the mountpoint to the value specified in `Config.Volumes`.
+   The other `mounts` fields are platform and context dependent, and thus are implementation-defined.
+   Note that the implementation of `Config.Volumes` need not use mountpoints, as it is effectively a mask of the filesystem.
+
+### `Config.ExposedPorts`
+
+The OCI runtime configuration does not provide a way of expressing the concept of "container exposed ports".
+However, converters SHOULD set the **org.opencontainers.image.exposedPorts** annotation, unless doing so will [cause a conflict](#annotations).
+
+**org.opencontainers.image.exposedPorts** is the list of values that correspond to the [keys defined for `Config.ExposedPorts`](config.md) (string, comma-separated values).
+
+## Annotations
+
+There are three ways of annotating an OCI image in this specification:
+
+1. `Config.Labels` in the [configuration](config.md) of the image.
+2. `annotations` in the [manifest](manifest.md) of the image.
+3. `annotations` in the [image index](image-index.md) of the image.
+
+In addition, there are also implicit annotations that are defined by this section which are determined from the values of the image configuration.
+A converter SHOULD NOT attempt to extract annotations from [manifests](manifest.md) or [image indices](image-index.md).
+If there is a conflict (same key but different value) between an implicit annotation (or annotation in [manifests](manifest.md) or [image indices](image-index.md)) and an explicitly specified annotation in `Config.Labels`, the value specified in `Config.Labels` MUST take precedence.
+
+A converter MAY add annotations which have keys not specified in the image.
+A converter MUST NOT modify the values of annotations specified in the image.
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ DOC_FILES := \
 	layer.md \
 	config.md \
 	annotations.md \
+	conversion.md \
 	considerations.md \
 	implementations.md
 
--- a/manifest.md
+++ b/manifest.md
@@ -4,7 +4,7 @@ There are three main goals of the Image Manifest Specification.
 The first goal is content-addressable images, by supporting an image model where the image's configuration can be hashed to generate a unique ID for the image and its components.
 The second goal is to allow multi-architecture images, through a "fat manifest" which references image manifests for platform-specific versions of an image.
 In OCI, this is codified in an [image index](image-index.md).
-The third goal is to be translatable to the [OCI Runtime Specification](https://github.com/opencontainers/runtime-spec).
+The third goal is to be [translatable](conversion.md) to the [OCI Runtime Specification](https://github.com/opencontainers/runtime-spec).
 
 This section defines the `application/vnd.oci.image.manifest.v1+json` [media type](media-types.md).
 For the media type(s) that this is compatible with see the [matrix](media-types.md#compatibility-matrix).


conversion.md
```diff
@@ -0,0 +1,89 @@
+# Conversion to OCI Runtime Configuration
+
+When extracting an OCI Image into an [OCI Runtime bundle][oci-runtime-bundle], two components of the orthogonal components of the extraction are relevant:
```
wking: “components of the orthogonal components of the extraction” → “orthogonal components”?

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+| ------------------- | --------------- | ----- |
+| `architecture`      | `platform.arch` |       |
+| `os`                | `platform.os`   |       |
+| `Config.WorkingDir` | `process.cwd`   |       |
+| `Config.Env`        | `process.env`   | 1     |
+| `Config.Entrypoint` | `process.args`  | 2     |
+| `Config.Cmd`        | `process.args`  | 2     |
+| `Config.Labels`     | `annotations`   | 3     |
+
+1. The converter MAY *prepend* additional environment variables to `process.env`, but it MUST NOT *append* values to the environment.
```
wking: From [POSIX][1]:

> There is no meaning associated with the order of strings in the environment. If more than one string in an environment of a process has the same name, the consequences are undefined.

So I think we should replace the append/prepend rule with one about not duplicating names.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+3. See the section on [annotations](#annotations) for caveats.
+
+## Parsed Fields
+
+Certain image configuration fields have a counterpart that must first be translated.
+A compliant configuration converter SHOULD parse all of these fields and set the corresponding fields in the generated runtime configuration:
+
+| Image Field         | Runtime Field    | Notes |
+| ------------------- | ---------------  | ----- |
+| `Config.User`       | `process.user.*` |       |
```
wking: Unless we have notes entries, I'd rather drop the column.  You have notes on this conversion below, so maybe add `[notes](#config-user)` or similar to the notes column here?

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+
+### `Config.User`
+
+When converting the value of `Config.User` a converter SHOULD resolve the user information using a method appropriate for the container's context.
+For Linux based systems this would involve parsing `/etc/passwd` from the extracted container's root filesystem to determine the values of `process.user.uid` and `process.user.gid`.
```
wking: The config docs allow [both numeric and named user/group values][1], and only mention `/etc/passwd` in [the context of default groups][2].  So I think we need to weaken this line to say “MAY involve parsing `/etc/passwd` and `/etc/group`” to allow for:

* Configs using numberic uid/gids (where there is no need for a lookup)
* Agents using other NSS backends (e.g. LDAP), since the config spec doesn't require a particular NSS backend.

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/config.md#L79
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/config.md#L80

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+
+In addition, a converter SHOULD set the value of `process.user.additionalGids` to a value corresponding to the user described by `Config.User`.
```
wking: Probably “corresponding to the user” → “corresponding to the group”.

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+For Linux based systems this would require parsing `/etc/group` and determining the group memberships of the user specified in `process.user.uid`.
+
+If `Config.User` is not defined, the converter MUST choose a default user appropriate for the container's context.
```
wking: MUST + “appropriate” combines a strict requirement with a very wiggly target.  I'd rather say:

> If `Config.User` is not set, the converted `process.user` value is [implementation-defined][1].

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/README.md#L28

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+If `Config.User` does not correspond to a user in the container's context, the convert MUST return an error.
+
+## Optional Fields
+
+Certain image configuration fields are not considered to
```
wking: This sentence is incomplete.

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+A compliant configuration converter SHOULD provide a way for users to extract these fields into the generated runtime configuration:
+
+| Image Field           | Runtime Field      | Notes |
+| --------------------- | ------------------ | ----- |
+| `Config.Memory`       | platform dependant | 1     |
```
wking: “dependant” → “dependent”

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+| `Config.MemorySwap`   | platform dependant | 1     |
+| `Config.CPUShares`    | platform dependant | 1     |
+| `Config.ExposedPorts` | none               | 2     |
+| `Config.Volumes`      | `mounts`           | 3     |
+
+1. The meaning of these fields is platform dependant, and runtimes MAY ignore these fields.
```
wking: “runtimes” → “converters”

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+   On Linux these image fields correspond to the runtime fields `linux.resources.memory.limit`, `linux.resources.memory.swap`, and `linux.resources.cpu.shares` respectively.
+2. The runtime configuration does not have a corresponding field for this image field.
+   Converters SHOULD use this as a hint to allow a container to provide a network service from the corresponding container-side port.
```
wking: Do we want to SHOULD an annotations key for this information?

conversion.md
```diff
@@ -0,0 +1,89 @@
...
+3. The converter MUST set the `destination` of the mountpoint to the value specified in `Config.Volumes`.
+   The other fields are platform and context dependant, though the converter SHOULD NOT allow volumes to be include in newly created [layers](layer.md).
```
wking: “include” → “included”.

And “SHOULD NOT allow volumes to be included in the newly created layers” sounds like it may belong [here][1], and not in the conversion spec.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc3/layer.md#determining-changes

conversion.md
cyphar: No, because on Linux `additionalGids` is defined by group association of a user. Maybe "corresponding to the value in `Config.User`" if you want to make it more generic.

conversion.md
cyphar: Maybe, but since we haven't defined such an annotation in the `org.opencontainers.` namespace I'm a bit iffy about it.

conversion.md
wking: Ah, right.  Maybe there is wording that makes “this is an NSS lookup in the container context” more obvious, so folks don't trip into the same hole I just did.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+## Annotations
+
+There are three ways of annotating an OCI image in this specification:
+
+1. `Config.Labels` in the [configuration](config.md) of an image.
+2. `annotations` in the [manifest](manifest.md) of an image.
+3. `annotations` in the [manifest list](manifest-list.md) of an image.
+
+A converter SHOULD extract all three of these sets of annotations to `annotations` in the runtime configuration.
+However, since there may be conflicts between different sets of annotations (same key but different value), the order of precedence SHOULD be in the following (ascending) order:
```
stevvooe: Why this ordering?

To tell you the truth, this is why have been against adding annotations everywhere, especially when there isn't easy ways to distinguish runtime annotations/labels from those inherited from the source image.

To tell you the truth, this should probably be the opposite order. Manifests and Manifest-List tend not to survive to the runtime portion and are more oriented towards distribution.

cyphar: The ordering was based on what I thought made sense (the different blob layers are like onions, with each descriptor dereference going down a layer). But I wasn't sure why there were multi-level annotations in the first place (I assumed it was historical).

The other reason is that it means that if you share the same config between different manifests it's possible to override the config's annotations without needing to modify it -- if the ordering was in the reverse order then you wouldn't be able to do anything similar (you'd have to store two copies of the config).

wking: > To tell you the truth, this should probably be the opposite order. Manifests and Manifest-List tend not to survive to the runtime portion and are more oriented towards distribution.

These are in order of ascending precedence, and while that's a somewhat unusual way to list precedence, it does mean that `Config.Labels` has the highest precedence.  I think that's what you're asking for too.

wking: > The other reason is that it means that if you share the same config between different manifests it's possible to override the config's annotations without needing to modify it…

Heh, so I read this completely wrong.  Whatever we settle on, I suggest a clarifying example to avoid future confusion like mine.

wking: > I'd also understand if we don't include the annotations from manifests at all (which would be an even better idea because it would mean that the conversion is entirely done with v1.Image and doesn't require knowing what the manifest was).

This sounds reasonable to me.

cyphar: @stevvooe @philips What do you think? Should we not require that runtimes include Manifest and ManifestList annotations in the runtime bundle?

cyphar:
@stevvooe Can you take another look? I've switched around the wording and changed my annotation conversion description so that *only* `Config.Labels` MUST be extracted (and it takes precedence over any of the implicit conversions).

stevvooe:
@cyphar Could update this to reflect the removal of `Memory`, et. al.?

cyphar:
Yup, it's on my list of things to do. If I get the chance I'll do it this weekend or on Monday if I'm too busy.

cyphar:
@stevvooe Updated.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+When extracting an OCI Image into an [OCI Runtime bundle][oci-runtime-bundle], two orthogonal components of the extraction are relevant:
+
+1. Extraction of the root filesystem from the set of [filesystem layers](layers.md).
+2. Conversion of the [image configuration blob](config.md) to an [OCI Runtime configuration blob][oci-runtime-config].
+
+This section defines how to convert an `application/vnd.oci.image.config.v1+json` blob to an [OCI runtime configuration blob][oci-runtime-config] (the latter component of extraction).
+The former component of extraction is defined [elsewhere](layers.md) and is orthogonal to configuration of a runtime bundle.
+
+[oci-runtime-bundle]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/bundle.md
+[oci-runtime-config]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md
```
wking: runtime-spec has cut an rc3.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+If `Config.User` does not correspond to a user in the container's context, the converter MUST return an error.
+
+## Optional Fields
+
+Certain image configuration fields are not applicable to all conversion usecases, and thus are optional for configuration converter to implement.
+A compliant configuration converter SHOULD provide a way for users to extract these fields into the generated runtime configuration:
+
+| Image Field           | Runtime Field      | Notes |
+| --------------------- | ------------------ | ----- |
+| `Config.ExposedPorts` | none               | 1     |
```
wking: In the verbatim section, you use `annotations` as the runtime field in cases like this.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+| `Config.Volumes`      | `mounts`           | 2     |
+
+1. The runtime configuration does not have a corresponding field for this image field.
+   However, converters SHOULD set the [`org.opencontainers.imageSpec.exposedPorts` annotation](#config.exposedports).
```
wking: Maybe "... annotation to the JSON serialization of the `ExpopsedPorts` value?  And copy footnote 1 about precedence from the verbatim table.  Then we could drop the `ExposedPorrs` section below without losing specificity.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+1. `Config.Labels` in the [configuration](config.md) of an image.
+2. `annotations` in the [manifest](manifest.md) of an image.
+3. `annotations` in the [manifest list](manifest-list.md) of an image.
+
+In addition, there are also implicit annotations that are defined by this section which are determined from the values of the image configuration.
+A converter SHOULD NOT attempt to extract annotations from [manifests](manifest.md) or [manifest lists](manifest-list.md).
+If there is a conflict (same key but different value) between an implicit annotation and an explicitly specified annotation in `Config.Labels`, the value specified in `Config.Labels` MUST take precedence.
+
+A converter MAY add annotations which have keys not specified in the image.
+A converter MUST NOT modify the values of annotations specified in an image.
```
wking: "an image" -> "the image" to match the preceding sentence and be more accurate.

conversion.md
wking: Scratch this, because the JSON for ExposedPorts is wonky.  Your current handling is fine, although coping down your precedence footnote is probably still a good idea for consistency.

conversion.md
```diff
@@ -0,0 +1,99 @@
...
+Certain image configuration fields have an identical counterpart in the runtime configuration.
+A compliant configuration converter MUST extract the following fields verbatim to the corresponding field in the generated runtime configuration:
+
+| Image Field         | Runtime Field   | Notes |
+| ------------------- | --------------- | ----- |
+| `architecture`      | `platform.arch` |       |
+| `os`                | `platform.os`   |       |
+| `author`            | `annotations`   | 1,2   |
+| `created`           | `annotations`   | 1,3   |
+| `os`                | `platform.os`   |       |
```
wking: Duplicate entry for `os`.

conversion.md
```diff
@@ -0,0 +1,99 @@
...
+
+1. The runtime configuration does not have a corresponding field for this image field.
+   However, converters SHOULD set the [`org.opencontainers.imageSpec.exposedPorts` annotation](#config.exposedports).
+2. If a converter implements conversion for this field, it MUST set the `destination` of the mountpoint to the value specified in `Config.Volumes`.
+   The other `mounts` fields are platform and context dependent, and thus are implementation-defined.
+
+### `Config.ExposedPorts`
+
+The OCI runtime configuration does not provide a way of expressing the concept of "container exposed ports".
+However, converters SHOULD provide a hint to the runtime what the set of container exposed ports are by using the **org.opencontainers.imageSpec.exposedPorts** annotation, unless doing so will [cause a conflict](#annotations).
```
wking: "provide a hint to the runtime what the set of container exposed ports are by using" -> "set"?

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+
+The method of parsing the above image fields are described in the following sections.
+
+### `Config.User`
+
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are numeric (`uid` or `gid`) then the values MUST be copied verbatim to `process.user.uid` and `process.user.gid` respectively.
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are not numeric (`user` or `group`) then a converter SHOULD resolve the user information using a method appropriate for the container's context.
+For Linux and Solaris based systems, this MAY involve resolution through NSS or parsing `/etc/passwd` from the extracted container's root filesystem to determine the values of `process.user.uid` and `process.user.gid`.
+
+In addition, a converter SHOULD set the value of `process.user.additionalGids` to a value corresponding to the user described by `Config.User`.
```
stevvooe: This doesn't sound right. `Config.User` should have a colon separated user and primary group. If there is no colon, the primary group is the user id. I don't think there is any support for supplementary (somehow, these where NIH'd to "additional") in the image format.

I am not sure why runtime-spec diverged from this. Having a separate field doesn't really provide much and makes it clear what is primary version. For example, in swarmkit, we have `User` and `Groups`, where user can be user or `user:group` and `Groups` is always supplementary.

I am not sure about the utility of supplementary groups in the image format.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+| `author`            | `annotations`   | 1,2   |
+| `created`           | `annotations`   | 1,3   |
+| `Config.WorkingDir` | `process.cwd`   |       |
+| `Config.Env`        | `process.env`   | 4     |
+| `Config.Entrypoint` | `process.args`  | 5     |
+| `Config.Cmd`        | `process.args`  | 5     |
+| `Config.Labels`     | `annotations`   | 6     |
+
+1. If a user has explicitly specified this annotation with `Config.Labels`, then the value specified in this field takes lower [precedence](#annotations) and the converter MUST instead use the value from `Config.Labels`.
+2. This value of this field MUST be set as the value of `org.opencontainers.authors` in `annotations`.
```
stevvooe: Let's separate the active fields from the annotation fields.

Also, we need to be careful about the impact of picking up image labels into runtime annotations. I know I am having trouble convincing everyone this isn't a good idea, but it is a security hole waiting to happen.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+The method of parsing the above image fields are described in the following sections.
+
+### `Config.User`
+
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are numeric (`uid` or `gid`) then the values MUST be copied verbatim to `process.user.uid` and `process.user.gid` respectively.
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are not numeric (`user` or `group`) then a converter SHOULD resolve the user information using a method appropriate for the container's context.
+For Linux and Solaris based systems, this MAY involve resolution through NSS or parsing `/etc/passwd` from the extracted container's root filesystem to determine the values of `process.user.uid` and `process.user.gid`.
+
+In addition, a converter SHOULD set the value of `process.user.additionalGids` to a value corresponding to the user described by `Config.User`.
+For Linux and Solaris based systems, this MAY involve resolution through NSS or parsing `/etc/group` and determining the group memberships of the user specified in `process.user.uid`.
```
stevvooe: Unices?

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+A compliant configuration converter SHOULD provide a way for users to extract these fields into the generated runtime configuration:
+
+| Image Field           | Runtime Field      | Notes |
+| --------------------- | ------------------ | ----- |
+| `Config.ExposedPorts` | `annotations`      | 1     |
+| `Config.Volumes`      | `mounts`           | 2     |
+
+1. The runtime configuration does not have a corresponding field for this image field.
+   However, converters SHOULD set the [`org.opencontainers.imageSpec.exposedPorts` annotation](#config.exposedports).
+2. If a converter implements conversion for this field, it MUST set the `destination` of the mountpoint to the value specified in `Config.Volumes`.
```
stevvooe: The `Config.Volumes` field doesn't necessarily require a mount. That is really an implementation detail.

`Config.Volumes` really just marks a section of the file system as not part of the primary container filesystem.

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+| `Config.Env`        | `process.env`   | 4     |
+| `Config.Entrypoint` | `process.args`  | 5     |
+| `Config.Cmd`        | `process.args`  | 5     |
+| `Config.Labels`     | `annotations`   | 6     |
+
+1. If a user has explicitly specified this annotation with `Config.Labels`, then the value specified in this field takes lower [precedence](#annotations) and the converter MUST instead use the value from `Config.Labels`.
+2. This value of this field MUST be set as the value of `org.opencontainers.authors` in `annotations`.
+3. This value of this field MUST be set as the value of `org.opencontainers.created` in `annotations`.
+4. The converter MAY add additional entries to `process.env` but it MUST NOT add entries that have variable names present in `Config.Env`.
+5. If both `Config.Entrypoint` and `Config.Cmd` are specified, the converter MUST append the value of `Config.Cmd` to the value of `Config.Entrypoint` and set `process.args` to that combined value.
```
stevvooe: MUST is too strong here when taking user input for entrypoint and args.

conversion.md
wking: > I don't think there is any support for supplementary (somehow, these where NIH'd to "additional") in the image format.

The config doesn't support it directly, which is why the following line suggests a way to extract this from the root filesystem or other source (“…MAY involve resolution through NSS or parsing `/etc/group`…”).  Resolving this via the root filesystem is the same type of thing that this conversion doc [suggest for non-numeric users and groups][1], although @jonboulle recently [questioned the usefulness of generic NSS vs. requiring `/etc/passwd` / `/etc/group`][2].

[1]: https://github.com/cyphar/image-spec/blame/2acf3a6b8cbe2404c6d0087771fbfda5c50a677a/conversion.md#L53-L54
[2]: https://github.com/opencontainers/image-spec/pull/492#discussion_r94411639

cyphar: @stevvooe To clarify, "corresponding to the user" refers to the concept of a user in the image as opposed to the user **string**. The utility is so that groups like `video`, `dialout` and `audio` all still work inside containers.

I actually completely agree with what you said in your first paragraph. The point is that the resolution of supplementary groups (either through NSS or `/etc/group`) is **recommended**.

conversion.md
cyphar: I'll change it to Unix-like

conversion.md
cyphar: Okay, but is there a way of phrasing it in terms of the runtime specification to guarantee this? Maybe there is another way of pasting together a filesystem, but I'm having trouble thinking of one at the moment.

conversion.md
cyphar: My main issue with entrypoint and cmd is that it's not really clear anywhere (outside of the Docker documentation) what their individual purposes are. The semantics also are quite hazy if you read `config.md` (and I think there's some hard-to-parse sentences there too).

conversion.md
stevvooe: I don't have a good suggestion here. The volume really acts as almost a "mask" of the filesystem. The implementation of the mask can vary.

conversion.md
stevvooe: @cyphar `Entrypoint` and `Cmd` are indeed a bit of a mess. We've shored this up _slightly_ in the docker documentation.

If you can find a way to word it such that implementations may substitute either `Entrypoint` or `Cmd` with user input, that would shore up the problem with this statement.

wking: > If you can find a way to word it such that implementations may substitute either Entrypoint or Cmd with user input, that would shore up the problem with this statement.

Without an API spec to pin down “user input”, I think this is going to be hard to do.  Can the user-input additions be a follow-up step?  So:

    image config →{this translation}→ runtime config →{user input}→ runtime config

Then whether the translator handles this translation correctly can be validated without worrying about user input, and whether the user-input injection happens correctly is up to whoever specified the user-input API.

stevvooe: @wking Not really.

wking: If you need to know where the break between the entrypoint and command is, you can turn it around and have:

    image config →{user input}→ image config →{this translation}→ runtime config

Then @cyphar's current append wording matches what is [listed here][1] for the case where both are [arrays of strings][2].  Or is image spec supposed to support the [shell form][3] versions too?

[1]: https://docs.docker.com/engine/reference/builder/#/understand-how-cmd-and-entrypoint-interact
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/config.md#L110
[3]: https://docs.docker.com/engine/reference/builder/#/shell-form-entrypoint-example

conversion.md
```diff
@@ -0,0 +1,98 @@
...
+| ------------------- | ---------------  |
+| `Config.User`       | `process.user.*` |
+
+The method of parsing the above image fields are described in the following sections.
+
+### `Config.User`
+
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are numeric (`uid` or `gid`) then the values MUST be copied verbatim to `process.user.uid` and `process.user.gid` respectively.
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are not numeric (`user` or `group`) then a converter SHOULD resolve the user information using a method appropriate for the container's context.
+For Linux and Solaris based systems, this MAY involve resolution through NSS or parsing `/etc/passwd` from the extracted container's root filesystem to determine the values of `process.user.uid` and `process.user.gid`.
```
cyphar: I don't see a compelling reason to disallow NSS. The point of this section is to say that the mapping of `Config.User` -> `{uid, gid}` is implementation defined, with some hints about how it should be done.

conversion.md
cyphar: @wking 

> Or is image spec supposed to support the shell form versions too?

Absolutely not. IMO those were a bad decision which Docker has had to stick with because of backwards compatibility issues.

cyphar: @stevvooe I'll give it a shot in terms of wording.

conversion.md
```diff
@@ -0,0 +1,109 @@
...
+   Note that the implementation of `Config.Volumes` need not use mountpoints, as it is effectively a mask of the filesystem.
+
+### `Config.ExposedPorts`
+
+The OCI runtime configuration does not provide a way of expressing the concept of "container exposed ports".
+However, converters SHOULD set the **org.opencontainers.imageSpec.exposedPorts** annotation, unless doing so will [cause a conflict](#annotations).
+
+**org.opencontainers.imageSpec.exposedPorts** is a list of comma-separated values that correspond to the [keys defined for `Config.ExposedPorts`](config.md).
+
+## Annotations
```
runcom: as pointed out in #541 - can we have an optional, but still `org.opencontainers.*` prefixed, annotation which specifies the signal the image author wants the container to be kill'ed?

stevvooe: This should have already been here. Is it missing from the config?

https://github.com/docker/docker/blob/master/api/types/container/config.go#L59

cyphar:
I would prefer if we can get this into one of the next milestones, @opencontainers/image-spec-maintainers. WDYT?

cyphar:
/ping @stevvooe You said you'd take another look at this?

conversion.md
```diff
@@ -0,0 +1,109 @@
...
+
+There are three ways of annotating an OCI image in this specification:
+
+1. `Config.Labels` in the [configuration](config.md) of the image.
+2. `annotations` in the [manifest](manifest.md) of the image.
+3. `annotations` in the [manifest list](manifest-list.md) of the image.
```
jonboulle: manifest list -> image index

conversion.md
```diff
@@ -0,0 +1,109 @@
...
+When extracting an OCI Image into an [OCI Runtime bundle][oci-runtime-bundle], two orthogonal components of the extraction are relevant:
+
+1. Extraction of the root filesystem from the set of [filesystem layers](layers.md).
+2. Conversion of the [image configuration blob](config.md) to an [OCI Runtime configuration blob][oci-runtime-config].
+
+This section defines how to convert an `application/vnd.oci.image.config.v1+json` blob to an [OCI runtime configuration blob][oci-runtime-config] (the latter component of extraction).
+The former component of extraction is defined [elsewhere](layers.md) and is orthogonal to configuration of a runtime bundle.
+The values of runtime configuration properties not specified by this document are implementation-defined.
+
+[oci-runtime-bundle]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/bundle.md
```
jonboulle: probably bump to latest rc

conversion.md
```diff
@@ -0,0 +1,109 @@
...
+## Annotations
+
+There are three ways of annotating an OCI image in this specification:
+
+1. `Config.Labels` in the [configuration](config.md) of the image.
+2. `annotations` in the [manifest](manifest.md) of the image.
+3. `annotations` in the [manifest list](manifest-list.md) of the image.
+
+In addition, there are also implicit annotations that are defined by this section which are determined from the values of the image configuration.
+A converter SHOULD NOT attempt to extract annotations from [manifests](manifest.md) or [manifest lists](manifest-list.md).
```
jonboulle: manifest list -> image index

conversion.md
```diff
@@ -0,0 +1,109 @@
...
+| `architecture`      | `platform.arch` |       |
+| `os`                | `platform.os`   |       |
+| `Config.WorkingDir` | `process.cwd`   |       |
+| `Config.Env`        | `process.env`   | 1     |
+| `Config.Entrypoint` | `process.args`  | 2     |
+| `Config.Cmd`        | `process.args`  | 2     |
+
+1. The converter MAY add additional entries to `process.env` but it MUST NOT add entries that have variable names present in `Config.Env`.
+2. If both `Config.Entrypoint` and `Config.Cmd` are specified, the converter MUST append the value of `Config.Cmd` to the value of `Config.Entrypoint` and set `process.args` to that combined value.
+   However, if a user has specified an alternate value for `Config.Entrypoint` or `Config.Cmd`, the user provided value SHOULD supercede the configuration-defined value in the above computation of `process.args`.
```
jonboulle: prefer "supersede"

cyphar:
Rebased, addressed @jonboulle's comments and added a conversion for `StopSignal`.

jonboulle:
LGTM. Thanks @cyphar

cyphar:
Squashed everything for proper review. Sorry for the bump @jonboulle.

jonboulle:
:+1:

conversion.md
```diff
@@ -0,0 +1,111 @@
+# Conversion to OCI Runtime Configuration
+
+When extracting an OCI Image into an [OCI Runtime bundle][oci-runtime-bundle], two orthogonal components of the extraction are relevant:
+
+1. Extraction of the root filesystem from the set of [filesystem layers](layers.md).
```
stevvooe: How will we verify implementations that have either the bundle or config opaque to the end user?

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Verbatim Fields
+
+Certain image configuration fields have an identical counterpart in the runtime configuration.
+Some of these are purely annotation-based fields, and have been extracted into a [separate subsection](#annotation-fields).
+A compliant configuration converter MUST extract the following fields verbatim to the corresponding field in the generated runtime configuration:
+
+| Image Field         | Runtime Field   | Notes |
+| ------------------- | --------------- | ----- |
+| `architecture`      | `platform.arch` |       |
```
stevvooe: Is this some sort of field path syntax?

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+
+| Image Field         | Runtime Field   | Notes |
+| ------------------- | --------------- | ----- |
+| `author`            | `annotations`   | 1,2   |
+| `created`           | `annotations`   | 1,3   |
+| `Config.Labels`     | `annotations`   | 4     |
+| `Config.StopSignal` | `annotations`   | 1,5   |
+
+1. If a user has explicitly specified this annotation with `Config.Labels`, then the value specified in this field takes lower [precedence](#annotations) and the converter MUST instead use the value from `Config.Labels`.
+2. The value of this field MUST be set as the value of `org.opencontainers.authors` in `annotations`.
```
stevvooe: Just make a separate table heading for this.

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+| Image Field         | Runtime Field    |
+| ------------------- | ---------------  |
+| `Config.User`       | `process.user.*` |
+
+The method of parsing the above image fields are described in the following sections.
+
+### `Config.User`
+
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are numeric (`uid` or `gid`) then the values MUST be copied verbatim to `process.user.uid` and `process.user.gid` respectively.
+If the values of [`user` or `group`](config.md#properties) in `Config.User` are not numeric (`user` or `group`) then a converter SHOULD resolve the user information using a method appropriate for the container's context.
```
stevvooe: This is fairly complex to perform outside the container's context. Wouldn't it be easier to set the direct values in the runtime and have it perform the correct uid lookups before exec?

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+For Unix-like systems, this MAY involve resolution through NSS or parsing `/etc/passwd` from the extracted container's root filesystem to determine the values of `process.user.uid` and `process.user.gid`.
+
+In addition, a converter SHOULD set the value of `process.user.additionalGids` to a value corresponding to the user in the container's context described by `Config.User`.
```
stevvooe: Again, this is very complex to do correctly outside the container's context.

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+These fields all affect the `annotations` of the runtime configuration, and are thus subject to [precedence](#annotations).
+
+| Image Field         | Runtime Field   | Notes |
+| ------------------- | --------------- | ----- |
+| `author`            | `annotations`   | 1,2   |
+| `created`           | `annotations`   | 1,3   |
+| `Config.Labels`     | `annotations`   | 4     |
+| `Config.StopSignal` | `annotations`   | 1,5   |
+
+1. If a user has explicitly specified this annotation with `Config.Labels`, then the value specified in this field takes lower [precedence](#annotations) and the converter MUST instead use the value from `Config.Labels`.
```
stevvooe: Does this differentiate between the time the image was created and the container was created? What is the point in making it look like the container was created at the same time as the image? Does the container have a different created time on it or is that part of the runtime information?

stevvooe:
@cyphar I mostly have a few questions about implementing certain aspects around user conversion. Specifically, resolving user and group names outside of the context of a container correctly is very hard, especially when you start involving nss and pam. I'm wondering if the runtime should do this pre-exec to remove that burden from the converter.

conversion.md
cyphar: Yeah, it's in `platform/arch`. I wasn't sure how to reference it, so I felt the `.` dereferencing style was clear enough.

conversion.md
cyphar: In the same way that we would have to test runtime implementations that similarly make the bundle or config opaque to the user. This document only concerns tools that implement this conversion (that intend to actually also implement either a runtime or to produce a `config.json` to the user).

conversion.md
cyphar: I assume you're referring to `org.opencontainers.created`. This is actually a piece of metadata from the image which isn't really represented in the runtime specification (the time when the image was created) because the runtime doesn't care about images.

However, information about when the container is created is already provided through the `state` API for OCI runtimes -- which is orthogonal to the container metadata.

conversion.md
cyphar: The OCI runtime configuration only accepts fully resolved `uid`s, `gid`s and `additionalGids` for several reasons. As a result, the runtime simply can't do the resolution and it needs to be done during the conversion from `Image` config to `config.json`.

However, I'm not sure I agree that (for the simple case) it is hard to implement. I've intentionally left this section vague when the values of `Config.User` are **non-numeric**, because at a certain level a runtime could implement whatever the hell they want there.

Though it should be noted that `libcontainer/user` is an implementation of `/etc/passwd` and `/etc/group` parsing which implements this already.

conversion.md
coolljt0725: link broken, `layers.md` should be `layer.md`

conversion.md
```diff
@@ -0,0 +1,111 @@
+# Conversion to OCI Runtime Configuration
+
+When extracting an OCI Image into an [OCI Runtime bundle][oci-runtime-bundle], two orthogonal components of the extraction are relevant:
+
+1. Extraction of the root filesystem from the set of [filesystem layers](layers.md).
+2. Conversion of the [image configuration blob](config.md) to an [OCI Runtime configuration blob][oci-runtime-config].
+
+This section defines how to convert an `application/vnd.oci.image.config.v1+json` blob to an [OCI runtime configuration blob][oci-runtime-config] (the latter component of extraction).
+The former component of extraction is defined [elsewhere](layers.md) and is orthogonal to configuration of a runtime bundle.
```
coolljt0725: same here, `layers.md` -> `layer.dm`

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+| Image Field         | Runtime Field   | Notes |
+| ------------------- | --------------- | ----- |
+| `architecture`      | `platform.arch` |       |
+| `os`                | `platform.os`   |       |
+| `Config.WorkingDir` | `process.cwd`   |       |
+| `Config.Env`        | `process.env`   | 1     |
+| `Config.Entrypoint` | `process.args`  | 2     |
+| `Config.Cmd`        | `process.args`  | 2     |
+
+1. The converter MAY add additional entries to `process.env` but it MUST NOT add entries that have variable names present in `Config.Env`.
```
coolljt0725: but user can override the env exists in `Config.Env`, right?

conversion.md
stevvooe: > This document only concerns tools that implement this conversion (that intend to actually also implement either a runtime or to produce a config.json to the user).

In practice, this conversion will be performed by systems that will make this opaque. The advice here applies, but may not be verifiable without an introspection mechanism (read: leaky abstraction). For example, we'd like to be able to pass this spec through several systems before writing out the resulting config.json.

conversion.md
stevvooe: @cyphar I'm just worried about reproducing https://linux.die.net/man/3/getpwnam behavior outside the container, when the runtime might be best positioned to do this correctly.

@crosbymichael Any thoughts here?

conversion.md
stevvooe: @cyphar Should we prefix the namespace with `org.opencontainers.image` to make this clear?

@vbatts ?

vbatts:
per the discussion, this needs to be the minimum viable path to a testable valid image config to a testable valid runtime config. Though not to say that two different implementations will result in the same runtime config result.

conversion.md
cyphar: Yeah, good idea.

vbatts:
Fixes #87

jonboulle:
> this needs to be the minimum viable path to a testable valid image config to a testable valid runtime config. 

what does this imply for the PR? (i.e. what needs to be done to move forward?) Or is this just more of a general high level comment?

jonboulle:
bump @vbatts @stevvooe ?

stevvooe:
> bump @vbatts @stevvooe ?

@jonboulle Sorry for not responding earlier.

My current concern is that the language may be too narrow.

Let's get more eyes on this. @tonistiigi @dmcgowan @aaronlehmann

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+1. `Config.Labels` in the [configuration](config.md) of the image.
+2. `annotations` in the [manifest](manifest.md) of the image.
+3. `annotations` in the [image index](image-index.md) of the image.
+
+In addition, there are also implicit annotations that are defined by this section which are determined from the values of the image configuration.
+A converter SHOULD NOT attempt to extract annotations from [manifests](manifest.md) or [image indices](image-index.md).
+If there is a conflict (same key but different value) between an implicit annotation and an explicitly specified annotation in `Config.Labels`, the value specified in `Config.Labels` MUST take precedence.
+
+A converter MAY add annotations which have keys not specified in the image.
+A converter MUST NOT modify the values of annotations specified in the image.
```
stevvooe: What if a converter wants to modify the exposed ports?

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+1. The runtime configuration does not have a corresponding field for this image field.
+   However, converters SHOULD set the [`org.opencontainers.imageSpec.exposedPorts` annotation](#config.exposedports).
+2. If a converter implements conversion for this field using mountpoints, it SHOULD set the `destination` of the mountpoint to the value specified in `Config.Volumes`.
+   The other `mounts` fields are platform and context dependent, and thus are implementation-defined.
+   Note that the implementation of `Config.Volumes` need not use mountpoints, as it is effectively a mask of the filesystem.
+
+### `Config.ExposedPorts`
+
+The OCI runtime configuration does not provide a way of expressing the concept of "container exposed ports".
+However, converters SHOULD set the **org.opencontainers.imageSpec.exposedPorts** annotation, unless doing so will [cause a conflict](#annotations).
```
stevvooe: `org.opencontainers.image`.

conversion.md
dmcgowan: > but nothing is stopping a converter from running a container (with a partially generated config.json) to figure out the UID/GID.

This seems pretty ugly, forces the converter to mount the rootfs just to resolve. What is the more common value for `Config.User`, name:group or uid:gid?

conversion.md
cyphar: Should we allow converters to append to the list of exposed ports? Allowing removal wouldn't be a good idea IMO because it would mean that images might not have ports exposed as they wanted.

conversion.md
cyphar: @dmcgowan `name:group` is used very extensively in Docker images. You don't **need** to resolve NSS -- I expect that most converters would just parse `/etc/passwd` and `/etc/group`. Hell, the converter doesn't even have to support `name:group`.

conversion.md
stevvooe: Shouldn't operators be in control of exposed ports?

conversion.md
stevvooe: In general, the fields can be overridden as a user sees fit. I am not sure if this document reflects that intent.

conversion.md
stevvooe: > You don't need to resolve NSS -- I expect that most converters would just parse /etc/passwd and /etc/group

This might be a no go for certain organizations, although I haven't seen this problem in container deployments. Why do you make the assertion that NSS isn't required?

conversion.md
jonboulle: IMO, we're in a slightly strange position here where we want to ensure some consistency in the conversion process, but also ultimately allow executors to generally do (override) whatever they want at runtime. 

I see two possible paths out of this:
i) add a general note about this, separating the conversor and operator phases - i.e. while a converter MUST follow the MUSTs in the described conversion process, _after_ the conversion has happened, any subsequent tooling (like an executor) can do whatever it likes
ii) move to softer ("SHOULD") wording for all of the individual prescriptions in this document

cyphar: I was about to say what you just said @jonboulle. My inclination is that i) is better than ii) because ii) means that we can no longer have the concept of "the runtime is allowed to add stuff given these restrictions" not to mention that there is some SHOULD wording in this document which is a guideline while the MUST wording is much more clearly "if you don't do this, you're going to massively confuse users".

jonboulle: Yep, good point; I suggest we move forward with i) unless anyone has a better idea.

stevvooe: > Yep, good point; I suggest we move forward with i) unless anyone has a better idea.

What is wrong with making this part of the specification informational? Sometimes it is better to do nothing than to do something that is sub-optimal. In that vein, ii) is the much better option. I see no reason to force implementers to follow these rules, _before_, _during_ or _after_ the conversion process. Doing so makes the specification much too narrow.

In the case of exposed ports, it is particularly problematic. Why force a "compliant" implementation to label a config with these ports if the networking stack completely ignores this value and doesn't expose these ports? It makes no sense.

cyphar: Because if we make it only informational, then a user of the specification doesn't have any meaningful guarantees about what their image config is going to be converted to. If we make this purely informational, an implementation could just ignore the entire document (or misread it and our automated compliance code wouldn't check this) and users would be out of luck when it comes to taking an image an converting it to a runtime configuration.

If you want to loosen the requirements on specific properties, that's fine. But I don't understand wanting to make it so that there are no requirements on implementations to do anything meaningful when converting an image config to a runtime config. What benefit would a document like this be if it's not actually specifying anything?

jonboulle: @opencontainers/image-spec-maintainers

stevvooe: @cyphar Please answer this question:

> Why force a "compliant" implementation to label a config with these ports if the networking stack completely ignores this value and doesn't expose these ports?

The compliant result would both be wrong and confusing for users. Is this not what we are trying to avoid?

vbatts:
reading over this, i like the approach.
LGTM

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+
+1. Extraction of the root filesystem from the set of [filesystem layers](layers.md).
+2. Conversion of the [image configuration blob](config.md) to an [OCI Runtime configuration blob][oci-runtime-config].
+
+This section defines how to convert an `application/vnd.oci.image.config.v1+json` blob to an [OCI runtime configuration blob][oci-runtime-config] (the latter component of extraction).
+The former component of extraction is defined [elsewhere](layers.md) and is orthogonal to configuration of a runtime bundle.
+The values of runtime configuration properties not specified by this document are implementation-defined.
+
+[oci-runtime-bundle]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc4/bundle.md
+[oci-runtime-config]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc4/config.md
```
wking: Bump these to rc5?  Or should we just wait until runtime-spec 1.0, and then bump these then (in a follow-up PR if this lands before runtime-spec cuts a 1.0)?

conversion.md
```diff
@@ -0,0 +1,111 @@
...
+| `architecture`      | `platform.arch` |       |
+| `os`                | `platform.os`   |       |
+| `Config.WorkingDir` | `process.cwd`   |       |
+| `Config.Env`        | `process.env`   | 1     |
+| `Config.Entrypoint` | `process.args`  | 2     |
+| `Config.Cmd`        | `process.args`  | 2     |
+
+1. The converter MAY add additional entries to `process.env` but it MUST NOT add entries that have variable names present in `Config.Env`.
+2. If both `Config.Entrypoint` and `Config.Cmd` are specified, the converter MUST append the value of `Config.Cmd` to the value of `Config.Entrypoint` and set `process.args` to that combined value.
+   However, if a user has specified an alternate value for `Config.Entrypoint` or `Config.Cmd`, the user provided value SHOULD supersede the configuration-defined value in the above computation of `process.args`.
```
wking: I don't think we need an `Entrypoint`/`Cmd`-specific call out for these sort of overrides, or for the earlier `Env` augmentation.  Can we bundle this into the (planned?) [general note][1] about “when you're not asked to do a strict conversion, you can make whatever changes you want”?

[1]: https://github.com/opencontainers/image-spec/pull/492#discussion_r109197328

conversion.md
wking: > Just make a separate table heading for this.

We probably still want this new column [before folks LGTM][1] and merge this PR, although follow-up PRs would work too.

[1]: https://github.com/opencontainers/image-spec/pull/492#issuecomment-291311776

conversion.md
crosbymichael: This does not make sense.  A user can specify runtime options that overwrite image options.

jonboulle:
@crosbymichael did you actually read the discussion in https://github.com/opencontainers/image-spec/pull/492#discussion_r109197328 ?

crosbymichael:
@jonboulle yes, but the document that is being reviewed is still incorrect.

jonboulle:
So you discount that discussion (which specifically addressed your concerns) to instead cast general doubt upon the PR? 

I might invite you to re-read the agreed-upon UX here: https://github.com/opencontainers/image-spec#running-an-oci-image

I am going to be very frank: I am tired of the number of image-spec discussions that are turning into anyone-with-a-docker.com-email-address disputing what is generally at least loose consensus among the rest of the image-spec community. Whose interests do you have in mind here?

cyphar:
@crosbymichael 

> For one, I don't see why you feel like this is something critical to specify, forcing all implementations to follow this conversion in order to be compliant. Personally, it not the image spec's concern how I consume the content, its mine.

Right. But "it's mine" refers to the user, and if we don't provide the user any guarantees of how their image will actually be run, then what is the point of having any fields in the image configuration -- an implementation could just do whatever they like with that information and still be compliant with the image spec. If you agree that's a bad thing, then we have to have some sort of language in the spec that defines what parts of an image configuration are necessary to convert.

Now, things are complicated by the fact that most implementations allow you to out-of-band change some parameters of the image. My opinion on this topic is that we have a caveat on this entire document that states "this document's semantics can be superseded if the user explicitly changes parts of the configuration" -- which is something that should go without saying in the first place, but we might as well include it. I don't see what the benefit is of having an entirely optional document (sure, it might mean that some implementations will implement it, but if it's optional then we won't compliance test it and then misunderstandings by implementations won't be considered "non-compliance").

> Second, with the env example that I commented on, it makes so sense what so ever.

As I pointed out above, the point of the wording in the document is that an implementation is allowed to add their own environment variables automatically (without the user's consent) as long as the user's defined environment variables aren't modified. If a user explicitly says "hey, I want this variable to be X instead of Y" then obviously that supersedes the value in the image. Why is this even a point of contention? Am I misunderstanding the conversation, or did I not make this clear enough in the various discussions above in this issue?

I'm trying to make it so that as a user, I (somewhat) know what an image-spec implementation is going to do when I tell it to give me a runtime-spec bundle. Without that, I have no guarantees that my images will be extracted and executed in any similar way between different implementations (thus, there's an interoperability problem).

I'm starting to wonder whether interoperability is a concern of Docker Inc? Specifically interoperability when it comes to user experience?

crosbymichael:
Keeping this technical, I provided a few examples where the document does not make sense and would hurt existing implementations.  The wording in this document is for explicit with the `MUST`s and it adds confusion to how things are mapped and what is required from an implementation or not. 

The discussion is good but the content of the PR is that will be merged and it still adds to confusion.

cyphar:
@crosbymichael How would you feel about a document-wide caveat saying that if an implementation allows users to specify out-of-band modifications then those modifications trump the conversion semantics for the related keys?

The only reason I don't like dropping `MUST`s from this document is that it then means that implementations could just implement a conversion that is blatantly incorrect and we couldn't say "this is not compliant" because it would be compliant.

crosbymichael:
@cyphar I think that would help.  We also need to define who the user is because there can be systems like a PaaS where its not exactly an end consumer that is creating the container based on the image but a more automated system that is the consumer of the image/runtime.

wking:
>  We also need to define who the user is…

As far as the image-spec is concerned, the user is “whichever process is invoking the image-spec unpacking logic”.  So whatever would get compliance-tested is “the image-spec unpacking logic”, and the next level up (which would not get compliance-tested) is “the user”.  In runtime-spec, the language for this is the “[runtime caller][1]”.

[1]: https://github.com/opencontainers/runtime-spec/blame/v1.0.0-rc5/runtime.md#L110

philips:
I agree that there are essentially two use cases:

1) Run this image: `runner /path/to/image-bundle`
2) Run this image and user explicitly overrides defaults: `runner /path/to/image-bundle --environment-file=/tmp/myenvs`

I think the contention on this thread might come down to not having a distinction between things that a converter can do without additional user overrides and not. So, perhaps we should break 1 and 2 down into two concepts:

1. converter relying on the OCI image for all defaults
2. converter relying on a combination of the OCI image defaults and externally provided default overrides

Then we can draw this distinction in the introduction to the document and just blanket say: "A converter MUST rely on the OCI image configuration to build the OCI runtime configuration; this will create the "default generated runtime configuration". This "default generated runtime configuration" MAY be overridden or combined with externally provided inputs from the caller. For example an environment variable may be added, removed, or changed by external input.`

crosbymichael:
@philips perfect.

Having the clear distinction between the two makes things much more flexible.  Having a strict default as well as acknowledging that the default is only half of the problem adds much more clarity for implementations.

The conversion tooling should always produce a consistent default output, however, after it hands the result off to the caller, things can and will be modified to fulfill a request.

cyphar:
@philips That sounds good to me. I can clean up this document with that in mind, I'll work on it as soon as I get the chance.

vbatts:
@cyphar bump. This is topic of interest, and while they're not breaking changes for the spec, it is something we would like to get settled for the v1.0.0-rc6

cyphar:
@vbatts Yes, sorry. I have a draft of this I'm going to push in the next few days. I hope to push it before DockerCon, but we'll see how that goes.

conversion.md
cyphar: @stevvooe Effectively what this section is saying (and maybe you can come up with better wording than me) is that:

1. If the `Config.User` is numerical (`uid:gid` or similar) then `uid` and `gid` in the runtime configuration must be set to the numerical values.

2. If the `Config.User` is not numerical, then the converter may set `uid` and `gid` to any value. However, the **recommendation** is that the converter attempt to resolve the string into a reasonable `uid` and `gid` in the container's context. I don't want to require this (for many reasons, mostly because it would tie `config.json` to the `rootfs` far too much).

conversion.md
cyphar: @stevvooe 

> The compliant result would both be wrong and confusing for users. Is this not what we are trying to avoid?

I don't agree with what the user's expectation should be. Users should expect that the annotations are set, regardless of whether the runtime supports the feature.

If it's ignored, it's ignored and that's a limitation of the runtime that should be documented. But if we allow converters to drop or otherwise mutate information that an image author explicitly specified, I don't like where that line of reasoning is heading -- it would mean that you couldn't be sure (as a user) that a converter won't suddenly stop converting fields like this (meaning that if you had a runtime that **does** support such a feature, it won't be used).

The purpose of the converted annotations is to allow the runtime **if they support it** to be able to fulfil the wishes of the image author. Runtimes are required by the runtime spec to ignore annotations they don't understand, so there should be no concern with a converter generating annotations that certain runtimes may not implement.

cyphar:
I've pushed my attempt at his wording. We can discuss the details at depth at DockerCon (outside the f2f obviously so we don't dominate the session).

cyphar:
I've pushed my attempt at @philips's wording. We can discuss the details at depth at DockerCon (outside the f2f obviously so we don't dominate the session).

The biggest difference from @philips's proposal is that I make a distinction between "externally provided input" (the implication being that it comes from a user) and "converter implementation-defined defaults". I anticipate this will cause tension because of how ill-defined these terms are. Here is my *intention* with this distinction so we can decide whether we agree with the intention and then we can work on the specific wording:

I anticipate that many converters / runtimes are going to want to specify their own extensions and defaults, this might include things like `cri-o`'s annotations that are used for implementing the CRI. Also there might be other examples (such as additions to the environment variables as well). We obviously don't want to restrict that freedom. However, at the same time, I don't want to open the door to converters just blatantly ignoring user-specified (or image-author-specified) options and using their own settings instead. Hence why there are restrictions for "implementation-defined defaults". On the other hand, as we've discussed at length above, such restrictions don't make sense for user-specified "externally provided input" and thus there are no restrictions for such options.

In my mind (and I'm hoping there's some reasonable wording that can represent this intent), the way I'm thinking of these requirements is that user-specified options are like overriding the **source** of the conversion. And implementation-defined defaults can't override anything that is **explicitly** specified in the "source".

/cc @opencontainers/image-spec-maintainers

vbatts:
@cyphar your intentions SGTM

vbatts:
nit: the conversation.md ought to be added to the list in `Makefile` for including in the pdf/html

cyphar:
@vbatts In which case this needs some review. I've gotten some suggestions from @vrothberg who was working on implementing this in `umoci`, but I would really appreciate a proper detailed review of the current state.

conversion.md
```diff
@@ -0,0 +1,119 @@
...
+
+A converter MUST rely on the OCI image configuration to build the OCI runtime configuration as described by this document; this will create the "default generated runtime configuration".
+
+The "default generated runtime configuration" MAY be overridden or combined with externally provided inputs from the caller.
+In addition, a converter MAY have its own implementation-defined defaults and extensions which MAY be combined with the "default generated runtime configuration".
+The restrictions in this document refer only to combining implementation-defined defaults with the "default generated runtime configuration".
+Externally provided inputs are considered to be a modification of the `application/vnd.oci.image.config.v1+json` used as a source, and such modifications have no restrictions.
+
+For example, externally provided inputs MAY cause an environment variable to be added, removed or changed.
+However, an implementation-default default MUST NOT result in an environment variable being removed or changed.
```
vbatts: sentence kind of stutters, but i think I understand it.

conversion.md
crosbymichael: uhh, hello?

vbatts: i took this as the "converter" is different from the runtime

conversion.md
crosbymichael: @vbatts you may understand it but it makes no sense, a user can override default values if they see fit.

stevvooe:
@cyphar Could you address? https://github.com/opencontainers/image-spec/pull/492/files#diff-8aafbe2c3690162540381b8cdb157112R40

conversion.md
crosbymichael: And what is the differences from a consumer/user of the spec and a "converter"?

vbatts:
there are a couple nits to get fixed, but LGTM

crosbymichael:
Its looking at a lot better but I'm wondering how you will identify what a converter is and what an end user is?

crosbymichael:
Is there going to be a package that does this mapping for us? 

```go

import "image-spec/converter"

runtime, err := converter.Generate(image)
```

cyphar:
> Is there going to be a package that does this mapping for us?

Yes, it currently exists inside `umoci` (`github.com/openSUSE/umoci/oci/config/convert`). Currently @vrothberg is cleaning it up so that it matches this document.

> Its looking at a lot better but I'm wondering how you will identify what a converter is and what an end user is?

From a testing / conformance perspective I imagine we would only want to test that any "converter defaults" don't affect image configuration conversion in the fields specified by an image. The main difference between the two is that when a user specifies something you can consider it as replacing the corresponding field in the original image, while converter defaults/extensions cannot override things set in an image.

vbatts:
@cyphar could you update the couple of nits? 😸

vbatts:
@cyphar bump. please :heart:

conversion.md
cyphar: Sorry for not commenting on this directly here (GitHub didn't show the comment box). I did address it in comments in the main thread. In addition the new wording mentions this precise example:

```
+For example, externally provided inputs MAY cause an environment variable to be added, removed or changed.
+However, an implementation-default default MUST NOT result in an environment variable being removed or changed.
```

cyphar:
@vbatts I was having trouble tracking what issues have not already been addressed. Here is my understanding of the current issues:

* `Config.User` (@stevvooe) appears to have been addressed. The concern was the difficulty in requiring converters to implement `getpwnam(3)` but my response was that implementing `user:group` support is only a recommendation (and the method of resolving them is also only a recommendation) and only numeric `uid:gid` forms MUST be implemented verbatim.

* The worries about `Config.Env` and other user-changable options (@crosbymichael + @stevvooe) should be resolved by the new wording which was sent after the last comments on those issues. There hasn't been much discussion in this thread recently so I don't know if anyone actually objects to @philips's proposal (and my extensions to it).

The one final point which we discussed in the call and also mentioned by @crosbymichael is

> [...] I'm wondering how you will identify what a converter is and what an end user is?

The distinction in this document is not between end users and converters. It's between implementation-defined defaults and externally provided inputs. In the basic case there is no distinction, but for the spec the difference is quite important.

An implementation-defined default is something that we can test by providing a bunch of images that contain various bits of metadata and our testing will only check the outputs that were actually specified in the image (the image is the source of truth). Any extra fields or whatever don't matter. But if the converter changes fields that we specified in the image, that's where non-compliance kicks in.

When it comes to externally defined input, the way I'm thinking of it is that when an externally provided input is specified it is applied to the source image as though the source image always had the value specified. For a more concrete example, imagine that `umoci` had some flags to modify the configuration. The following two examples would be identical.

```
% # unpack the image, with an externally provided value for Config.User
% umoci unpack --image theimage:latest --override=config.user="someone:else" bundle
```

```
% # modify the image and change Config.User
% umoci config --config.user="someone:else" --image theimage:latest
% umoci unpack --image theimage:latest bundle
```

But the important point is that a converter wouldn't be able to change `Config.User` without being told to by the user.

And please note that this entire document is describing the "default generated runtime configuration". The point is that if a converter decides to start overriding things it's okay as long as they specify that they aren't outputting the default generated runtime configuration (and maybe we should require that implementations provide a way to use the "default generated runtime configuration" -- I'm not sure).

conversion.md
```diff
@@ -0,0 +1,119 @@
...
+| Image Field           | Runtime Field      | Notes |
+| --------------------- | ------------------ | ----- |
+| `Config.ExposedPorts` | `annotations`      | 1     |
+| `Config.Volumes`      | `mounts`           | 2     |
+
+1. The runtime configuration does not have a corresponding field for this image field.
+   However, converters SHOULD set the [`org.opencontainers.image.exposedPorts` annotation](#config.exposedports).
+2. If a converter implements conversion for this field using mountpoints, it SHOULD set the `destination` of the mountpoint to the value specified in `Config.Volumes`.
+   The other `mounts` fields are platform and context dependent, and thus are implementation-defined.
+   Note that the implementation of `Config.Volumes` need not use mountpoints, as it is effectively a mask of the filesystem.
```
runcom: @cyphar could you help me fully understand point 2 for `Config.Volumes`; the first line reads:

> If a converter implements conversion for this field using mountpoints, it SHOULD set the `destination` of the mountpoint to the value specified in `Config.Volumes`.

what does that mean? are there any other way to implement conversion for this field other than mountpoints? if yes, which ones?
How's the expected flow to implement the field with mountpoints for runtimes? create a host dir for that volume and bind mount it in the container?

> Note that the implementation of `Config.Volumes` need not use mountpoints, as it is effectively a mask of the filesystem.

as a non-native speaker this seems in contrast with the first sentence. It now reads "don't use mountpoints!"

cyphar: > are there any other way to implement conversion for this field other than mountpoints?

Yes, you could just copy the contents of the volume into the container. Hopefully (see #496) the purpose of volumes is to restrict how implementations should handle diff layers (allow external data that is not snapshotted).

In the case of umoci I could envisage this all being done through manifests and black-holing certain directories so umoci will simply ignore them.

> How's the expected flow to implement the field with mountpoints for runtimes?

I only mention the destination. In particular what this is meant to mean is that the value of `Config.Volumes` should be a mountpoint (though there are reasons to not do that). The source of the mountpoint, the type of filesystem and so on can be whatever you want (it could be a `tmpfs` for example or NFS).

> as a non-native speaker this seems in contrast with the first sentence. It now reads "don't use mountpoints!"

The first sentence says you SHOULD, but the second sentence says you don't **have** to. To be fair, maybe I shouldn't repeat that point but I felt worried people would make the same assumption you did.

runcom: > Yes, you could just copy the contents of the volume into the container.

alright, I guess what's still not clear to me is the definition of a volume which this doc is missing. What's a volume then? some directory/mountpoint on the host or somewhere else?
from that sentence, as you said, it can be anything.

stevvooe: Should this also stipulate that data from the image may be copied into the volume?

conversion.md
stevvooe: @cyphar This, as I read it, would prevent one from having an implementation-default that replaces security sensitive environment variables. For example, it might make sense to remove environment variables that configure user shell configuration from a remotely acquired image.

conversion.md
cyphar: No, because that's forbidden by the definition of `Config.Volume` (https://github.com/opencontainers/image-spec/blame/master/config.md#L152):

> If a file or folder exists within the image with the same path as a data volume, that file or folder will be replaced by the data volume and never be merged.

cyphar: "Mask" is the runtime-spec terminology for this. Basically it's a path that won't be recursed into.

stevvooe: This is behavior that docker allows today: effectively, the contents of a volume can be seeded with the contents of the image.

cyphar: Okay, but that's not what the spec currently allows (I don't agree with that design either, but let's discuss that somewhere else). Would you mind if we handle that in a separate PR (it's got nothing to do with the config generation and more to do with extraction surely).

vbatts:
discussed on a call between @stevvooe and I just now. @stevvooe to make more commentary from his perspective.

conversion.md
stevvooe: @cyphar I'll file an issue and not hold up this PR further on this matter.

conversion.md
```diff
@@ -0,0 +1,119 @@
...
+
+A converter MUST rely on the OCI image configuration to build the OCI runtime configuration as described by this document; this will create the "default generated runtime configuration".
+
+The "default generated runtime configuration" MAY be overridden or combined with externally provided inputs from the caller.
+In addition, a converter MAY have its own implementation-defined defaults and extensions which MAY be combined with the "default generated runtime configuration".
+The restrictions in this document refer only to combining implementation-defined defaults with the "default generated runtime configuration".
+Externally provided inputs are considered to be a modification of the `application/vnd.oci.image.config.v1+json` used as a source, and such modifications have no restrictions.
+
+For example, externally provided inputs MAY cause an environment variable to be added, removed or changed.
+However an implementation-defined default MUST NOT result in an environment variable being removed or changed.
```
stevvooe: I think if we relax this to a `SHOULD NOT`, we can go forward here.

cyphar: Presumably you also want me to change this line too:

> 1. The converter MAY add additional entries to `process.env` but it MUST NOT add entries that have variable names present in `Config.Env`.

?

stevvooe: @cyphar Yep!

cyphar:
Alright, fixed @stevvooe @vbatts. PTAL.

stevvooe:
LGTM

cyphar:
Sorry, I just squashed it. Can you LGTM again?

stevvooe:
LGTM

vbatts:
LGTM

I'll open PRs for the two left over touch-ups (https://github.com/opencontainers/image-spec/pull/492/files#r104840624 & https://github.com/opencontainers/image-spec/pull/492/files#r104840585)

Dfurb87:
V

PR #495, opened 2016-12-14
Remove Memory MemorySwap CpuShares from config

coolljt0725:
As I said in https://github.com/opencontainers/image-spec/pull/371#issuecomment-252411562,
I'd like to remove these resource limit from image-spec.

On the on hand, if we have this, we should have limit
the min and max size of these value. For example, in
docker, there are some min and max limit for some
resource, see https://github.com/docker/docker/blob/master/daemon/daemon_unix.go#L48

ping @vbatts  @stevvooe 

### DIFF
--- a/config.md
+++ b/config.md
@@ -18,7 +18,7 @@ Using a layer-based or union filesystem such as AUFS, or by computing the diff f
 
 ### Image JSON
 
-Each image has an associated JSON structure which describes some basic information about the image such as date created, author, as well as execution/runtime configuration like its entrypoint, default arguments, CPU/memory shares, networking, and volumes.
+Each image has an associated JSON structure which describes some basic information about the image such as date created, author, as well as execution/runtime configuration like its entrypoint, default arguments, networking, and volumes.
 The JSON structure also references a cryptographic hash of each layer used by the image, and provides history information for those layers.
 This JSON is considered to be immutable, because changing it would change the computed [ImageID](#imageid).
 Changing it means creating a new derived image, instead of changing the existing image.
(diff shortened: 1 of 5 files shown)


cyphar:
We can't remove them, it'll break compability with Docker.

stevvooe:
The interesting code is in https://github.com/docker/docker/blob/master/daemon/commit.go#L198 and https://github.com/docker/docker/blob/master/daemon/commit.go#L27. Pay attention to which fields are actually merged.

It would be helpful if you could do the analysis for which fields actually end up on disk. As far as I can tell, it is defined by `BuildFromConfig` and that merge function.

stevvooe:
Adding to the above, from my inspection, these fields don't land on disk.

Also, note that there are interesting lines, like https://github.com/docker/docker/blob/master/daemon/commit.go#L70, which don't seem correct.

philips:
@stevvooe As the resident expert on keeping Docker compatibility what is your recommendation?

coolljt0725:
@stevvooe  These fields does land on disk on prev v1.6 docker,  see https://github.com/docker/docker/blob/v1.6.0/runconfig/merge.go#L14 and  https://github.com/docker/docker/blob/v1.6.0/runconfig/config.go#L15 , https://github.com/docker/docker/blob/v1.6.0/image/image.go#L31  but after https://github.com/docker/docker/pull/10298 which landed in v1.7 docker, all cgroup field are move to 
`hostConfig`, `hostConfig` are not commit to image.

stevvooe:
LGTM

After doing some further research, I found these fields are from before the Config/HostConfig split. In practice, these fields are really host-specific.

Their appearance in 1.6 will be a non-factor for modern Docker deployments. If we convert them into the post-1.9 formats, they will be ignored and not written to disk.

cyphar:
I'm :+1: if they aren't actually used (they always seemed quite an odd addition).

philips:
LGTM. Thank you @stevvooe !!!

jonboulle:
yay

cyphar:
/me will update #492 to remove these.

PR #500, opened 2016-12-14
README: Link to specs-go and schema and declare Go-compat policy

wking:
The links help with discoverability, otherwise folks reading the README might not notice that we provided these resources in addition to the spec itself.

The Go-compat policy formalizes the rough sketch [here][1].  By declaring a Go-compat policy, folks who have Go troubles can tell without testing whether the image-spec tooling *should* work for their Go environment.  And if/when it does not, they can see whether image-spec is interested in patches or not.  For example, if the tooling breaks on Go 1.5, we don't care or want some awkward workaround.  But if it breaks on Go 1.6 we do care and want a patch.   And if someone wants to genericize our test suite to run on both 1.6 and 1.7, we want a patch for that too.

And I'm just guessing at 1.6.  Maybe the repo-policy is to support 1.5+?  Something else?

[1]: https://github.com/opencontainers/image-spec/pull/487#issuecomment-265876398

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,9 @@ The OCI Image Format project creates and maintains the software shipping contain
 
 The specification can be found [here](spec.md).
 
+This repository also provides [Go types](specs-go), [intra-blob validation tooling, and JSON Schema](schema).
+The Go types and validation should be compatible with the current Go release; earlier Go releases are not supported.
+
 Additional documentation about how this group operates:
 
 - [Code of Conduct](https://github.com/opencontainers/tob/blob/d2f9d68c1332870e40693fe077d311e0742bc73d/code-of-conduct.md)


wking:
Travis failure is [a Docker timeout while fetching the pandoc image][1].  Kicking Travis should fix it.

[1]: https://travis-ci.org/opencontainers/image-spec/builds/184090973#L333

stevvooe:
This policy should be the most recent version of Go, for now. We already have a lot of versioning to deal with and having to worry about Go laggards will put too much burden on the project.

wking:
On Fri, Dec 16, 2016 at 11:38:15AM -0800, Stephen Day wrote:
> This policy should be the most recent version of Go, for now.

Updated with 41b47f6 → 6191aff.

I personally hope the “for now” here goes away quickly (in a
subsequent PR).  Dropping existing consumers as soon as a new version
of Go is released doesn't give me that “we care about backwards
compat” feeling ;).  And we don't have all that much Go code, so it
seems unlikely that the spec-provided Go *has* to live on the bleeding
edge.

stevvooe:
@wking Go isn't really like other languages. There is really no excuse to not upgrade. We can do build flags, but, this case, it was just test features.

wking:
On Tue, Dec 20, 2016 at 12:15:53PM -0800, Stephen Day wrote:
> Go isn't really like other languages. There is really no excuse to
> not upgrade.

This seems unlikely to me ;), see the caveats in [1].  But I'm trying
to punt whether we *should* have backwards compat to a later issue;
this PR just documents the *current* policy [2].

[1]: https://golang.org/doc/go1compat#expectations
[2]: https://github.com/opencontainers/image-spec/pull/500#issuecomment-267722130

jonboulle:
There are absolutely valid reasons not to immediately upgrade Go when a new version is out. Pretty sure Docker has been in this situation before :-). In any case, I'm "okay" with this change - it's vague enough to leave it as caveat lector.

jonboulle:
LGTM

stevvooe:
@jonboulle 

> Pretty sure Docker has been in this situation before

Yes, I believe there was some problems moving to Go 1.5 or 1.6. But we _never_ forced this requirement on our dependencies because we vendored.

I would say most recent minus one would be acceptable, as well.

philips:
LGTM

PR #501, opened 2016-12-15
Descriptor annotations

vbatts:
this has come up in discussion and is needed for adding discoverable arbitrary data to the descriptor references. Especially in making an index for image layout.

In DRY'ing up the references to annotations. This makes the defined keys in a very logical place rather than the bottom of one of the other docs.

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -1,0 +1,23 @@
+# Annotations
+Several components of the specification, like [Image Manifests](manifest.md) and [Descriptors](descriptor.md), feature an optional annotations property, whose format is common and defined in this section.
+
+This property contains arbitrary metadata.
+
+## Rules
+
+Annotations MUST be a key-value map where both the key and value MUST be strings.
+While the value MUST be present, it MAY be an empty string.
+Keys MUST be unique within this map, and best practice is to namespace the keys.
+Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications and extensions.
+If there are no annotations then this property MUST either be absent or be an empty map.
+Consumers MUST NOT generate an error if they encounter an unknown annotation key.
+
+## Pre-Defined Annotation Keys
+
+This specification defines the following annotation keys, intended for but not limited to manifest list and image manifest authors:
+* **org.opencontainers.created** date on which the image was built (string, date-time as defined by [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6)).
+* **org.opencontainers.authors** contact details of the people or organization responsible for the image (freeform string)
+* **org.opencontainers.homepage** URL to find more information on the image (string, a URL with scheme HTTP or HTTPS)
+* **org.opencontainers.documentation** URL to get documentation on the image (string, a URL with scheme HTTP or HTTPS)
+* **org.opencontainers.source** URL to get source code for the binary files in the image (string, a URL with scheme HTTP or HTTPS)
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,7 @@ DOC_FILES := \
 	manifest-list.md \
 	layer.md \
 	config.md \
+	annotations.md \
 	canonicalization.md
 
 FIGURE_FILES := \
--- a/descriptor.md
+++ b/descriptor.md
@@ -40,6 +40,11 @@ The following fields contain the primary properties that constitute a Descriptor
   Each entry MUST conform to [RFC 3986][rfc3986].
   Entries SHOULD use the `http` and `https` schemes, as defined in [RFC 7230][rfc7230-s2.7].
 
+- **`annotations`** *string-string map*
+
+    This OPTIONAL property contains arbitrary metadata for this descriptor.
+    This OPTIONAL property MUST use the [annotation rules](annotations.md#rules).
+
 ### Reserved
 
 The following field keys are reserved and MUST NOT be used by other specifications.
--- a/descriptor.md
+++ b/descriptor.md
@@ -118,9 +118,11 @@ The following example describes a [_Manifest_](manifest.md#image-manifest) with 
   "size": 7682,
   "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270"
 }
+```
 
 In the following example, the descriptor indicates that the referenced manifest is retrievable from a particular URL:
 
+```json,title=Content%20Descriptor&mediatype=application/vnd.oci.descriptor.v1%2Bjson
 {
   "mediaType": "application/vnd.oci.image.manifest.v1+json",
   "size": 7682,
--- a/manifest-list.md
+++ b/manifest-list.md
@@ -67,15 +67,9 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 - **`annotations`** *string-string map*
 
     This OPTIONAL property contains arbitrary metadata for the manifest list.
-    Annotations MUST be a key-value map where both the key and value MUST be strings.
-    While the value MUST be present, it MAY be an empty string.
-    Keys MUST be unique within this map, and best practice is to namespace the keys.
-    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
-    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications.
-    If there are no annotations then this property MUST either be absent or be an empty map.
-    Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
+    This OPTIONAL property MUST use the [annotation rules](annotations.md#rules).
 
-    See [Pre-Defined Annotation Keys](manifest.md#pre-defined-annotation-keys).
+    See [Pre-Defined Annotation Keys](annotations.md#pre-defined-annotation-keys).
 
 ### Extensibility
 Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown property.
--- a/manifest.md
+++ b/manifest.md
@@ -62,15 +62,9 @@ Unlike the [Manifest List](manifest-list.md), which contains information about a
 - **`annotations`** *string-string map*
 
     This OPTIONAL property contains arbitrary metadata for the image manifest.
-    Annotations MUST be a key-value map where both the key and value MUST be strings.
-    While the value MUST be present, it MAY be an empty string.
-    Keys MUST be unique within this map, and best practice is to namespace the keys.
-    Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
-    Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications.
-    If there are no annotations then this property MUST either be absent or be an empty map.
-    Implementations that are reading/processing the image manifest MUST NOT generate an error if they encounter an unknown annotation key.
+    This OPTIONAL property MUST use the [annotation rules](annotations.md#rules).
 
-    See [Pre-Defined Annotation Keys](#pre-defined-annotation-keys).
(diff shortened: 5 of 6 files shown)


annotations.md
```diff
@@ -0,0 +1,25 @@
...
+Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications and extensions.
+If there are no annotations then this property MUST either be absent or be an empty map.
+Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
+
+See [Pre-Defined Annotation Keys](manifest.md#pre-defined-annotation-keys).
+
+## Pre-Defined Annotation Keys
+
+This specification defines the following annotation keys, which MAY be used by manifest list and image manifest authors:
```
wking: Is there a reason to restrict the scope of the pre-defined keys to manifests and manifest lists?  For example, `org.opencontainers.source` seems like it would be a good match for a descriptor referencing layer tar archive.  If that's too big a shift for this PR, I'm fine spinning it off into a follow-up issue/PR.

wking:
Oh, you'll also want `Makefile` and spec.md-ToC updates to pull in the new file.  See [here][1] for a pre-`spec.md` example.

[1]: https://github.com/opencontainers/image-spec/pull/340/files

annotations.md
```diff
@@ -0,0 +1,25 @@
+# Annotations
+This OPTIONAL property MUST use the [annotation rules](annotations.md#rules).
```
jonboulle: imho something like "Several components of the specification, like Image Manifests and Descriptors, feature an optional _annotations_ property, whose format is common and defined in this section"

jonboulle:
one suggestion, looks good overall!

vbatts:
updated. PTAL

annotations.md
```diff
@@ -0,0 +1,25 @@
...
+
+Annotations MUST be a key-value map where both the key and value MUST be strings.
+While the value MUST be present, it MAY be an empty string.
+Keys MUST be unique within this map, and best practice is to namespace the keys.
+Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications and extensions.
+If there are no annotations then this property MUST either be absent or be an empty map.
+Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
+
+See [Pre-Defined Annotation Keys](manifest.md#pre-defined-annotation-keys).
```
wking: You removed this `manifest.md` section, and the `annotations.md` section if right after this line.  So I think we should drop this line.

annotations.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Pre-Defined Annotation Keys
+
+This specification defines the following annotation keys, intended for but not limited to manifest list and image manifest authors:
+* **org.opencontainers.created** date on which the image was built (string, date-time as defined by [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6)).
+* **org.opencontainers.authors** contact details of the people or organization responsible for the image (freeform string)
+* **org.opencontainers.homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
```
wking: I think we should either drop "must be a" here or add it to `org.opencontainers.created`.  I'm in favor of dropping it.

philips:
@vbatts please fix the one nit and then we can LGTM this.

annotations.md
```diff
@@ -0,0 +1,25 @@
...
+If there are no annotations then this property MUST either be absent or be an empty map.
+Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
+
+See [Pre-Defined Annotation Keys](manifest.md#pre-defined-annotation-keys).
+
+## Pre-Defined Annotation Keys
+
+This specification defines the following annotation keys, intended for but not limited to manifest list and image manifest authors:
+* **org.opencontainers.created** date on which the image was built (string, date-time as defined by [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6)).
+* **org.opencontainers.authors** contact details of the people or organization responsible for the image (freeform string)
```
stevvooe: How would multiple authors be specified?

annotations.md
```diff
@@ -0,0 +1,25 @@
...
+* **org.opencontainers.homepage** URL to find more information on the image (string, must be a URL with scheme HTTP or HTTPS)
+* **org.opencontainers.documentation** URL to get documentation on the image (string, must be a URL with scheme HTTP or HTTPS)
+* **org.opencontainers.source** URL to get source code for the binary files in the image (string, must be a URL with scheme HTTP or HTTPS)
```
stevvooe: Do we require a licenses field?

cyphar: Yeah, as per the discussion in #506, we should add an `org.opencontainers.licenses` field. Because licensing is complicated, we could allow a free-form string but the string SHOULD be made up of SPDX identifiers like GPL-2.0 or MIT.

gregkh: No, that's not going to work at all, you can't properly describe the license for a whole layer in any form of a string that could be standardized or parsed. SPDX is great for describing the individual licenses of things, but not for a collection of things, which almost always has an arbitrary license (example, what's the license, in a simple string, for a Ubuntu base layer?)

gregkh:
As much as I feel "this way lies madness", I understand the need/wish to have arbitrary annotations to help describe things, so it is good to describe a way to have this.

LGTM.

caniszczyk:
Can we also target this towards RC4?

annotations.md
wking: > what's the license, in a simple string, for a Ubuntu base layer?

This is still a pain, but it's becoming almost bearable as [Debian's copyright-file spec][1] gains ground.  Walking the package list (e.g. [yakkety-20170104][2]), looking up the associated copyright file (e.g. [util-linux_2.28.2-1ubuntu1][3]) and collecting the license information will result in something like this quasi-[SPDX license expression][5]:

> Artistic AND BSD-2-Clause AND BSD-3-Clause AND BSD-3-Clause-with-weird-numbering AND BSD-4-Clause AND BSD-4-Clause-POWERDOG AND BZIP AND CC0 AND CC0-1.0 AND DONT-CHANGE-THE-GPL AND Expat AND ( Expat OR GPL-1+ OR Artistic ) AND GAP AND GAP~FSF AND GFDL-1.3+ AND GPL AND ( GPL-1+ OR Artistic ) AND GPL-2 AND GPL-2+ AND ( GPL-2+ OR Artistic ) AND ( GPL-2+ OR LGPL-3+ ) AND ( GPL-2+ WITH libtool exception ) AND GPL-2.0+ AND GPL-2.1+ AND GPL-3 AND GPL-3+ AND ( GPL-3+ OR BSD-3-Clause ) AND ( GPL-3+ WITH the GCC Runtime Library Exception 3.1 ) AND ( GPL-3+-WITH-BISON-EXCEPTION ) AND HPND AND ( HSIEH-DERIVATIVE OR HSIEH-BSD OR LGPL-2.1 ) AND INNER-NETv2.0 AND ISC AND ISOC-rfc WITH disclaimer AND LGPL AND LGPL-2 AND LGPL-2+ AND ( LGPL-2+ OR BSD-2-Clause OR MIT ) AND LGPL-2.0+ AND LGPL-2.1 AND LGPL-2.1+ AND LGPL-3+ AND ( LGPL-3+ OR GPL-2+ ) AND LGPL-3.1 AND MACHv3.0-modified AND MIT AND PCRE AND PD AND REGCOMP AND RFC-Reference AND S2P AND SDBM-PUBLIC-DOMAIN AND SMAIL AND Spencer-94 AND TEXT-TABS AND TinySCHEME AND Unicode AND X11 AND ZLIB AND attribution AND bzip2-1.0.6 AND permissive AND public-domain AND public-domain-md5 AND public-domain-s-s-d AND verbatim AND ( verbatim OR GFDL-1.3+ ) AND ( verbatim WITH some-TeX-stuff ) AND more because many packages don't use the copyright-file spec yet…

Besides being incomplete, the license tagging isn't particularly consistent (and [SPDX 2.6 doesn't name][4] everthing we need).  But that can get fixed on a per-project basis for parse-ability and a per-license basis for registration (and/or we could follow Debian and give a standard way to inline any unregistered licenses).

> You can't properly describe the license for a whole layer…

It would be a lot easier to handle this if we had per-file granularity for the license metadata.  But with the current OCI image spec, layers are the closest we get to per-file.  I don't think bumping the licensing information up into *higher levels* is going to make it any easier to set the appropriate data when you publish an image (although it may make it slightly easier to discover the data when you read the image, since you'd only have to look in one place).

[1]: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
[2]: https://partner-images.canonical.com/core/yakkety/20170104/ubuntu-yakkety-core-cloudimg-amd64.manifest
[3]: http://changelogs.ubuntu.com/changelogs/pool/main/u/util-linux/util-linux_2.28.2-1ubuntu1/copyright
[4]: https://spdx.org/licenses/
[5]: https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60

stevvooe: > It would be a lot easier to handle this if we had per-file granularity for the license metadata.

This is wildly outside of scope for the OCI image format.

What is wrong with having something where you take a layer digest and get back the licensing information about that digest? Data normalization is a good thing and applying it here is a no brainer.

Trying to make the format do everything is just going to make it bad at everything.

wking: > What is wrong with having something where you take a layer digest and get back the licensing information about that digest?

I'm not arguing for changing the model here.  Per-descriptor metadata for licensing seems like the best way to handle licensing with the current OCI image modeling.  I don't see a need to [push this metadata out of band][1], and I don't see a benefit to [documenting licensing in the config][2] instead of in the digest linking to the licensed content.

A per-layer out of band licensing registry might be a useful way to share data among image producers (as a build-time optimization), but I think having licensing metadata in the Merkle DAG is a better experience for image consumers.

[1]: https://github.com/opencontainers/image-spec/pull/506#issuecomment-268085581
[2]: https://github.com/opencontainers/image-spec/pull/506#issuecomment-273641172

stevvooe: @wking 

> I think having licensing metadata in the Merkle DAG is a better experience for image consumers.

I don't see how this can possibly work without making a format that has an entry for every file.

Either way, let's not let this point block the PR. We can get annotations in, then have another PR to discuss standard annotation license key.

jonboulle:
lgtm

jonboulle:
@gregkh bot didn't pick up your approval properly

gregkh:
lgtm

There, that better?

gregkh:
lgtm

jonboulle:
Oh, oops. I forgot you aren't actually a maintainer of the spec...

gregkh:
ah, that makes sense why it wasn't working for me :)

vbatts:
nits fixed and rebased. PTAL

stevvooe:
LGTM

annotations.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Rules
+
+Annotations MUST be a key-value map where both the key and value MUST be strings.
+While the value MUST be present, it MAY be an empty string.
+Keys MUST be unique within this map, and best practice is to namespace the keys.
+Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
+Keys using the `org.opencontainers` namespace are reserved and MUST NOT be used by other specifications and extensions.
+If there are no annotations then this property MUST either be absent or be an empty map.
+Implementations that are reading/processing manifest lists MUST NOT generate an error if they encounter an unknown annotation key.
```
wking: Is this restriction specific to manifest lists?  Maybe it should read:

> Consumers MUST NOT generate an error if they encounter an unknown annotation key.

or something similarly generic.

jonboulle: +1

vbatts: done

jonboulle:
lgtm

stevvooe:
LGTM

PR #504, opened 2016-12-18
config: forbid snapshotting of data volumes

### LINKED ISSUE #496: layer: make clear that diffs should not cross mountpoints
From #492, I noticed that there was a concern about whether `Config.Volume`s should be included in layer changesets. IMO the answer is clearly no, but we have to make a bunch of other decisions:

1. Should layer changesets ever cross a mountpoint? My gut says no, but we need to make this explicit (and what mountpoint we should start at). There's also a platform-dependence issue.

2. While #492 is defining the conversion of `Volume` -> `mounts` it's not a **hard** requirement, which means that either we have to i) talk about `Volumes` in `layer.md` b) make the requirement hard.

WDYT? Especially @vbatts who has had strong opinions about changesets in the past. :wink:

cyphar:
This is necessary in order to make sure that unpackers all have sane
behaviour when it comes to handling image repacking and layer
generation. Since data volumes are generally bindmounts from sources
external to the image, it is not a good idea to snapshot said data --
and thus we should forbid it.

Closes #496

### DIFF
--- a/config.md
+++ b/config.md
@@ -149,7 +149,8 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
    - **Volumes** *object*, OPTIONAL
 
      A set of directories which SHOULD be created as data volumes in a container running this image.
-     If a file or folder exists within the image with the same path as a data volume, that file or folder will be replaced by the data volume and never be merged.
+     If a file or folder exists within the image with the same path as a data volume, that file or folder MUST be replaced by the data volume and never be merged.
+     When a new layer is created from a container that has data volumes, any data within the data volume SHOULD NOT be included in the changeset layer.
      **NOTE:** This JSON structure value is unusual because it is a direct JSON serialization of the Go type `map[string]struct{}` and is represented in JSON as an object mapping its keys to an empty object.
 
    - **WorkingDir** *string*, OPTIONAL


stevvooe:
Do we need language clarifying that "copying up" from the layer before into the volume is allowed?

cyphar:
So you want me to change the previously line, because currently copy-up is forbidden:

> If a file or folder exists within the image with the same path as a data volume, that file or folder is replaced with the data volume and is never merged.

stevvooe:
@cyphar In Docker today, you can create an image with a volume such that it will "copy up" _before_ running. If the new copy is modified, it won't be included in the image.

jonboulle:
this raises the point that "data volumes" are not defined in image-spec (they presumably rely on the [Docker concept](https://docs.docker.com/engine/tutorials/dockervolumes/#/data-volumes) but don't even link there). Can we add some more verbiage here describing what it means?

cyphar:
@stevvooe Alright, I'll update the wording to include a note saying that `copyup` is permitted by a runtime (though I don't like that we talk about runtimes here...), just that diff generation is not permitted. The thing I don't like about this is that it's a very runtime-specific thing, and honestly I don't like how Docker has two different volume modes (`--volumes-from` and `-v`).

@jonboulle Yeah, I'll add some words about it, but I won't reference the Docker docs because they specifically refer to the Docker implementation. Something like:

> Data volumes are directories within the rootfs of an image which are intended to store external (possibly persistent) data in the rootfs of a container based on the image. Any changes to a data volume will not be included in further layers based off the image.

wking:
On Wed, Dec 21, 2016 at 05:10:11AM -0800, Aleksa Sarai wrote:
> > Data volumes are directories within the rootfs of an image which
> > are intended to store external (possibly persistent) data in the
> > rootfs of a container based on the image. Any changes to a data
> > volume will not be included in further layers based off the image.

I still prefer focusing on mount points [1], but if we want to go down
the “data volumes” path we probably need:

* To link this sentence's “Data volumes” with the config's ‘Volumes’
  if we ever grow a volume generator
  (e.g. opencontainers/image-tools#8).

* To replace “further layers based off the image” with “the layer”.
  The data-volume masking also applies to the initial layer, so
  “further” doesn't always apply.  And “the image” is a pretty fuzzy
  term.  The important context is “the updated rootfs, the previous
  rootfs snapshot (if any), and the set of data-volume directories”,
  but that's a bit of a mouthful ;).

[1]: https://github.com/opencontainers/image-spec/issues/496#issuecomment-267006130

cyphar:
@wking 

> I still prefer focusing on mount points [1], but if we want to go down
the “data volumes” path we probably need:

> * To link this sentence's “Data volumes” with the config's ‘Volumes’
>  if we ever grow a volume generator
>  (e.g. opencontainers/image-tools#8).

This is already done, `Config.Volume` is defined as "data volumes.

> * To replace “further layers based off the image” with “the layer”.
>  The data-volume masking also applies to the initial layer, so
>  “further” doesn't always apply.  And “the image” is a pretty fuzzy
>  term.  The important context is “the updated rootfs, the previous
>  rootfs snapshot (if any), and the set of data-volume directories”,
>  but that's a bit of a mouthful ;).

I will use the term "layer changeset based on the image". The term "image" is not fuzzy -- you are creating a layer based on an image. As for the context, I'm not sure what you mean by your comment.

wking:
On Wed, Dec 21, 2016 at 12:32:42PM -0800, Aleksa Sarai wrote:
> > * To link this sentence's “Data volumes” with the config's
> >  ‘Volumes’ if we ever grow a volume generator
> >  (e.g. opencontainers/image-tools#8).
> 
> This is already done, `Config.Volume` is defined as "data volumes.

I expect layer implementors to read layer.md, but:

  $ git grep -i 'data\|volume' origin/pr/504 -- layer.md
  …no hits…

Since this PR is restricting layer generation, I think it has to land
wording about the new restriction in [1].  And it would be nice if the
config and layer references cross-linked each other to make
discovering the relationship easier.

> > * To replace “further layers based off the image” with “the layer”.
> >  The data-volume masking also applies to the initial layer, so
> >  “further” doesn't always apply.  And “the image” is a pretty fuzzy
> >  term.  The important context is “the updated rootfs, the previous
> >  rootfs snapshot (if any), and the set of data-volume directories”,
> >  but that's a bit of a mouthful ;).
>
> I will use the term "layer changeset based on the image". The term
> "image" is not fuzzy -- you are creating a layer based on an
> image.

Always?  Why can't you just create a layer in a vacuum?

The current API in flight with opencontainers/image-tools#8 takes
arguments for the child directory and parent directory.  There's no
reference to anything more image-y.  If this PR lands, that API will
need a new argument.  What will that new argument look like?
‘--config PATH_TO_IMAGE_CONFIG_JSON’?  ‘--data-volumes DIR[,DIR...]’?
‘--image WHATEVER_AN_IMAGE_IS’?

> As for the context, I'm not sure what you mean by your
> comment.

Just “these are the parameters for the layer-generation algorithm”.
I'm not clear on what “an image” is, but I'm pretty sure it has more
information than you need for layer generation.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc3/layer.md#determining-changes

cyphar:
> Since this PR is restricting layer generation, I think it has to land wording about the new restriction in [1].

Maintainers asked me to put it in `config.md` but I can include it there too. Problem is that it has a bunch of consequences for cross-linking.

> Always?  Why can't you just create a layer in a vacuum?

You can, but then there's two cases:

1. You have a configuration, but there's no layers. In which case you just follow the configuration. This is like doing `umoci init` then `umoci new`.

2. You have no configuration, and you're just creating a tar archive of a directory. In which case, there's no `Config.Volume` for you to worry about.

So okay, not "always" but the case where you don't have an image configuration is effectively just creating a tar archive...

wking:
On Wed, Dec 21, 2016 at 12:49:41PM -0800, Aleksa Sarai wrote:
> > Since this PR is restricting layer generation, I think it has to
> > land wording about the new restriction in [1].
>
> Maintainers asked me to put it in `config.md` but I can include it
> there too. Problem is that it has a bunch of consequences for
> cross-linking.

Which is why I prefer focusing on bind mounts ;).  But if we want to
link the config to the layer-generation spec via a “data volumes”
idea, I think we need the cross-linking to make the relationship very
obvious.

> 1. You have a configuration, but there's no layers. In which case
>    you just follow the configuration. This is like doing `umoci
>    init` then `umoci new`.

Does that touch layers at all?  If not, how does it relate to this PR?

> 2. You have no configuration, and you're just creating a tar archive
>    of a directory. In which case, there's no `Config.Volume` for you
>    to worry about.
 
And:

3. You have a set of data volume paths (which you got from somewhere)
   and a rootfs and you want to make the initial layer.  This is like
   tar --exclude …, possibly modulo mount-point stubs.

4. You have a set of data volume paths (which you got from somewhere),
   a previous snapshot, and a rootfs and want to make the initial
   layer.  This touches both [1] and the data-volume exclusion you're
   proposing here.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc3/layer.md#determining-changes

cyphar:
> Which is why I prefer focusing on bind mounts ;).

You could implement volumes without bindmounts or mounts at all, just by excluding changes to the relevant directory. Should you? Probably not, but defining that a volume **must** be a mount is a bit too strong, and is also restrictive on implementors.

Not to mention that `umoci` (as an example) won't use bindmounts because they require either running containers or having root. So there's an example of an existing implementation where bindmount wording won't work.

> And:

Both of your examples (3 and 4) are effectively saying "I have an image configuration somewhere that I'm generating a changeset for, but I'm not going to give you the configuration for some reason". I don't understand what the usecase is that you're fighting for this distinction between an image configuration and the process for generating layers. You need to have `DiffIDs` in order to extract the layers (well it's actually `Layers` in the manifest), you need to have `Volume` in order to know which directories are to be ignored.

How about I write a PR with some actual language and then we can discuss that, as opposed to commenting on a single phrase in a paragraph that I wrote off the top of my head (as an example of kind-of what I will write) before going to bed? I mean, ultimately I will probably choose wording that doesn't have "the base image" purely because I would then have to define that as well. But please let me actually write something before you start getting pedantic about wording that I wrote as an example... :weary:

wking:
On Wed, Dec 21, 2016 at 01:37:44PM -0800, Aleksa Sarai wrote:
> How about I write a PR with some actual language and then we can
> discuss that…

+1

philips:
@cyphar What is going on with this?

RobDolinMS:
@stevvooe Has concerns about this potentially conflicting with existing use cases and implementations.

vbatts:
more from the call yesterday, i feel like this is like trying to pin down the exact flags to use with the `find` command (i.e which files to match on; whether to stay on the device or traverse mounts; etc). If there is verbiage added, it should be more instructive like "it's recommended to not include child content from volume paths in an image layer". Which is still a little confusing, and should be clarified that it does not mean that content can still be _copied_ from the volume to facilitate the building of the layer.

stevvooe:
@vbatts @cyphar How do we move this one forward?

After some thought, I think there needs to be clear recommendation about what layers mean, but not necessarily how they are created. Generically, one should mask off volume declarations in the image diff creation process, but that seems like it could be purely up to the user.

Perhaps, this PR would work if it was a SHOULD.

vbatts:
i think i'm inclined to say this should be a SHOULD as well. because as a general rule this is the case, though sure-enough there would be some feature that asks to override and include a default fileset in where the volume would be, so if the volume is not present then the app "does the right thing" or some non-sense, and the MUST would make it non-compliant.

cyphar:
Alright, I'm going to bump this one with SHOULD so we can get it into `1.0.0` (this is quite important IMO even though it's only informative -- because of how it plays with #492).

cyphar:
Updated, PTAL @stevvooe @vbatts. I also switched a `will` to a `MUST` but I can drop that if you want.

stevvooe:
@cyphar I still believe this language to be too strong. It is imposing requirements on build process that doesn't really exist. OCI doesn't really create any requirements for how one image is related to another.

The way this is written, it also opens up the possibility for a Volume to be a file, which hasn't really ever existed.

cyphar:
> OCI doesn't really create any requirements for how one image is related to another.

True, but we need to encode some semantic meaning in Config.Volumes so that it actually does mask diffs. We could make it only affect extraction but that won't actually help solve the whole "volume containing a secret getting snapshotted" problem which this current PR solves. In particular I want to make it so that the whole idea that "`docker commit` doesn't cross volume boundaries" is actually enforced by implementations.

I won't lie, I actually missed this issue when implementing `umoci` (and I have yet to fix it because I want this language to be specified first).

> it also opens up the possibility for a Volume to be a file, which hasn't really ever existed.

Personally I don't see that as a negative. :wink: But we could deny that if you really want. Ultimately "data volume" is so ill-specified in this spec I really don't know what an implementation should reasonably do in the absence of prior implementations to compare against.

vbatts:
@cyphar this language prohibits actual use-cases i'm aware of. Not for copying/commiting FROM a volume, but rather copying TO a volume
https://github.com/moby/moby/issues/17470#issuecomment-152167622

```shell
[vbatts@bananaboat] /tmp/tmp.GhZqUJ$ docker build -t f .
Sending build context to Docker daemon 3.584 kB
Step 1 : FROM fedora
 ---> 15895ef0b3b2
Step 2 : RUN mkdir /data && touch /data/README
 ---> Running in 3c2c12dd5864
 ---> 5b1376684e24
Removing intermediate container 3c2c12dd5864
Step 3 : VOLUME /data
 ---> Running in 4c5f67eeee33
 ---> 9a9ec38af080
Removing intermediate container 4c5f67eeee33
Successfully built 9a9ec38af080
[vbatts@bananaboat] /tmp/tmp.GhZqUJ$ docker run -v /data/ --rm -it f sh
sh-4.3# ls /data/
README
```
and this makes a new `docker volume ls` visible volume.

I realize this is confusing language to define in the spec, but your proposed language prohibits this existing functionality. :-\
I'm inclined to close this and we can revisit for a future version.

cyphar:
> this language prohibits actual use-cases i'm aware of

The old language does too. If we want to change it we should, but in any case we still need to fix up this wording anyway because the old text is problematic **already**.

vbatts:
fair

cyphar:
Closing in favour of #694.

Dfurb87 (commented):
- - [ ][]()

PR #509, opened 2016-12-26
compatibility_test.go:change manifest to manifestlist

q384566678:

### DIFF
--- a/schema/manifest_backwards_compatibility_test.go
+++ b/schema/manifest_backwards_compatibility_test.go
@@ -45,13 +45,13 @@ func convertFormats(input string) string {
 
 func TestBackwardsCompatibilityManifestList(t *testing.T) {
 	for i, tt := range []struct {
-		manifest string
+		manifestlist string
 		digest   string
 		fail     bool
 	}{
 		{
 			digest: "sha256:219f4b61132fe9d09b0ec5c15517be2ca712e4744b0e0cc3be71295b35b2a467",
-			manifest: `{
+			manifestlist: `{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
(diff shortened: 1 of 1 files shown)


q384566678:
@stevvooe @vbatts @jbouzane @jonboulle PTAL

schema/manifest_backwards_compatibility_test.go
```diff
@@ -110,14 +110,14 @@ func TestBackwardsCompatibilityManifestList(t *testing.T) {
 			fail: false,
 		},
 	} {
-		sum := sha256.Sum256([]byte(tt.manifest))
+		sum := sha256.Sum256([]byte(tt.manifestlist))
```
stevvooe: While we're in here, let's start using the digest package to perform these kinds of operations.

q384566678: The digest open is manifeslist, so I think there is nothing wrong with the use of manifestlist here.

vbatts:
looks good. please rebase

RobDolinMS:
@stevvooe Are you OK with @q384566678's response?

schema/manifest_backwards_compatibility_test.go
stevvooe: That is not what I said. I meant that if you're in the process of cleaning up, it would be better to use the go-digest package.

I've filed https://github.com/opencontainers/image-spec/issues/513 to take care this, so don't worry about it.

stevvooe:
@RobDolinMS No but I guess we can merge anyways. I've filed https://github.com/opencontainers/image-spec/issues/513 to handle the concern.

q384566678:
updated. PTAL

stevvooe:
LGTM

philips:
LGTM

PR #510, opened 2016-12-29
Modify the test's config type

q384566678:
When I execute the test command, I get the following warning, so I made the following changes.
`$ go test`
`warning: config sha256:5359a4f250650c20227055957e353e8f8a74152f35fe36f00b6b1f9fc19c8861 has an unknown media type: application/octet-stream`

### DIFF
--- a/schema/manifest_backwards_compatibility_test.go
+++ b/schema/manifest_backwards_compatibility_test.go
@@ -50,13 +50,13 @@ func TestBackwardsCompatibilityManifestList(t *testing.T) {
 		fail     bool
 	}{
 		{
-			digest: "sha256:e588eb8123f2031a41f2e60bc27f30a4388e181e07410aff392f7dc96b585969",
+			digest: "sha256:219f4b61132fe9d09b0ec5c15517be2ca712e4744b0e0cc3be71295b35b2a467",
 			manifest: `{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": [
       {
-         "mediaType": "application/vnd.docker.distribution.manifest.v1+json",
+         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
          "size": 2094,
          "digest": "sha256:7820f9a86d4ad15a2c4f0c0e5479298df2aa7c2f6871288e2ef8546f3e7b6783",
          "platform": {
@@ -65,7 +65,7 @@ func TestBackwardsCompatibilityManifestList(t *testing.T) {
          }
       },
       {
-         "mediaType": "application/vnd.docker.distribution.manifest.v1+json",
+         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
          "size": 1922,
          "digest": "sha256:ae1b0e06e8ade3a11267564a26e750585ba2259c0ecab59ab165ad1af41d1bdd",
          "platform": {
@@ -77,7 +77,7 @@ func TestBackwardsCompatibilityManifestList(t *testing.T) {
          }
       },
       {
-         "mediaType": "application/vnd.docker.distribution.manifest.v1+json",
+         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
          "size": 2084,
          "digest": "sha256:e4c0df75810b953d6717b8f8f28298d73870e8aa2a0d5e77b8391f16fdfbbbe2",
          "platform": {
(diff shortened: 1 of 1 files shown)


schema/manifest_backwards_compatibility_test.go
```diff
@@ -136,14 +136,14 @@ func TestBackwardsCompatibilityManifest(t *testing.T) {
 		//
 		// curl -L -H "Authorization: Bearer ..." -H \
 		// "Accept: application/vnd.docker.distribution.manifest.v2+json" \
-		// https://registry-1.docker.io/v2/library/docker/manifests/sha256:888206c77cd2811ec47e752ba291e5b7734e3ef137dfd222daadaca39a9f17bc
+		// https://registry-1.docker.io/v2/library/docker/manifests/sha256:d901d5244c94db8151415bf3a1d3a0b10e7468e4997786b01c287cbcbaeca446
```
wking: Have you tested this?  I expect the `curl` call will not succeed.  I think the right solution to this problem it to make these child-media-type checks optional, with something like a `strict` option.  More background in opencontainers/image-tools#66, #403, and #452.

q384566678: I have tested the changes and the results are correct：

`go test`
`PASS`
`ok  	_/home/zhouhao/image-spec/schema	0.033s`

wking: On Tue, Jan 03, 2017 at 05:00:57PM -0800, Zhou Hao wrote:
> `go test`
> `PASS`

I agree that the *test pass* with your 36073936, but the *point* of these tests is to show that the current OCI spec is compatible with existing Docker-hub blob instances with minimal media type replacements.  If you change the Docker-hub blobs locally, you break that.  Using the token fetching comment from #370:

    $ TOKEN=$(curl https://auth.docker.io/token\?service\=registry.docker.io\&scope\=repository:library/docker:pull  | jq -r .token)
    $ curl -sILH "Authorization: Bearer ${TOKEN}" -H "Accept: application/vnd.docker.distribution.manifest.v2+json" https://registry-1.docker.io/v2/library/docker/manifests/sha256:888206c77cd2811ec47e752ba291e5b7734e3ef137dfd222daadaca39a9f17bc | head -n1
    HTTP/1.1 200 OK
    $ curl -sILH "Authorization: Bearer ${TOKEN}" -H "Accept: application/vnd.docker.distribution.manifest.v2+json" https://registry-1.docker.io/v2/library/docker/manifests/sha256:d901d5244c94db8151415bf3a1d3a0b10e7468e4997786b01c287cbcbaeca446 | head -n1
    HTTP/1.1 404 Not Found

The tests in this backwards-compat file need to be blobs you can successfully pull from the Docker hub, not ones that 404.

q384566678: Why do I enter the results of both cases are  `HTTP/1.1 200 Connection established`

wking: The "Connection established" line is probably [from a proxy][1].  You'll see the 404 if you drop the `head` calls.

[1]: http://stackoverflow.com/questions/16965530/what-to-do-with-extra-http-header-from-proxy

q384566678: However, when the modified mediatype, we must then modify the digest. This will resolve the warning

wking: > However, when the modified mediatype, we must then modify the digest. This will resolve the warning

To satisfy the spirit of these compat tests, you cannot modify the source content (or its digest) at all.  The point is that these examples are caught-in-the-wild blobs, and that after flushing that wild-caught content through `convertFormats` you get a valid OCI blob.  The compatibility claim will be less convincing if it becomes “we edit the wild blob by hand and then flush it through `convertFormats` so we're compatible” ;).

And you *do* get a valid OCI blob (because we only [SHOULD `application/vnd.oci.image.config.v1+json` for a manifest's `config` descriptor][1]).  The “has an unknown media type” [error you get][2] is because the current tests are stricter than the spec and provide no way to [select between “MUST all compliant handlers support this blob?” and “is this blob valid?”][3].

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc3/manifest.md#L40
[2]: https://github.com/opencontainers/image-spec/pull/510#issue-197963184
[3]: https://github.com/opencontainers/image-spec/pull/510#discussion_r94497050

q384566678: thanks,updated.

philips:
LGTM

vbatts:
LGTM

PR #511, opened 2017-01-06
schma: add make fmt

q384566678:
I saw `make fmt` command in the runtime - spec , I think this is a very good method, should increase the feature in the iamge - spec. 

### DIFF
--- a/schema/Makefile
+++ b/schema/Makefile
@@ -1,0 +1,10 @@
+
+default: help
+
+help:
+	@echo "Usage: make <target>"
+	@echo
+	@echo " * 'fmt' - format the json with indentation"
+
+fmt:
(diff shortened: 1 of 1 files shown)


schema/Makefile
```diff
@@ -0,0 +1,10 @@
+
+default: help
+
+help:
+	@echo "Usage: make <target>"
+	@echo
+	@echo " * 'fmt' - format the json with indentation"
+
+fmt:
+	for i in *.json ; do jq --indent 4 -M . "$${i}" > xx && cat xx > "$${i}" && rm xx ; done
```
wking: There would be no `*.json` churn if you dropped the `--indent 4` and stuck with jq's default two-space indent.

q384566678: updated, thanks.

stevvooe:
Let's not make a separate Makefile for this.

PR #514, opened 2017-01-19
specs-go/v1: use digest type on descriptor

stevvooe:

### DIFF
--- a/specs-go/v1/descriptor.go
+++ b/specs-go/v1/descriptor.go
@@ -14,13 +14,15 @@
 
 package v1
 
+import digest "github.com/opencontainers/go-digest"
+
 // Descriptor describes the disposition of targeted content.
 type Descriptor struct {
 	// MediaType contains the MIME type of the referenced object.
 	MediaType string `json:"mediaType"`
 
 	// Digest is the digest of the targeted content.
-	Digest string `json:"digest"`
+	Digest digest.Digest `json:"digest"`
 
 	// Size specifies the size in bytes of the blob.
 	Size int64 `json:"size"`


specs-go/v1/descriptor.go
```diff
@@ -14,13 +14,15 @@
 
 package v1
 
+import digest "github.com/opencontainers/go-digest"
```
xiekeyang: `import "github.com/opencontainers/go-digest"` is OK?

stevvooe: Go tool does this automatically.

wking:
The Travis errors are all git-validation looking at ancient commits
[1].  This looks a lot like [2].  And while I'm not sure if
opencontainers/runtime-spec#216 will fix this particular issue, the
TRAVIS_COMMIT_RANGE echo it adds will help debug things like this.

[1]: https://travis-ci.org/opencontainers/image-spec/builds/193543329#L303
[2]: https://github.com/opencontainers/runtime-spec/pull/660#issuecomment-273562233

stevvooe:
@caniszczyk @vbatts @philips Any idea what is going on here? What is this second builder "push" doing?

stevvooe:

caniszczyk:
why is the git-validation DCO check failing? cc: @vbatts

caniszczyk:
also the DCO check is completed via PullApprove now so things are somewhat redundant imho

stevvooe:
@caniszczyk I have no clue. `continuous-integration/travis-ci/push` doesn't seem to be showing up on other PRs.

wking:
On Mon, Jan 23, 2017 at 03:27:15PM -0800, Chris Aniszczyk wrote:
> also the DCO check is completed via PullApprove now so things are
> somewhat redundant imho

Some of the failures are for trailing whitespace (e.g. [1]), so I
think we want to keep the git-validation checks (they are also
something folks can run locally).  The problem is that Travis isn't
running git-validation on the right commit set, although I'm not sure
why not.  Echoing TRAVIS_COMMIT_RANGE (or just running ‘env’?) would
help figure out why the commit set is wrong [2].

@stevvooe, the /push tests are for the branch itself, and the /pr
tests are for the branch as merged into master [3,4].  Maybe we should
be using the epoch commit [5] for the push tests?

[1]: https://travis-ci.org/opencontainers/image-spec/builds/193543329#L304
[2]: https://github.com/opencontainers/image-spec/pull/514#issuecomment-273980810
[3]: http://stackoverflow.com/a/41534988
[4]: https://docs.travis-ci.com/user/pull-requests/#How-Pull-Requests-are-Tested
[5]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc3/Makefile#L42

wking:
On Mon, Jan 23, 2017 at 03:37:54PM -0800, W. Trevor King wrote:
> Maybe we should be using the epoch commit [5] for the push tests?

Potential fix along these lines filed as #521, although there's no
/push test there.  I suspect the /push test here is because you're
using a branch inside this repo (vs. filing the PR from a branch in
your personal fork).

stevvooe:
> I suspect the /push test here is because you're
using a branch inside this repo (vs. filing the PR from a branch in
your personal fork).

This may be it. My `stevvooe` remote was pointing at main repo. Re-opening this PR against that.

PR #515, opened 2017-01-19
config: Simplify the paragraph contrasting diffIDs with layer digests

wking:
@stevvooe and I [couldn't agree on more detailed wording][1], and this paragraph was never normative anyway.

Spun off from #388 starting [here][2].  The new wording does not mention media types at all, so there's no reason to tangle it up with the rest of #388.

CC @stevvooe.

[1]: https://github.com/opencontainers/image-spec/pull/388#discussion_r96765401
[2]: https://github.com/opencontainers/image-spec/pull/388/files#r96978832

PR #518, opened 2017-01-22
layer is compressed tar format

xiekeyang:

### DIFF
--- a/config.md
+++ b/config.md
@@ -12,7 +12,7 @@ This specification uses the following terms:
 ### [Layer](layer.md)
 
 Image filesystems are composed of *layers*.
-Each layer represents a set of filesystem changes in a tar-based [layer format](layer.md), recording files to be added, changed, or deleted relative to its parent layer.
+Each layer represents a set of filesystem changes in a compressed tar-based [layer format](layer.md), recording files to be added, changed, or deleted relative to its parent layer.
 Layers do not have configuration metadata such as environment variables or default arguments - these are properties of the image as a whole rather than any particular layer.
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -12,7 +12,7 @@ This specification uses the following terms:
 ### [Layer](layer.md)
 
 Image filesystems are composed of *layers*.
-Each layer represents a set of filesystem changes in a tar-based [layer format](layer.md), recording files to be added, changed, or deleted relative to its parent layer.
+Each layer represents a set of filesystem changes in a compressed tar-based [layer format](layer.md), recording files to be added, changed, or deleted relative to its parent layer.
```
wking: This is currently covered [here][1], and will no longer be true if/when #388 lands.  In any event, compression seems like a property of the later media type, and not a fundamental property of “being a layer”.

[1]: https://github.com/opencontainers/image-spec/blame/d1c7054fae33dd186c78fb126d5c613b233aa5eb/layer.md#L11

xiekeyang:
@wking You are right, this should be closed.

xiekeyang:
@wking I have a question.
> compression seems like a property of the later media type, and not a fundamental property of “being a layer”.

Then, if this comment are a little conflict to https://github.com/opencontainers/image-spec/pull/515, as `Do not confuse DiffIDs with layer digests, often referenced in the manifest, which are digests over compressed content.`?

If above words in #515 should be modified?

wking:
> If above words in #515 should be modified?

Those words are fine now.  I'm [restoring][1] @stevvooe's "uncompressed" wording in #388.

[1]: https://github.com/opencontainers/image-spec/pull/515#discussion_r96984598

PR #519, opened 2017-01-22
Add imagelayout validator and ImageLayoutFile const

coolljt0725:
We had added image layout schema, but didn't add image layout validator.
Add imagelayout validator and ImageLayoutFile const, so image tool and other
tools can use image -spec to validate the layout of the oci image.

### DIFF
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -26,6 +26,7 @@ const (
 	MediaTypeManifest     Validator     = v1.MediaTypeImageManifest
 	MediaTypeManifestList Validator     = v1.MediaTypeImageManifestList
 	MediaTypeImageConfig  Validator     = v1.MediaTypeImageConfig
+	TypeImageLayout       Validator     = v1.ImageLayoutFile
 	MediaTypeImageLayer   unimplemented = v1.MediaTypeImageLayer
 )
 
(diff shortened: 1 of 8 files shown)


philips:
LGTM

schema/schema.go
```diff
@@ -26,6 +26,7 @@ const (
 	MediaTypeManifest     Validator     = v1.MediaTypeImageManifest
 	MediaTypeManifestList Validator     = v1.MediaTypeImageManifestList
 	MediaTypeImageConfig  Validator     = v1.MediaTypeImageConfig
+	TypeImageLayout       Validator     = v1.ImageLayoutFile
```
stevvooe: O, this is weird. All of these should start with `Validator`.

wking: > All of these should start with Validator.

Or we could drop them and use a map: #403.

stevvooe: #403 is never going to land. This effectively uses a map internally anyways. See my example.

coolljt0725:
@stevvooe 
> O, this is weird. All of these should start with Validator.

Added a commit to address this.

coolljt0725:
ping @philips  @stevvooe

q384566678:
You have to modify other test files.

stevvooe:
LGTM

Please fix the build. ;)

coolljt0725:
@stevvooe  @philips  fixed the build :)

vbatts:
LGTM

stevvooe:
LGTM

PR #520, opened 2017-01-23
Fix jq usage in example

coolljt0725:

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -70,7 +70,7 @@ Those tags will often be represented in an image-layout repository with matching
 ### Example Ref
 
 ```
-$ cat ./refs/v1.0 | jq
+$ cat ./refs/v1.0 | jq .
 {
   "size": 4096,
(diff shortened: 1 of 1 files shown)


image-layout.md
```diff
@@ -70,7 +70,7 @@ Those tags will often be represented in an image-layout repository with matching
 ### Example Ref
 
 ```
-$ cat ./refs/v1.0 | jq
+$ cat ./refs/v1.0 | jq .
```
wking: At least with jq 1.5, the `.` filter seems to be the default, so the old versions worked fine.  I haven't checked to see if all recent versions of jq are like this or not.  If you need `jq .` but get an error on a bare `jq`, which version of jq are you running?

coolljt0725: @wking  I use jq 1.3.  
ok https://github.com/opencontainers/image-spec/blob/master/HACKING.md#json-schema-formatting  requires for jq 1.5, missing this part.

PR #521, opened 2017-01-23
Makefile: Use 'ifdef TRAVIS_COMMIT_RANGE' for git-validation

wking:
Only use the auto-ranging when Travis tells us what the range is.  Use our `EPOCH_TEST_COMMIT`-based range in all other cases.

This may fix the issue we've been having with #514.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ before_install:
 install: true
 
 script:
+  - env | grep TRAVIS_
   - make .gitvalidation
   - make lint
   - make check-license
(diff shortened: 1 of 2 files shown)


vbatts:
LGTM

.travis.yml
```diff
@@ -20,6 +20,7 @@ before_install:
 install: true
 
 script:
+  - env | grep TRAVIS_
```
jonboulle: what is this?

wking: > what is this?

A way to see what Travis-specific variables were set for a given build.

jonboulle:
lgtm

PR #480, opened 2016-12-02
Replace ISO-8601 references with RFC-3339

aecolley:
Fixes #389

PR #529, opened 2017-01-25
config: Require RFC 3339's more restricted version of ISO 8601

### LINKED ISSUE #389: Prefer RFC-3339 over ISO-8601
The image config's `created` property is defined as anything compatible with [ISO-8601](http://www.iso.org/iso/home/standards/iso8601.htm), but this should be restricted to date/time formats compatible with [section 5.6 of RFC-3339](https://tools.ietf.org/html/rfc3339#section-5.6).

The RFC-3339 format is compatible with, but stricter than, ISO-8601 format. It removes much ambiguity and undesired variability, such as 2-digit years or the time "24:00". It is also available freely on the Internet, unlike the ISO standard.

wking:
This brings us into compliance with JSON Schema, whose [current validation draft requires RFC 3339 for date-times][1].  I've modeled the reference phrasing on the draft JSON Schema's date-time definition.

Fixes #389.
Obsoletes #480 (which still hasn't grown a Signed-off-by).

[1]: https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-7.3.1

### DIFF
--- a/config.md
+++ b/config.md
@@ -51,7 +51,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
 
 - **created** *string*, OPTIONAL
 
-  An ISO-8601 formatted combined date and time at which the image was created.
+  An combined date and time at which the image was created, formatted as defined by [RFC 3339, section 5.6][rfc3339-s5.6].
 
 - **author** *string*, OPTIONAL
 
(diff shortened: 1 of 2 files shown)


stevvooe:
LGTM

vbatts:
LGTM

PR #530, opened 2017-01-26
.travis.yml: Rebuild schema-fs in the Travis tests

wking:
And error out if that fails.  This would have caught a bug where #444 passes but 111a99d (which merged #444) fails after a schema-fs rebuild.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,11 +16,13 @@ before_install:
   - go get -u github.com/alecthomas/gometalinter
   - gometalinter --install
   - go get -t -d ./...
+  - make schema-fs
 
 install: true
 
 script:
   - env | grep TRAVIS_
+  - git diff --exit-code
   - make .gitvalidation
   - make lint
   - make check-license
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,10 @@ endif
 
 .PHONY: install.tools
 
-install.tools: .install.gitvalidation .install.glide .install.glide-vc
+install.tools: .install.esc .install.gitvalidation .install.glide .install.glide-vc
+
+.install.esc:
+	go get -u github.com/mjibson/esc
 
 .install.gitvalidation:
 	go get -u github.com/vbatts/git-validation
--- a/schema/fs.go
+++ b/schema/fs.go
@@ -205,7 +205,7 @@ var _escData = map[string]*_escFile{
 	"/config-schema.json": {
 		local:   "config-schema.json",
 		size:    774,
-		modtime: 1485388791,
+		modtime: 0,
 		compressed: `
 H4sIAAAJbogA/5SRvW7rMAyFdz+F4WS8ju7QKWsfoEPHooMqUzEDWFRJZgiKvHv1EzcxEBTuEsSH/M4R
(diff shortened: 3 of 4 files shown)


wking:
Ok, I think 2d54423c has everything we need.  The tests are [still failing][1], but that's catching the breakage from #411 which is addressed in #531.  Once #531 lands, I'll rebase this PR on top.

[1]: https://travis-ci.org/opencontainers/image-spec/builds/195426527#L415

jonboulle:
Ready for rebase

wking:
On Fri, Jan 27, 2017 at 02:59:20AM -0800, Jonathan Boulle wrote:
> Ready for rebase

Rebased with 2d54423 → 4112ba1.

wking:
Hmm, looks like there's some [unstable compression][1].  I wonder if we can turn that off…

[1]: https://travis-ci.org/opencontainers/image-spec/builds/195901232#L355

vbatts:
I just fixed this in #533 in a nicer way that doesn't require travis.

stevvooe:
> I just fixed this in #533 in a nicer way that doesn't require travis.

Want to pull this out into a separate PR or do we want to push #533 through?

stevvooe:
@wking @vbatts Also, closing this, since @vbatts has a solution.

wking:
On Fri, Jan 27, 2017 at 12:39:26PM -0800, Vincent Batts wrote:
> I just fixed this in #533 in a nicer way that doesn't require
> travis.

The point of this PR is to test (in Travis) whether fs.go is up to
date.  I don't see anything that sounds similar in #533.  Maybe you're
thinking about #531 (which caught the spec validation up with #411)?

vbatts:
@wking https://github.com/opencontainers/image-spec/pull/533/files#diff-b67911656ef5d18c4ae36cb6741b7965R79 sorry. I should've included a link.

By running `make validate-examples` whether local or on travis, it will ensure they're fresh.

wking:
On Fri, Jan 27, 2017 at 12:51:31PM -0800, Vincent Batts wrote:
> @wking
> https://github.com/opencontainers/image-spec/pull/533/files#diff-b67911656ef5d18c4ae36cb6741b7965R79
> sorry. I should've included a link.
> 
> By running `make validate-examples` whether local or on travis, it
> will ensure they're fresh.

But the copy in version control could be stale (as it was in the wake
of #411).  Presumably, fs.go is only in version control to save some
network access for folks using schema/ as a library.  Those folks will
probably not be running ‘make schema-fs’ on their vendored package, so
we need a plan to keep them working.  This PR (once we figured out the
unstable compression [1]) would have ensured fs.go was always current
in version control.

An alternative would be to take your approach from #533 (ignoring the
version-controlled fs.go in CI), but we'd have to remember to re-build
and commit schema-fs for each release.

A third option would be to drop http.FileSystem and hit the network
for the JSON Schema files when validating (or to work out an
alternative method to cache these locally).  More on this in #150,
which I still think needs to be re-opened in image-spec (although the
current position there is to allow both http.FileSystem and network
access, and with that we'd still need to keep fs.go up to date).

[1]: https://github.com/opencontainers/image-spec/pull/530#issuecomment-275688678

PR #531, opened 2017-01-26
schema: Drop self-describing media types from manifest(-list) tests

wking:
Catch up with #411 (and fixing test errors after rebuilding `schema/fs.go`).

Ideally we'd be testing “is `schema/fs.go` current?” in Travis, and I'm trying to handle that in #530, but it's tricky.  We can land these test fixes in the meantime, without blocking on more robust CI.

PR #533, opened 2017-01-26
image-layout: ./refs/ -> index.json

vbatts:
For media that pertains to an image, like AppStream metadata, may need to accompany a manifest-list, but is not a descriptor that points to a manifest (i.e. and XML file).

This enables non-manifest descriptors. 

From the 2017-01-25 call we agreed to keep the manifest list as the entry point for accessing objects. The index concept is only a more generic use of the manifest-list. https://github.com/opencontainers/image-spec/pull/438#issuecomment-275252735

This change uses the manifest-list as an index that allows implementations to work around needing to walk/traverse the `./refs/` directories.

Obsoletes #438


vbatts:

image-layout.md
```diff
@@ -156,4 +161,48 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+      },
+      "annotations": {
+        "org.opencontainers.ref.name": "v1.0"
+      }
+    }
+  ],
+  "elements": [
+    {
+      "mediaType": "application/xml",
+      "Asize": 7143,
```
stevvooe: `Asize`?

image-layout.md
```diff
@@ -156,4 +161,48 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+      "platform": {
+        "architecture": "amd64",
+        "os": "linux"
+      },
+      "annotations": {
+        "org.opencontainers.ref.name": "v1.0"
+      }
+    }
+  ],
+  "elements": [
```
stevvooe: `elements`?

stevvooe:
> needing to walk/traverse the ./refs

Doesn't this obsolete `./refs`?

Were we going to change the name of `manifest-list` or do you want to do that in a follow up?

image-layout.md
```diff
@@ -156,4 +161,48 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 [tar stream]
 ```
 
+## index.json
+
+This OPTIONAL file is for use indexing references and elements of the image-layout, where walking the `refs` and `blobs` is sub-optimal.
```
wking: The index file has no effect on whether you need to or can walk `blobs`.  Performance issues aside, it is a drop-in replacement for `refs`, except that [the previous ref names are gone][1].  So I think we should remove the `blobs` reference from this line.

[0]: https://github.com/opencontainers/image-spec/pull/438#issue-187078763
[1]: https://github.com/opencontainers/image-spec/pull/438#issuecomment-264035961

image-layout.md
```diff
@@ -156,4 +161,48 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+It is intended to be generated from the content in an image-layout after any changes.
+
+From the base properties of the [manifest-list](manifest-list.md), this object will us
+
+### Index Example
+
+```json,title=Manifest%20List&mediatype=application/vnd.oci.image.manifest.list.v1%2Bjson
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.list.v1+json",
```
wking: This line can go away (see #531 and #411).

manifest-list.md
```diff
@@ -11,7 +11,8 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 - **`schemaVersion`** *int*
 
   This REQUIRED property specifies the image manifest schema version.
-  For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker. The value of this field will not change. This field MAY be removed in a future version of the specification.
+  For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker.
+  The value of this field will not change. This field MAY be removed in a future version of the specification.
```
wking: If you're trying to break this into one sentence per line, you'll want to put another newline in between “change.” and “This”.  But you can probably make this PR without touching the `schemaVersion` definition at all.

manifest-list.md
```diff
@@ -108,5 +116,5 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 }
 ```
 
-[runtime-platform2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc2/config.md#platform
+[runtime-platform2]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config.md#platform
```
wking: If we bump this to rc3, we probably want to update the [config reference as well][1].  But we can probably leave the rc2 → rc3 bump out of this PR.

[1]: https://github.com/opencontainers/image-spec/blame/eee8749ab8dd46b26778f6d97a9d59e60aefe5d0/config.md#L232

schema/content-descriptor.json
```diff
@@ -23,6 +23,10 @@
         "type": "string",
         "format": "uri"
       }
+    },
+    "annotations": {
+      "id": "https://opencontainers.org/schema/image/descriptor/annotations",
+      "$ref": "defs-image.json#/definitions/annotations"
```
wking: This catches us up with #501 (and it needs an `fs.go` bump too).  With the spec change already hashed out in #501, this can probably get spun off into its own PR and landed faster than the index pivot as a whole.

schema/defs-image.json
```diff
@@ -11,14 +11,47 @@
       "type": "string",
       "pattern": "^[a-z0-9_+.-]+:[a-f0-9]+$"
     },
+    "elementDescriptor": {
```
wking: Do you need this type (which drops `urls`?).  I'd expect you could just use:

in the `elements` array (or whatever that ends up being called).

image-layout.md
vbatts: Oh hah. I was messing with the schema validation... Clearly it works :-/

image-layout.md
vbatts: Elements of the index. Elements of the image. 
"Channels" felt too much like providing _alternate_ paths or manifests, which is not the point. "descriptors" seems obvious but also a bit redundant.

image-layout.md
vbatts: Yeah... I was wrestling with the schema validation which is requiring it

schema/defs-image.json
vbatts: I tried that to no avail. Clearly there is some schema attention needed

manifest-list.md
vbatts: Missed this. I'll pick it up in a second

vbatts:
@stevvooe I reckon is does obsolete the `./refs`, but not intentionally.
Also I will get to the rename, but want Ed to get this up for review first

image-layout.md
vbatts: Fair.
For the name, that is why I made up the example
```json
      "annotations": {
        "org.opencontainers.ref.name": "v1.0"
      }   
```

wking: > For the name, that is why I made up the example [`org.opencontainers.ref.name`]

That works as a ref-name replacement, but if we actually want to use it (e.g. in image-tools), I think we want to define it [here][1].

[1]: https://github.com/opencontainers/image-spec/blob/72a840ec12140beaf29c043826a3d701d0820ee7/annotations.md#pre-defined-annotation-keys

image-layout.md
vbatts: fixed

image-layout.md
vbatts: yes. I didn't do that because it was not my intention to anihilate the `./refs/` directory. 
@stevvooe thoughts?

image-layout.md
vbatts: fixed

manifest-list.md
vbatts: i've updated the other references now.

schema/content-descriptor.json
vbatts: i've updated the fs.go now. But what needs to get spun off?

wking: Nothing *needs* to get spun off.  But adding `annotations` here is independent of the rest of this PR, so you could spin this whole hunk off into a separate PR if you wanted more, smaller PRs (vs. fewer, bigger PRs).

image-layout.md
wking: > I didn't do that because it was not my intention to anihilate the `./refs/` directory.

Define something like `org.opencontainers.ref.name` lets consumers use the index to access content by name (e.g. `oci-unpack --ref v1.0 foo/ bar/`) vs. having callers access it by annotation (e.g. `oci-unpack --annotation org.opencontainers.ref.name=v1.0 foo/ bar/`).  But being able to use the index to access content by name doesn't mean you have to drop `refs/` (which is another way to access content by name).

schema/content-descriptor.json
vbatts: oh right. It could be independent. This PR makes an example of it though. So I fixed what I saw.

image-layout.md
vbatts: updated.

vbatts:
@stevvooe I think doing the manifest-list name and media-type change can happen in a separate thread than this PR.

vbatts:
Alrighty, y'all. PTAL

image-layout.md
stevvooe: Ok, `extensions`?

(let's explore a few...)

schema/fs.go
```diff
@@ -202,129 +202,348 @@ func _escFSMustString(useLocal bool, name string) string {
...
+H4sIAAAJbogA/wrLzJ0fmueS/5SBSaMjRGHmQ/EFfl9eT6850djLIrZo9xKlhtQ9qr06a//Zz2RgYFBl
+YGAIUAADpcwUJSsFpYySkoJiK339/ILUvOT8vJLEzLzUomK9/KJ0/eLkjNTcRP3M3MT0VP3cxLzMtNTi
+Et2czOIS/cS8vPySxJLM/LxiJR0GBgYZBgYGUwYGBkYGBgZmBghgRsIMEV25F1lAskwM8RegCpgYEAAk
+JQXj/P///z9YZBTQCER05R5jgYTwVwlo0MszMDAooohIw0RMG8GRhhy7DNAIBKljZRiNNDoAeKQxQqJI
+EhpFqtCoYQVl61poXDGhaWZGyqWjcUUpgIcklO0IZZfB2BFduReg5R04rkDxJIsUV1QvhE0bn68CBAAA
+///GLb+iDAYAAA==
+`,
+	},
+
+	"/.defs-image.json.swp": {
```
wking: Local editor stuff?  Maybe we should `-include=.*\.json` instead of our [current `-ignore=.*go`][1]?

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc3/schema/gen.go#L21

image-layout.md
wking: My personal favorite is `descriptors`.  It's boring and redundant, but obviously correct and it is very clearly not implying any semantic meaning about the content.

manifest-list.md
```diff
@@ -11,7 +11,9 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 - **`schemaVersion`** *int*
 
   This REQUIRED property specifies the image manifest schema version.
-  For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker. The value of this field will not change. This field MAY be removed in a future version of the specification.
+  For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker.
+  The value of this field will not change
```
wking: nit: missing a trailing period.

manifest-list.md
```diff
@@ -64,6 +66,13 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
         This OPTIONAL property specifies an array of strings, each specifying a mandatory CPU feature (for example `sse4` or `aes`).
 
+- **`elements`** *array of objects*
+
+  This OPTIONAL property contains a list of non-manifest objects.
```
wking: In the “using the object as an index” use case, the entries here are *likely* to be manifests and manifest lists.  In fact, folks using a manifest-list to switch between platforms could have an index where multiple `elements` (or whatever we end up calling them) have the same `org.opencontainers.ref.name` but differ in other metadata.  That would shorten index → manifest-list → manifest to index → manifest.  So I'd recommend something more generic, like:

> This OPTIONAL property contains an array of descriptors.  No restrictions on the referenced content are implied.

image-layout.md
wking: > updated.

You're still referencing `blobs` in this line (which is what [started][1] the sub-thread on this line).

[1]: https://github.com/opencontainers/image-spec/pull/533/files#r98119434

image-layout.md
```diff
@@ -27,23 +29,26 @@ The image layout MUST also contain an `oci-layout` file:
...
-./oci-layout
-./refs
+$ find . -type f
+./index.json
 ./refs/v1.0
 ./refs/stable-release
+./oci-layout
+./blobs/sha256/3588d02542238316759cbf24502f4344ffcc8a60c803870022f335d1390c13b4
+./blobs/sha256/4b0bc1c4050b03c95ef2a8e36e25feac42fd31283e8c30b3ee5df6b043155d3c
+./blobs/sha256/7968321274dc6b6171697c33df7815310468e694ac5be0ec03ff053bb135e768
```
wking: Probably keep this alphabetized.

vbatts: fair, but it's the output of `find . -type f`

schema/fs.go
vbatts: oh noes

vbatts: the -include did not do it. had to add another -ignore. fixed now.

manifest-list.md
vbatts: fixed

image-layout.md
vbatts: updated the blobs reference too.

manifest-list.md
vbatts: there is still the `manifests` field. Why would manifest and manifest-lists be put into this `elements` array?

vbatts:
updated.

manifest-list.md
wking: > there is still the `manifests` field. Why would manifest and manifest-lists be put into this `elements` array?

Maybe I'm not understanding right.  I'd assumed that `elements` would be used when folks were looking for a single-file `refs/` alternative and that `manifests` would only be used for traditional manifest-list stuff (e.g. selecting the best manifest based on platform matching) during some hopefully-short deprecation phase.  In that case, the index would have an `elements` entry pointing at a manifest whenever you create a named ref for a manifest.  And eventually, folks would only use `elements` and we could drop `manifests`.

But it sounds like you are intending all descriptors pointing at `application/vnd.oci.image.manifest.v1+json` go into `manifests` and everything else would go into `elements`?  That works, I guess, although I'm not sure how you'd identify non-OCI types which count as manifests (e.g. a third-party `application/vnd.foo.manifest`).

schema/fs.go
wking: > the -include did not do it.

It worked for me (although I had to add a trailing `$` to my [earlier guess][1].  I've filed the change as #538.

[1]: https://github.com/opencontainers/image-spec/pull/533#discussion_r98296739

schema/defs-image.json
```diff
@@ -78,6 +77,17 @@
               }
             }
           }
+        },
+        "annotations": {
+          "id": "https://opencontainers.org/schema/image/annotations",
```
wking: This has the same `id` as the defintion below ([here][1] in master).  It would be more DRY to reference that here:

    "annotations": {
      "$ref": "#/definitions/annotations"
    }

[1]: https://github.com/opencontainers/image-spec/blob/00850eca2ab993e282a4921f8b7000b2fcbd26fa/schema/defs-image.json#L84

Makefile
```diff
@@ -76,12 +76,15 @@ $(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: $(DOC_FILES) $(FIGURE_FILES)
 	ls -sh $(shell readlink -f $@)
 endif
 
-validate-examples:
+validate-examples: schema/fs.go
```
wking: Travis [doesn't actually call `validate-examples`][1], and it's not a dependency of the `test` target either.  If we're trying to rebuild `schema/fs.go` before any tests, you probably need to have:

    test: schema/fs.go

as well.

[1]: https://github.com/opencontainers/image-spec/blob/00850eca2ab993e282a4921f8b7000b2fcbd26fa/.travis.yml

Makefile
```diff
@@ -76,12 +76,15 @@ $(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: $(DOC_FILES) $(FIGURE_FILES)
...
 	go test -run TestValidate ./schema
 
-schema-fs:
+schema/fs.go: $(wildcard schema/*.json) schema/gen.go
+	cd schema && printf "%s\n\n%s\n" "$$(cat ../.header)" "$$(go generate)" > fs.go
+
+.PHONY: schema-fs
```
wking: Quite a few `.PHONY` entries are collected at the bottom of the Makefile if you want to put this entry there.  On the other hand, there are still a few single-entry `.PHONY`s scattered throughout the Makefile, so there is precedent if you want to keep it here.

Makefile
```diff
@@ -76,12 +76,15 @@ $(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: $(DOC_FILES) $(FIGURE_FILES)
 	ls -sh $(shell readlink -f $@)
 endif
 
-validate-examples:
+validate-examples: schema/fs.go
 	go test -run TestValidate ./schema
 
-schema-fs:
+schema/fs.go: $(wildcard schema/*.json) schema/gen.go
```
wking: I'm not sure how well make will handle this in a fresh Git checkout.  Maybe Travis should touch `schema/gen.go` to ensure a newer-than-`fs.go` mtime?

image-layout.md
vbatts: understandable

manifest-list.md
vbatts: So, the `manifests` array is intended for `application/vnd.oci.image.manifest.v1+json` and `application/vnd.oci.image.manifest.list.v1+json`. Any distinguishing of non-OCI manifest lists ... I'm not real sure. This field is also what Docker uses presently for their manifest-list and therefore needs some backwards-compat consideration. 

This `elements` (or whatever) is intended for additional metadata like AppStream or other relevant repo-metadata (say XML or sqlite objects) that are facilitating this image-layout. 

I would think it much cleaner to just have a single field like `descriptors` where they all live. Then this desire to distinguish goes away.

image-layout.md
```diff
@@ -156,4 +161,43 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 [tar stream]
 ```
 
+## index.json
+
+This OPTIONAL file is for use indexing references and elements of the image-layout, where walking the `refs` is sub-optimal.
+It is intended to be generated from the content in an image-layout after any changes.
+
+From the base properties of the [manifest-list](manifest-list.md), this object will us
```
runcom: `will us`?

vbatts: @wking it _wills_ you to do ... things ...

/me fixes it

Makefile
vbatts: i didn't notice the phony's getting grouped. I'll update mine. I'm used to just scattering them so you can clearly read the phony next to the target it is operating on.

Makefile
vbatts: presumably, a fresh checkout wouldn't need a fresh schema. the schema-fs ought to be commited when the JSON schema changes.

Makefile
vbatts: then i'll add validate-examples as a travis target.

schema/defs-image.json
vbatts: yes

Makefile
wking: > the schema-fs ought to be commited when the JSON schema changes.

But that doesn't always happen (it didn't in #411).  Making sure it *does* always happen was what I was trying to do with #530.  I'd thought the approach you were aiming for here was to always rebuild in CI so it didn't matter if the committed copy was stale.

Makefile
wking: > then i'll add validate-examples as a travis target.

That will double those tests, since `make test`, which we already call, also runs the example tests.  Other `make test` tests depend on `fs.go` as well, so I think you can leave Travis alone and add the `fs.go` dependency to `test`.

image-layout.md
vbatts: Must have had some copy-pasta or accidental delete. I'm wondering where this thought went and what I intended that sentence to be..

Makefile
vbatts: right. If travis runs the validate-examples, then it would fail CI for their PR, then they update schema-fs, commit and re-push.

Makefile
vbatts: oh i see. ok.

vbatts: maybe `schema/fs.go` doesn't even need to be commited then ... only generated. It often causes a lot of senseless rebasing.

vbatts: https://github.com/opencontainers/image-spec/pull/543

vbatts:
Updated. PTAL

manifest-list.md
stevvooe: > I would think it much cleaner to just have a single field like descriptors where they all live. Then this desire to distinguish goes away.

I'm not sure we can make this work. Migrating these top-level types are a huge pain and I'm not sure we gain anything except naming. I'd prefer to call this `extensions`, as they seem like extensions.

philips:
Yes, this should remove refs directory. @vbatts

manifest-list.md
philips: I don't follow why these things can't just go into the manifests list? We already have a namespaced mediaType for all of the descriptors so adding an extension field doesn't really buy much AFAICS.

philips:
Overall LGTM, I am confused on why we need extensions when mediaType is already namespaced.

wking:
> I am confused on why we need extensions when mediaType is already namespaced.

My understanding is that having `manifests` and something else (like `extentions`) is to support legacy manifest-list consumers which ignore the descriptor `mediaType`s.  More discussion around that [on IRC yesterday][1].

[1]: http://ircbot.wl.linuxfoundation.org/eavesdrop/%23opencontainers/%23opencontainers.2017-01-31.log.html#t2017-01-31T18:54:13

image-layout.md
```diff
@@ -156,4 +161,44 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+```json,title=Manifest%20List&mediatype=application/vnd.oci.image.manifest.list.v1%2Bjson
+{
+  "schemaVersion": 2,
+  "manifests": [
+    {
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
+      "size": 7143,
+      "digest": "sha256:e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7fc331f",
+      "annotations": {
+        "org.opencontainers.ref.name": "v1.0"
```
runcom: is this mandatory? walking the reference dir and using the filename as the reference name worked but if this annotation isn't provided, what should implementos use as "the reference name"?

image-layout.md
```diff
@@ -27,23 +29,26 @@ The image layout MUST also contain an `oci-layout` file:
     - The `imageLayoutVersion` value will align with the OCI Image Specification version at the time changes to the layout are made, and will pin a given version until changes to the layout are required
 - It MAY include additional fields
 
+The image layout MAY also contain an `index.json` file:
+
+- It MUST be a JSON object
+- It MUST have the base properties of [manifest-list](manifest-list.md).
+- See [index.json](#index.json) section
```
runcom: the `.` (dot) in the link isn't working, the works w/o the dot

image-layout.md
wking: > walking the reference dir and using the filename as the reference name worked but if this annotation isn't provided, what should implementos use as "the reference name"?

Walking the reference directory [didn't work over bare HTTP][1], and we're likely [dropping `refs/`][2] anyway.  I still don't think this annotation needs to be required though.  If it's not set, the entry just doesn't have a “reference name”.  You may still be able to address it through other means (e.g. maybe it's the only entry with a given media type, or it has some other annotation that you can use to identify it).

[1]: https://github.com/opencontainers/image-spec/pull/438#issue-187078763
[2]: https://github.com/opencontainers/image-spec/pull/533#issuecomment-276618123

runcom: how would you refer to a given revision of an image w/o a ref name?

wking: > how would you refer to a given revision of an image w/o a ref name?

* By index in the array (e.g. `manifests[0]`), or
* By another annotation (e.g. `com.example.foo.bar="docs"`), or
* By media type (e.g. `application/vnd.example.docs+xml`).

Consumers which were based on referencing solely via `org.opencontainers.ref.name` would just ignore (and be unable to address) entries which didn't set that annotation.

runcom: @vbatts @stevvooe how is this working with docker or other means (skopeo) to load OCI images and give a reference (tag) to them?

wking: While we're figuring out the naming semantics, we probably want to also address whether the ref name MUST be unique or SHOULD be unique or MAY be duplicated.  And for all of that, within what scope (the descriptor array?   The referencing blob?).

manifest-list.md
```diff
@@ -36,7 +38,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
   - **`platform`** *object*
 
-    This REQUIRED property describes the platform which the image in the manifest runs on.
+    This OPTIONAL property describes the platform which the image in the manifest runs on.
```
qianzhangxa: Why do we make it as an optional property? Without this object, if a manifest list have multiple manifests, how can we know which platform (e.g., amd64, ppc64le) each manifest is for?

stevvooe: For those cases, it will be present. There is a lot of discussion above, but the idea is that there might be companion data that is _not_ platform-specific.

@vbatts We may want to bolster this with something saying "This property SHOULD be present if its target is platform-specific." Maybe, even a must?

philips:
LGTM

@vbatts can you document the IRC discussion so we remember why we did the extensions field? https://github.com/opencontainers/image-spec/pull/533#issuecomment-276681899

stevvooe:
> @vbatts can you document the IRC discussion so we remember why we did the extensions field? #533 (comment)

I thought we arrived at just having it all in a `manifests` field.

image-layout.md
vbatts: @runcom the `org.opencontainers.ref.name` will act as the pointers from the descriptors in the list.
This will not be _that_ different from the manifest.json in the current `docker save ...` layout.

image-layout.md
vbatts: thanks

manifest-list.md
vbatts: alrighty

vbatts:
@philips after more discussion, the current behavior of the `manifests` field is such that if an unknown `mediaType` is encountered, then ignore it. Which is good for future proofing and adding manifest-prime mediaType.
So, rather than adding a new field with messy semantics, we're going to just keep the `manifests` field of descriptors and adding some clarification on ignoring unknown types.

vbatts:
alrighty. rebased on master. removed the `elements` field. removed the `./refs` directory. clarified the behavior on an unknown `mediaType`.

PTAL

image-layout.md
```diff
@@ -3,47 +3,45 @@
...
+    - Directory MUST exist and MAY be empty
+    - See [blobs](#blobs) section
+- `oci-layout` file
+    - It MUST exist
+    - It MUST be a JSON object
+    - It MUST contain an `imageLayoutVersion` field
+    - See [oci-layout file](#oci-layout-file) section
+    - It MAY include additional fields
+- `index.json` file
+    - It MUST exist
```
wking: This doesn't match your “OPTIONAL index.json” commit message and PR summaries.  You'll probably want to update for consistency.

Personally, I have no problem with making `index.json` optional (it feels like a bigger requirement than an empty `refs` directory for folks who have no references).

image-layout.md
```diff
@@ -156,4 +127,79 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+{
+    "imageLayoutVersion": "1.0.0"
+}
+```
+
+## index.json file
+
+This REQUIRED file is the entry point for references and descriptors of the image-layout.
+The [manifest-list](manifest-list.md) is an multi-descriptor entry point.
+It is intended to be generated from the content in an image-layout after any changes.
```
wking: This line makes less sense to me now that we're dropping `refs/`.  Can we drop it?

And one line up, “an multi-descriptor” → “a multi-descriptor”.

image-layout.md
vbatts: I'll update the message and summaries. It is a simple enough file to be there and it is logically consistent to have it be required if the ./refs/ directory was required.

image-layout.md
vbatts: fixed

vbatts:
updated. PTAL

stevvooe:
LGTM

image-layout.md
```diff
@@ -156,4 +127,78 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
...
+
+## index.json file
+
+This REQUIRED file is the entry point for references and descriptors of the image-layout.
+The [manifest-list](manifest-list.md) is a multi-descriptor entry point.
+
+This index provides an established path (`/index.json`) to have an entry point for an image-layout and to discover auxiliary descriptors.
+
+No semantic restriction is given for the "org.opencontainers.ref.name" annotation of descriptors.
+In general the `mediaType` of each [descriptor][descriptors] object in the `manifests` field will be either `application/vnd.oci.image.manifest.list.v1+json` or `application/vnd.oci.image.manifest.v1+json`
```
stevvooe: Can we remove the part about `application/vnd.oci.image.manifest.list.v1+json`? I don't see a use case for it and it just makes things complicated and unbounded.

In fact, if we ban, we can guarantee constant time processing for manifest resolution.

philips:
LGTM

image-layout.md
vbatts: It needs to be there, because an image might have a ref/tag that points to a manifest-list, say for the "v1.0.0" or "stable" and has multiple platforms.

vbatts:
Cool. I'm going to merge this!

image-layout.md
stevvooe: Should we encourage merging and flattening them in the index?

vbatts: interesting thought, thought that would mean possibly having duplicate ref.name's and the only difference is the `platform` field. That sound a tad messy, no?

stevvooe: @vbatts Just a thought. There are some implications to that. I think this is fine.

vbatts: @stevvooe right on. Thankfully, leaving like it is currently will not exclude folks that want to flatten in the way you described.

PR #543, opened 2017-01-30
schema: make fs.go dynamic

vbatts:
the schema-fs is needed dynamically and should be generated as such.
Further, it is often a tidious cause of rebasing.
Therefore, just make it generated everytime and not even managed in git.


wking:
On Mon, Jan 30, 2017 at 09:47:59AM -0800, Vincent Batts wrote:
> the schema-fs is needed dynamically and should be generated as such.
> Further, it is often a tidious cause of rebasing.  Therefore, just
> make it generated everytime and not even managed in git.
> …
>     M schema/fs.go (278)

As of d8a9fea6, you still have it in Git.  I agree that there are not
bumping fs.go with each PR would be nice because it would avoid those
rebases.  However, I think consumers of this library will need a
current fs.go when they vendor us.  In the “rarely-bump fs.go”
approach I sketched out in [1], we'd keep fs.go in version control but
only bump it as part of cutting a release.  Maybe that's what you're
aiming for here (in which case, you probably want to drop your local
fs.go bump ;).

[1]: https://github.com/opencontainers/image-spec/pull/530#issuecomment-275779569

vbatts:
i forgot about the imports and go get. this is less than ideal.

wking:
On Tue, Jan 31, 2017 at 10:28:08AM -0800, Vincent Batts wrote:
> i forgot about the imports and go get. this is less than ideal.

Does that mean we're coming back around towards the approach I floated
in #530?  If so, I can get back to work on figuring how to work around
the compression issue [1].

[1]: https://github.com/opencontainers/image-spec/pull/530#issuecomment-275688678

PR #544, opened 2017-01-30
config: add StopSignal

### LINKED ISSUE #541: Stop signal
Who is going to define the stop signal for a container? Is this an image thing or a runtime thing?

runcom:
Close #541 

@stevvooe @vbatts PTAL I probably forgot something

### DIFF
--- a/config.md
+++ b/config.md
@@ -124,6 +124,10 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
      If there are no labels then this property MAY either be absent or an empty map.
      Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown labels key.
 
+  - **StopSignal** *string*, OPTIONAL
+
+    The field contains the system call signal that will be sent to the container to exit. The signal can be a signal name in the format `SIGNAME`, for instance `SIGKILL` or `SIGRTMIN+3`.
+
 - **rootfs** *object*, REQUIRED
 
    The rootfs key references the layer content addresses used by the image.
--- a/schema/config-schema.json
+++ b/schema/config-schema.json
@@ -80,6 +80,9 @@
               "type": "null"
             }
           ]
+        },
+        "StopSignal": {
+          "type": "string"
         }
       }
     },
--- a/schema/config_test.go
+++ b/schema/config_test.go
@@ -164,6 +164,7 @@ func TestConfig(t *testing.T) {
             "/var/job-result-data": {},
             "/var/log/my-app-logs": {}
         },
+        "StopSignal": "SIGKILL",
         "WorkingDir": "/home/alice",
         "Labels": {
             "com.example.project.git.url": "https://example.com/project.git",
--- a/specs-go/v1/config.go
+++ b/specs-go/v1/config.go
@@ -41,6 +41,9 @@ type ImageConfig struct {
 
 	// Labels contains arbitrary metadata for the container.
 	Labels map[string]string `json:"labels,omitempty"`
+
+	// StopSignal contains the system call signal that will be sent to the container to exit.
+	StopSignal string `json:"StopSignal,omitempty"`
 }
 
 // RootFS describes a layer content addresses


stevvooe:
LGTM

I am not sure if we should add `StopTimeout`, as well. `StopSignal` is fairly portable, but the timeout can be deployment specific.

wking:
On Mon, Jan 30, 2017 at 12:26:08PM -0800, Antonio Murdaca wrote:
>     M config.md (4)
>     M specs-go/v1/config.go (3)

This also needs a JSON Schema change and fs.go rebuild, although that
can happen as a follow-up PR if folks prefer.

I'd rather address this with an annotation [1], because then we don't
have to add property-specific documentation for how this should be
translated into the runtime config (in flight as #492).  But
documenting translation for another property isn't that hard, so it's
not a big deal either way.

[1]: https://github.com/opencontainers/image-spec/issues/541#issuecomment-275955519

runcom:
Updated schema-fs as well

jonboulle:
> I am not sure if we should add StopTimeout, as well. StopSignal is fairly portable, but the timeout can be deployment specific.

what exactly are the criteria here? this seems a bit loose; I thought we were going from a versioned docker config type?

philips:
@runcom can you please list the valid strings in this field? And ideally write a test case?

runcom:
@philips sure

stevvooe:
@jonboulle Portability is just a measure of whether config is made with respect to the internals of the image or the host environment. Specifying the stop signal for a process is going to operate regardless of host machine (on the same os), whereas stop timeout may be very specific to how a machine is configured or what it is running. From a docker perspective, stop timeout exists in the image config, but there could be an argument that this is host specific, hence not included here automatically. Granted, I am favoring caution here, as it is easy to add but impossible to remove.

It would be good for someone else to lodge an informed opinion here.

runcom:
@philips updated

config.md
```diff
@@ -124,6 +124,10 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
      If there are no labels then this property MAY either be absent or an empty map.
      Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown labels key.
 
+  - **StopSignal** *string*, OPTIONAL
+
+    The field contains the system call signal that will be sent to the container to exit. The signal can be a valid unsigned number that matches a position in the kernel’s syscall table, for instance `9`, or a signal name in the format `SIGNAME`, for instance `SIGKILL`.
```
wking: Can we require the named version (e.g. `SIGKILL`) and not allow numbers (e.g. `9`)?  A fair number of names are [portable across POSIX systems][1] (and [a handful are required by ISO C][1]), but POSIX has very little to say about the semantics for a particular signal number (e.g. it does not require `SIGKILL` to be 9).  And the Linux man pages [has][2]:

> Because the range of available real-time signals varies according to the glibc threading implementation… programs should *never refer to real-time signals using hard-coded numbers*, but instead should always refer to real-time signals using the notation `SIGRTMIN+n`…

And [more generally][3]

> Several signal numbers are architecture-dependent, as indicated in the "Value" column.

So using numbers is less portable for no obvious benefit (or maybe there is a benefit, and I'm just not seeing it?).

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/signal.h.html
[2]: https://git.kernel.org/cgit/docs/man-pages/man-pages.git/tree/man7/signal.7?h=man-pages-4.09#n371
[3]: https://git.kernel.org/cgit/docs/man-pages/man-pages.git/tree/man7/signal.7?h=man-pages-4.09#n224

runcom: I'm fine using named signals - @stevvooe wdyt?

jonboulle:
@stevvooe sorry, my question was imprecise - I meant criteria with respect to what fields we add to this config object. e.g. I thought this was basically https://github.com/docker/docker/blob/master/image/spec/v1.md#image-json-description , except now we're extending beyond that.

config.md
stevvooe: We use the posix constants: https://github.com/docker/docker/blob/dad8cbfc2d6db45331be2cbd8f9997ab0851fccf/container/container_unit_test.go#L29, as far as I can tell. These may be slightly platform specific, so I am not sure what constraints to apply, if any at all.

vbatts:
god i hate this `schema/fs.go` file. It causes senseless number of rebases. :frowning_face: 

Can you rebase please?

runcom:
@vbatts done!

vbatts:
LGTM

stevvooe:
@runcom Please rebase!

runcom:
@stevvooe done

philips:
I still feel we shouldn't allow numbers unless there is a really good reason.

stevvooe:
> I still feel we shouldn't allow numbers unless there is a really good reason.

@runcom Is this easy to do? If not, let me know and we can merge this and open an issue.

runcom:
> @runcom Is this easy to do? If not, let me know and we can merge this and open an issue.

I think it's fine, this is the spec and it's just wording afaict. I can drop the number stuff in the stop signal paragraph I guess.

runcom:
@philips @stevvooe updated

runcom:
BTW, this is the original conversation when stopsignal was added in docker https://github.com/docker/docker/pull/15307#issuecomment-128387259 and Dan raised the issue that matching syscall names against a table is often outdated. Not sure that's a real argument though, but still, something to consider when deciding to use numbers as well.

stevvooe:
> BTW, this is the original conversation when stopsignal was added in docker docker/docker#15307 (comment) and Dan raised the issue that matching syscall names against a table is often outdated. Not sure that's a real argument though, but still, something to consider when deciding to use numbers as well.

I wonder if we could look these up dynamically.

stevvooe:
@runcom Looks like we require another rebase. Sorry.

runcom:
Rebased

stevvooe:
LGTM

philips:
LGTM

PR #545, opened 2017-01-31
docs: image base reference

### LINKED ISSUE #247: Images missing from HTML version of spec
Images are not rendering in the HTML released version of the Image Format spec (v0.4.0.)  

The root cause MIGHT be that the HTML src= attributes seem to use relative URLs instead of absolute URLs.:

```
<p>The following figure shows how the above media types reference each other:</p>
<p><img src="img/media-types.png" /></p>
```

NOTE: Images ARE rendering in PDF version of spec :)

### LINKED ISSUE #435: Version ought to be included in the docs
Presently the tagged version of the specification is only available in `./specs-go/version.go`, but the generated PDF or HTML docs do not hold any reference to the version.

Perhaps either generate a *.md file with this info and/or include header/footer with the version info.

vbatts:
Fixes #247
Fixes #435

this makes an ephemeral header that gets included into the html doc that
bases the images path to pull from the same tag on github.

CC @RobDolinMS @wking 

### DIFF
--- a/.tool/genheader.go
+++ b/.tool/genheader.go
@@ -1,0 +1,54 @@
+// Copyright 2017 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"text/template"
+
+	specs "github.com/opencontainers/image-spec/specs-go"
+)
+
+var headerTemplate = template.Must(template.New("gen").Parse(`<title>image-spec {{.Version}}</title>
+<base href="https://raw.githubusercontent.com/opencontainers/image-spec/{{.Branch}}/">`))
+
+type Obj struct {
+	Version string
+	Branch  string
+}
+
+func main() {
+	obj := Obj{
+		Version: specs.Version,
+		Branch:  specs.Version,
+	}
+	if strings.HasSuffix(specs.Version, "-dev") {
+		cmd := exec.Command("git", "log", "-1", `--pretty=%H`, "HEAD")
+		var out bytes.Buffer
+		cmd.Stdout = &out
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
+		}
+
+		obj.Branch = strings.Trim(out.String(), " \n\r")
+	}
+	headerTemplate.Execute(os.Stdout, obj)
+}
--- a/Makefile
+++ b/Makefile
@@ -68,12 +68,15 @@ $(OUTPUT_DIRNAME)/$(DOC_FILENAME).pdf: $(DOC_FILES) $(FIGURE_FILES)
 	$(PANDOC) -f markdown_github -t latex --latex-engine=xelatex -o $(PANDOC_DST)$@ $(patsubst %,$(PANDOC_SRC)%,$(DOC_FILES))
 	ls -sh $(shell readlink -f $@)
 
-$(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: $(DOC_FILES) $(FIGURE_FILES)
+$(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: header.html $(DOC_FILES) $(FIGURE_FILES)
 	@mkdir -p $(OUTPUT_DIRNAME)/ && \
 	cp -ap img/ $(shell pwd)/$(OUTPUT_DIRNAME)/&& \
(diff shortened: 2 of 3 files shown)


.tool/genheader.go
```diff
@@ -0,0 +1,20 @@
...
+	specs "github.com/opencontainers/image-spec/specs-go"
+)
+
+var headerTemplate = template.Must(template.New("gen").Parse(`<title>image-spec {{.}}</title>
+<base href="https://raw.githubusercontent.com/opencontainers/image-spec/{{.}}/">`))
+
+func main() {
+	version := specs.Version
+	if strings.HasSuffix(specs.Version, "-dev") {
+		version = "master"
```
wking: Why not just keep the `*-dev` version?

Makefile
```diff
@@ -65,16 +65,19 @@ $(OUTPUT_DIRNAME)/$(DOC_FILENAME).pdf: $(DOC_FILES) $(FIGURE_FILES)
 else
 $(OUTPUT_DIRNAME)/$(DOC_FILENAME).pdf: $(DOC_FILES) $(FIGURE_FILES)
 	@mkdir -p $(OUTPUT_DIRNAME)/ && \
-	$(PANDOC) -f markdown_github -t latex -o $(PANDOC_DST)$@ $(patsubst %,$(PANDOC_SRC)%,$(DOC_FILES))
+	$(PANDOC) -f markdown_github -t latex -o $(PANDOC_DST)$@ $(patsubst %,$(PANDOC_SRC)%,$(DOC_FILES)) && \
```
wking: No need for `&&` chaining.  Make will die if any individual command fails.

Makefile
```diff
@@ -65,16 +65,19 @@ $(OUTPUT_DIRNAME)/$(DOC_FILENAME).pdf: $(DOC_FILES) $(FIGURE_FILES)
...
 	ls -sh $(shell readlink -f $@)
 
-$(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: $(DOC_FILES) $(FIGURE_FILES)
+$(OUTPUT_DIRNAME)/$(DOC_FILENAME).html: header.html $(DOC_FILES) $(FIGURE_FILES)
 	@mkdir -p $(OUTPUT_DIRNAME)/ && \
 	cp -ap img/ $(shell pwd)/$(OUTPUT_DIRNAME)/&& \
-	$(PANDOC) -f markdown_github -t html5 -o $(PANDOC_DST)$@ $(patsubst %,$(PANDOC_SRC)%,$(DOC_FILES))
+	$(PANDOC) -f markdown_github -t html5 -H $(patsubst %,$(PANDOC_SRC)%,header.html) --standalone -o $(PANDOC_DST)$@ $(patsubst %,$(PANDOC_SRC)%,$(DOC_FILES)) && \
```
wking: `$(PANDOC_SRC)header.html` should work fine, and it reads more clearly (to me, anyway).

.tool/genheader.go
wking: > Why not just keep the *-dev version?

Ah, because it won't work for `<base>`.  Maybe use `specs.Version` directly in `<title>`, and then use the HEAD commit hash in `<base>` when `specs.Version` ends with `-dev`.

vbatts: lemme take a stab at that.

Makefile
vbatts: i don't recall that to be the case. It's not like there is an implied `set -e` inside the target

.tool/genheader.go
vbatts: updated.

Makefile
vbatts: fair. copy pasta. fixed

Makefile
wking: > i don't recall that to be the case. It's not like there is an implied `set -e` inside the target

    $ cat Makefile
    foo:
            false
            echo "still going?"
    $ make
    false
    Makefile:2: recipe for target 'foo' failed
    make: *** [foo] Error 1

If that's not convincing enough, I can try to hunt down some docs for you, just let me know.

vbatts: heh. fair. my memory fails me on the need for painful amounts of new-line escapes for long one-liners. No biggie.

vbatts:
updated. PTAL.

wking:
On Wed, Feb 01, 2017 at 02:39:02PM -0800, Vincent Batts wrote:
> updated. PTAL.

cec7700 is still using ‘master’ for *-dev versions.  I'm happy to take
a stab at my HEAD approach [1] in a PR against this one (or a
follow-up PR?) if that would be helpful.

[1]: https://github.com/opencontainers/image-spec/pull/545#discussion_r98728930

vbatts:
oh oh. You're saying to get the commit and reference the precise remote head commit. hrm.

vbatts:
fixed. Now a `-dev` version is pin the `<base>` at that HEAD commit that the docs were built from (which pre-supposes that the commit exists on github.com, but so it goes).

wking:
On Thu, Feb 02, 2017 at 08:11:47AM -0800, Vincent Batts wrote:
> … which pre-supposes that the commit exists on github.com…

Yeah, but that seems more likely than assuming a similar-enough image
will still be in master.

Anyhow, 7692fe47 looks good to me.

philips:
LGTM

jonboulle:
needs rebase

stevvooe:
@vbatts Please rebase!

vbatts:
rebased, but will need to rebase again on once master is fixed ...

vbatts:
rebased

stevvooe:
LGTM

jonboulle:
LGTM, lol, whatever

vbatts:
@jonboulle you mad

jonboulle:
@vbatts dat html/pdf generation 
wut

PR #547, opened 2017-02-02
descriptor.md:fix info

q384566678:

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -19,7 +19,7 @@ The following fields contain the primary properties that constitute a Descriptor
 
 - **`mediaType`** *string*
 
-  This REQUIRED property contains the media type of the referenced content.
+  This OPTIONAL property contains the media type of the referenced content.
   Values MUST comply with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2].
 
   The OCI image specification defines [several of its own MIME types](media-types.md) for resources defined in the specification.


descriptor.md
```diff
@@ -19,7 +19,7 @@ The following fields contain the primary properties that constitute a Descriptor
 
 - **`mediaType`** *string*
 
-  This REQUIRED property contains the media type of the referenced content.
+  This OPTIONAL property contains the media type of the referenced content.
```
wking: This should be REQUIRED.  Without it, a consumer will not know how to interpret the referenced content.

stevvooe:
Closing because this is wrong. MediaType is required on descriptors.

@q384566678 I am really having trouble making sense of your PRs. Could you please take the time to link to an issue or describe why you are proposing a change?

q384566678:
@stevvooe I made this changes because of [this](https://github.com/opencontainers/image-spec/blame/master/specs-go/v1/descriptor.go#L23).Perhaps I do not have a good understanding of this problem.

stevvooe:
@q384566678 What do you mean by that? That is just a link to a line of code.

wking:
> What do you mean by that? That is just a link to a line of code.

Maybe that having `omitempty` for a required property is strange.  I suggest dropping `omitempty` (and keeping descriptor `mediaType` required).

q384566678:
> Maybe that having omitempty for a required property is strange. I suggest dropping omitempty (and keeping descriptor mediaType required).

Agree,
Before I look at this code, `omitempty` let me mistakenly believe that it is an optional field, so it caused the wrong PR.

wking:
> … `omitempty` let me mistakenly believe that it is an optional field…

I've filed #553 removing it.

stevvooe:
> Maybe that having omitempty for a required property is strange. I suggest dropping omitempty (and keeping descriptor mediaType required).

This is terrible practice. `omitempty` and a field requirement are not related.

PR #548, opened 2017-02-02
validate.go: add the verification content

q384566678:
According to [this](https://github.com/opencontainers/image-spec/blame/master/manifest.md#L57-60)  modified.

### DIFF
--- a/schema/validator.go
+++ b/schema/validator.go
@@ -111,7 +111,9 @@ func validateManifestDescendants(r io.Reader) error {
 
 	for _, layer := range header.Layers {
 		if layer.MediaType != string(v1.MediaTypeImageLayer) &&
-			layer.MediaType != string(v1.MediaTypeImageLayerNonDistributable) {
+			layer.MediaType != string(v1.MediaTypeImageLayerGzip) &&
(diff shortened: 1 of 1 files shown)


q384566678:
Here I have a doubt that there is no reason for the validation of the mediatype field, because the mediatype field is optional and not always present in the descriptor.

wking:
48fcec2 looks good to me.

stevvooe:
It might be better to combine this into a function that tells whether or not a mediatype is a layer, although, this validation may be too strict.

q384566678:
> It might be better to combine this into a function that tells whether or not a mediatype is a layer, although, this validation may be too strict.

@stevvooe The judgment here is only a warning, not an error. So it can not be said that the conditions are not layers.So I do not think your idea is good.

vbatts:
LGTM

cyphar:
@wking As @q384566678 said, this is only a warning -- the discussion inside opencontainers/image-tools#111 is not completely valid (this schema validator is not meant to error on unknown layer types).

q384566678:
ping @philips @jonboulle @stevvooe

jonboulle:
lgtm

stevvooe:
As @cyphar said, this is probably too aggressive.

PR #549, opened 2017-02-02
descriptor_test.go:fix test

q384566678:


wking:
This seems like it should be part of #547.  But I think descriptors
should REQUIRED mediaType [1], in which case we'd need neither.

[1]: https://github.com/opencontainers/image-spec/pull/547#discussion_r99169640

schema/content-descriptor.json
```diff
@@ -26,7 +26,6 @@
     }
   },
   "required": [
-    "mediaType",
```
stevvooe: Why are you removing this? `mediaType` is a requirement for descriptors.

stevvooe:
This is also incorrect. Closing.

schema/descriptor_test.go
```diff
@@ -24,7 +24,7 @@ import (
 func TestDescriptor(t *testing.T) {
 	for i, tt := range []struct {
 		descriptor string
-		fail     bool
+		fail       bool
```
q384566678: @stevvooe Can you please reopen this PR? I will delete the wrong changes, only to retain this one format changes.thanks.

stevvooe: Just open another PR and be clear about what you're actually fixing.

PR #552, opened 2017-02-03
schema/manifest_backwards_compatibility_test.go: drop mediatype

q384566678:

### DIFF
--- a/schema/manifest_backwards_compatibility_test.go
+++ b/schema/manifest_backwards_compatibility_test.go
@@ -49,10 +49,9 @@ func TestBackwardsCompatibilityManifestList(t *testing.T) {
 		fail         bool
 	}{
 		{
-			digest: "sha256:219f4b61132fe9d09b0ec5c15517be2ca712e4744b0e0cc3be71295b35b2a467",
+			digest: "sha256:ce939cd86094ec89abba1168d55257dfe74781f438916737d5c0905d7afb8803",
 			manifestlist: `{
    "schemaVersion": 2,
-   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": [
       {
          "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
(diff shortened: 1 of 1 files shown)


schema/manifest_backwards_compatibility_test.go
```diff
@@ -49,10 +49,9 @@ func TestBackwardsCompatibilityManifestList(t *testing.T) {
 		fail         bool
 	}{
 		{
-			digest: "sha256:219f4b61132fe9d09b0ec5c15517be2ca712e4744b0e0cc3be71295b35b2a467",
+			digest: "sha256:ce939cd86094ec89abba1168d55257dfe74781f438916737d5c0905d7afb8803",
 			manifestlist: `{
    "schemaVersion": 2,
-   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
```
wking: I don't think we want these, because (as we've [covered before][1]) `manifest_backwards_compatibility_test.go` is about “are we compatible with caught-in-the-wild blobs blobs?” and making local adjustments like this (even to a field which we will be ignoring in the OCI framework) defeats that purpose.

[1]: https://github.com/opencontainers/image-spec/pull/510#discussion_r94909550

q384566678: For this I am a bit confused, if you can not modify the examples of these tests, these original test data is how to get?

wking: > …if you can not modify the examples of these tests, these original test data is how to get?

It would be nice if all of these blobs had notes on their provenance (like we have [here][1]).  And for the second example you touch in this PR, [we do have those notes][2].  For the other blobs (including the first one we touch in this PR), we're left to trust the initial authors that they found these somewhere and haven't manipulated them too much.  So you can add new tests here by capturing new blobs in the wild, and we can drop blobs here if we no longer want to test backwards-compat with an old blob.  But we don't want to change the existing blobs by hand.

And all of this is just for `manifest_backwards_compatibility_test.go`.  In other files you can manipulate as much as you like.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc4/schema/manifest_backwards_compatibility_test.go#L198-L202
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc4/schema/manifest_backwards_compatibility_test.go#L135

Issue #467, opened 2016-11-18
Define image-related action nomenclature


stevvooe:
The image-tools repository has a few different terms that are being applied to image layouts that seems unintuitive or confusing. To ensure that tools have operational parity, it seems like a solid idea to define the terminology for this common operations.

Specifically, we should define the high-level terms for "unpack", "create", "pack" (?).

This will make it clear what functionality the specification actually provides.

wking:
On Fri, Nov 18, 2016 at 12:30:13PM -0800, Stephen Day wrote:
> Specifically, we should define the high-level terms for "unpack",
> "create", "pack", "verify" (?).

+1 to defining these sorts of terms.

The spec already uses “applying changesets” [1] for the operation that
image-tools describes as “unpack”.  And it uses something like
“creating” [2] or “representing changes” [3] for the inverse.
(Un)packing sounds more specific to me, so I'm in favor of adopting
(and defining) (un)packing in the spec.

I'm not aware of clear wording in the spec analagous to image-tool's
“create” (and the spec is weak on config translation in general, see
#454).

“verify” seems pretty clear to me, and the spec's key words [4] and
in-flight compliance language (#432) cover most of what I'd like to
see there.

[1]: https://github.com/opencontainers/image-spec/blob/de440dbfe41f35eb03117204bc99eeb0e6a186b8/layer.md#applying-changesets
[2]: https://github.com/opencontainers/image-spec/blob/de440dbfe41f35eb03117204bc99eeb0e6a186b8/layer.md#creating
[3]: https://github.com/opencontainers/image-spec/blob/de440dbfe41f35eb03117204bc99eeb0e6a186b8/layer.md#representing-changes
[4]: https://github.com/opencontainers/image-spec/blob/de440dbfe41f35eb03117204bc99eeb0e6a186b8/README.md#overview

wking:
I've taken a stab at this in #554.

vbatts:
so we want to go with "pack" and "unpack" vs "create" and "apply"?
honestly, cleaning up the current wording is a separate task from switching to new terms. These new terms are more confusing in the sense of sounding like regular tar archives, which they aren't.

wking:
On Wed, Mar 08, 2017 at 11:18:56AM -0800, Vincent Batts wrote:
> so we want to go with "pack" and "unpack" vs "create" and "apply"?
> honestly, cleaning up the current wording is a separate task from
> switching to new terms. These new terms are more confusing in the
[...]

Issue #415, opened 2016-10-25
ensure that layer unpacking algorithms are windows compatible


stevvooe:
Recently, there has been a lot of activity in [`layer.md`](https://github.com/opencontainers/image-spec/blob/master/layer.md) around the unpacking procedure. I am not sure if we've fully evaluated this in the context of windows containers and other platforms.

As a strawman, there is a lot of discussion around links. How are these correctly handled in NTFS?

@jstarks @RobDolinMS

RobDolinMS:
Thanks @stevvooe.  I've sent an email ping to @jhowardmsft and @jstarks to encourage them to take a look.

philips:
@RobDolinMS ping.

stevvooe:
@jhowardmsft @jstarks PTAL

AkihiroSuda:
* some notes about sockets, block devices, char devices, and FIFOs seem needed
* It should be clarified that the content of rootfs is extracted under the root of the `%SystemDrive%`?
(Is there any standard about `%SystemDrive%`? Is it guaranteed to be always `C:`?)

AkihiroSuda:
PTAL @jhowardmsft @jstarks

vbatts:
can i test this on a vm? Is windows 7 suitable enough?

vbatts:
like fifos, symlinks, sockets, character devices. These are platform specific.
The assumption on the current doc, that is readying for v1, is that it is unless otherwise specified targeting POSIX. As windows specific changes come in, we'll just denote those platform dependent differences, so that a post-v1 addition will not break the current POSIX expectations.

_(i realize it's not strictly POSIX or the "POSIX is dead" or whatever, but it most closely matches what happens here)_

jonboulle:
POSIX is dead?? no-one told me!!

jonboulle:
(I would also really like to hear more input from our MS friends on how the current layer specification is working out for them)

vbatts:
> POSIX is dead?? no-one told me!!

2016 article that recently floated around
http://www.cs.columbia.edu/~vatlidak/resources/POSIXmagazine.pdf
Take it with a grain of salt ...

stevvooe:
PTAL @RobDolinMS @jstarks @jhowardmsft

philips:
PTAL @RobDolinMS @jstarks @jhowardmsft

philips:
@RobDolinMS @jstarks @jhowardmsft This is a v1.0 blocking issue.

[...]

PR #172, opened 2016-07-18
Export types

runcom:
/cc @vbatts @stevvooe @philips 

This is another version https://github.com/opencontainers/image-spec/pull/147 w/o using any code generation tool - as discussed over there

Close https://github.com/opencontainers/image-spec/pull/147

PR #553, opened 2017-02-03
specs-go/v1/descriptor: Drop 'omitempty' from mediaType

wking:
This is a required property, so `omitempty` serves no purpose.

### DIFF
--- a/specs-go/v1/descriptor.go
+++ b/specs-go/v1/descriptor.go
@@ -20,7 +20,7 @@ import digest "github.com/opencontainers/go-digest"
 // This structure provides `application/vnd.oci.descriptor.v1+json` mediatype when marshalled to JSON
 type Descriptor struct {
 	// MediaType is the media type of the object this schema refers to.
-	MediaType string `json:"mediaType,omitempty"`
+	MediaType string `json:"mediaType"`
 
 	// Digest is the digest of the targeted content.
 	Digest digest.Digest `json:"digest"`


philips:
LGTM

stevvooe:
`omitempty` has no bearing on whether the field is required or not. They are completely unrelated. It just means that if there is an empty string, you don't emit the field. Either way, the descriptor is still invalid.

I am in favor of leaving `omitempty` in place for all fields.

wking:
On Fri, Feb 03, 2017 at 02:28:12PM -0800, Stephen Day wrote:
> It just means that if there is an empty string, you don't emit the
> field. Either way, the descriptor is still invalid.

Agreed.

> I am in favor of leaving `omitempty` in place for all fields.

Your “either way, the descriptor is still invalid” argument will break
down if there is a required property where Go's zero-value is legal.
For example, Descriptor.Size may be zero (and it's not omitempty [1]).

I don't see a point to having ‘omitempty’ where you never intend it to
be used, but *shrug*, it's your project.

[1]: https://github.com/opencontainers/image-spec/blob/ed18de10b0d7e34076e1557bdcd6f500504a6673/specs-go/v1/descriptor.go#L29

stevvooe:
@wking Look, I am just trying to reduce churn in areas that don't need changes. How bout someone tries to tackle action problems?

https://github.com/opencontainers/image-spec/issues/467
https://github.com/opencontainers/image-spec/issues/415

This doesn't even address a real problem. Is there a bug? Did this break something? No, it is just a waste of time.

> For example, Descriptor.Size may be zero (and it's not omitempty [1]).

What exactly is the point of a zero-sized target object? In fact, there can only be one valid item with zero-sized target and that `Digest` is `sha256:e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`. What, then, is the media type of said thing?

Let's get focused on a 1.0 and stop faffing about.

wking:
On Fri, Feb 03, 2017 at 03:44:23PM -0800, Stephen Day wrote:
> This doesn't even address a real problem. Is there a bug? Did this
> break something? No, it is just a waste of time.

This is minor polishing to reduce confusion like [1] and save
maintainers time in the long run (by not having to field PRs like
#547).

> What exactly is the point of a zero-sized target object?

I dunno, but nobody feels strongly enough about it to make it illegal
(and I think not jumping through hoops to make things illegal is a
good thing).

> Let's get focused on a 1.0 and stop faffing about.

I didn't expect this removal to be so contentious ;).  But I'll PR
some language for #467 to rebalance my karma (although it's not
milestoned 1.0.0, perhaps an oversight?).

[1]: https://github.com/opencontainers/image-spec/pull/547#issuecomment-277153830

stevvooe:
> This is minor polishing to reduce confusion like [1] and save
maintainers time in the long run (by not having to field PRs like
#547).

How is this the cause of #547? The problem with #547 exists between a keyboard and a chair.

> I didn't expect this removal to be so contentious ;). 

It is just unnecessary.

wking:
On Fri, Feb 03, 2017 at 04:43:04PM -0800, Stephen Day wrote:
> > This is minor polishing to reduce confusion like [1] and save
> > maintainers time in the long run (by not having to field PRs like
> > #547).
>
> How is this the cause of #547? The problem with #547 exists between
> a keyboard and a chair.

Excepting repository curruption and similar, *all* problems are
between a keyboard and a chair ;).

> > I didn't expect this removal to be so contentious ;).
>
> It is just unnecessary.

*This same change* landed without contention in #410 (for the
since-removed Versioned.MediaType, which was also required at that
point [1]).  There was backing discussion around that change in [2].
Looking back at the history of Descriptor.MediaType, we have it
landing without omitempty in #172, and the omitempty getting injected
without discussion in #411.  I'm not sure what @vbatts' motivation was
for adding it; maybe he can weigh in here?  And grepping master for
‘omitempty’ shows Descriptor.MediaType as the *only* omitempty
property that is REQUIRED in the associated Markdown (all of the other
omitempty properties are explicitly OPTIONAL).

So is this PR going to have a big impact on compiled Go code?  No.
But I don't think I had reason to expect this PR would have taken any
more maintainer time than a glance at the diff, an LGTM in a comment,
and a tap on the merge button.

[1]: https://github.com/opencontainers/image-spec/pull/410/files#diff-ba085f9a0c64bb6a91f374aa38b15393L25
[2]: https://github.com/opencontainers/image-spec/commit/4267a18d918c4d87bb7423b9f1d6ae72b9ea9e76#commitcomment-19497879

stevvooe:
@wking Unfortunately, had I seen #411, I would have rejected it.

`omitempty` ensures that we have clean output that reduces errors in consumers. `omitempty` has nothing to do with REQUIRED vs OPTIONAL. Stop making that false equivalence.

I am closing this and I hope someone will take the time to revert #411 so I don't have to.

Issue #557, opened 2017-02-06
Revert #410


stevvooe:
https://github.com/opencontainers/image-spec/pull/410

`omitempty` has nothing to do with REQUIRED vs OPTIONAL fields.

wking:
Cross-linking earlier discussion in #553.

philips:
@stevvooe in #553 you are saying we should omitempty everywhere. So you would be ok with #410 if it was just doing omitempty but didn't mention optional/required?

stevvooe:
@philips We should not emit fields that contain zero-values. It's wasteful and distracting.

There are certain exceptions, but, in general, use of `omitempty` should be the rule.

wking:
On Mon, Feb 06, 2017 at 02:20:19PM -0800, Stephen Day wrote:
> There are certain exceptions, but, in general, use of `omitempty`
> should be the rule.

If the exceptions are “when the zero-value is valid” (based on [1]).
Then:

  $ git grep '^        [A-Z]' specs-go/ | grep -v omitempty

turns up the following fields (in master, as of ed18de10) which could
pick up omitempty:

* RootFS.type (REQUIRED, MUST be ‘layers’)
* RootFS.DiffIDs (REQUIRED, corresponsed to manifest's ‘layers’ which
  REQUIRES at least one entry with #407 in flight).
* Image.Architecture (REQUIRED, empty string not explicitly illegal,
  but it probably should be).
* Image.OS (REQUIRED, empty string not explicitly illegal, but it
  probably should be).
* Descriptor.Digest (REQUIRED, empty string missing algorithm,
  separator, and hash).
* ImageLayout.Version (REQUIRED, empty string will never be valid
  unless image-spec cuts an empty-string release).
* Manifest.Layers (REQUIRES at least one entry with #407 in flight).
* Platform.Architecture (REQUIRED, empty string not explicitly
  illegal, but it probably should be).
* Platform.OS (REQUIRED, empty string not explicitly illegal, but it
  probably should be).
* Versioned.SchemaVersion (REQUIRED, must be 2).

I still don't see a point to adding omitempty to a REQUIRED property,
but if that's what you want to do, the above properties are probably
what you'd do it to.

Manifest.Manifests is REQUIRED, although without a requirement to have
at least one value I'm not sure why.  In any case, as long as it's
required and allows an empty array we don't want ‘omitempty’ there.

[1]: https://github.com/opencontainers/image-spec/pull/553#issuecomment-277379924

[...]

PR #558, opened 2017-02-08
Delete the omitempty in the field

q384566678:
Solve #557 .
@stevvooe  I can't revert #410 ，because the content have changed since it was merged. So I  open this PR to restore it.

### DIFF
--- a/specs-go/v1/config.go
+++ b/specs-go/v1/config.go
@@ -19,28 +19,28 @@ import "time"
 // ImageConfig defines the execution parameters which should be used as a base when running a container using an image.
 type ImageConfig struct {
 	// User defines the username or UID which the process in the container should run as.
-	User string `json:"User,omitempty"`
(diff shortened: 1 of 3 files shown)


stevvooe:
Sorry, I must have been confused. Fields _should_ generally have `omitempty`. The following are missing `omitempty`:

https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest.go#L20
https://github.com/opencontainers/image-spec/blob/master/specs-go/v1/manifest_list.go#L23

I am really sorry. I misread the original changes.

q384566678:
Uh, I think I'm a bit confused.How to determine specific circumstances?

stevvooe:
@q384566678 We can close this. We have `omitempty` on the correct fields.

Again, apologies. I got confused when discussing this earlier.

PR #560, opened 2017-02-08
schema: validation: correctly validate gzip layer mediatypes

cyphar:
When the change was made to make gzip optional for layers, the validator
was not updated -- causing completely valid layers to cause warnings
during validation.

Fixes: aad7f240f0c54 ("media-types: Define layer media types with and without '+gzip'")


cyphar:
Can we _please_ be a bit more careful?

/cc @wking

wking:
Duplicates #548?

vbatts:
it is a dup of #548

PR #564, opened 2017-02-09
move defs-config.json to config-schema.json

xiekeyang:
The schema items of `config`, `rootfs` and `history`, to be config
properties, are and will only referred by config-schema.json. So they
had better to be moved to config-schema.json.


vbatts:
I suppose that is more readable.

LGTM

philips:
LGTM

stevvooe:
LGTM

PR #567, opened 2017-02-13
schema: urls optimization

q384566678:


schema/defs-image.json
```diff
@@ -11,6 +11,14 @@
       "type": "string",
       "pattern": "^[a-z0-9_+.-]+:[a-f0-9]+$"
     },
+    "urls": {
```
stevvooe: This should be centralized in the descriptor type, not the image type.

vbatts:
LGTM

stevvooe:
Needs a rebase.

wking:
@jbouzane, I see you [approved this PR][1], but PullApprove doesn't count that as a LGTM [unless you use their v2 config][2] (which we aren't).  If folks want to switch to a v2 config, I've taken a stab at one in opencontainers/project-template#29 and am happy to land any pieces of that in whatever repository folks want if they don't want to land it in project-template.  Until that happens, you should probably stick to [LGTM/Rejected/etc.][3] comments if you want PullApprove to notice.

[1]: https://github.com/opencontainers/image-spec/pull/567#pullrequestreview-25901352
[2]: https://github.com/pullapprove/support/issues/98#issuecomment-256448778
[3]: https://github.com/opencontainers/image-spec/blob/039bc0c964d6074350d0195b75262a889ee7e391/.pullapprove.yml#L2-L3

wking:
Needs a rebase (likely because #539 has landed, so we want to drop [this][1]).

[1]: https://github.com/opencontainers/image-spec/pull/567/files#diff-6f11672cc3c0cd49052d926e2ca5a82aR29

q384566678:
updatede,PTAL

jonboulle:
lgtm

vbatts:
LGTM

PR #568, opened 2017-02-13
schema: delete content that is not used

q384566678:
The content is not called or used, and I think they may be superfluous.


stevvooe:
This looks like it was copied over from the runtime-spec.

@vbatts Is there anything here that was useful?

philips:
LGTM

stevvooe:
LGTM

jonboulle:
lgtm

PR #569, opened 2017-02-13
manifestlist_test: fix test

q384566678:


stevvooe:
@q384566678 We really need these PRs to start including context as to why this should fail.

Please explain why you think this change should go through.

q384566678:
@stevvooe Here I just want to add a validation, that is, in the presence of the platform, os and arch must exist. Because platform is an optional field, so here is a list of circumstances.

vbatts:
LGTM

stevvooe:
LGTM

PR #571, opened 2017-02-13
Makefile: suitable "default" target

vbatts:
I like that there is a "help" target, but the default ought to do
sensible things. :-)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ DOC_FILENAME	?= oci-image-spec
 
 EPOCH_TEST_COMMIT ?= v0.2.0
 
-default: help
+default: check-license lint test
 
 help:
 	@echo "Usage: make <target>"


stevvooe:
LGTM

Makefile
```diff
@@ -40,7 +40,7 @@ DOC_FILENAME	?= oci-image-spec
 
 EPOCH_TEST_COMMIT ?= v0.2.0
 
-default: help
+default: check-license validate-examples lint test
```
wking: `validate-examples` is a subset of `test`, so we can drop it from the list of prereqs here.

vbatts: hm. i forget that.

vbatts:
updated. PTAL

philips:
LGTM

stevvooe:
LGTM

PR #576, opened 2017-02-17
image-index: delete the extra fields

q384566678:
I do not understand why this field is added and it is not used.


wking:
On Thu, Feb 16, 2017 at 10:33:00PM -0800, Zhou Hao wrote:
> I do not understand why this field is added and it is not used.

‘elements’ was part of the discussion in #533 and #546.  @vbatts meant
to drop it (see [1]), but looks like this bit slipped through.  +1 to
removing it.

[1]: https://github.com/opencontainers/image-spec/pull/533#issuecomment-277355950

stevvooe:
Please rebase.

q384566678:
rebased. PTAL

stevvooe:
LGTM

q384566678:
@vbatts @jonboulle @philips PTAL

jonboulle:
LGTM

PR #579, opened 2017-02-22
fix error valid target in image-layout

xiekeyang:
In layout.md, The validated target should be image layout itself, not only image index.

I'd like to extract and land this firstly, while #452 is in WIP. 

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -134,7 +134,7 @@ The `imageLayoutVersion` value will align with the OCI Image Specification versi
 
 ### oci-layout Example
 
-```json
+```json,title=OCI%20Layout&mediatype=oci%2Dlayout
 {
     "imageLayoutVersion": "1.0.0"
 }


image-layout.md
```diff
@@ -134,7 +134,7 @@ The `imageLayoutVersion` value will align with the OCI Image Specification versi
 
 ### oci-layout Example
 
-```json
+```json,title=OCI%20Layout&mediatype=oci%2Dlayout
```
wking: I'd [still][1] rather have “oci-layout file” (or similar) as the title and a real media type.

[1]: https://github.com/opencontainers/image-spec/pull/452#discussion_r102088422

xiekeyang: > I'd still rather have “oci-layout file” (or similar) as the title and a real media type.
it likely make no much sense. Refer to `manifest.md` or so.

wking: > > I'd still rather have “oci-layout file” (or similar) as the title and a real media type.
>
> it likely make no much sense. Refer to manifest.md or so.

`manifest.md` has `title=Manifest`, which makes sense, because the JSON it's titling is a manifest.  Here, the JSON we're titling is an oci-layout file, so I think we want `title=oci-layout%20file`.

xiekeyang:
@opencontainers/image-spec-maintainers This might be ready to merged, PTAL.

vbatts:
I'm not opposed, though `oci-layout` doesn't have a mediaType yet. Perhaps we need this for validating them and this example as well.

vbatts:
oh we do! I was searching for "oci-layout" and missed `./schema/image-layout-shema.json`. I'll add more to that description.

LGTM

image-layout.md
stevvooe: How is `mediatype=oci%2Dlayout` a valid mediatype?

xiekeyang: @stevvooe 
Sorry for late response. This is because of https://github.com/opencontainers/image-spec/blob/master/schema/spec_test.go#L70. It means that only those who have `mediatype` JSON key will be validated. Any arbitrary string here will not influence it to be joined to spec-test function. I use `oci-layout`, as to close to description in `image-layout-schema.json`.
If we need a nicer name here?

jonboulle:
lgtm, will let @stevvooe ack/nack tho

image-layout.md
stevvooe: I see. We should probably note somewhere that this _is not_ a mediatype.

stevvooe:
LGTM

stevvooe:
@xiekeyang Follow this up with a note somewhere that `oci-layout` is not a mediatype.

xiekeyang:
@stevvooe got it. i would present another patch for this.

PR #580, opened 2017-02-23
descriptor-test: add test

q384566678:


wking:
a2dab39c7 looks good to me.

We probably also want a test with a valid URL to ensure that passes,
but I don't care if that happens in this PR or in followup work.

q384566678:
@vbatts @jonboulle @stevvooe @philips PTAL

stevvooe:
LGTM

philips:
LGTM

PR #582, opened 2017-02-23
image-layout: make index.json refnames unique

### LINKED ISSUE #581: index.json: make org.opencontainers.ref.name unique?
While trying to implement `index.json` parsing in `umoci` I've come to a bit of a worrying realisation. There isn't really any explanation of how cases where `org.opencontainers.ref.name` is not unique should be handled. While I understand the case where you have multiple `application/vnd.oci.image.manifest.v1+json` descriptors with different `platform` entries (though I think that doesn't make sense in light of the fact that `application/vnd.oci.image.index.v1+json` has `platform` entries too), it doesn't really make sense to me in the general case to allow `org.opencontainers.ref.name`s to be duplicated in `index.json`.

So here's two things that would be nice if we could make `org.opencontainers.ref.name`s unique. This includes the multiple-`platform` case I mention above because it starts to get a bit hairy -- and honestly you should just be using `application/vnd.oci.image.index.v1+json` for that.

Sorry to stir this pot again, I've only just hit this issue trying to fix up the `oci/cas` implementation in `umoci`.

#533

cyphar:
Without this change, a generic image handling tool would not be able to
handle all cases with duplicate org.opencontainers.ref.name annotations,
due to ambiguity about which descriptor is the "best match".

While it has been floated that since the annotation is not actually
mandatory implementations should have alternative methods of referencing
entries in an image-index, this is quite a frustrating issue for
implementations that just want to use org.opencontainers.ref.name in the
"common" case.

Fixes #581 

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -147,7 +147,8 @@ The [image index](image-index.md) is a multi-descriptor entry point.
 
 This index provides an established path (`/index.json`) to have an entry point for an image-layout and to discover auxiliary descriptors.
 
-No semantic restriction is given for the "org.opencontainers.ref.name" annotation of descriptors.
+If set, the value of the `org.opencontainers.ref.name` annotation in each descriptor MUST be unique to the `index.json`.
+No semantic restriction is given for the `org.opencontainers.ref.name` annotation of descriptors.
 In general the `mediaType` of each [descriptor][descriptors] object in the `manifests` field will be either `application/vnd.oci.image.index.v1+json` or `application/vnd.oci.image.manifest.v1+json`.
 Future versions of the spec MAY use a different mediatype (i.e. a new versioned format).
 An encountered `mediaType` that is unknown SHOULD be safely ignored.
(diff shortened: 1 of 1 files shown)


coolljt0725:
https://github.com/opencontainers/image-spec/blob/master/annotations.md#rules says that `Keys MUST be unique within this map` ,  does it address your issue?

gao-feng:
@coolljt0725 I thought they are in different `manifests` array element,
https://github.com/opencontainers/image-spec/blob/master/image-layout.md#index-example


LGTM

mikebrow:
@coolljt0725 he's asking for the values to be unique for that particular key (annotation) for entries in an image index.

q384566678:
looks good to me.

jonboulle:
LGTM

stevvooe:
Again, validating over annotations is a non-starter.

This is the exact kind of scope creep that is a concern when adding annotations. Now we have fields that are first-class in the annotations. Nice job.

But either, even though there are major objections in #581, just submit a PR and shove it through. This is open source after all. /s

stevvooe:
Reject

stevvooe:
Actually, I am closing this until we can have further discussion on #581. This is fairly dishonest to move forward with a PR while something is still contentious.

cyphar:
@stevvooe Sorry, I opened the PR and issue at the same time (6 days ago) -- that's how I've always created issues. I didn't mean to seem dishonest (I assumed you'd seen this PR). My apologies.

PR #583, opened 2017-02-23
Use `REQUIRED properties` instead of `base properties`

coolljt0725:
`base properties` is not quite clear because we don't
define which property is the base properties and which
is not the base properties. Use `REQUIRED properties`
to make it more clear.

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -25,7 +25,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
   This REQUIRED property contains a list of [manifests](manifest.md) for specific platforms.
   While this property MUST be present, the size of the array MAY be zero.
 
-  Each object in `manifests` has the base properties of [descriptor](descriptor.md) with the following additional properties and restrictions:
+  Each object in `manifests` has the REQUIRED properties of [descriptor](descriptor.md) with the following additional properties and restrictions:
 
   - **`mediaType`** *string*
 
--- a/image-layout.md
+++ b/image-layout.md
@@ -26,8 +26,7 @@ The image layout is as follows:
     - It MAY include additional fields
 - `index.json` file
     - It MUST exist
-    - It MUST be a JSON object
-    - It MUST have the base properties of an [image index](image-index.md).
+    - It MUST be an [image index](image-index.md) JSON object.
     - See [index.json](#indexjson-file) section
 
 ## Example Layout


image-index.md
```diff
@@ -25,7 +25,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
   This REQUIRED property contains a list of [manifests](manifest.md) for specific platforms.
   While this property MUST be present, the size of the array MAY be zero.
 
-  Each object in `manifests` has the base properties of [descriptor](descriptor.md) with the following additional properties and restrictions:
+  Each object in `manifests` has the REQUIRED properties of [descriptor](descriptor.md) with the following additional properties and restrictions:
```
mikebrow: That's not the point. The objects in the manifests list each contain a set of descriptor properties. Some of these are REQUIRED but not all.

Suggest: "Each object in `manifests` includes a set of [descriptor properties](descriptor.md#properties) with the following additional properties and restrictions:"

image-layout.md
```diff
@@ -27,7 +27,7 @@ The image layout is as follows:
 - `index.json` file
     - It MUST exist
     - It MUST be a JSON object
-    - It MUST have the base properties of an [image index](image-index.md).
+    - It MUST have the REQUIRED properties of an [image index](image-index.md).
```
mikebrow: Suggest: "It MUST include [image index properties](image-index.md)."

wking: How about consolidating the last two lines:

> - `index.json` file
>     - It MUST exist.
>     - It MUST be an [image index](image-index.md) JSON object.

Or alternatively:

> - `index.json` file
>     - It MUST exist.
>     - It MUST be a valid [`application/vnd.oci.image.index.v1+json`](image-index.md).

mikebrow: prefer @wking 's first option

image-index.md
coolljt0725: @mikebrow  is `a set of`  also not clear on which property is required and which is not requred?

image-layout.md
coolljt0725: I'm ok with @wking 's first option

vbatts:
is this PR waiting on updates, @coolljt0725 ?

coolljt0725:
@vbatts  sorry for the late response, updated

vbatts:
LGTM

image-index.md
mikebrow: Yes. The point is, we are repeating the REQUIRED statement which is located in the forward reference document.

wking: +1 to @mikebrow's wording.  We're also including OPTIONAL descriptor properties in this extension, not just the REQUIRED properties.

PR #591, opened 2017-03-08
Add some restriction to "org.opencontainers.ref.name"

coolljt0725:
A common use case of `org.opencontainers.ref.name`
is representing a `tag` for a container image. In docker,
there is some restriction of the tag see
https://github.com/docker/docker/blob/master/vendor/github.com/docker/distribution/reference/regexp.go#L37.
Without these restrictions, I'm afraid the tag of
oci image is invalid for docker. These restrictions used to exist but removed in
https://github.com/opencontainers/image-spec/pull/533, removed it by accidents or by intention?

ping @stevvooe  @vbatts

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -148,6 +148,7 @@ The [image index](image-index.md) is a multi-descriptor entry point.
 This index provides an established path (`/index.json`) to have an entry point for an image-layout and to discover auxiliary descriptors.
 
 No semantic restriction is given for the "org.opencontainers.ref.name" annotation of descriptors.
+The value of "org.opencontainers.ref.name" annotation is a string which MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
 In general the `mediaType` of each [descriptor][descriptors] object in the `manifests` field will be either `application/vnd.oci.image.index.v1+json` or `application/vnd.oci.image.manifest.v1+json`.
 Future versions of the spec MAY use a different mediatype (i.e. a new versioned format).
 An encountered `mediaType` that is unknown SHOULD be safely ignored.


lucab:
Cross-reference https://github.com/opencontainers/image-spec/issues/581#issuecomment-284777108

image-layout.md
```diff
@@ -148,6 +148,7 @@ The [image index](image-index.md) is a multi-descriptor entry point.
 This index provides an established path (`/index.json`) to have an entry point for an image-layout and to discover auxiliary descriptors.
 
 No semantic restriction is given for the "org.opencontainers.ref.name" annotation of descriptors.
+The value of "org.opencontainers.ref.name" annotation is a string which MUST NOT include characters outside of the set of "A" to "Z", "a" to "z", "0" to "9", the hyphen `-`, the dot `.`, and the underscore `_`.
```
lucab: The original regexp also has some more constraints on corner-cases which are allowed by your wording:
 * empty string
 * arbitrarily-long string (> 128 chars)
 * symbol-only string (eg. `..`)
 * string starting with a symbol (eg. `.hidden`)

vbatts:
I'm not sure this is always wanted though. It had been a request in the past for nested names, like including the "/" but not possible due to file names. It there are hard requirement for these restrictions?

stevvooe:
We'll have to rethink restrictions here. These were really to support mapping to arbitrary filesystems. We have a lot more options, now that this is just a json string.

Recently, I have done some research in specification names in the context of [containered](https://github.com/docker/containerd/pull/612/files#diff-d653f2397c86d971763dd0decf06ce72R9). This format provides a superset of docker+rkt(based on examples) and is intended to be a lot less opinionated. It is based around schema-less URIs. Now, I am not suggesting we adopt this, but I think the approach of restricting this to the URI charset may be viable for this use case.

stevvooe:
@coolljt0725 Could you please open an issue so we can discuss the way forward before we go ahead with a PR?

stevvooe:
Closing in favor of addressing full scope of problem in #599.

PR #538, opened 2017-01-28
schema/gen: Include .json instead of excluding .go

wking:
To avoid accidentally including [`.swp`, `.un~`, etc.][1]  Test with:

    $ touch schema/.defs-image.json.swp
    $ touch schema/.content-descriptor.json.un~
    $ touch schema/foojson
    $ make schema-fs
    $ grep '"/' schema/fs.go
            "/config-schema.json": {
            "/content-descriptor.json": {
            "/defs-config.json": {
            "/defs-image.json": {
            "/defs.json": {
            "/image-layout-schema.json": {
            "/image-manifest-schema.json": {
            "/manifest-list-schema.json": {
            "/": {
                    local: "/",

CC @vbatts.

[1]: https://github.com/opencontainers/image-spec/pull/533#discussion_r98296739

PR #605, opened 2017-03-09
schema/fs: Drop *.go files

wking:
With:

    $ go get -u github.com/mjibson/esc
    $ make schema/fs.go

now that #538 has landed.


vbatts:
while i like this idea, this also means the code can not be `go get`ten and vendor'ed, right?

wking:
On Thu, Mar 09, 2017 at 11:07:37AM -0800, Vincent Batts wrote:
> while i like this idea, this also means the code can not be `go
> get`ten and vendor'ed, right?

This has no impact on ‘go get’, since all the schema/*.go is still
there.  All I'm doing is removing *.go which had been bundled into
fs.go.  All we use fs.go for is the *.json files, and those are
still there [1].

[1]: https://github.com/wking/image-spec/blob/5e79bb5ba4c9ab85f6e115372cb9608750dceab4/schema/fs.go#L205

vbatts:
k. Prolly should add it to `.gitignore` too

wking:
> Prolly should add it to .gitignore too

?  `fs.go` is still in version control after this commit.  I an only removing some unused content from `fs.go`.  What would I add to `.gitignore`.

vbatts:
i misunderstood. I thought (hoped?) you meant that schema/fs.go is removed and only generated as needed.

wking:
> ...only generated as needed

Well, only generated when `.json` changes (#538) should help a bit ;).

jonboulle:
LGTM

> I thought (hoped?) you meant that schema/fs.go is removed and only generated as needed.

we could try this elsewhere but this change is fine for now

vbatts:
LGTM

PR #608, opened 2017-03-10
config: minor cleanup

jonboulle:
The DiffID exploration doesn't parse as cleanly as it could to me and I'd like
to work on improving it. Starting with some hopefully uncontroversial tweaks.

### DIFF
--- a/config.md
+++ b/config.md
@@ -26,7 +26,7 @@ Changing it means creating a new derived image, instead of changing the existing
 ### Layer DiffID
 
 A layer DiffID is the digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., `sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9`.
-Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using tar-split to save the tar headers.
+Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using [tar-split][] to save the tar headers.
 
 NOTE: Do not confuse DiffIDs with [layer digests](manifest.md#image-manifest-property-descriptions), often referenced in the manifest, which are digests over compressed or uncompressed content.
 
@@ -276,3 +276,4 @@ Here is an example image configuration JSON document:
 
 [rfc3339-s5.6]: https://tools.ietf.org/html/rfc3339#section-5.6
 [runtime-platform]: https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config.md#platform
+[tar-split]: https://github.com/vbatts/tar-split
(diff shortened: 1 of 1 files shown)


wking:
[Missing signoffs][1]?

[1]: https://travis-ci.org/opencontainers/image-spec/builds/209703563#L346

config.md
```diff
@@ -26,7 +26,7 @@ Changing it means creating a new derived image, instead of changing the existing
 ### Layer DiffID
 
 A layer DiffID is the digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., `sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9`.
-Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using tar-split to save the tar headers.
+Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using [tar-split][tar-split] to save the tar headers.
```
wking: nit: You can just use `[tar-split][]` to take advantage of the [implicit link name shortcut][1].  Removing the redundant link name makes this line slightly easier to read, but it's obviously not a big deal either way.

[1]: https://daringfireball.net/projects/markdown/syntax.php#link

jonboulle: updated

config.md
```diff
@@ -26,7 +26,7 @@ Changing it means creating a new derived image, instead of changing the existing
 ### Layer DiffID
 
 A layer DiffID is the digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., `sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9`.
-Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using tar-split to save the tar headers.
+Layers must be packed and unpacked reproducibly to avoid changing the layer DiffID, for example by using [tar-split][] to save the tar headers.
```
vbatts: does that empty bracket work?

vbatts: i guess so

vbatts:
LGTM

config.md
jonboulle: https://github.com/opencontainers/image-spec/pull/608#discussion_r105465003

stevvooe:
LGTM

Need to address https://github.com/opencontainers/image-spec/pull/608#discussion_r105802346 separately. This should not be a must.

config.md
jonboulle: I suppose (?) as written it's not supposed to be an imperative must (i.e. "to avoid changing the headers, it is necessary for layers to be packed reproducibly" rather than "implementers must pack reproducibly") but I think we should probably standardise on RFC2119 language to be clearer. #611

PR #609, opened 2017-03-10
descriptor: recognise sha512, tweak wording

jonboulle:
As previously discussed in #589, implementations may wish to use SHA-512 for
performance/security reasons, and since our endorsed digest package also
supports it, let's add the algorithm identifier to the recognised ones in the
spec.

At the same time, we keep wording to clarify (and indeed mandate) that SHA-256
MUST be supported by implementations and is preferred for interoperability
reasons.

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -59,10 +59,10 @@ Extended _Descriptor_ field additions proposed in other OCI specifications SHOUL
 
 The _digest_ property of a Descriptor acts as a content identifier, enabling [content addressability](http://en.wikipedia.org/wiki/Content-addressable_storage).
 It uniquely identifies content by taking a [collision-resistant hash](https://en.wikipedia.org/wiki/Cryptographic_hash_function) of the bytes.
-If the identifier can be communicated in a secure manner, one can retrieve the content from an insecure source, calculate the digest independently, and be certain that the correct content was obtained.
+If the digest can be communicated in a secure manner, one can retrieve the content from an insecure source, recalculate the digest independently, and be certain that the correct content was obtained.
 
-The value of the digest property, the _digest string_, is a serialized hash result, consisting of an _algorithm_ portion and a _hex_ portion.
-The algorithm identifies the methodology used to calculate the digest; the hex portion is the lowercase hex-encoded result of the hash.
+The value of the digest property is a string consisting of an _algorithm_ portion (the "algorithm identifier") and a _hex_ portion.
+The algorithm identifier specifies the cryptographic hash function used to calculate the digest; the hex portion is the lowercase hex-encoded result of the hash.
 
 The digest string MUST match the following grammar:
 
@@ -74,20 +74,17 @@ hex         := /[a-f0-9]+/
 
 Some example digest strings include the following:
 
-digest                                                                  | algorithm           |
+digest string                                                           | algorithm           |
 ------------------------------------------------------------------------|---------------------|
 sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b | [SHA-256](#sha-256) |
 
-* Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the digest.
+* Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the digest string.
 * Before calculating the digest, the size of the content SHOULD be verified to reduce hash collision space.
 * Heavy processing before calculating a hash SHOULD be avoided.
-* Implementations MAY employ some canonicalization of the underlying content to ensure stable content identifiers.
+* Implementations MAY employ [canonicalization](canonicalization.md) of the underlying content to ensure stable content identifiers.
 
-### Algorithms
+### Digest calculations
 
-While the _algorithm_ component of the digest does allow one to utilize a wide variety of algorithms, compliant implementations SHOULD use [SHA-256](#sha-256).
-
-Let's use a simple example in pseudo-code to demonstrate a digest calculation:
 A _digest_ is calculated by the following pseudo-code, where `H` is the selected hash algorithm, identified by string `<alg>`:
 ```
 let ID(C) = Descriptor.digest
@@ -97,7 +94,7 @@ let verified = ID(C) == D
 ```
 Above, we define the content identifier as `ID(C)`, extracted from the `Descriptor.digest` field.
 Content `C` is a string of bytes.
-Function `H` returns the hash of `C` in bytes and is passed to function `EncodeHex` to obtain the _digest_.
+Function `H` returns the hash of `C` in bytes and is passed to function `EncodeHex` and prefixed with the algorithm to obtain the digest.
 The result `verified` is true if `ID(C)` is equal to `D`, confirming that `C` is the content identified by `D`.
 After verification, the following is true:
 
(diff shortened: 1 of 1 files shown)


descriptor.md
```diff
@@ -61,10 +61,10 @@ Extended _Descriptor_ field additions proposed in other OCI specifications SHOUL
 
 The _digest_ property of a Descriptor acts as a content identifier, enabling [content addressability](http://en.wikipedia.org/wiki/Content-addressable_storage).
 It uniquely identifies content by taking a [collision-resistant hash](https://en.wikipedia.org/wiki/Cryptographic_hash_function) of the bytes.
-If the identifier can be communicated in a secure manner, one can retrieve the content from an insecure source, calculate the digest independently, and be certain that the correct content was obtained.
+If the digest can be communicated in a secure manner, one can retrieve the content from an insecure source, recalculate the digest independently, and be certain that the correct content was obtained.
 
-The value of the digest property, the _digest string_, is a serialized hash result, consisting of an _algorithm_ portion and a _hex_ portion.
-The algorithm identifies the methodology used to calculate the digest; the hex portion is the lowercase hex-encoded result of the hash.
+The value of the digest property, the _digest string_, is a serialized hash result, consisting of an _algorithm_ portion (the "algorithm identifier") and a _hex_ portion (the digest).
```
wking: I would prefer we refer to the whole string as the “digest” and the part after the semicolon as the “hash” or the “hex” to match our [definition][1] and [implementation][2].  Calling the hash/hex portion the “digest” if confusing if the whole thing is so similar (the “digest string”) and is represented by the `Digest` type.

[1]: https://github.com/opencontainers/image-spec/blame/37f2a1b20c1f59e9a7dcdbf7a9fe50144912960c/descriptor.md#L72-L74
[2]: https://github.com/opencontainers/go-digest/blob/v1.0.0-rc0/digest.go#L30-L36

jonboulle: It's hard to find consistency in the terminology use in this area, but I agree this can be improved; see  23af8344414c

vbatts:
LGTM

descriptor.md
```diff
@@ -76,20 +76,17 @@ hex         := /[a-f0-9]+/
 
 Some example digest strings include the following:
 
-digest                                                                  | algorithm           |
+digest string                                                           | algorithm           |
```
wking: With 23af834 moving us towards consistently using [digest/algorithm (identifier)/hex][1] matching our ABNF, I think we may want to stick to “digest” instead of using “digest string”.  I don't mind if existing instances of “digest string” are changed to “digest” in this PR or not, but I think this line (and the “Before consuming…” line below) should be left alone in this PR.

[1]: https://github.com/opencontainers/image-spec/blame/52d4836391a210e5af79afd70cadd8643829f325/descriptor.md#L72

jonboulle: It's fine like this in context. If you want then submit a follow up that changes those AND the preceding two references.

wking:
> @jbouzane [approved these changes an hour ago][1]

This is [another][2] ignored-by-PullApprove approval ;).  Did you want to LGTM this PR?

[1]: https://github.com/opencontainers/image-spec/pull/609#pullrequestreview-27199819
[2]: https://github.com/opencontainers/image-spec/pull/567#issuecomment-285215840

vbatts:
needs a rebase

jonboulle:
rebased

vbatts:
LGTM

jonboulle:
bump (before something happens that requires another rebase!)

philips:
LGTM

stevvooe:
Would have liked to have reviewed this one before going forward...

descriptor.md
```diff
@@ -59,10 +59,10 @@ Extended _Descriptor_ field additions proposed in other OCI specifications SHOUL
 
 The _digest_ property of a Descriptor acts as a content identifier, enabling [content addressability](http://en.wikipedia.org/wiki/Content-addressable_storage).
 It uniquely identifies content by taking a [collision-resistant hash](https://en.wikipedia.org/wiki/Cryptographic_hash_function) of the bytes.
-If the identifier can be communicated in a secure manner, one can retrieve the content from an insecure source, calculate the digest independently, and be certain that the correct content was obtained.
+If the digest can be communicated in a secure manner, one can retrieve the content from an insecure source, recalculate the digest independently, and be certain that the correct content was obtained.
```
stevvooe: Changing to `digest` here is incorrect. We are trying to communicate the concept of calculating a common identifier.

descriptor.md
stevvooe: This is just referred to as a "digest". String here is redundant.

descriptor.md
```diff
@@ -74,20 +74,17 @@ hex         := /[a-f0-9]+/
...
+digest string                                                           | algorithm           |
 ------------------------------------------------------------------------|---------------------|
 sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b | [SHA-256](#sha-256) |
 
-* Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the digest.
+* Before consuming content targeted by a descriptor from untrusted sources, the byte content SHOULD be verified against the digest string.
 * Before calculating the digest, the size of the content SHOULD be verified to reduce hash collision space.
 * Heavy processing before calculating a hash SHOULD be avoided.
-* Implementations MAY employ some canonicalization of the underlying content to ensure stable content identifiers.
+* Implementations MAY employ [canonicalization](canonicalization.md) of the underlying content to ensure stable content identifiers.
```
stevvooe: This implies that the canonicalization is limited to those described in the document. Implementations may employ any kind of canonicalization they want in the generation of content.

descriptor.md
```diff
@@ -74,20 +74,17 @@ hex         := /[a-f0-9]+/
...
 
-### Algorithms
+### Digest calculations
 
-While the _algorithm_ component of the digest does allow one to utilize a wide variety of algorithms, compliant implementations SHOULD use [SHA-256](#sha-256).
```
stevvooe: > While the _algorithm_ component of the digest does allow one to utilize a wide variety of algorithms, compliant implementations SHOULD use those specified here.

descriptor.md
```diff
@@ -107,21 +104,29 @@ D == ID(C) == '<alg>:' + EncodeHex(H(C))
...
 
 #### SHA-256
 
-[SHA-256](https://tools.ietf.org/html/rfc4634#page-7) is a collision-resistant hash function, chosen for ubiquity, reasonable size and secure characteristics.
+[SHA-256][rfc4634-s4.1] is a collision-resistant hash function, chosen for ubiquity, reasonable size and secure characteristics.
 Implementations MUST implement SHA-256 digest verification for use in descriptors.
 
+#### SHA-512
+
+[SHA-512][rfc4634-s4.2] is a collision-resistant hash function which [may be more perfomant][sha256-vs-sha512] than [SHA-256](#sha-256) on some CPUs.
```
stevvooe: This implies that performance is a good reason to select one digest algorithm over another, when that is not the case at all. The most important factors in the selection of a digest algorithm is that it is common to all implementations.

In practice, the use of `sha512` will likely cause the introduction of incompatible images.

stephenrwalli:
This request is a late change as you try to close the Image Spec. 
It is defined in implementation specific language about performance. 
This essentially looks like folks trying to ensure their implementation specific change is allowed by the specification, while possibly preventing other implementation changes in the space. 
It's a product versus specification timing issue. 

If you are focused on finishing a spec I would suggest you either:
1. Try to keep it pure. Mandate the simple. Say nothing about other possible implementations.
2. Accept this is a space with a lot of change still possible. Mandate the simple. Specifically call out that implementations MAY do other things, thereby warning implementers and users alike that this space can't be guaranteed.

descriptor.md
stevvooe: Nevermind, this moved down.

vbatts:
@stephenrwalli I get calling out that implementations MAY choose others, but this is not a late change. Calling out a hash like sha512 has been a couple year lingering conversation. Besides sha256 isn't immediately going away. This is not a product vs. spec issue.

PR #610, opened 2017-03-13
define media type for oci-layout

xiekeyang:
Following #579.

To add a new specific media type for oci-layout, serving over its transportation.

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -133,7 +133,7 @@ This section defines the `application/vnd.oci.layout.header.v1+json` [media type
 
 ### oci-layout Example
 
-```json,title=OCI%20Layout&mediatype=oci%2Dlayout
+```json,title=OCI%20Layout&mediatype=application/vnd.oci.layout.header.v1%2Bjson
 {
     "imageLayoutVersion": "1.0.0"
 }
--- a/image-layout.md
+++ b/image-layout.md
@@ -129,6 +129,7 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 
 This JSON object serves as a marker for the base of an Open Container Image Layout and to provide the version of the image-layout in use.
 The `imageLayoutVersion` value will align with the OCI Image Specification version at the time changes to the layout are made, and will pin a given version until changes to the image layout are required.
+This section defines the `application/vnd.oci.layout.header.v1+json` [media type](media-types.md).
 
 ### oci-layout Example
 
--- a/.tool/lint
+++ b/.tool/lint
@@ -20,5 +20,5 @@ for d in $(find . -type d -not -iwholename '*.git*' -a -not -iname '.tool' -a -n
 		--disable=gas \
 		--cyclo-over=35 \
 		--tests \
-		--deadline=10s "${d}"
+		--deadline=15s "${d}"
 done
--- a/media-types.md
+++ b/media-types.md
@@ -3,6 +3,7 @@
 The following media types identify the formats described here and their referenced resources:
 
 - `application/vnd.oci.descriptor.v1+json`: [Content Descriptor](descriptor.md)
+- `application/vnd.oci.layout.header.v1+json`: [OCI Layout](image-layout.md#oci-layout-file)
 - `application/vnd.oci.image.index.v1+json`: [Image Index](image-index.md)
 - `application/vnd.oci.image.manifest.v1+json`: [Image manifest](manifest.md#image-manifest)
 - `application/vnd.oci.image.config.v1+json`: [Image config](config.md)
--- a/schema/schema.go
+++ b/schema/schema.go
@@ -22,12 +22,12 @@ import (
 
 // Media types for the OCI image formats
 const (
-	ValidatorMediaTypeDescriptor  Validator     = v1.MediaTypeDescriptor
-	ValidatorMediaTypeManifest    Validator     = v1.MediaTypeImageManifest
-	ValidatorMediaTypeImageIndex  Validator     = v1.MediaTypeImageIndex
-	ValidatorMediaTypeImageConfig Validator     = v1.MediaTypeImageConfig
-	ValidatorTypeImageLayout      Validator     = v1.ImageLayoutFile
-	ValidatorMediaTypeImageLayer  unimplemented = v1.MediaTypeImageLayer
+	ValidatorMediaTypeDescriptor   Validator     = v1.MediaTypeDescriptor
+	ValidatorMediaTypeLayoutHeader Validator     = v1.MediaTypeLayoutHeader
+	ValidatorMediaTypeManifest     Validator     = v1.MediaTypeImageManifest
+	ValidatorMediaTypeImageIndex   Validator     = v1.MediaTypeImageIndex
+	ValidatorMediaTypeImageConfig  Validator     = v1.MediaTypeImageConfig
+	ValidatorMediaTypeImageLayer   unimplemented = v1.MediaTypeImageLayer
 )
 
 var (
@@ -37,11 +37,11 @@ var (
 
 	// specs maps OCI schema media types to schema files.
 	specs = map[Validator]string{
-		ValidatorMediaTypeDescriptor:  "content-descriptor.json",
-		ValidatorMediaTypeManifest:    "image-manifest-schema.json",
-		ValidatorMediaTypeImageIndex:  "image-index-schema.json",
-		ValidatorMediaTypeImageConfig: "config-schema.json",
-		ValidatorTypeImageLayout:      "image-layout-schema.json",
+		ValidatorMediaTypeDescriptor:   "content-descriptor.json",
+		ValidatorMediaTypeLayoutHeader: "image-layout-schema.json",
+		ValidatorMediaTypeManifest:     "image-manifest-schema.json",
+		ValidatorMediaTypeImageIndex:   "image-index-schema.json",
+		ValidatorMediaTypeImageConfig:  "config-schema.json",
 	}
 )
 
--- a/specs-go/v1/mediatype.go
+++ b/specs-go/v1/mediatype.go
@@ -18,6 +18,9 @@ const (
 	// MediaTypeDescriptor specifies the media type for a content descriptor.
 	MediaTypeDescriptor = "application/vnd.oci.descriptor.v1+json"
 
+	// MediaTypeLayoutHeader specifies the media type for the oci-layout.
+	MediaTypeLayoutHeader = "application/vnd.oci.layout.header.v1+json"
+
 	// MediaTypeImageManifest specifies the media type for an image manifest.
 	MediaTypeImageManifest = "application/vnd.oci.image.manifest.v1+json"
 


vbatts:
or should we just give it a mediaType, @stevvooe ?

jonboulle:
That seems simpler than going through these contortions.

stevvooe:
@vbatts @jonboulle I agree that is simpler.

We do need to be careful about contexts under which a mediatype is encountered. For example, I would never expect to encounter this in the course of arbitrary descriptors, so we may want to add some clarification that you only need to handle these when it is an oci layout file.

xiekeyang:
@vbatts @jonboulle @stevvooe 

I change the clarification on https://github.com/opencontainers/image-spec/pull/610/commits/d678b9c26401a0b808836d410583f82e721d04b5

I prefer to naming the file as `oci-layout`, instead of providing a media type. In layout use case of a variety of different transport mechanisms, we check other JSON objects through media type because they are CAS blobs, but image-layout is always on the top of image dir, so that users can address it through name of `oci-layout`. And I think it might be applicable on local or repository endpoint.

If it is reasonable?

image-layout.md
```diff
@@ -20,6 +22,7 @@ The image layout is as follows:
     - See [blobs](#blobs) section
 - `oci-layout` file
     - It MUST exist
+    - Its name MUST be `oci-layout`
```
wking: I think this is already covered by the earlier `- oci-layout file` line.  Note that we have no `Its name MUST be blobs` entry under `- blobs directory` above (for the same reason).

image-layout.md
```diff
@@ -2,6 +2,8 @@
 
 The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
 This layout MAY be used in a variety of different transport mechanisms: archive formats (e.g. tar, zip), shared filesystem environments (e.g. nfs), or networked file fetching (e.g. http, ftp, rsync).
+This section defines none of media type for image layout.
```
jonboulle: is this necessary? its inclusion just confuses me. a layout as described here couldn't really have a media type

image-layout.md
jonboulle: +1

xiekeyang:
@jonboulle @wking 
I had wanted to present 3 supplements according to previous confusion and discussion.
1. `oci-layout` is unchangeable. (It seems redundant to add any comment, so that I removed it)
2. This section defines none of media type. I think it might make sense, for this JSON object has seem to confuse developers try to find its media type. Then I want to clear it here.
3. How image provider or consumer to address/handle `oci-layout`, this seems to be acceptable.

xiekeyang:
@jonboulle @wking These clarification might be not graceful, could you suggest some better words? Thanks.

wking:
On Tue, Mar 14, 2017 at 06:50:47PM -0700, xiekeyang wrote:
> @jonboulle @wking These clarification might be not graceful, could
> you suggest some better words? Thanks.

I don't think it's a good idea to declare that there is *no* media
type for oci-layout.  If you're serving an image-layout over HTTP,
you'll want to put something in the Content-Type header for that file.
If we don't give it a more specific media type, then it will default
to being application/json [1].  So there are two options for the type:

a. Declare a specific media type for the oci-layout file
   (e.g. application/vnd.oci.image.layout.v1+json).
b. Use application/json.

With (a) we can continue to use our current media-type based spec
validation [2,3] without trouble.  With (b), we have two validation
options:

i. Use a dummy “media type” like oci-layout [4,5].
ii. Add a new field (independent of or supporting the media type) to
    select which schema should be used to validate the content
    (e.g. json,title=OCI%20Layout&mediatype=application/json&schema=oci%2Dlayout).

I prefer (a), (b.ii) is ok, and (b.i) (the route taken in #519, #579, and
this PR as of c0362a34) just seems confusing to me.

So if you want to make it clear that we are declaring no special media
type for this file (b), I'd suggest a line like:

  This specification does not declare a specific media type for
  `oci-layout`, so the media type for `oci-layout` is the generic
  [`application/json`][1].

But if we go that route, I'd strongly recommend we stop pretending to
be using media types to select the appropriate schema for validation
[7].  And that we do clearly document the identifiers that we do use
[3], whatever we end up calling them (“schema identifiers”?).

[1]: https://tools.ietf.org/html/rfc7159#section-1.2
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/schema/spec_test.go#L80
[3]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/schema/schema.go#L23-L30
[4]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/schema/schema.go#L29
[5]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/specs-go/v1/layout.go#L19
[6]: https://github.com/opencontainers/image-spec/pull/579/files#diff-99bfcb610f238b942638d49d1ff2eaf4R137
[7]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/schema/schema.go#L23

xiekeyang:
@wking I'm not opposing specific oci-layout media type strongly. I'm just considering https://github.com/opencontainers/image-spec/pull/610#issuecomment-286373425 . 
Spec declares that `This layout MAY be used in a variety of different transport mechanisms`. In this case if it really need media type? I suspect it. For validation purpose, I have said some times that we should improve validation code, instead set up a media type. (Maybe in future we will need specific media type, I'm not sure.)

So I like the language:
```
This specification does not declare a specific media type for `oci-layout`, so it is the generic [`application/json`](https://tools.ietf.org/html/rfc7159#section-1.2).
```

This patch I would like to achieve in spec, then I will follow up patch on validation part for this.

xiekeyang:
Updated in https://github.com/opencontainers/image-spec/pull/610/commits/70d7a61a6d750e4b8174eb3c3b27552196a57c79

jonboulle:
I'm lost. I thought we established we would just add the media type.

xiekeyang:
According to most people's suggestion, this add specific media type for image layout in https://github.com/opencontainers/image-spec/pull/610/commits/b7e8d5c0a91fb8c2e83f5208ed16d9fbe876d476 

After that, image layout key word in schema should be updated accordingly, which I fix in https://github.com/opencontainers/image-spec/pull/610/commits/b8e832dc319c6aafdb60dfba12f806fffb582c68 . Here I didn't remove `ImageLayoutFile` even it is not used currently. That is because the name (oci-layout) will be use by image-tools or customers when they pack the OCI image.

During making lint I encounter failure. It not caused the above commit, but failure under any commit. So I've to extent the deadline of lint fro 10s to 15s, in https://github.com/opencontainers/image-spec/pull/610/commits/eb5fcf82b0f98d1365523fb914dab0735e2af5c7 .

@opencontainers/image-spec-maintainers PTAL, thank so much!

image-layout.md
```diff
@@ -2,6 +2,7 @@
 
 * The OCI Image Layout is a slash separated layout of OCI content-addressable blobs and [location-addressable](https://en.wikipedia.org/wiki/Content-addressable_storage#Content-addressed_vs._location-addressed) references (refs).
 * This layout MAY be used in a variety of different transport mechanisms: archive formats (e.g. tar, zip), shared filesystem environments (e.g. nfs), or networked file fetching (e.g. http, ftp, rsync).
+* This section defines the `application/vnd.oci.image.layout.v1+json` [media type](media-types.md).
```
jonboulle: I think this line should be moved down [here](https://github.com/opencontainers/image-spec/pull/610/files#diff-99bfcb610f238b942638d49d1ff2eaf4R129) to the oci-layout-file section. Otherwise it seems to imply that the entire document is talking about this media type.

xiekeyang: > I think this line should be moved down here to the oci-layout-file section. Otherwise it seems to imply that the entire document is talking about this media type.

I'm a little doubting on it. It seems had better to be aligned with other spec docs (manifest, index, config..), in which `define ... media type` is put on the first section.

jonboulle: I see what you mean, but would also argue that it's more appropriate in those other documents because there the bulk of the text is describing the type, whereas in this case it's only a small part of the document. But ¯\_(ツ)_/¯ , other maintainers can weigh in

stevvooe: Also, if we created a mediatype for the layout archive, it would probably be this, so this is very confusing.

stevvooe:
Media type needs to be registered in https://github.com/opencontainers/image-spec/blob/master/media-types.md, as well.

xiekeyang:
@stevvooe 
> Media type needs to be registered in https://github.com/opencontainers/image-spec/blob/master/media-types.md, as well.

I've registered it in https://github.com/opencontainers/image-spec/pull/610/commits/b7e8d5c0a91fb8c2e83f5208ed16d9fbe876d476#diff-691102624fa3cbd2452c05ada4a3f91cR6 . Do you mean that way?

xiekeyang:
@opencontainers/image-spec-maintainers  I rename the oci-layout media type to `application/vnd.oci.layout.version.v1+json`. That express precisely the oci-layout itself, to avoid confusing customers. PTAL, thanks. Related changeset is done in `media-type.md`, `schema` and `spec/v1`.

stevvooe:
@xiekeyang Thanks! `application/vnd.oci.layout.version.v1+json` is much clearer. Perhaps, `pplication/vnd.oci.layout.header.v1+json` might be even more accurate?

@jonboulle @vbatts Thoughts?

xiekeyang:
> Perhaps, pplication/vnd.oci.layout.header.v1+json might be even more accurate?

I like it! related name is updated, PTAL.

jonboulle:
sure, this is fine with me.

jonboulle:
lgtm

image-layout.md
```diff
@@ -129,10 +129,11 @@ $ cat ./blobs/sha256/e692418e4cbaf90ca69d05a66403747baa33ee08806650b51fab815ad7f
 
 This JSON object serves as a marker for the base of an Open Container Image Layout and to provide the version of the image-layout in use.
 The `imageLayoutVersion` value will align with the OCI Image Specification version at the time changes to the layout are made, and will pin a given version until changes to the image layout are required.
+This section defines the `application/vnd.oci.layout.header.v1+json` [media type](media-types.md).
 
 ### oci-layout Example
 
-```json,title=OCI%20Layout&mediatype=oci%2Dlayout
+```json,title=OCI%20Layout&mediatype=application/vnd.oci.layout.header.v1%2Bjson
```
vbatts: will the `TestValidateImageLayout` in `./schema/spec_test.go` test this now?

wking: We have been testing this example since #579:

```
$ git describe
v1.0.0-rc5-68-gf59a917
$ sed -i 's/imageLayoutVersion/iLV/' image-layout.md
$ go test ./schema/spec_test.go
{
    "iLV": "1.0.0"
}
 ---
--- FAIL: TestValidateImageLayout (0.00s)
        spec_test.go:190: invalid       oci-layout      OCI Layout
        spec_test.go:100: imageLayoutVersion: imageLayoutVersion is required
…
```

What this PR does is give us a better identifier for this validator (`application/vnd.oci.layout.header.v1+json`) than the current one (`oci-layout`).

vbatts: right on

vbatts:
LGTM

PR #611, opened 2017-03-14
config: change reproducibility "MUST" to "SHOULD"

jonboulle:
As noted in
https://github.com/opencontainers/image-spec/pull/608#discussion_r105802346 , we shouldn't be mandating that implementers use something like
tar-split to facilitate reproducible DiffIDs, but rather recommend it.

### DIFF
--- a/config.md
+++ b/config.md
@@ -26,7 +26,7 @@ Changing it means creating a new derived image, instead of changing the existing
 ### Layer DiffID
 
 A layer DiffID is the digest over the layer's uncompressed tar archive and serialized in the descriptor digest format, e.g., `sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9`.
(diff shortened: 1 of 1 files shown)


vbatts:
fair, though it doesn't say they _must_ use tar-split, it is just an example right?

and what's this even saying is that implied by having a tar archive that matches checksum, so how that is achieved is up to the implementer, right?

jonboulle:
@vbatts see my comment at https://github.com/opencontainers/image-spec/pull/608#discussion_r105804305 , this is probably just the ambiguity of natural language. Nevertheless seems cleaner if we can standardise on the RFC2119-style

stevvooe:
LGTM

> fair, though it doesn't say they must use tar-split, it is just an example right?

Agreed.

@vbatts We could probably relax this altogether. In fact, if the _exact_ `DiffID` is not reproduced, everything will still work.

vbatts:
LGTM

PR #614, opened 2017-03-14
s/should/SHOULD/ in a few spots, also a s/should/must/

erikh:
The capslock changes speak for themselves. I tried to avoid situations where deliberately capitalizing SHOULD were too prescriptive.

As for the s/should/MUST/, I think it's appropriate here as there will never be a situation where aborting on invalid keys is a thing that is allowed. I hope you agree that this is an appropriate change.

Thanks!


jonboulle:
LGTM overall, but the changes to config.md are really crying out for something like #492 to land, so we can clarify the image<->runtime relationship.

vbatts:
LGTM

Issue #588, opened 2017-03-01
index.json: improve implementors note to better represent algorithms necessary


cyphar:
As discussed in the OCI call, currently the implementors note for `index.json` doesn't fully encapsulate the reason for the more complex `Index` data structure and the ways in which implementations should handle the `n`-tall tree that it can represent.

The main points that need to be clarified are:

* That the UX for selecting what manifest is being operated on is _not_ defined by the spec, and that implementations are free to either fail when they can't uniquely resolve a manifest, or provide some sort of indication to the user that 

* The platform matching / annotation matching algorithm necessary to match and flatten the tree structure of descriptors and then reduce the set of descriptors to those that match a particular query. The SQL analogy would probably be something we should mention.

@stevvooe Stepped up to work on this one.

stevvooe:
Based on my research over the last week, I think we can address this by pulling up the platform descriptions into the descriptor to get rid of the type hierarchy of descriptors which is awkward. This is a Go type change and a slight reorganization of the types definition which is completely compatible, but it would be nice to have that correct for 1.0.

Let me know what you think about this or if more details are needed.

cyphar:
I agree with moving platform to descriptor annotations (now that we have them). Though we still would need to add some implementor notes about how they are expected to handle the hierarchy of annotations that you could run into and that the recommended referencing implementation would involve effectively tag matching (more of a search engine and less of a simple dereferencer).

I've been working on this on the `umoci` side and I'll admit it's quite difficult coming up with an interface that is symmetric when it comes to referencing "from" and "to" an image (you are effectively binding the interface quite tightly with the concept of annotations which is a spec detail that I'm not sure users should care about). But we'll see how it goes.

stevvooe:
> I agree with moving platform to descriptor annotations (now that we have them). 

I think we'd have to support both platform field and annotations, in practice, but the simplification seems straightforward.

> I've been working on this on the umoci side and I'll admit it's quite difficult coming up with an interface that is symmetric when it comes to referencing "from" and "to" an image (you are effectively binding the interface quite tightly with the concept of annotations which is a spec detail that I'm not sure users should care about). But we'll see how it goes.

I'll see if I can look deeper into platform qualified pull to give some better guidance here, as promised.

qianzhangxa:
@stevvooe Did you mean adding `Platform` into `Descriptor` as its field and then we do not need `ManifestDescriptor` anymore?

[...]

PR #620, opened 2017-03-20
specs-go/v1: unify the descriptor type

### LINKED ISSUE #618: merge "ManifestDescriptor" and "Descriptor" types
Currently, we have `ManifestDescriptor` and `Descriptor` types. The only difference is the inclusion of the platform field. For making generic processors that filter based on platform type, this creates a number of complexities (see https://github.com/docker/containerd/pull/638 for an example of this pattern).

We'll simply need to merge these types in the specification and struct definitions.

stevvooe:
Unifying the descriptor type makes it much easier to handler descriptors
in generic manner. This change ensures that a single deserialization can
be done a descriptor that may be processed in separate contexts.

It is unclear whether or not the specification should be pedantic about
this, as well. As it is written, these fields are only valid when being
referred through a manifest, but this distinction is only required
during validation. It may be possible to lift the platform definition to
the descriptor type, but this isn't strictly necessary.

Addresses #588.
Closes #618.

### DIFF
--- a/specs-go/v1/descriptor.go
+++ b/specs-go/v1/descriptor.go
@@ -17,7 +17,8 @@ package v1
 import digest "github.com/opencontainers/go-digest"
 
 // Descriptor describes the disposition of targeted content.
-// This structure provides `application/vnd.oci.descriptor.v1+json` mediatype when marshalled to JSON
+// This structure provides `application/vnd.oci.descriptor.v1+json` mediatype
+// when marshalled to JSON.
 type Descriptor struct {
 	// MediaType is the media type of the object this schema refers to.
 	MediaType string `json:"mediaType,omitempty"`
@@ -33,4 +34,35 @@ type Descriptor struct {
 
 	// Annotations contains arbitrary metadata relating to the targeted content.
 	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Platform describes the platform which the image in the manifest runs on.
+	//
+	// This should only be used when referring to a manifest.
+	Platform *Platform `json:"platform,omitempty"`
+}
+
+// Platform describes the platform which the image in the manifest runs on.
+type Platform struct {
+	// Architecture field specifies the CPU architecture, for example
+	// `amd64` or `ppc64`.
+	Architecture string `json:"architecture"`
+
+	// OS specifies the operating system, for example `linux` or `windows`.
+	OS string `json:"os"`
+
+	// OSVersion is an optional field specifying the operating system
+	// version, for example `10.0.10586`.
+	OSVersion string `json:"os.version,omitempty"`
+
+	// OSFeatures is an optional field specifying an array of strings,
+	// each listing a required OS feature (for example on Windows `win32k`).
+	OSFeatures []string `json:"os.features,omitempty"`
+
+	// Variant is an optional field specifying a variant of the CPU, for
+	// example `ppc64le` to specify a little-endian version of a PowerPC CPU.
+	Variant string `json:"variant,omitempty"`
+
+	// Features is an optional field specifying an array of strings, each
+	// listing a required CPU feature (for example `sse4` or `aes`).
+	Features []string `json:"features,omitempty"`
 }
--- a/specs-go/v1/index.go
+++ b/specs-go/v1/index.go
@@ -16,47 +16,13 @@ package v1
 
 import "github.com/opencontainers/image-spec/specs-go"
 
-// Platform describes the platform which the image in the manifest runs on.
-type Platform struct {
-	// Architecture field specifies the CPU architecture, for example
-	// `amd64` or `ppc64`.
-	Architecture string `json:"architecture"`
-
-	// OS specifies the operating system, for example `linux` or `windows`.
-	OS string `json:"os"`
-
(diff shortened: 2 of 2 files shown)


erikh:
Would it make sense to make this a well-known annotation instead? Seems like a good use for the field. Would require moving the annotations type to `map[string]interface{}` of course though, which might have consequences I haven't foreseen.

specs-go/v1/descriptor.go
```diff
@@ -33,4 +34,35 @@ type Descriptor struct {
...
+	// OSVersion is an optional field specifying the operating system
+	// version, for example `10.0.10586`.
+	OSVersion string `json:"os.version,omitempty"`
+
+	// OSFeatures is an optional field specifying an array of strings,
+	// each listing a required OS feature (for example on Windows `win32k`).
+	OSFeatures []string `json:"os.features,omitempty"`
+
+	// Variant is an optional field specifying a variant of the CPU, for
+	// example `ppc64le` to specify a little-endian version of a PowerPC CPU.
```
AkihiroSuda: PowerPC -> POWER?

wking:
Reposted from an emailed comment, which GitHub seems to have eaten:

On Mon, Mar 20, 2017 at 07:01:12PM -0700, Erik Hollensbe wrote:
> Would it make sense to make this a well-known annotation instead?

+1 to this.  With platform-ness being just one way to select a given child, and the existing annotation org.opencontainers.image.ref.name being another parallel/complimentary way, I think moving the platform information into annotations makes sense.

@stevvooe, you'd seemed [to be on board with platform annotations earlier][1].  Has something changed?  We can always continue to inject the old platform properties if we want backwards compatibility with existing Docker manifest-list engines, although we may not be able to only use annotations if we require forward compatibility for existing Docker manifest-list blobs.

> Seems like a good use for the field. Would require moving the annotations type to `map[string]interface{}` of course though, which might have consequences I haven't foreseen.

There's some background on string vs. `interface{}` values in #44 if that helps orient you.

[1]: https://github.com/opencontainers/image-spec/issues/581#issuecomment-282883666

specs-go/v1/descriptor.go
stevvooe: This was copy pasted. If you want to change it, please do so in a follow up.

stevvooe:
> Reposted from an emailed comment, which GitHub seems to have eaten

@wking GitHub probably ate this because you are supposed to be blocked. Please do not comment on my pull requests. I have asked you, kindly, more than once.

> With platform-ness being just one way to select a given child, and the existing annotation org.opencontainers.image.ref.name being another parallel/complimentary way, I think moving the platform information into annotations makes sense.

Ref name isn't really binding in the way that platform is. Platform is just pulled up from the underlying config and can be validated. Ref name is a hint that shouldn't be trusted and can be ignored. This PR brings them to the same level so they can be used in filtration logic, but that doesn't point to their equivalence.

> Would it make sense to make this a well-known annotation instead? Seems like a good use for the field. Would require moving the annotations type to map[string]interface{} of course though, which might have consequences I haven't foreseen.

I don't see any reason to break from the structured approach. Having such a large change before 1.0 would be detrimental to the release and not provide value. Furthermore, annotations are not a dumping ground for stuff; they are meant for non-binding metadata. If we can provide structure for something, we absolutely should.

Note that this PR doesn't change the context under which the platform field can be used. It only unifies the Go type. From the perspective of the specification, the `Platform` field can only be used with indexes. Moving platform to annotations or allowing platform descriptions outside of this context are outside the scope of this PR. This PR is only present to make the descriptor type easier to work with.

erikh:
Fair enough.

I hope my query wasn't a bother! I just thought it was a reasonable
question.

-Erik

stevvooe:
> I hope my query wasn't a bother! I just thought it was a reasonable
question.

No, it is a completely reasonable question. I hope the clarification was helpful.

qianzhangxa:
I guess we need another PR to update specification accordingly to make it consistent with the changes in this PR.

vbatts:
LGTM

q384566678:
If such changes, calls to [descriptor](https://github.com/opencontainers/image-spec/blame/master/specs-go/v1/manifest.go#L25) will be no errors in the `manifest`?

specs-go/v1/descriptor.go
stevvooe: @AkihiroSuda What was the PR that updates this? This all gets moved from index.go to descriptor.go, so we need to make sure it gets rebased appropriately.

stevvooe:
@opencontainers/image-spec-maintainers Rebased after the pointer change in to `Platform`. PTAL

vbatts:
LGTM

specs-go/v1/descriptor.go
AkihiroSuda: IIUC you question, this comment about ppc64le was introduced in https://github.com/opencontainers/image-spec/pull/172 . 

@runcom What was the intention to add `ppc64le`? Is there any implementation using this field (suppose not)?
(related: https://github.com/opencontainers/image-spec/pull/632)

jonboulle:
lgtm

PR #623, opened 2017-03-23
Makefile: fix src path for local pandoc builds

vbatts:
found that the the `/` redundant, and makes the path `/header.html` for local pandoc builds.


jonboulle:
LGTM ok mr batts

stevvooe:
LGTM

PR #627, opened 2017-03-29
README: Document the alternate-week meeting times

wking:
Cherry-pick opencontainers/runtime-spec#740 over now that it has landed (and still ~45 minutes before today's meeting starts ;).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# https://tools.ietf.org/html/rfc5545#section-3.1
+*.ics text eol=crlf
--- a/README.md
+++ b/README.md
@@ -73,10 +73,16 @@ When in doubt, start on the [mailing-list](#mailing-list).
 
 ## Weekly Call
 
-The contributors and maintainers of all OCI projects have a weekly meeting Wednesdays at 2:00 PM (USA Pacific).
+The contributors and maintainers of all OCI projects have a weekly meeting on Wednesdays at:
+
+* 8:00 AM (USA Pacific), during [odd weeks][iso-week].
+* 5:00 PM (USA Pacific), during [even weeks][iso-week].
+
+There is an [iCalendar][rfc5545] format for the meetings [here](meeting.ics).
+
 Everyone is welcome to participate via [UberConference web][UberConference] or audio-only: +1-415-968-0849 (no PIN needed).
 An initial agenda will be posted to the [mailing list](#mailing-list) earlier in the week, and everyone is welcome to propose additional topics or suggest other agenda alterations there.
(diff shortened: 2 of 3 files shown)


vbatts:
Cherry pick indeed, it references the runtime-spec ;-)

wking:
On Mon, Apr 03, 2017 at 04:30:32PM -0700, Vincent Batts wrote:
> Cherry pick indeed, it references the runtime-spec ;-)

Oops :p.  I updated the meeting.ics entries with 20673c6 → 3f7694b.
The runtime-spec wiki link stays, because image-spec doesn't have a
wiki (and the current image-spec master also links to the runtime-spec
wiki for minutes).

wking:
On Mon, Apr 03, 2017 at 04:56:56PM -0700, W. Trevor King wrote:
> I updated the meeting.ics entries…

Thinking this over a bit more, I'm not quite sure how we want to
handle the iCalendar file.  PRODID is supposed to be globally unique
[1], and I'm currently using the same PRODID for both runtime-spec and
in this PR.  Do we expect folks to track both of these iCalendar files
at the same time?  Will they want to see a single OCI meeting, or
separate (overlapping) meetings for runtime-spec and image-spec?  I
think the best route forward as long as there's a shared meeting is
keep the ICS in a shared location (project-template?  Outside of
version control on www.opencontainers.org?  In runtime-spec?) and have
other projects link to it there.  If/when we split into image- and
runtime-specific meetings, we can fork that central iCalendar file
into spec-specific ones.  Thoughts?

[1]: https://tools.ietf.org/html/rfc5545#section-3.7.3

vbatts:
closing for now as this is too variadic. perhaps the meetings should just be in a wiki or something more flexible

wking:
On Fri, May 19, 2017 at 10:56:58AM -0700, v1.0.0.batts wrote:
> closing for now as this is too variadic.

Hopefully the timing will settle down again soon.

> perhaps the meetings should just be in a wiki or something more
> flexible

There is already an entry in [1], although that's stale with the 17:00
reference.  However, entries in the image-spec repo itself show
clearly that the image-spec maintainers have approved a given time.
That's harder to establish with an open wiki or an opencontainers.org
entry.

And as it stands, the image-spec master is a stale reference to
“always 14:00 Pacific” [2].  If the image-spec maintainers aren't
interested in keeping that up to date, they should replace the current
reference with a punt to somewhere…

[1]: https://www.opencontainers.org/community
[2]: https://github.com/opencontainers/image-spec/blob/5ebcc54f1725d37e8a8dbdebe1931df46d486736/README.md#weekly-call

PR #629, opened 2017-03-30
config.md: Sentence modification 

q384566678:

### DIFF
--- a/config.md
+++ b/config.md
@@ -59,7 +59,7 @@ Defining `|` as a binary application operator, the root filesystem may be `A|B|C
 While it is implied that `C` is only useful when applied to `A|B`, the identifier `C` is insufficient to identify this result, as we'd have the equality `C = A|B|C`, which isn't true.
 
 The main issue is when we have two definitions of `C`, `C = C` and `C = A|B|C`.
-If this is true (with some handwaving), `C = x|C` where `x = any application` must be true.
+If this is true (with some handwaving), `C = x|C` where `x = any application`.
 This means that if an attacker can define `x`, relying on `C` provides no guarantee that the layers were applied in any order.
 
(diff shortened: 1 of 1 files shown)


config.md
```diff
@@ -59,7 +59,7 @@ Defining `|` as a binary application operator, the root filesystem may be `A|B|C
 While it is implied that `C` is only useful when applied to `A|B`, the identifier `C` is insufficient to identify this result, as we'd have the equality `C = A|B|C`, which isn't true.
 
 The main issue is when we have two definitions of `C`, `C = C` and `C = A|B|C`.
-If this is true (with some handwaving), `C = x|C` where `x = any application` must be true.
+If this is true (with some handwaving), `C = x|C` where `x = any application` MUST be true.
```
wking: This is not a [compliance-requirement][1], so I don't think we need `MUST`.  It's intended to be a mathematical assertion, which has no compliance impact (i.e. there will be no certification testing to check this).  Perhaps we should rephrase the sentence to avoid “must”?  Maybe:

> If this is true (with some handwaving), then `C = x|C` is true where `x = any application`.

That would be the narrow fix, although I'm still partial to #593 as a broader simplification.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/spec.md#notational-conventions

stevvooe:
This isn't a binding specification statement.

config.md
stevvooe: Just `is also true` would work here.

jonboulle: IMO combine the two:

> If this is true (with some handwaving), then C = x|C is true where x = any application.

q384566678: updated.

jonboulle:
LGTM

stevvooe:
LGTM

PR #630, opened 2017-03-31
specs-go/v1/config.go: Capitalize `Labels`

erikh:
This field was not similar to the other fields and causes problems when dealing with using this struct to convert between docker v1 and oci v1 repositories (I cheat a little by doing this, I realize it's not a long term solution).

Still, this seems out of place and I wanted to fix it while I noticed.


erikh:
gah stupid github edit

jonboulle:
LGTM, nice catch!

stevvooe:
LGTM

Issue #626, opened 2017-03-29
specs-go/v1/config: `omitempty` is ineffective on Created


lucab:
https://github.com/opencontainers/image-spec/pull/534 changed `Created` from string to time.Time field. This neutered the `omitempty` property on it, as a missing field is now populated as the default time.Time value (0000-01-01).

Ref: https://github.com/golang/go/issues/11939

/cc @jlbutler as #534 author

jonboulle:
@lucab how about submitting a patch to pointer-ify this one?

lucab:
Ack, up at https://github.com/opencontainers/image-spec/pull/633.

PR #534, opened 2017-01-26
specs-go: Image.Created uses a string type

jlbutler:
Modify History.Created and Image.Created to be of type time.Time so that creation times may be validated, compared, etc.

PR #633, opened 2017-04-03
specs-go/v1/config: pointer-ify optional "Created" field

lucab:
PR #534 changed Created from string to time.Time field.
This neutered the `omitempty` property on it, as a missing
field is now populated as the default time.Time value.

This commit turns the field into a pointer, in order to keep
the empty-value semantic for `omitempty`.

Fixes https://github.com/opencontainers/image-spec/issues/626

Signed-off-by: Luca Bruno <luca.bruno@coreos.com>

### DIFF
--- a/specs-go/v1/config.go
+++ b/specs-go/v1/config.go
@@ -62,7 +62,7 @@ type RootFS struct {
 // History describes the history of a layer.
 type History struct {
 	// Created is the combined date and time at which the layer was created, formatted as defined by RFC 3339, section 5.6.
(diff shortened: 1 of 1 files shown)


jonboulle:
LGTM in itself; did you check for any other references to this field?

lucab:
I grep'd this repo before commiting, and I haven't see anything else. The original PR also only touched those two lines. I don't know if there are any other repos dependent on this.

stevvooe:
LGTM

Issue #485, opened 2016-12-08
Suggestion: refer to label-schema.org for examples of label usage


lizrice:
Over at [label-schema.org](http://label-schema.org) we've been reaching a consensus about some [common keys for container labels](http://label-schema.org/rc1/) which could be referred to by different tools. 

I'm opening this issue to start a discussion about whether the OCI image spec could in some way reference label-schema - perhaps as an example of label key usage? If this is something folks are willing to see added to the spec, I'm more than happy to make a PR. 

To be clear: we're not trying to suggest that a container has to have any of these labels! We're envisaging a future where tools extract metadata from a container, so having common keys could save users from having to add duplicate labels with the same information. A reference from the OCI image spec could really help establish the conventions. 

[Here's a list](https://microbadger.com/labelschema) of public container images published on Docker Hub that are using label-schema labels today.

vbatts:
This SGTM. The conventions laid out in label-schema.org are completely compatible with both the labels in the OCI Image config.md, as well as annotations et al, and to my knowledge do not conflict anywhere.

jonboulle:
+1. Honestly I'd like to see this in the spec itself (of course always as
optional guidelines, never mandatory), but I suspect not all might agree
with that sentiment.

wking:
On Thu, Dec 08, 2016 at 11:57:08AM -0800, Jonathan Boulle wrote:
> Honestly I'd like to see this in the spec itself (of course always
> as optional guidelines, never mandatory), but I suspect not all
> might agree with that sentiment.

The spec already has a stub like this [1], and we may need to keep
critical keys in-spec (e.g. if we use one for naming [2]).  But I
think it's good to push as much as possible (including the ones we
define now [1]) to a separately versioned entity.  Then you can drop
an ill-conceived label (and major bump the label spec) without having
to major bump the core image spec.  I'm agnostic about whether those
external label specs live under the OCI umbrella or not.

[1]: https://github.com/opencontainers/image-spec/blob/656fb2f82121e8bd611b5aae4a6a3471e1bfe078/manifest.md#pre-defined-annotation-keys
[2]: https://github.com/opencontainers/image-spec/pull/438#issuecomment-264042207

stevvooe:
[...]

PR #635, opened 2017-04-03
config: reference common label conventions

vbatts:
This has come up a couple of times on the weekly discussion, but the PR
had not been followed through with.

label-schema.org is a convention that follows the practices outlined in
OCI. Provide here only as a reference example of label usage.

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -13,6 +13,7 @@ This property contains arbitrary metadata.
 * Keys using the `org.opencontainers.image` namespace are reserved for use in the OCI Image Specification and MUST NOT be used by other specifications and extensions, including other OCI specifications.
 * If there are no annotations then this property MUST either be absent or be an empty map.
 * Consumers MUST NOT generate an error if they encounter an unknown annotation key.
+* There are community conventions available for image labels and annotations, for example [Label Schema](http://label-schema.org/rc1/).
 
 ## Pre-Defined Annotation Keys
 


vbatts:
cc @lizrice

wking:
Fixes #485?

I'd rather see the reference land in [1].  I'm not sure why #501
didn't setup a link to annotations.md for Labels, but maybe we want to
do that?

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/annotations.md

jonboulle:
Why not annotations too?

vbatts:
Nothing stopping that, but the convention is based on image labels

jonboulle:
I suppose (IMHO) we still haven't clearly described the use cases for one vs the other.

stevvooe:
@vbatts I've filed #636 to bring the OCI annotations/labels to feature parity with label-schema, minus the `docker/rkt.cmd` stuff.

In general, these style of annotations work best when sitting on the actual image config, rather than manifests or indexes. They can be "pulled up", which makes them indexed, in a sense, but pushing them down to the image config minimizes the possibility of them being lost in transit.

It would be good to add this line:

> `Labels` SHOULD use the annotations defined in [annotations.md](annotations.md).

vbatts:
That's cool, though it still of benefit for such a community convention to
have reference of it's usage. No?

stevvooe:
@vbatts Yes, but it seems better if we just have them defined here. I am not saying remove the reference, just point them towards using the annotations. #636 ensures we have all the equivalent annotations from label-schema.

In that vein, I see this PR being the following:

> Labels SHOULD use the annotations defined in annotations.md.
> There are community conventions available for image labels, for example [Label Schema](http://label-schema.org/rc1/).

vbatts:
@stevvooe I see. Perhaps I'll do that in a separate PR because the label rules could just point to ./annotations.md#rules

wking:
> Yes, but it seems better if we just have them defined here.  I am not saying remove the reference, just point them towards using the annotations. #636 ensures we have all the equivalent annotations from label-schema.

I don't see the point to having OCI-named “equivalent annotations” for keys that don't directly impact image-spec or runtime-spec use-cases.  So if we expect to have image-spec or image-tools code acting on the keys, having an OCI-named key makes sense.  But I don't see any OCI-repo use-case for “packaged sofware version” (and [the goal seems to be just human-readable anyway][1]), so adding OCI-named keys seems like a needless community fragmentation.  Why would users want to use the OCI-named keys when they could use the existing [`org.label-schema.version`][2], etc.?  And having to support both in tooling seems like tedious busywork for no practical gain.

[1]: https://github.com/opencontainers/image-spec/pull/636#discussion_r109793089
[2]: http://label-schema.org/rc1/#build-time-labels

vbatts:
@wking conversation for the _other_ thread (#636), but agreed

vbatts:
rebased (moved stuff to `annotations.md` to keep DRY)

stevvooe:
@vbatts Does this need to be modified after #636? Do you want to carry this for the compatibility table?

vbatts:
@stevvooe i reckon a compatibility table would cover any references that this PR would be attempting to achieve.

PR #638, opened 2017-04-05
schema: Update the json file

### LINKED ISSUE #618: merge "ManifestDescriptor" and "Descriptor" types

q384566678:
Update the corresponding json file and spec file according to #620 

Fixes #618

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -25,7 +25,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
   This REQUIRED property contains a list of [manifests](manifest.md) for specific platforms.
   While this property MUST be present, the size of the array MAY be zero.
 
-  Each object in `manifests` has the base properties of [descriptor](descriptor.md) with the following additional properties and restrictions:
+  Each object in `manifests` has the base properties of [descriptor](descriptor.md) with the following requirements and restrictions:
 
   - **`mediaType`** *string*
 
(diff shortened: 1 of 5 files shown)


image-index.md
```diff
@@ -43,32 +43,6 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
     This OPTIONAL property describes the platform which the image in the manifest runs on.
```
jonboulle: line 28 should be changed from "with the following additional properties and restrictions" to "with the following requirements and restrictions" or similar

stevvooe:
Platform shouldn't be valid on descriptors outside usage on the index. While we've changed the Go datatype to make processing easier, the specification still reads correctly.

stevvooe:
I missed https://github.com/opencontainers/image-spec/pull/638#pullrequestreview-31021778.

> Perhaps it would make more sense to leave platform where it is, and instead simply augment the descriptor description here to mention that "descriptors pointing to certain types may include extended fields" or similar ..?

This is probably the correct approach here.

qianzhangxa:
Agree, that should be the correct approach.

image-index.md
q384566678: update, what do you think of https://github.com/opencontainers/image-spec/pull/638/commits/3e7b8d00a1c7b5c8561c1fbcbdc8a437f6c82850

stevvooe:
I'm going to go ahead and close this favor of @jonboulle's suggestion in https://github.com/opencontainers/image-spec/pull/638#pullrequestreview-31021778.

qianzhangxa:
But based on @jonboulle's comment, we need to augment the descriptor description, so we still need another PR to update `descriptor.md`?

stevvooe:
@qianzhangxa Yes, but this PR needs to be closed.

PR #639, opened 2017-04-05
config: DRY up labels with annotations reference

vbatts:
from
https://github.com/opencontainers/image-spec/pull/635#issuecomment-291634824
it was clear that all the rules for `Labels` are the same as
annotations, so let's DRY that up a bit.

### DIFF
--- a/config.md
+++ b/config.md
@@ -160,12 +160,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
    - **Labels** *object*, OPTIONAL
 
      The field contains arbitrary metadata for the container.
-     Labels MUST be a key-value map where both the key and value MUST be strings.
-     Keys MUST be unique within this map, and best practice is to namespace the keys.
-     Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
-     Keys using the `org.opencontainers.image` namespace are reserved and MUST NOT be used by subsequent specifications.
-     If there are no labels then this property MAY either be absent or an empty map.
-     Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown labels key.
+     This OPTIONAL property MUST use the [annotation rules](annotations.md#rules).
 
   - **StopSignal** *string*, OPTIONAL
 


vbatts:
cc @stevvooe how's this?

jonboulle:
oh, hmm, seems reasonable enough

config.md
```diff
@@ -160,12 +160,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
    - **Labels** *object*, OPTIONAL
 
      The field contains arbitrary metadata for the container.
-     Labels MUST be a key-value map where both the key and value MUST be strings.
-     Keys MUST be unique within this map, and best practice is to namespace the keys.
-     Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
-     Keys using the `org.opencontainers.image` namespace are reserved and MUST NOT be used by subsequent specifications.
-     If there are no labels then this property MAY either be absent or an empty map.
-     Implementations that are reading/processing this configuration file MUST NOT generate an error if they encounter an unknown labels key.
+     This OPTIONAL property MUST use the [annotation rules](annotations.md#rules).
```
wking: While the other annotations references (e.g. [here][1]) use the “This OPTIONAL property…” pattern, `config.md` puts the OPTIONAL/REQUIRED-ness [in the same line as the property name][2] (while [`manifest.md` does not][3]).  My preferred larger fix would be to follow runtime-spec and consistently declare OPTIONAL/REQUIRED-ness in the same line as  the property name, but if that's too big to bite off here, I'd rather reword this line to not repeat OPTIONAL.  Something like:

> This property MUST use the [annotation rules](annotations.md#rules).

would be fine.

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc5/manifest.md#L69
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc5/config.md#L117
[3]: https://github.com/opencontainers/image-spec/blame/v1.0.0-rc5/manifest.md#L66

stevvooe:
LGTM

config.md
jonboulle: follow up?

jonboulle:
lgtm

PR #641, opened 2017-04-06
spec: consistent wording

jonboulle:
Chasing #621, standardise on case/punctuation/articles

Signed-off-by: Jonathan Boulle <jonathanboulle@gmail.com>

PR #245, opened 2016-09-01
[ReadMe] Remove extraneous "should"

RobDolinMS:
Signed-off-by: Rob Dolin <robdolin@microsoft.com>

PR #110, opened 2016-06-01
[ReadMe] remove "should" from FAQ question

RobDolinMS:
Replaces "should strive" below with "strives" since "should" could be construed to have specific meaning per RFC2119

> Q: What happens to AppC or Docker Image Formats?
> A: ... The OCI Image Format project should strive to provide a dependable open specification ...

Signed-off-by: Rob Dolin <RobDolin@microsoft.com>

Issue #109, opened 2016-06-01
[Manifest] Unclear if "should" intended as "SHOULD"


RobDolinMS:
>size int
> ...
>If the length of the retrieved content does not match the specified length, the content should not be trusted.

If intended as a recommendation, use "SHOULD" (ALL CAPS.)  
If not intended as a recommendation invoke RFC meaning, consider alternate wording.

stevvooe:
I'm not even sure this is correct. The size is really there to support preallocation. Content verification should be done through digest.

philips:
I think it is correct. This may protect against future length extension attacks.

stevvooe:
Ok, this should be SHOULD, then.

I'm going to add [this](https://en.wikipedia.org/wiki/Length_extension_attack) to #111, as this question comes up all the time.

philips:
confirmed fixed in #111

PR #643, opened 2017-04-07
spec: Remove DNS-naming and signing entries from high-level components

wking:
We don't have these yet, and can restore the entries if/when we land spec sections describing them.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -57,8 +57,11 @@ The high-level components of the spec include:
 * [Filesystem Layer](layer.md) - a changeset that describes a container's filesystem
 * [Image Configuration](config.md) - a document determining layer ordering and configuration of the image suitable for translation into a [runtime bundle][runtime-spec]
 * [Descriptor](descriptor.md) - a reference that describes the type, metadata and content address of referenced content
-* Signatures that are based on signing image content address (optional OCI layer)
-* Naming that is federated based on DNS and can be delegated (optional OCI layer)
+
+Future versions of this specification may include the following OPTIONAL features:
+
+* Signatures that are based on signing image content address
+* Naming that is federated based on DNS and can be delegated
 
 [c99-unspecified]: http://www.open-std.org/jtc1/sc22/wg14/www/C99RationaleV5.10.pdf#page=18
 [runtime-spec]: https://github.com/opencontainers/runtime-spec


stevvooe:
LGTM

However, I thought this was part of a road map thing. Do we have it represented anywhere?

wking:
> Do we have it represented anywhere?

#11, #22, and https://www.opencontainers.org/about/oci-scope-table

coolljt0725:
@wking  conflict, needs a rebase :)

jonboulle:
I don't feel extremely strongly about this, but why not just add a note saying "work has not yet started"?

vbatts:
we had roadmapped these. It's more of a "work has not yet started" like @jonboulle said

wking:
On Mon, Apr 10, 2017 at 05:08:59AM -0700, v1.0.0.batts wrote:
> we had roadmapped these. It's more of a "work has not yet started"
> like @jonboulle said

Right, and I think we should remove them from the spec as long as they
are roadmap-only.  For example, the README currently links to
milestones for roadmap information [1], although in this case tags are
probably more appropriate [2,3].  I'm fine adding links to those
labels (or just link to [4]?) in an out-of-spec location (like the the
README), but I don't think we need or want to track roadmap
information in the spec itself.

Rebased around #641 with 15c38f9 → 8eef994.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/README.md#roadmap
[2]: https://github.com/opencontainers/image-spec/labels/component%2Fnaming%20spec
[3]: https://github.com/opencontainers/image-spec/labels/component%2Fsigning%20spec
[4]: https://github.com/opencontainers/image-spec/labels

philips:
+1 to work has not started language.

wking:
On Wed, Apr 19, 2017 at 11:42:59AM -0700, Brandon Philips wrote:
> +1 to work has not started language.

Is that “and I require the not-started language to be in spec.md and
not in the “Roadmap” README section [1]?  Or are you ok with me
removing the entries from spec.md if I add not-started language to
that README section?

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/README.md#roadmap

jonboulle:
Since the roadmap to 1.0 doesn't reference these, please just add work-not-yet-started here in this list.

wking:
On Thu, Apr 20, 2017 at 04:09:32AM -0700, Jonathan Boulle wrote:
> Since the roadmap to 1.0 doesn't reference these, please just add
> work-not-yet-started here in this list.

I see no reason for that README section to be 1.0-specific, especially
since the only accurate part of the current roadmap body is the link
to milestones (which is not 1.0-specific).  But whatever, it seems
like the maintainers really want to keep these two particular roadmap
entries in the spec itself, so in 8eef994 → d67c709 I've restored them
with a “we may get to these at some point” caveat.

vbatts:
LGTM

vbatts:
review please

spec.md
```diff
@@ -57,8 +57,11 @@ The high-level components of the spec include:
 * [Filesystem Layer](layer.md) - a changeset that describes a container's filesystem
 * [Image Configuration](config.md) - a document determining layer ordering and configuration of the image suitable for translation into a [runtime bundle][runtime-spec]
 * [Descriptor](descriptor.md) - a reference that describes the type, metadata and content address of referenced content
-* Signatures that are based on signing image content address (optional OCI layer)
-* Naming that is federated based on DNS and can be delegated (optional OCI layer)
+
+Future versions of this specification may include the following OPTIONAL features:
```
wking: @RobDolinMS might be concerned about this lowercase “may” (e.g. #109, #110, #245).  Do folks have other ideas for rephrasing this sentence to avoid “may”?  I don't want to use MAY, because this is about the spec roadmap, not [about conformance with this spec version][1].  Perhaps:

> The following OPTIONAL features are under consideration for future versions of this specification:

?

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.0-rc5/spec.md#notational-conventions

jonboulle:
lgtm

bclairw-commits:
> @wking conflict, needs a rebase :)

If needed, read some route paths it gave ethical hacking its path function for wht I read

PR #644, opened 2017-04-07
spec: Drop "one or more" from index description

wking:
Fixing overly-strict wording from #641 (see [this comment][1]).  We [explicitly allow `manifests` to be empty][2], which allows you to do things like distribute an image-layout which only provides CAS blobs.

[1]: https://github.com/opencontainers/image-spec/pull/641#discussion_r110429162
[2]: https://github.com/opencontainers/image-spec/blame/eb02c88ef382324a1572fa8c66f836a00d865a3f/image-index.md#L26


vbatts:
LGTM

PR #645, opened 2017-04-10
address duplicate REQUIREMENT

mikebrow:
We are repeating the REQUIRED statement here in image-index.md which is properly located in the forward reference document descriptor.md. All upper case terms hold special meaning see  https://tools.ietf.org/html/rfc2119

Additionally as @wking points out, we're also including OPTIONAL descriptor properties in the extension, not just the REQUIRED properties.

Signed-off-by: Mike Brown <brownwm@us.ibm.com>


stevvooe:
LGTM

wking:
Chasing #583.

vbatts:
LGTM

PR #646, opened 2017-04-11
Drop "one or more" from index description, follow up #644

coolljt0725:


vbatts:
LGTM

PR #647, opened 2017-04-11
spec: Add "NOT RECOMMENDED" to RFC 2119 keywords

wking:
Catching up with [erratum 499][1].

[1]: https://www.rfc-editor.org/errata_search.php?rfc=2119


jonboulle:
lgtm

stevvooe:
I'd like to point out that "NOT RECOMMENDED" is actually two words.

LGTM

PR #648, opened 2017-04-19
image-layout.md: fix example

Mashimiao:
Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>

I think put `application/vnd.oci.image.index.v1+json` in manifests is not correct, we should fix it.
And mediaType of manifest can't be application/xml

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -163,15 +163,8 @@ Those tags will often be represented in an image-layout repository with matching
 ```json,title=Image%20Index&mediatype=application/vnd.oci.image.index.v1%2Bjson
 {
   "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
-    {
-      "mediaType": "application/vnd.oci.image.index.v1+json",
-      "size": 7143,
-      "digest": "sha256:0228f90e926ba6b96e4f39cf294b2586d38fbb5a1e385c05cd1ee40ea54fe7fd",
-      "annotations": {
-        "org.opencontainers.image.ref.name": "stable-release"
-      }
-    },
     {
       "mediaType": "application/vnd.oci.image.manifest.v1+json",
       "size": 7143,
@@ -185,7 +178,7 @@ Those tags will often be represented in an image-layout repository with matching
       }
     },
     {
-      "mediaType": "application/xml",
+      "mediaType": "application/vnd.oci.image.manifest.v1+json",
       "size": 7143,
       "digest": "sha256:b3d63d132d21c3ff4c35a061adf23cf43da8ae054247e32faa95494d904a007e",
       "annotations": {


image-layout.md
```diff
@@ -163,16 +163,9 @@ Those tags will often be represented in an image-layout repository with matching
 ```json,title=Image%20Index&mediatype=application/vnd.oci.image.index.v1%2Bjson
 {
   "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.index.v1+json",
```
vbatts: this field is not being used, but only on descriptors

image-layout.md
```diff
@@ -185,7 +178,7 @@ Those tags will often be represented in an image-layout repository with matching
       }
     },
     {
-      "mediaType": "application/xml",
```
vbatts: this example is illustrating that other mediatypes can be included that are not just OCI mediatyps

vbatts:
perhaps the confusion here is the explanation of the example is not clear enough? all the items being fixed here are intended to be as they were. I'm going to close this issue. Thanks for the effort though.

image-layout.md
```diff
@@ -163,16 +163,9 @@ Those tags will often be represented in an image-layout repository with matching
...
+  "mediaType": "application/vnd.oci.image.index.v1+json",
   "manifests": [
     {
-      "mediaType": "application/vnd.oci.image.index.v1+json",
-      "size": 7143,
-      "digest": "sha256:0228f90e926ba6b96e4f39cf294b2586d38fbb5a1e385c05cd1ee40ea54fe7fd",
-      "annotations": {
-        "org.opencontainers.image.ref.name": "stable-release"
-      }
-    },
```
vbatts: this descriptor is here because the index can point to other image indexes. that is a feature.

Mashimiao: If so, I think we need fix mediatype description of manifest in image-index.md. It does not say `application/vnd.oci.image.index.v1+json` is supported.

wking: > I think we need fix mediatype description of manifest in image-index.md. It does not say `application/vnd.oci.image.index.v1+json` is supported.

That's because [compliant consumers don't *need* to be able to handle `application/vnd.oci.image.index.v1+json` at all][1].  And we do [require consumers to ignore unknown media types][2]

[1]: https://github.com/opencontainers/image-spec/blame/a82c09852aba4978e3d16bb0cd583d5e2f4a63c9/image-index.md#L4
[2]: https://github.com/opencontainers/image-spec/blame/a82c09852aba4978e3d16bb0cd583d5e2f4a63c9/image-index.md#L39

Mashimiao: @wking  I think you said conflicts with @vbatts said. He said that's a feature. If it is not, I think there is no need to fix because spec actually says ignore unknown media types

wking: Thinking about this again, I would like to see `application/vnd.oci.image.index.v1+json` added to the media types which MUST be supported here.  Any consumer which gets to this point already understands the type (because they're reading one), so (for those implementations) an `application/vnd.oci.image.index.v1+json` reference will not meet the “unknown media-type” ignore condition.  The MUST entry will make it very clear (although I think it is already true) that compliant image-spec consumers will either:

1. Support `application/vnd.oci.image.index.v1+json` including `application/vnd.oci.image.index.v1+json` entries in `manifests`.
2. Not support `application/vnd.oci.image.index.v1+json` at all.

I don't see how “supports `application/vnd.oci.image.index.v1+json` but *not* `application/vnd.oci.image.index.v1+json` entries in `manifests`” is possible.

PR #650, opened 2017-04-24
image-index: define platform.variant for arm

AkihiroSuda:
Please refer to https://github.com/opencontainers/image-spec/pull/632 for previous discussion.

In this PR, I avoided to use Linux-specific uname values for ease of porting to non-Linux.
https://github.com/opencontainers/image-spec/pull/632#discussion_r109226040

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -63,7 +63,16 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     - **`variant`** *string*
 
-        This OPTIONAL property specifies the variant of the CPU, for example `armv6l` to specify a particular CPU variant of the ARM CPU.
+        This OPTIONAL property specifies the variant of the CPU.
+        Image indexes SHOULD use, and implementations SHOULD understand, values listed in the following table.
+        When the variant of the CPU is not listed in the table, values are implementation-defined and SHOULD be submitted to this specification for standardization.
+
+        | ISA/ABI         | `architecture` | `variant`   |
+        |-----------------|----------------|-------------|
+        | ARM 32-bit, v6  | `arm`          | `v6`        |
+        | ARM 32-bit, v7  | `arm`          | `v7`        |
+        | ARM 32-bit, v8  | `arm`          | `v8`        |
+        | ARM 64-bit, v8  | `arm64`        | `v8`        |
 
     - **`features`** *array of strings*
 
--- a/schema/backwards_compatibility_test.go
+++ b/schema/backwards_compatibility_test.go
@@ -49,7 +49,7 @@ func TestBackwardsCompatibilityImageIndex(t *testing.T) {
 		fail       bool
 	}{
 		{
-			digest: "sha256:219f4b61132fe9d09b0ec5c15517be2ca712e4744b0e0cc3be71295b35b2a467",
+			digest: "sha256:d0ed7cfe33821cb6a15624486e650149e92fff3192ff2014bda0c4b0206c1aa2",
 			imageIndex: `{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
@@ -91,7 +91,7 @@ func TestBackwardsCompatibilityImageIndex(t *testing.T) {
          "platform": {
             "architecture": "arm",
             "os": "linux",
-            "variant": "armv7"
+            "variant": "v7"
          }
       },
       {
@@ -101,7 +101,7 @@ func TestBackwardsCompatibilityImageIndex(t *testing.T) {
          "platform": {
             "architecture": "arm64",
             "os": "linux",
-            "variant": "armv8"
+            "variant": "v8"
          }
       }
    ]
--- a/specs-go/v1/descriptor.go
+++ b/specs-go/v1/descriptor.go
@@ -59,7 +59,7 @@ type Platform struct {
 	OSFeatures []string `json:"os.features,omitempty"`
 
 	// Variant is an optional field specifying a variant of the CPU, for
-	// example `ppc64le` to specify a little-endian version of a PowerPC CPU.
+	// example `v7` to specify ARMv7 when architecture is `arm`.
 	Variant string `json:"variant,omitempty"`
 
 	// Features is an optional field specifying an array of strings, each


AkihiroSuda:
CI failed but locally passes
https://travis-ci.org/opencontainers/image-spec/builds/225111253
```
$ make lint
checking lint
WARNING: deadline exceeded by linter staticcheck on schema (try increasing --deadline)
WARNING: deadline exceeded by linter errcheck on schema (try increasing --deadline)
make: *** [lint] Error 2
The command "make lint" exited with 2.
```

image-index.md
```diff
@@ -63,7 +63,9 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     - **`variant`** *string*
 
-        This OPTIONAL property specifies the variant of the CPU, for example `armv6l` to specify a particular CPU variant of the ARM CPU.
+        This OPTIONAL property specifies the variant of the CPU.
+        When `architecture` is `arm`, values for arch SHOULD use, and runtimes SHOULD understand, arch entries listed in the Go Language document for [`$GOARM`](go-arm). i.e. `5`, `6`, or `7`.
```
wking: “values for arch” → “values for `variant`”.  And maybe “arch entries” → “architecture versions” to match Go's wording.

AkihiroSuda: done

schema/backwards_compatibility_test.go
```diff
@@ -100,8 +100,7 @@ func TestBackwardsCompatibilityImageIndex(t *testing.T) {
          "digest": "sha256:fb2fc0707b86dafa9959fe3d29e66af8787aee4d9a23581714be65db4265ad8a",
          "platform": {
             "architecture": "arm64",
-            "os": "linux",
-            "variant": "armv8"
```
mattspencer-arm: I would like to see "variant": "8" here for completeness.

schema/backwards_compatibility_test.go
```diff
@@ -91,7 +91,7 @@ func TestBackwardsCompatibilityImageIndex(t *testing.T) {
          "platform": {
             "architecture": "arm",
```
mattspencer-arm: I understand that "arm" is what gets expressed by GOARCH, which is why it is used here.  But would it not be better to disambiguate the property by calling it "arm32"?

schema/backwards_compatibility_test.go
AkihiroSuda: I don't think we should add it until upstream Go adds "8".

Note that there is no guarantee Go adds "8".
It can be "8-foo" depending on the future extension, or maybe completely different name depending on the future ARM versioning scheme.

schema/backwards_compatibility_test.go
AkihiroSuda: This is defined by runtime-spec, and I don't think we can change anymore.
https://github.com/opencontainers/runtime-spec/blob/v1.0.0-rc3/config.md#platform

schema/backwards_compatibility_test.go
mattspencer-arm: A question then - is the OCI specification subordinate to go with respect to the value of attributes here?  Would this not make the implementation of this specification sensitive to changes in go?
I thought that the metadata in this specification was to define the compatibility of the images - in which case we ought to be complete in our definition here.  The full definition in this instance would be { "arch": "arm64", "variant": "8" }?

schema/backwards_compatibility_test.go
mattspencer-arm: That's fine, and as a result of a historical decisions.
ARM would rather see the use of arm32 to reference their 32 bit instruction set rather than the company name - which could be considered ambiguous?

schema/backwards_compatibility_test.go
AkihiroSuda: @stevvooe WDYT?

AkihiroSuda: another idea is to use Linux `utsname` (`armv8l`, `armv7l`) across all the platforms, although Linux-specific https://github.com/opencontainers/image-spec/pull/632#discussion_r109226040

mattspencer-arm: I have been chatting with the kernel/toolchain and disto teams here within ARM, and the consensus is that there is currently no reliable way of runtime detecting the underlying architecture.

To quote our kernel team when asked about runtime detection
"As an example, Linux reports ARMv7 for ARM11MPCore or ARM1176 (and I think newer revisions of ARM1136). The architecture versions are just some milestones or new baselines without a dedicated ID. The architected CPUID registers only expose features and not architecture versions. However, since some features are "back-ported" to previous
architecture versions, the kernel gets confused (the ARMv7 MMU features have been built into ARM11MPCore, hence Linux reporting v7).

The "armv8l" is _only_ reported by an AArch64 kernel when running a task with the PER_LINUX32 personality. Otherwise the utsname is "aarch64". If you run an AArch32 kernel on ARMv8, it will report armv7l.

I doubt the above helps but I think we should just avoid such run-time
detection at the architecture version. If the hardware does not provide
this information, the OS cannot reliably infer it."

I know this is not very helpful at the moment, but it is a statement of where we are today.  There is talk about ARM potentially investing time in 'lscpu' with the help of a LUT to correctly identify architecture.

stevvooe:
After looking into this further, mapping `variant` to `GOARM` is incorrect. The `variant` field must have the abi+isa within it. For example, `arm64ipl32` would be the `variant`.

AkihiroSuda:
@mattspencer-arm PTAL about @stevvooe 's comment?
https://github.com/opencontainers/image-spec/pull/650#issuecomment-298084072

Also, I think it would be better to add an implementator's note about how to check supported `abi` and `isa`.
With regarding to `arm64ilp32`, https://wiki.linaro.org/Platform/arm64-ilp32 might be useful, but couldn't get what is the standard way

schema/backwards_compatibility_test.go
stevvooe: > The full definition in this instance would be { "arch": "arm64", "variant": "8" }?

This example was only effort put forth on this particular pull request. The actual usage current looks like this:

```
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v1+json",
         "size": 2084,
         "digest": "sha256:07ebe243465ef4a667b78154ae6c3ea46fdb1582936aac3ac899ea311a701b40",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "armv7"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v1+json",
         "size": 2090,
         "digest": "sha256:fb2fc0707b86dafa9959fe3d29e66af8787aee4d9a23581714be65db4265ad8a",
         "platform": {
            "architecture": "arm64",
            "os": "linux",
            "variant": "armv8"
         }
      }
```

schema/backwards_compatibility_test.go
stevvooe: > ARM would rather see the use of arm32 to reference their 32 bit instruction set rather than the company name - which could be considered ambiguous?

Please take this up with the Go team.

AkihiroSuda:
Can we remove and reserve this field as in platform.features #672?

cc @estesp

AkihiroSuda:
Uh, I just received email from stevvooe to make this forward, +1

https://groups.google.com/a/opencontainers.org/forum/m/#!topic/dev/Eht-i5UbUvo

stevvooe:
@AkihiroSuda Was just coming here to post that!

stevvooe:
@AkihiroSuda Would you mind updating this PR in the vein of that proposal? They are quite compatible, but I'd like to create a table, as is in the proposal. I'll paste that table into #622.

AkihiroSuda:
done

image-index.md
```diff
@@ -63,7 +63,18 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
...
 
-        This OPTIONAL property specifies the variant of the CPU, for example `armv6l` to specify a particular CPU variant of the ARM CPU.
+        This OPTIONAL property specifies the variant of the CPU.
+        Image indexes SHOULD use, and implementations SHOULD understand, values listed in the following table.
+        When the variant of the CPU is not listed in the table, values are implementation-defined and SHOULD be submitted to this specification for standardization.
+
+        | ISA/ABI                              | `architecture` | `variant`   |
+        |--------------------------------------|----------------|-------------|
+        | ARM 32-bit, v5 and earlier           | Unsupported    | Unsupported |
+        | ARM 32-bit, v6 (non-Raspberry Pi 1)  | Unsupported    | Unsupported |
```
stevvooe: Leave these lines off. Nothing is explicitly unsupported.

image-index.md
```diff
@@ -63,7 +63,18 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
...
+        | ARM 32-bit, v6 (Raspberry Pi 1)      | `arm`          |`v6`         |
```
stevvooe: Leave "Raspberry Pi 1". That was just a note in the proposal.

AkihiroSuda:
done

mattspencer-arm:
Can I suggest that we don't use the word 'Unsupported' but replace it with 'Out of scope'.  I think this better reflects the position?


/Matt

________________________________

image-index.md
mattspencer-arm: I agree - but I would replace 'unsupported' with 'out of scope' to make it explicit that this spec does not cover older ARM ISAs.

stevvooe:
> Can I suggest that we don't use the word 'Unsupported' but replace it with 'Out of scope'.  I think this better reflects the position?

I don't think these lines are even needed. If someone wants to build ARM support for pre-v6, they are within their right, but it won't be specified and if it conflicts, it conflicts.

@AkihiroSuda Let's just remove those lines.

AkihiroSuda:
already removed

image-index.md
```diff
@@ -63,7 +63,16 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
...
+        This OPTIONAL property specifies the variant of the CPU.
+        Image indexes SHOULD use, and implementations SHOULD understand, values listed in the following table.
+        When the variant of the CPU is not listed in the table, values are implementation-defined and SHOULD be submitted to this specification for standardization.
+
+        | ISA/ABI         | `architecture` | `variant`   |
+        |-----------------|----------------|-------------|
+        | ARM 32-bit, v6  | `arm`          | `v6`        |
+        | ARM 32-bit, v7  | `arm`          | `v7`        |
+        | ARM 32-bit, v8  | `arm`          | `v8`        |
+        | ARM 64-bit, v8  | `arm64`        | `v8`        |
```
hqhq: If we scope arch and variant with this table, and remove features, we can't identify platforms with ilp32 support, are there other issues tracing this?

And is `variant` really ARM specific property?

stevvooe: Please see https://golang.org/src/go/build/syslist.go for extra architecture values. I think ilp32 will be arm64p32. It is not here because it has low adoption at this point.

> And is variant really ARM specific property?

It is not. Where did you see this implied?

stevvooe:
LGTM

vbatts:
LGTM

image-index.md
imrehg: Would it be possible to include a `v5` variant? armv5/armel is still supported by go, and would be great to cover all options (and it's in use for some devices at us at [resin.io](https//resin.io))

Issue #599, opened 2017-03-08
spec: define charset for "org.opencontainers.ref.name"


stevvooe:
Currently, there is no restriction on the character set for reference names. We'll need something open but define some common bounds to make this.

I suggested restricting this to the URI character set, in https://github.com/opencontainers/image-spec/pull/591#issuecomment-285153069.

Problem called out in #591.

stevvooe:
cc @q384566678

vbatts:
cc @alexlarsson, would the URI character set work for you as well?

jonboulle:
To be very explicit, is this a change in direction from the position expressed [here](https://github.com/opencontainers/image-spec/issues/581#issuecomment-282074320) and [here](https://github.com/opencontainers/image-spec/issues/581#issuecomment-283434330)?

stevvooe:
@jonboulle Not exactly and there is some subtlety here.

From the perspective of the specification, annotations are opaque. They will be read and propagated without modification or verification.

This does not mean that the specifier of the annotation can't make requirements that a given annotation has a structure that is internal to that annotation. Practically, this means that when an annotation producer or consumer reads an annotation, it may be generated or validated in a specified format.

Let me know if this distinction is unclear. It seems like splitting hairs, but it is to avoid having the annotations "leak" into the first-class field space.

q384566678:
I think that as long as the string type of standard can be, do not need more norms.

erikh:
https://en.m.wikipedia.org/wiki/Punycode Is a method of encoding non-latin encodings into urls in a way that's easy to both parse and display in all instances. Maybe that's a good solution since it has a limited input set?

erikh:
It's used in many internationalized urls, so it's not just a spec for spec's sake. :D

erikh:
https://godoc.org/golang.org/x/net/idna Is a golang library to manage the conversion.

AkihiroSuda:
How to tell punycode string from ASCII string?
(Is is valid to use "xn--" prefix for non-domain strings?)

Why not use percent-encoded UTF-8?

erikh:
no argument from me re: percent encoded utf-8

erikh:
So I did some reading tonight and discovered these. With the punycode suggestion I was attempting to find a word-safe set of characters to use and realized later that the unicode standard already does this for us elsewhere. :)

http://unicode.org/reports/tr18/#Categories

Which are supported by most regex implementations these days, since unicode classes are well-defined.

[...]

PR #654, opened 2017-04-24
schema: allow compound algorithm specifiers in digests

stevvooe:
While we currently have support for only `sha256` in digests, OCI image
processors may encounter other kinds of "compound" digests, such as
`tarsum+sha256`. This change allows OCI validation stacks to correctly
validate these digest algorithms, then let them report failure based on
the lack of algorithm support.

Future cases include allow different encoding, such as a `sha256+b64`
and others. While this future proofs the field against this, this isn't
an endorsement to ride out and do this today. This just provides some
algorithmic agility that may be required later.

Note that this brings digest in line with what is supported across the
docker ecosystem today.


AkihiroSuda:
Do we need other than `+`?

AkihiroSuda:
Also, can we allow non-hexadecimal digest values?
e.g.
`multihash+base58:QmRZxt2b1FVZPNqd8hsiykDL3TdBDeTSPX9Kv46HmX4Gx8`

schema/manifest_test.go
```diff
@@ -191,6 +191,54 @@ func TestManifest(t *testing.T) {
...
+      "digest": "sha256+foo-bar:c86f7763873b6c0aae22d963bab59b4f5debbed6685761b5951584f6efb0633b"
+    },
+    {
+      "mediaType": "application/vnd.oci.image.config.v1+json",
+      "size": 1470,
+      "digest": "sha256.foo-bar:c86f7763873b6c0aae22d963bab59b4f5debbed6685761b5951584f6efb0633b"
+    }
+  ]
+}
+`,
```
q384566678: Need  add `fail: false,`.

stevvooe: Actually, you don't. Go initializes values to their zero value. The zero value for a boolean is false.

stevvooe:
> Also, can we allow non-hexadecimal digest values?
e.g.
multihash+base58:QmRZxt2b1FVZPNqd8hsiykDL3TdBDeTSPX9Kv46HmX4Gx8

Agreed, but I'll do this in a follow up.

vbatts:
This lgtm, though there is no wording around this structure and what the separators indicate. If a value of `sha256+farts+farts+farts:0228f90e926ba6b96e4f39cf294b2586d38fbb5a1e385c05cd1ee40ea54fe7fd` is encountered, should it be attempted as `sha256` or just errored out if I don't have a hash of `sha256+farts+farts+farts`?

stevvooe:
The algorithms are treated as hardcoded strings. The allowance of these characters is just for stylization and this PR is just reserving the ability to do this in the future. 

`sha256+farts+farts+farts` would have to be specific algorithm registered with the digester, __contained in at least three levels of `farts`__, by convention. From the perspective of the digest dispatch in `go-digest`, `sha256+farts+farts` and `sha256+farts+farts+farts` are just strings without any relation.

AkihiroSuda:
What is the expected meaning of `[._-]`?
Does it differ from `+`?

vbatts:
right on. A brief explanation of that, as well as updating the regexp, in https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests-and-verification would be helpful then.

stevvooe:
@vbatts Widened the character set and updated the specification to match the goal.

descriptor.md
```diff
@@ -61,22 +61,30 @@ The _digest_ property of a Descriptor acts as a content identifier, enabling [co
...
 
 The digest string MUST match the following grammar:
 
 ```
-digest      := algorithm ":" hex
-algorithm   := /[a-z0-9_+.-]+/
-hex         := /[a-f0-9]+/
+digest      := algorithm ":" encoded
+algorithm   := /[a-z0-9]+(?:[+._-][a-z0-9]+)*/
+encoded     := /[a-zA-Z0-9]+/
```
wking: If the intention is to allow [filesystem safe base 64][1], you'll want to add `_=-` to your `encoded` regexp.  If the intention is to allow [vanilla base 64][2], you'll want to add `+/=`.

[1]: https://tools.ietf.org/html/rfc4648#section-5
[2]: https://tools.ietf.org/html/rfc4648#section-4

vbatts: oh right. hrm.

vbatts:
apart from allowing the charset for base64 in the `encoded` section, this LGTM

descriptor.md
stevvooe: `=` actually isn't needed for `base64`. I'll add `_-`.

wking: > `=` actually isn't needed for base64. I'll add `_-`.

[From the RFC][1]:

> Implementations MUST include appropriate pad characters at the end of encoded data unless the specification referring to this document explicitly states otherwise.

So if you do not allow `=` than all specs that use base 64 will need that explicit statement.  This spec doesn't currently reference RFC 4648, so we don't need that explicit statement, but I see no reason to forbid `=` in the `encoded` part.

[1]: https://tools.ietf.org/html/rfc4648#section-3.2

stevvooe: @wking The padding is recoverable from the length of the string. It is not needed in this case or most uses of base64. In fact, the RFC says exactly that. Please be clear that _we are not adding support for base64 encoded digests_. We are ensuring that they are supportable in the future.

stevvooe:
@vbatts Updated.

erikh:
can you see my comments on this subject on #599? Maybe we can unify this.

descriptor.md
wking: > We are ensuring that they are supportable in the future.

Right, but without allowing trailing `=`, you're allowing them only when:

* The alg defines the length of the unencoded hash, and
* The alg spec explicitly states that pad characters MUST NOT be appended.

stevvooe:
@erikh I'm not sure that is fully related. Digest already has a history of a particular character set.

descriptor.md
```diff
@@ -61,22 +61,31 @@ The _digest_ property of a Descriptor acts as a content identifier, enabling [co
...
 
 The digest string MUST match the following grammar:
 
 ```
-digest      := algorithm ":" hex
-algorithm   := /[a-z0-9_+.-]+/
-hex         := /[a-f0-9]+/
+digest      := algorithm ":" encoded
+algorithm   := /[a-z0-9]+(?:[+._-][a-z0-9]+)*/
+encoded     := /[a-zA-Z0-9_-]+/
```
vbatts: base64 has `=` for padding as well

stevvooe: `=` is not required for `base64`, since the length is known.

vbatts: sorry, how is the length known for arbitrary encoded types?

stevvooe: The length of the digest is a part of the algorithm. Each algorithm has a fixed length.

stevvooe:

Let's move this forward so we can get the release out.

descriptor.md
vbatts: that feels like an assumption, especially as this is introducing arbitrary encoding types.

stevvooe: If you insist.

stevvooe: But it is not an assumption. The algorithm must encode the length.

vbatts: i must be missing something. i don't see that is not outlined anywhere.

stevvooe: [`crytpo.Hash`](https://golang.org/pkg/crypto/#Hash), identified by the algorithm, has a fixed length output. It is a property of cryptographic hash functions. I'm pushing an update that allows this anyway. Do you have any examples of cryptographic hash functions that aren't parameterized by configuration of the algorithm? Effectively, they have to comparable in the same domain and typically that doesn't exist.

stevvooe: Updated to allow it though.

stevvooe:
Updates:
- Allow `=` in encoded portion
- Decomposed regexp into components

descriptor.md
vbatts: right right, but `crypto.Hash` is not what we're binding ourselves to here. It's being opened up to allow arbitrary strings for the algorithm. Just saying that it's an assumption that the algorithm being used has a fixed size.
Thanks for updating. I'll take a look.

vbatts:
heh
LGTM

descriptor.md
```diff
@@ -61,22 +61,33 @@ The _digest_ property of a Descriptor acts as a content identifier, enabling [co
...
+Implementations SHOULD allow digests that are unsupported to pass validation if they comply with the above grammar.
+While `sha256` will only use hex encoded digests, support for separators in _algorithm_ and alpha numeric in _encoded_ is included to allow for future extension of digest support.
+As an example, we can parameterize the encoding and algorithm as `multihash+base58:QmRZxt2b1FVZPNqd8hsiykDL3TdBDeTSPX9Kv46HmX4Gx8`, which would be considered valid but unsupported by this specification.
+
 Some example digest strings include the following:
 
-digest                                                                  | algorithm           |
-------------------------------------------------------------------------|---------------------|
-sha256:6c3c624b58dbbcd3c0dd82b4c53f04194d1247c6eebdaab7c610cf7d66709b3b | [SHA-256](#sha-256) |
+digest                                                                     | algorithm           | Supported |
```
wking: “Supported” isn't particularly clear to me.  I think there are two interesting levels for algorithm identifiers in the spec:

* Registered (e.g. listed [here][1]).
* Required to be implemented (as we do for `sha256` [here][2]).

For example, `sha512` (in flight with #609) would be “registered” but would not be “required to be implemented”.

I'm in favor of staying DRY and dropping the “Supported” column here, but I'd also be ok with a “Registered” and/or “Required to be implemented” columns.  Without a clearer definition, I'm not in favor of a “Supported” column.

[1]: https://github.com/opencontainers/image-spec/blob/1eb142309125cdf562d7cca8c279836cd6c04140/descriptor.md#registered-identifiers
[2]: https://github.com/opencontainers/image-spec/blame/1eb142309125cdf562d7cca8c279836cd6c04140/descriptor.md#L123

descriptor.md
```diff
@@ -61,22 +61,33 @@ The _digest_ property of a Descriptor acts as a content identifier, enabling [co
...
 
 ```
-digest      := algorithm ":" hex
-algorithm   := /[a-z0-9_+.-]+/
-hex         := /[a-f0-9]+/
+digest                := algorithm ":" encoded
+algorithm             := algorithm-component [algorithm-separator algorithm-component]*
+algorithm-component   := /[a-z0-9]+/
+algorithm-separator   := /[+._-]/
+encoded               := /[a-zA-Z0-9=_-]+/
```
AkihiroSuda: Please consider adding the following sentence:

When `encoded` denotes a hex value, it MUST NOT (SHOULD NOT?) use upper characters, i.e. `/[A-F]/`. (when `algorithm` is `sha256`?)

descriptor.md
stevvooe: When this PR lands, you can submit a wildly complicated scheme to your liking. For now, this is quite clear.

descriptor.md
stevvooe: > When encoded denotes a hex value, it MUST NOT (SHOULD NOT?) use upper characters, i.e. /[A-F]/. (when algorithm is sha256?)

This constraint belongs to the algorithm, not this portion of the specification. If you want to qualify this, please submit a second PR.

descriptor.md
vbatts: heh. Though "support" has meaning in contexts. This is perhaps clear _enough_

vbatts:
pls2rebase

descriptor.md
wking: > This is perhaps clear enough…

It will become more clear once this is rebased around #609 (which has since landed), since we can see what value `sha512` gets in this column to distinguish “registered” from “required to be implemented”.

stevvooe:
> pls2rebase

Done.

descriptor.md
stevvooe: @AkihiroSuda Maybe, it could be part of the registered table?

descriptor.md
```diff
@@ -57,58 +57,71 @@ The following field keys are reserved and MUST NOT be used by other specificatio
...
 
-The digest string MUST match the following grammar:
+A digest string MUST match the following grammar:
 
 ```
-digest      := algorithm ":" hex
-algorithm   := /[a-z0-9_+.-]+/
-hex         := /[a-f0-9]+/
+digest                := algorithm ":" encoded
+algorithm             := algorithm-component [algorithm-separator algorithm-component]*
```
vbatts: maybe a nit, but does this grammar implies that there be only a single separator+component?

vbatts:
besides the nit question, i reckon this LGTM

vbatts:
silly pullapprove
:+1:

descriptor.md
stevvooe: `[]` is "optional" and `*` is zero or more. This should match the following productions (`A` = algorithm-component, `S` = algorithm-separator):

```
A
ASA
ASASA
```

The following would not be matched:

```
SA
ASAS
```

Put more succinctly, it allows a separator to appear sandwiched by algorithm-component, non-contiguously.

stevvooe:

philips:
LGTM, thanks @stevvooe

PR #655, opened 2017-04-25
descriptor: Formal ABNF for digest values

wking:
I'm not sure what the grammar language was before, but this commit converts it to ABNF and references the defining RFC.

The ABNF highlighting comes from [here][1] and [here][2].

[1]: https://help.github.com/articles/creating-and-highlighting-code-blocks/#syntax-highlighting
[2]: https://github.com/github/linguist/blob/v5.0.8/lib/linguist/languages.yml#L50-L56

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -64,12 +64,12 @@ If the identifier can be communicated in a secure manner, one can retrieve the c
 The value of the digest property, the _digest string_, is a serialized hash result, consisting of an _algorithm_ portion and a _hex_ portion.
 The algorithm identifies the methodology used to calculate the digest; the hex portion is the lowercase hex-encoded result of the hash.
 
-The digest string MUST match the following grammar:
+The digest string MUST match the following [ABNF][rfc5234]:
 
-```
-digest      := algorithm ":" hex
-algorithm   := /[a-z0-9_+.-]+/
-hex         := /[a-f0-9]+/
+```ABNF
+digest    = algorithm ":" hex
+algorithm = 1*(DIGIT / %x61-7A / "+" / "." / "_" / "-")  ; [0-9a-z+._-]+
+hex       = 1*(DIGIT / %x61-66)  ; [0-9a-f]+
 ```
 
 Some example digest strings include the following:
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
I prefer current notation over ABNF for readability, although little a bit informal

wking:
On Tue, Apr 25, 2017 at 10:01:53AM -0700, Akihiro Suda wrote:
> I prefer current notation over ABNF for readability, although little
> a bit informal

Enforcing a MUST based on informal notation seems like more trouble
than reading the ABNF ;).  And while the current syntax is more
compact, the ABNF RFC is about the size of the BRE spec [1] and it
makes it easy to compose a more complicated rule from simpler rules,
while BREs do not.  The current syntax may be Go's syntax [2], but
that looks underspecified to me unless you say “as implemented in Go
1.8” or some such.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/
[2]: https://golang.org/pkg/regexp/syntax/

stevvooe:
@wking Could you please stop filing PRs that invalidate others? Please submit and issue and discuss the change before submitting.

stevvooe:
> Enforcing a MUST based on informal notation seems like more trouble
than reading the ABNF ;).

Also, there is nothing informal about the syntax here. Regular expressions are well-defined and understood by a broader group of readers than ABNF.

vbatts:
this seems to be coming from left field and not needed. closing.

wking:
> Also, there is nothing informal about the syntax here. Regular expressions are well-defined…

The syntax grammar is informal because the current grammar language is undefined.  If we wanted to formally usie a known regex language (for example), we'd need to link to the spec defining it.

erikh:
https://en.m.wikibooks.org/wiki/Regular_Expressions/POSIX_Basic_Regular_Expressions
?

wking:
On Mon, May 01, 2017 at 01:04:12PM -0700, Erik Hollensbe wrote:
> https://en.m.wikibooks.org/wiki/Regular_Expressions/POSIX_Basic_Regular_Expressions
> ?

I'm fine linking [1] as an alternative to this PR, but with just BREs
you'd need to have a single regexp for the whole digest.  To decompose
it into parts (e.g. an algorithm and a hex/encoded part), you'd need
to use ABNF or some other syntax for defining concatenation.

But as it stands, the spec links nothing to define the syntax being
used.

[1]: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap09.html#tag_09_03

PR #660, opened 2017-04-28
Platform identifies the image runtime requirements

mattspencer-arm:
It should be clarified that the manifest is describing what is
'in the image', not what 'it will run on'.

Signed-off-by: Matt Spencer <matt.spencer@arm.com>


stevvooe:
LGTM

vbatts:
LGTM

PR #662, opened 2017-05-08
Add MUST NOT and SHALL NOT to conventions

Mashimiao:
MUST NOT and SHALL NOT means an absolute prohibition of the spec. They should be one of requirements to judge one implementation if it is compliant.

Signed-off-by: Ma Shimiao <mashimiao.fnst@cn.fujitsu.com>


vbatts:
LGTM

PR #667, opened 2017-05-10
descriptor: sha{256,512}: disallow /[A-F]/ characters

AkihiroSuda:
Disallowing upper characters will be useful for avoiding implementation bugs

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -76,6 +76,9 @@ algorithm-separator   := /[+._-]/
 encoded               := /[a-zA-Z0-9=_-]+/
 ```
 
+Note that _algorithm_ MAY impose algorithm-specific restriction on the grammar of the _encoded_ portion.
+See also [Registered Algorithms](#registered-identifiers).
+
 Some example digest strings include the following:
 
 digest                                                                    | algorithm           | Supported |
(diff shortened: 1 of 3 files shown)


descriptor.md
```diff
@@ -137,11 +140,17 @@ If a useful algorithm is not included in the above table, it SHOULD be submitted
 [SHA-256][rfc4634-s4.1] is a collision-resistant hash function, chosen for ubiquity, reasonable size and secure characteristics.
 Implementations MUST implement SHA-256 digest verification for use in descriptors.
 
+When the _algorithm identifier_ is `sha256`, the _encoded_ portion MUST match `/[a-f0-9]{64}/`.
+Note that `[A-F]` MUST NOT be used here.
```
wking: I'd rather stay DRY by defining a hex encoding with these restrictions, and then saying that `sha256` and `sha512` MUST be hex-encoded.

AkihiroSuda: I think hex-encoding is specific to algorithm.
Depending on the algorithm, hex-encoded digest can be `somealgorithm+uuid:f81d4fae-7dec-11d0-a765-00a0c91e6bf6`

erikh:
Should probably make the regex [:upper:] then; i imagine base64 will make
an appearance eventually.

descriptor.md
wking: >  I think hex-encoding is specific to algorithm.

This is true, but both of the current algorithms use the same hex encoding.  Defining that in one place doesn't block other algorithms from defining or using alternative hex encodings, but it does make it easier for other algorithms that want the current hex encoding.

AkihiroSuda:
@erikh 

base64 should use different algorithm identifier, so I think that is the out scope of this PR, which only focuses on `sha256` and `sha512`.

stevvooe:
@AkihiroSuda This looks good. Could you also make a PR to `go-digest`?

AkihiroSuda:
sure https://github.com/opencontainers/go-digest/pull/34

stevvooe:
LGTM

vbatts:
LGTM

PR #666, opened 2017-05-10
descriptor: digest example link should be SHA-512

stevvooe:
Signed-off-by: Stephen J Day <stephen.day@docker.com>

PR #668, opened 2017-05-10
descriptor: Replace "supported" with "registered"

wking:
Avoid adding a second word meaning the same thing (spun off from [here][1]).  I've consolidated around “registered” because it matches the existing “Registered algorithms” header.

Also some copy-edits for the sentences I've touched:

* “alpha numeric” -> “alphanumerics” (matching [Wikipedia][2]).
* “paramterize” -> “parameterize”
* “future extension of digest support” -> “extensions”, because folks may be taking advantage of the extended grammar *now*, without waiting for the image-spec to take advantage of the extended grammar.

[1]: https://github.com/opencontainers/image-spec/pull/654#discussion_r114640550
[2]: https://en.wikipedia.org/wiki/Alphanumeric

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -78,18 +78,18 @@ encoded               := /[a-zA-Z0-9=_-]+/
 
 Some example digest strings include the following:
 
-digest                                                                    | algorithm           | Supported |
---------------------------------------------------------------------------|---------------------|-----------|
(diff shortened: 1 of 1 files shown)


wking:
This will conflict with #666.  I'm happy to rebase after #666 lands.

vbatts:
LGTM but pls2rebase

wking:
Rebased around #666 with 723ddd8 → 0eee504.

stevvooe:
LGTM

vbatts:
LGTM

PR #226, opened 2016-08-31
Fix a typo and reference the RFC for defining standard DNS labels

philips:
Signed-off-by: Jason Bouzane <jbouzane@gmail.com>

PR #163, opened 2016-06-21
Fix a grammar mistake in serialization.md and don't imply _ is valid in dns hostnames

jbouzane:
- Fix a grammar error ("follow comply" -> "comply")
- Underscores are invalid in hostnames in DNS; fix the text not to imply they are valid.

Signed-off-by: Jason Bouzane jbouzane@gmail.com

PR #671, opened 2017-05-11
annotations: restrict character set of ref.name values

### LINKED ISSUE #599: spec: define charset for "org.opencontainers.ref.name"

vbatts:
Started from the URI charset, and ended up with a broader set and pattern from the docker tag. While not strictly the same, it's closer to the ideal set for containerd, flatpak and others.

Fixes #599

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -28,7 +28,16 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.revision** Source control revision identifier for the packaged software.
 * **org.opencontainers.image.vendor** Name of the distributing entity, organization or individual.
 * **org.opencontainers.image.licenses** License(s) under which contained software is distributed as an [SPDX License Expression][spdx-license-expression].
-* **org.opencontainers.image.ref.name** Name of the reference for a target (string). SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+* **org.opencontainers.image.ref.name** Name of the reference for a target (string).
+  * SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+  * Character set of the value SHOULD conform to alphanum of `A-Za-z0-9` and separator set of `-._:@/+`
+  * An EBNF'esque grammar + regular expression like:
+    ```
+    ref := component ["/" component]*
+    component := alphanum [separator alphanum]*
+    alphanum := /[A-Za-z0-9]+/
+    separator := /[-._:@+]/ | "--"
+    ```
 * **org.opencontainers.image.title** Human-readable title of the image (string)
 * **org.opencontainers.image.description** Human-readable description of the software packaged in the image (string)
 


erikh:
An unsolicited sgtm!

annotations.md
```diff
@@ -26,4 +26,6 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.revision** Source control revision identifier for packaged software.
 * **org.opencontainers.image.vendor** Name of the distributing entity, organization or individual.
 * **org.opencontainers.image.licenses** Comma-separated list of licenses under which contained software is distributed, in [SPDX Short identifier]https://spdx.org/licenses/] form.
-* **org.opencontainers.image.ref.name** Name of the reference for a target (string). SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+* **org.opencontainers.image.ref.name** Name of the reference for a target (string).
+  * SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+  * Character set of the value SHOULD conform to the [Uniform Resource Identifier (URI) RFC3986](https://tools.ietf.org/html/rfc3986).
```
wking: As this stands, it sounds like you're punting to the `URI` rule, but that requires a `scheme`, colon, and `hier-part`.  I doubt we want to recommend all names contain a colon.  Can we punt to `hier-part` instead?  Or maybe `segment-nz`?  @stevvooe mentioned “[schema-less URIs][1]”, perhaps he was thinking of `path-noscheme`?

[1]: https://github.com/opencontainers/image-spec/pull/591#issuecomment-285153069

erikh:
I wouldn't like that either; but it seems like this just covers the
character set?

wking:
On Thu, May 11, 2017 at 01:12:11PM -0700, Erik Hollensbe wrote:
> I wouldn't like that either; but it seems like this just covers the
> character set?

There are a number of charset rules in that ABNF (pchar, unreserved,
reserved, gen-delims, and sub-delims are all single-char choices).
Which one is this PR recommending?  The other rules look like they
have more structure, but the name values might be longer than a single
character, so a rule like segment-nz could cover a fair chunk of
useful names out of the box.  No slashes in segment-nz though, so if
you want those you have to go for something more complicated or
compose your own rule like:

  name  = 1*nchar
  nchar = pchar / reserved / gen-delims

annotations.md
stevvooe: This is quite clear that it only restricts it to the character set.

stevvooe:
LGTM

@vbatts Thank you for getting this done!

annotations.md
wking: > This is quite clear that it only restricts it to the character set.

Which character set is it restricting it to?  pchar?  unreserved?  Something else?  There are lots of possible choices in the linked collected ABNF.

AkihiroSuda:
```
  segment       = *pchar
   segment-nz    = 1*pchar
   segment-nz-nc = 1*( unreserved / pct-encoded / sub-delims / "@" )
                 ; non-zero-length segment without any colon ":"

   pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"

   query         = *( pchar / "/" / "?" )

   fragment      = *( pchar / "/" / "?" )

   pct-encoded   = "%" HEXDIG HEXDIG

   unreserved    = ALPHA / DIGIT / "-" / "." / "_" / "~"
   reserved      = gen-delims / sub-delims
   gen-delims    = ":" / "/" / "?" / "#" / "[" / "]" / "@"
   sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```

I guess `segment-nz-nc`? (excludes colon)

stevvooe:
Would it be better to just go with `\w+`?

AkihiroSuda:
`[\w-\.%]+` as far as I can come up with.

Also we need to look into what current Docker registry implementations allowing

vbatts:
@AkihiroSuda true true. cause this will map to tags

vbatts:
https://github.com/docker/distribution/blob/master/reference/regexp.go#L37

wking:
On Fri, May 12, 2017 at 06:42:47AM -0700, v1.0.0.batts wrote:
> https://github.com/docker/distribution/blob/master/reference/regexp.go#L37

Docker's current \w[\w.-]{0,127} is not enough space for some common
encodings.  Base 32 and 64 use = for padding [1], percent encoding
needs access to % [2].  Commonly-used encodings that will fit that
regexp are base 16 [3] and punycode [4].  You can use the ACE prefix
[5] as a way to label punycode, so I'd lean on RFC 3490 here and say
that names SHOULD be A-labels, U-labels, or NR-LDH labels.  There's
more discussion of RFC 3490 in #163 and #226.

[1]: https://tools.ietf.org/html/rfc4648
[2]: https://tools.ietf.org/html/rfc3986#section-2.1
[3]: https://tools.ietf.org/html/rfc4648#section-8
[4]: https://tools.ietf.org/html/rfc3492
     “… non-ASCII characters are represented by ASCII characters that
     are allowed in host name labels (letters, digits, and hyphens)."
[5]: https://tools.ietf.org/html/rfc3490#section-5

vbatts:
@AkihiroSuda this ABNF is terrible to read, but i think i agree on `segment-nz-nc` or even `segment-nz`, but then docker/moby would stop at the first occurence of an `:`

vbatts:
updated. PTAL.

stevvooe:
LGTM

annotations.md
```diff
@@ -26,4 +26,6 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.revision** Source control revision identifier for packaged software.
 * **org.opencontainers.image.vendor** Name of the distributing entity, organization or individual.
 * **org.opencontainers.image.licenses** Comma-separated list of licenses under which contained software is distributed, in [SPDX Short identifier]https://spdx.org/licenses/] form.
-* **org.opencontainers.image.ref.name** Name of the reference for a target (string). SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+* **org.opencontainers.image.ref.name** Name of the reference for a target (string).
+  * SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+  * Character set of the value SHOULD conform to the [Uniform Resource Identifier (URI) RFC3986](https://tools.ietf.org/html/rfc3986) [`segment-nz` (from appendix A)](https://tools.ietf.org/html/rfc3986#appendix-A).
```
wking: The canonical link for `segment-nz` is [section 3.3][1], appendix A just collects the ABNF together for ease-of-use.  If you want to keep the appendix reference, maybe “from appendix A” → “see appendix A”?

Also, `segment-nz` is a bigger charset than [Docker's tags][1].

[1]: https://github.com/opencontainers/image-spec/pull/671#issuecomment-301079786

[1]: https://tools.ietf.org/html/rfc3986#section-3.3

AkihiroSuda:
If we allow `@` here, implementations may cause security issues when parsing impl-specific ref string `example.com/foo/bar:baz@sha256:deadbeefdeadbeef`

AkihiroSuda:
So +1 for copypasting from the Docker registry implementation.

I'm not sure we should add % here.
One may put both `latest` and `l%d0%b0test` ( UI shows this as `lаtest` - Russian homoglyph) to single manifest, and may distribute malware with the latter ref name

vbatts:
@AkihiroSuda good points.

So we outline our own character set, or just leave this as `(string)`?

cyphar:
If we're going to restrict it, let's make it a URI fragment (the bit after `#`) which is defined in the URI RFC. The reason for this is that it will allow for URIs like `http://somedistributionthing#therefname` as a referencing style.

stevvooe:
Going from the ABNF, we have the following:

```
fragment      = *( pchar / "/" / "?" )
unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"
pchar         = unreserved / pct-encoded / sub-delims / ":" / "@"
pct-encoded   = "%" HEXDIG HEXDIG
sub-delims    = "!" / "$" / "&" / "'" / "(" / ")"
                 / "*" / "+" / "," / ";" / "="
```

In regexp, it is something like this:

```
fragment := /[A-Za-z0-9-._:@!$&"()*+,;=/?]+/
```

I would prefer something like this:

```
ref := /[A-Za-z0-9-._+]+/
```

The above is nearly identical to docker's tag, but allows `+`.

The alternative, with slashes, would be something like this:

```
ref := component ["/" component]*
component := alphanum [separator alphanum]*
alphanum := /[A-Za-z0-9]+/
separator := /[-._:@]/ | "--"
```

This allows docker references, docker tags, hostnames, punycode hosts, as well as a number of other things. The above grammar is regular, so it can just be a regexp. It does prevent some of the nastier parts of urls, like percent encoding.

cyphar:
I'm :+1: on anything that is a subset of `fragment` (so that tools can represent tags as fragments which feeds into my plans for distribution with https://github.com/cyphar/parcel).

vbatts:
@stevvooe cool that's pretty close. Is your example ABNF, regexp, or sttevveeoeooeoe expression? :stuck_out_tongue:

vbatts:
@stevvooe I don't get why for the separator to have `|"--"`

vbatts:
updated. PTAL

philips:
LGTM

stevvooe:
> @stevvooe I don't get why for the separator to have |"--"

Punycode!

> @stevvooe cool that's pretty close. Is your example ABNF, regexp, or sttevveeoeooeoe expression?

sttevveeoeooeoe expression. I think it might just be EBNF minus a colon and the regexp like character class syntax.

@vbatts The syntax proposed here allows things like `foo//bar` and `foo/-:yar`. Is that okay?

vbatts:
> Punycode!
Ok, well the regex provided allows for it, just not explicitly called out.

> @vbatts The syntax proposed here allows things like foo//bar and foo/-:yar. Is that okay?
terrible looking, but sure.

Also, as an example the flatpak usage looks like `app/org.my-app.App/x86_64/stable`.

stevvooe:
> Also, as an example the flatpak usage looks like `app/org.my-app.App/x86_64/stable`

How would such a reference be imported into Docker?

Typically, we have disallowed things like `foo/../bar` to prevent the security problems that might arise when these get cleaned before doing lookups on a filesystem.

vbatts:
@stevvooe there is no assumptions about how it would get loaded into docker.
As for the path cleaning, that is done on a tag? That affects filesystem lookups?

vbatts:
updated. please review

AkihiroSuda:
Can you please add:

Implementors' note: following strings conform to the recommendation but implementors might need consideration for security:
- foo/../../bar
- foo@sha256:deadbeefdeadbeefdeadbeef deadbeefdeadbeefdeadbeef..

vbatts:
@AkihiroSuda how come `foo/../../bar` is a security concern? Due to REST APIs?

stevvooe:
@vbatts Moving this off the rc6 milestone.

vbatts:
@stevvooe should there be language for implementations to use their own custom annotation for something like a docker:tag?

vbatts:
the concern here is that this field is most closely tied to the docker "tag" and that this current language is incompatible. It is not the intention to make something incompatible, but rather to describe the ideal for generic use.

wking:
On Wed, May 24, 2017 at 09:05:19AM -0700, v1.0.0.batts wrote:
> the concern here is that this field is most closely tied to the
> docker "tag" and that this current language is incompatible.

It's not compatible if you forbid encoding, but it could be compatible
if Docker implements OCI import/export by encoding/decoding ref.name
in Punycode [1] (possibly conditionally with the ACE prefix [2]).  But
then places that display the (encoded) Docker tag would have to be
updated too, or they could end up showing
‘xn--PorqunopuedensimplementehablarenEspaol-fmd56a’ instead of
‘PorquénopuedensimplementehablarenEspañol’ [3, example J].  That
helps, but Punycode doesn't address backslashes and such [3, example
S, and 4].  A non-Punycode Bootstring implemenation [5] may be able to
give more complete coverage (e.g. for flatpack's
‘app/org.my-app.App/x86_64/stable’ [6]), but getting support out of
the box would be harder.

So not trivial, but there is room for divergence between Docker's
current tag charset and what gets specified here if the Docker folks
are ok with encoding (when necessary).  I'm not sure if that is the
case or not.

[1]: https://tools.ietf.org/html/rfc3492
[2]: https://tools.ietf.org/html/rfc3490#section-5
[3]: https://tools.ietf.org/html/rfc3492#section-7.1
[4]: https://tools.ietf.org/html/rfc3492#section-5
     “The basic code points are the ASCII [ASCII] code points (0..7F)…”
[5]: https://tools.ietf.org/html/rfc3492#section-6
[6]: https://github.com/opencontainers/image-spec/pull/671#issuecomment-302508587

vbatts:
perhaps we(i) should include @stevvooe grammar here, 
```
ref := component ["/" component]*
component := alphanum [separator alphanum]*
alphanum := /[A-Za-z0-9]+/
separator := /[-._:@/+]/ | "--"
```

annotations.md
```diff
@@ -28,7 +28,16 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.revision** Source control revision identifier for the packaged software.
 * **org.opencontainers.image.vendor** Name of the distributing entity, organization or individual.
 * **org.opencontainers.image.licenses** License(s) under which contained software is distributed as a [SPDX License Expression][spdx-license-expression].
-* **org.opencontainers.image.ref.name** Name of the reference for a target (string). SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+* **org.opencontainers.image.ref.name** Name of the reference for a target (string).
+  * SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+  * Character set of the value SHOULD conform to alphanum of `A-Za-z0-9` and separator set of `-._:@/+`
+  * An EBNF + regular expression like:
+```
```
stevvooe: Indent this block so it hangs under the bullet point.

stevvooe:
LGTM

vbatts:
updated the commit to address @AkihiroSuda concerns

vbatts:
@opencontainers/image-spec-maintainers please review

annotations.md
```diff
@@ -28,7 +28,16 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.revision** Source control revision identifier for the packaged software.
 * **org.opencontainers.image.vendor** Name of the distributing entity, organization or individual.
 * **org.opencontainers.image.licenses** License(s) under which contained software is distributed as a [SPDX License Expression][spdx-license-expression].
-* **org.opencontainers.image.ref.name** Name of the reference for a target (string). SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+* **org.opencontainers.image.ref.name** Name of the reference for a target (string).
+  * SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
+  * Character set of the value SHOULD conform to alphanum of `A-Za-z0-9` and separator set of `-._:@/+`
+  * An EBNF + regular expression like:
```
wking: Why use EBNF (presumably [ISO/IEC 14977:1996][1]) instead of ABNF ([RFC 5234][2])?  ABNF is much more available than EBNF's [restrictively-licensed zipped PDF][3].  ABNF covers the same bits you need here ([`=` for rule definition][4] and [`/` for alternates][5]).  For folks who prefer to read EBNF, that spec defines `/` as an alternative definition-separator-symbol (with `|` being the canonical EBNF definition-separator-symbol).

[1]: https://www.iso.org/standard/26153.html
[2]: https://tools.ietf.org/html/rfc5234
[3]: http://standards.iso.org/ittf/PubliclyAvailableStandards/s026153_ISO_IEC_14977_1996%28E%29.zip
[4]: https://tools.ietf.org/html/rfc5234#section-2.2
[5]: https://tools.ietf.org/html/rfc5234#section-3.2

annotations.md
```diff
@@ -28,7 +28,16 @@ This specification defines the following annotation keys, intended for but not l
...
+    ```
+    ref := component ["/" component]*
```
wking: Both [ABNF][1] and [EBNF][2] (§4.3 and table 1) use `=` for rules, so I'd recommend `:=` → `=`.

Both [ABNF][3] and [EBNF][2] (§4.12) use a *prefixed* `*` for repetition (not a suffixed `*`), so I'd recommend `["/" component]*` → `*["/" component]`.

[1]: https://tools.ietf.org/html/rfc5234#section-2.2
[2]: http://standards.iso.org/ittf/PubliclyAvailableStandards/s026153_ISO_IEC_14977_1996%28E%29.zip
[3]: https://tools.ietf.org/html/rfc5234#section-3.6

annotations.md
wking: Ah, reason for using `|` for alternates is probably so we can use `/…/` for marking regexps.  Since regexps are part of neither ABNF nor EBNF, I'd rather just drop the regexps.  I think:

```ABNF
ref = component *("/" component)
component = alphanum *(separator alphanum)
alphanum = 1*(ALPHA DIGIT)
separator = "-" / "." / "_" / ":" / "@" / "+" / "--"
```

(which leans on the [core rules][1] for `ALPHA` and `DIGIT`) is about as readable as the EBNF-ish + regexp syntax you're using in 5d36c514, and it has the additional benefit of using a [vanilla syntax][2].

[1]: https://tools.ietf.org/html/rfc5234#appendix-B.1
[2]: https://tools.ietf.org/html/rfc5234

wking: In [XML 1.0's EBNF][1], this would be:

```
ref ::= component ("/" component)*
component ::= alphanum (separator alphanum)*
alphanum ::= [a-zA-Z0-9]+
separator ::= [-._:@+] | "--"
```

which is close to the current PR.  I think the XML EBNF spec is harder to read than the RFC 5234 or ISO/IEC 14977:1996 (although it's certainly shorter), and I'd rather lean on something stand-alone than a part of the XML spec, but even with those caveats I'd rather use the XML EBNF than rolling our own unspecified EBNF-ish.

[1]: https://www.w3.org/TR/REC-xml/#sec-notation

vbatts:
based on https://godoc.org/golang.org/x/exp/ebnf and https://github.com/github/linguist/blob/master/lib/linguist/languages.yml#L1030  going with EBNF would look like this:
```EBNF
letter = "A" | "B" | "C" | "D" | "E" | "F" | "G"
   | "H" | "I" | "J" | "K" | "L" | "M" | "N"
   | "O" | "P" | "Q" | "R" | "S" | "T" | "U"
   | "V" | "W" | "X" | "Y" | "Z" | "a" | "b"
   | "c" | "d" | "e" | "f" | "g" | "h" | "i"
   | "j" | "k" | "l" | "m" | "n" | "o" | "p"
   | "q" | "r" | "s" | "t" | "u" | "v" | "w"
   | "x" | "y" | "z" ;
digit = "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" ;
alphanum = letter | digit, { letter | digit } ;
separator = "-" | "." | "_" | ":" | "@" | "+" | "--" ;
component = alphanum, { separator | alphanum, alphanum } ;
ref = component, { "/", component } ;
```
This way we could have validation on the grammar, and pretty highlighting for github-displayed code blocks.

There are some ways that ABNF looks nicer, as it has default variables like `DIGIT` and `ALPHA`. Also it has linguist highlighting. Honestly, I'm just not down with the ABNF `*( ... )` and ` foo / bar` notation. :-\

vbatts:
rebased, and went with EBNF. Please review.

annotations.md
```diff
@@ -28,7 +28,25 @@ This specification defines the following annotation keys, intended for but not l
...
+  * An EBNF grammar like:
+    ```EBNF
+    letter = "A" | "B" | "C" | "D" | "E" | "F" | "G"
+       | "H" | "I" | "J" | "K" | "L" | "M" | "N"
+       | "O" | "P" | "Q" | "R" | "S" | "T" | "U"
+       | "V" | "W" | "X" | "Y" | "Z" | "a" | "b"
+       | "c" | "d" | "e" | "f" | "g" | "h" | "i"
+       | "j" | "k" | "l" | "m" | "n" | "o" | "p"
+       | "q" | "r" | "s" | "t" | "u" | "v" | "w"
+       | "x" | "y" | "z" ;
```
wking: The [Go docs][1] have:

    Production  = name "=" [ Expression ] "." .

which is using `.` (and not `;`) as the rule terminator.

[1]: https://godoc.org/golang.org/x/exp/ebnf

annotations.md
```diff
@@ -28,7 +28,25 @@ This specification defines the following annotation keys, intended for but not l
...
+    digit = "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" ;
+    alphanum = letter | digit, { letter | digit } ;
```
wking: The [Go docs][1] aren't particularly clear on this, but you seem to be using `,` for concatenation and they don't seem to use any character for concatenation.  I think it would be better to use:

    alphanum = (letter | digit) {letter | digit} .


[1]: https://godoc.org/golang.org/x/exp/ebnf

annotations.md
```diff
@@ -28,7 +28,25 @@ This specification defines the following annotation keys, intended for but not l
...
+    separator = "-" | "." | "_" | ":" | "@" | "+" | "--" ;
+    component = alphanum, { separator | alphanum, alphanum } ;
```
wking: This line will allow component to contain consecutive separators.  I think you want:

    word = alphanum {alphanum} .
    component = word {separator word} .

or some such.

vbatts:
fair, the golang EBNF is different than what things like wikipedia show...
To get something that validates with `ebnflint -start ref ...` would be:
```
<pre class="ebnf">
ref = component { ( "/" component ) } .
component = alphanum { ( separator | alphanum ) alphanum } .
alphanum = ( letter | digit ) { letter | digit } .
separator = "-" | "." | "_" | ":" | "@" | "+" | "--" .
letter = "A" | "B" | "C" | "D" | "E" | "F" | "G"
   | "H" | "I" | "J" | "K" | "L" | "M" | "N"
   | "O" | "P" | "Q" | "R" | "S" | "T" | "U"
   | "V" | "W" | "X" | "Y" | "Z" | "a" | "b"
   | "c" | "d" | "e" | "f" | "g" | "h" | "i"
   | "j" | "k" | "l" | "m" | "n" | "o" | "p"
   | "q" | "r" | "s" | "t" | "u" | "v" | "w"
   | "x" | "y" | "z" .
digit = "0" | "1" | "2" | "3" | "4" | "5" | "6" | "7" | "8" | "9" .
</pre>
```

wking:
On Thu, Jun 01, 2017 at 11:14:25AM -0700, v1.0.0.batts wrote:
> ref = component { ( "/" component ) } .

You don't need the parens here, because the braces are already
grouping their content.

> component = alphanum { ( separator | alphanum ) alphanum } .

This is overly complicated, because the alphanum case is:

  component = alphanum { alphanum alphanum } .

but alphanum already allows multiple characters.  So all you need is
the separator case [1]:

  component = alphanum { separator alphanum } .

[1]: https://github.com/opencontainers/image-spec/pull/671#discussion_r119664748

vbatts:
it's due to not ending the name with a seperator, but ending with an alphanum

stevvooe:
@vbatts I'd rather we merge this with the existing grammar and then make the change (or not) across the entire specification in a single pass. There are many other places in the specification that would need to be updated.

Looks like some of this syntax is picked up from the [PEG paper](http://www.romanredz.se/papers/FI2007.pdf). You can read up on PEGs at [this website](http://bford.info/packrat/). It is also based on the myriad combinator-based parsing libraries. PEGs are generally better for this as they don't allow ambiguity.

wking:
On Thu, Jun 01, 2017 at 12:54:58PM -0700, v1.0.0.batts wrote:
> it's due to not ending the name with a seperator, but ending with an
> alphanum

This:

  component = alphanum { separator alphanum } .

has to end in an alphanum (which I think is what you're aiming at
[1]).  You're allowed to append any number of ‘separator alphanum’
bits to the initial alphanum.

[1]: https://github.com/opencontainers/image-spec/pull/671#issuecomment-302251178

wking:
> Looks like some of this syntax is picked up from the [PEG paper][1].

The PEG papers seem more focused on the mathematical properties of PEGs and less focused on the specific instances (e.g. which character(s) to use for definitions).  But based on [this][2], the PEG formulation of the current rule is something like:

```
ref <- component ("/" component)*
component <- alphanum (separator / alphanum)*
alphanum <- (letter | digit)+
separator <- "--" / "-" / "." / "_" / ":" / "@" / "+"
letter <- [A-Za-z]
digit <- [0-9]
```

which looks a lot like the [ABNF formulation][3].  The PEG form is a bit less ambiguous, because you don't have the “is this next `-` part of the `--` or `-` separator?” issue.  But that's not a problem for us because we never allow consecutive separators and no other term could start or end with a hyphen.

Another issue with PEG (although no different from ABNF in this regard) is that it uses `/` for choices which @vbatts [does not like][4].  Of the currently-proposed syntaxes, only [Go's][5] and [XML's][6] use `|`.

[1]: http://www.romanredz.se/papers/FI2007.pdf
[2]: http://bford.info/pub/lang/peg.pdf
[3]: https://github.com/opencontainers/image-spec/pull/671#discussion_r119452588
[4]: https://github.com/opencontainers/image-spec/pull/671#issuecomment-305523424
[5]: https://godoc.org/golang.org/x/exp/ebnf
[6]: https://www.w3.org/TR/REC-xml/#sec-notation

vbatts:
git reflog to the rescue.
rolled back to the EBNF'esque + regex grammar.
Please review.

vbatts:
(hence, this whole thing of picking a particular grammar is a bit silly, as i've found that unless you define your own dialect in full, like XML have, then you can't just rely on saying EBNF, etc)

wking:
On Fri, Jun 02, 2017 at 11:51:51AM -0700, v1.0.0.batts wrote:
> (hence, this whole thing of picking a particular grammar is a bit
> silly, as i've found that unless you define your own dialect in
> full, like XML have, then you can't just rely on saying EBNF, etc)

There's no problem with referencing someone else who has defined a
whole dialect.  For example:

* Parsing Expression Grammars: A Recognition-Based Syntactic
  Foundation. (PDF). Bryan Ford, POPL, January 2004.
  http://bford.info/pub/lang/peg
  http://bford.info/pub/lang/peg.pdf

* RFC 5234 https://tools.ietf.org/html/rfc5234

* XML 1.0's EBNF https://www.w3.org/TR/REC-xml/#sec-notation

* ISO/IEC 14977:1996
  https://www.iso.org/standard/26153.html
  http://standards.iso.org/ittf/PubliclyAvailableStandards/s026153_ISO_IEC_14977_1996%28E%29.zip

* Go's EBNF https://godoc.org/golang.org/x/exp/ebnf

All of those are explicit about what characters to use for the various
grammer components, although some are more detailed and/or easier
reads than others.  There are also parser-generators to get you from
grammars defined in the various *BNF flavors to machine code.  For
example peg(1) [1] will compile my PEG formulation [2] (with a typo
fix) to C:

  $ cat ref.peg
  start <- component ("/" component)*
  component <- alphanum (separator alphanum)*
  alphanum <- (letter / digit)+
  separator <- "--" / "-" / "." / "_" / ":" / "@" / "+"
  letter <- [A-Za-z]
  digit <- [0-9]
  $ peg ref.peg >ref.c
  $ cat <<EOF >>ref.c
  > int main(void) {
  >   if (yyparse()) {
  >     printf("found a valid ref\n");
  >     if (fgetc(stdin) != EOF) {
  >       printf("some characters were not consumed\n");
  >       return 1;
  >     }
  >     return 0;
  >   }
  >   printf("did not find a valid ref\n");
  >   return 1;
  > }
  > EOF
  $ cc -o ref ref.c
  $ echo hello | ./ref
  found a valid ref
  $ echo hello/world | ./ref
  found a valid ref
  $ echo hello/world--foo | ./ref
  found a valid ref
  $ echo hello/world--foo--bar | ./ref
  found a valid ref
  $ echo hello/world--foo~bar | ./ref
  found a valid ref
  some characters were not consumed
  $ echo ~hello | ./ref
  did not find a valid ref

[1]: http://piumarta.com/software/peg/
[2]: https://github.com/opencontainers/image-spec/pull/671#issuecomment-305838174

stevvooe:
> (hence, this whole thing of picking a particular grammar is a bit silly, as i've found that unless you define your own dialect in full, like XML have, then you can't just rely on saying EBNF, etc)

I wish I could have put this as elegantly as you. This is basically why we just went with the syntax that is in place.

stevvooe:
LGTM

jonboulle:
LGTM conditional on something like #692 going in; we can definitely do better than "EBNF-esque"

PR #678, opened 2017-05-17
annotations: addressing comments

vbatts:
From: https://github.com/opencontainers/image-spec/issues/658#issuecomment-301494825

* Source is not just to binaries.
* the strings already say URL, so it's redundant
* git:// is a valid URL as well

Signed-off-by: Vincent Batts <vbatts@hashbangbash.com>

PR #680, opened 2017-05-18
annotations: Use SPDX License Expressions for licenses

wking:
Instead of comma-separated short identifiers, which have unclear semantics (are the delimiters `AND` or `OR`?).  I don't see any discussion of the syntax for this field in #636 (which landed it), but I'd floaded license expressions before in the sub-thread starting [here][2].  @gregkh had pushed back against my earlier proposal (licensing information on descriptors) [with][3]:

> No, that's not going to work at all, you can't properly describe the license for a whole layer in any form of a string that could be standardized or parsed. SPDX is great for describing the individual licenses of things, but not for a collection of things, which almost always has an arbitrary license (example, what's the license, in a simple string, for a Ubuntu base layer?)

But SPDX License Expressions are both more expressive and better defined than the current comma delimiters.  Everything you could have said with the comma-delimited string you can say more clearly with a SPDX License Expression.  And because the syntax is not OCI-specific, you're more likely to be able to find tooling that handles these values out of the box.

There are other annotation adjustments in flight with #678, but that's not currently touching the `licenses` entry.

[2]: https://github.com/opencontainers/image-spec/pull/501#discussion_r96761939
[3]: https://github.com/opencontainers/image-spec/pull/501#discussion_r97056075

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -25,7 +25,7 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.version** [Semantic versioning-compatible](http://semver.org/) version of the packaged software. The version MAY match a label or tag in the source code repository.
 * **org.opencontainers.image.revision** Source control revision identifier for the packaged software.
 * **org.opencontainers.image.vendor** Name of the distributing entity, organization or individual.
-* **org.opencontainers.image.licenses** Comma-separated list of licenses under which contained software is distributed, in [SPDX Short identifier](https://spdx.org/licenses/) form.
+* **org.opencontainers.image.licenses** License(s) under which contained software is distributed as a [SPDX License Expression][spdx-license-expression].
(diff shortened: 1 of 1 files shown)


vbatts:
I remember greg saying that. I'm not familiar with spdx-expressions and would wager that many would have to get familiarized with it.
All in all it is clearer than just comma-delimited.
:+1:

wking:
> I'm not familiar with spdx-expressions and would wager that many would have to get familiarized with it.

That SPDX appendix isn't very long ;).  Reading through it a few times seems better than inventing an OCI-specific syntax that is more limited.  If maintainers *want* to limit what you can say (for some reason), then making `licenses` ` AND ` (or ` OR `) delimeted would preserve the current semantics as a subset of SPDX License Expressions.

> All in all it is clearer than just comma-delimited.

So "I'm leaning towards LGTM but want to think about it more and/or hear what others think"?  Or is there some way I can adjust the wording to make you happier now?

vbatts:
:+1: == LGTM

philips:
LGTM

PR #690, opened 2017-05-31
image-index: add armv5 to the accepted ARM variants

imrehg:
Go still supports armv5, and should include all the accepted
architectures in the variants.

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -71,6 +71,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
         | ISA/ABI         | `architecture` | `variant`   |
         |-----------------|----------------|-------------|
+        | ARM 32-bit, v5  | `arm`          | `v5`        |
         | ARM 32-bit, v6  | `arm`          | `v6`        |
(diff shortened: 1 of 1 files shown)


vbatts:
i'm mostly indifferent, because nothing it stopping from using these values in these fields without it being in the spec.

stevvooe:
ARMv5 was left out due to the variation in architectures. I am unclear on the exact details, but I think we can let @mattspencer-arm comment further.

vbatts:
ping @mattspencer-arm

mattspencer-arm:
@vbatts I was under the impression that this was not a 'what is acceptable' list as details or architecture are pulled in from golang.  This list is just to represent how the data should be formatted.  To this end, I don't think we need to have a v5 line in there as the progression should be obvious?

vbatts:
@mattspencer-arm That is my opinion too. 

@imrehg not having this in the spec does not block `v5` as a value, and should not block the work that resin.io does. By choosing to support the `v5` arch requires maintaining the support matrix of the cpu features/variants. That is the part we are not detailing in this spec.

I'm going to close this for now as it is not required and not blocking any features.

PR #696, opened 2017-06-09
schema/validator: Punt sha256/sha512 hex-validation to go-digest

wking:
It's covered this since opencontainers/go-digest#34, so we don't need to do it locally.  Also, use a more specific check for ignoring unrecognized algorithms, because we don't want to return nil when go-digest notices an invalid encoding, etc.

This fixes issues like [this][1]:

    -- FAIL: TestDescriptor (0.09s)
      descriptor_test.go:300: test 14: expected validation failure true but got false, err <nil>

which have been affecting recent unit tests on unrelated PRs.

[1]: https://travis-ci.org/opencontainers/image-spec/builds/241019584#L373

### DIFF
--- a/schema/validator.go
+++ b/schema/validator.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"regexp"
 
 	digest "github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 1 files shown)


stevvooe:
LGTM

Wow, that code was wrong before.

vbatts:
since go-digest is under the oci org then it seems fine to delegate this check.
LGTM

PR #700, opened 2017-06-22
Makefile: use make builtin rather than shell

vbatts:
Since READLINK(1) varies across platforms, namely darwin version does
not have the `-f` flag, then don't rely on the shell. Especially since
gnu make already has this as a builtin function.


philips:
LGTM

PR #702, opened 2017-06-30
layer: clarify that not all file types need be supported

jstarks:
As I mentioned in #415, some file types (sockets, block devices, character devices, FIFOs) are not currently supported on Windows. This change intends to clarify that not all file types are supported on all platforms without being too prescriptive.


vbatts:
LGTM

PR #703, opened 2017-06-30
layer: document Windows-specific file attributes

jstarks:
Windows relies on additional file attributes to produce a working image. This change documents the encoding of the necessary fields (which seems particularly important because these vendor-specific extensions are not currently documented anywhere else).


jstarks:
(As discussed briefly in #415)

vbatts:
SGTM
LGTM

PR #711, opened 2017-07-16
conversion: add platform conversions

cyphar:
In b6d5a8cc1562 ("Change platform ref from runtime-spec"), the
conversion to runtime-spec for the "os" and "architecture" fields was
removed (as the fields had also been removed in runtime-spec). Re-add
the conversion as an annotation field rather than a verbatim field.

[Obviously this can't go into `1.0.0` due to the time window, but we can put it in `1.0.1`].

### DIFF
--- a/conversion.md
+++ b/conversion.md
@@ -44,15 +44,19 @@ These fields all affect the `annotations` of the runtime configuration, and are 
 
 | Image Field         | Runtime Field   | Notes |
 | ------------------- | --------------- | ----- |
-| `author`            | `annotations`   | 1,2   |
-| `created`           | `annotations`   | 1,3   |
+| `os`                | `annotations`   | 1,2   |
+| `architecture`      | `annotations`   | 1,3   |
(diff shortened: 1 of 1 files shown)


cyphar:
/ping @opencontainers/image-spec-maintainers -- this really should've gone into `1.0.1`...

stevvooe:
Nothing controversial here, but can these be `MUST` after a 1.0 release?

philips:
I don't think they can be a MUST.

cyphar:
Are we not allowed to MUST new features in new releases? Also, this document's "MUST" language refers to the "default generated runtime configuration" which in principle could be modified to remove these annotations -- so a runtime that does the wrong thing here isn't non-compliant.

cyphar:
@philips @stevvooe ?

stevvooe:
LGTM

@cyphar Not sure if these can be a MUST, still.

PR #270, opened 2016-09-07
remove temporary directory for unit test with TestUnpackLayerDuplicateEntries

TrumanLing:
Signed-off-by: Ling FaKe <lingfake@huawei.com>

PR #716, opened 2017-07-28
clarification about nested index

AkihiroSuda:


jonjohnsonjr:
What is meant by required? For whom?

There is pretty specific language here:
https://github.com/opencontainers/image-spec/blob/master/image-index.md

I don't see any reason to forbid an image index referencing another image index, but I'm curious what the use case would be.

AkihiroSuda:
https://github.com/opencontainers/image-spec/blob/9e2ed30b411505b510b4ab080c4035d71bd56b44/image-layout.md says "This REQUIRED file"

jonjohnsonjr:
Ah, thanks! My apologies, please ignore me :)

I guess I don't quite understand what is meant by "image providers" here:
> While the use of an image index is OPTIONAL for image providers, image consumers SHOULD be prepared to process them.

Does that just refer to distribution in general? (I don't mean to hijack your PR but it seems like the `<<optional>>` bit traces back to this sentence, so it might be worth clarifying/removing as part of this PR?)

jonboulle:
I also don't understand the use case for an image index to reference another image index. Notably this is not called out here in the actual index specification: https://github.com/opencontainers/image-spec/blame/master/image-index.md#L35

> https://github.com/opencontainers/image-spec/blame/9e2ed30b411505b510b4ab080c4035d71bd56b44/image-layout.md#L148

I'm wondering if this line snuck in accidentally as the original context is very different:
https://github.com/opencontainers/image-spec/commit/9bb56d8ea0a0eff3e051ec9c737347f691208445#diff-99bfcb610f238b942638d49d1ff2eaf4R46

Here it refers to blobs in the ref directory, not the actual refs listed in the index file.

AkihiroSuda:
@jonboulle

Yes, I'm +1 for removing nested index if it was introduced by accident.
WDYT about necessity of index?

jonboulle:
I don't see a need for it but could be proven wrong.

AkihiroSuda:
opened #718 and #719

AkihiroSuda:
#718 and #719 were rejected.

Let me reopen this PR with updated docs.

AkihiroSuda:
@jonboulle @stevvooe @vbatts @philips

image-index.md
```diff
@@ -2,6 +2,7 @@
 
 The image index is a higher-level manifest which points to specific [image manifests](manifest.md), ideal for one or more platforms.
 While the use of an image index is OPTIONAL for image providers, image consumers SHOULD be prepared to process them.
+Also, the top-level index (`index.json`) is a REQUIRED file in [the image layout](image-layout.md).
```
wking: I'd rather not have this line.  This file/section is about defining the `application/vnd.oci.image.index.v1+json` media type, and should not be about where that media type is used.

I'd also encourage dropping the previous line for the same reason.  Instead, requirements about what providers may/should/must supply and what consumers may/should/must consume should be handled in the referencing location.  For example, you can see that [`application/vnd.oci.image.index.v1+json` consumers MUST handle `application/vnd.oci.image.manifest.v1+json` in the context of `manifests`][1].  And you'd see that consumers of image-layout v1.0.0 MUST handle `application/vnd.oci.image.index.v1+json` in the context of `index.json` around [here][2] (what I think you should be adding in this PR).  And you'd see that consumers need not handle v1.0.0 image layouts [here][3] (which I think you should also be adding in this PR).

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0/image-index.md#L33-L35
[2]: https://github.com/opencontainers/image-spec/blame/v1.0.0/image-layout.md#L29
[3]: https://github.com/opencontainers/image-spec/blame/v1.0.0/image-layout.md#L1-L10

image-layout.md
```diff
@@ -149,6 +149,7 @@ This index provides an established path (`/index.json`) to have an entry point f
 * Future versions of the spec MAY use a different mediatype (i.e. a new versioned format).
 * An encountered `mediaType` that is unknown SHOULD be safely ignored.
 
+Note that while `index.json` is a REQUIRED file, the use of an image index is OPTIONAL for image providers.
```
wking: I'd shift this line up to the [opening section][1] and rephrase it to something like:

> Consumers MAY support OCI Image Layout and MAY also support other mechanisms for distributing image content.

The optionality of image-layout support is important enough that I'd rather not bury it 150 lines into the file ;).

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0/image-layout.md#L1-L10

image-layout.md
```diff
@@ -199,7 +200,11 @@ Those tags will often be represented in an image-layout repository with matching
 }
 ```
 
-This illustrates an index that provides two named manifest references and an auxiliary mediatype for this image layout.
+This illustrates an index that provides two named references and an auxiliary mediatype for this image layout.
 
+The first named reference (`"stable-release"`) points to another index that might contain multiple references with several platform specification and annotations.
```
wking: Backticks *and* quotes seems like overkill.  Can we use `stable-release` here (and `v1.0` below)?  Those are the names; the quotes are just part of representing strings in JSON.

Also probably “several platform sepecification” → “distinct platforms” or some such.

image-layout.md
```diff
@@ -199,7 +200,11 @@ Those tags will often be represented in an image-layout repository with matching
...
+Note that the `org.opencontainers.image.ref.name` annotation SHOULD only be considered valid when on descriptors on `index.json`.
```
wking: Maybe link to `annotations.md` for this so folks can see that we're referencing a requirement defined there.

image-layout.md
stevvooe: We already state this in a clear manner in another doc.

stevvooe:
@AkihiroSuda It would be good if you could file an issue stating the problem before filing PRs. I am not sure that these are correct and they seem redundant with other statements that are already present.

What is the problem that you are trying to address here?

image-index.md
stevvooe: Yes, this doesn't belong in this file.

AkihiroSuda:
Opened #270 for the problem: The fact that "Image Layout is OPTIONAL" is not explicitly documented and causing confusion about necessity of `index.json`.

I modified this PR to limit its scope to nested index examples. PTAL.

image-layout.md
```diff
@@ -199,7 +199,11 @@ Those tags will often be represented in an image-layout repository with matching
 }
 ```
 
-This illustrates an index that provides two named manifest references and an auxiliary mediatype for this image layout.
+This illustrates an index that provides two named references and an auxiliary mediatype for this image layout.
 
+The first named reference (`stable-release`) points to another index that might contain multiple references with distinct platforms and annotations.
+Note that the [`org.opencontainers.image.ref.name` annotation](annotations.md) SHOULD only be considered valid when on descriptors on `index.json`.
```
AkihiroSuda: Note: This redundant sentence is intentionally added for better human understandability.

image-layout.md
```diff
@@ -199,7 +199,11 @@ Those tags will often be represented in an image-layout repository with matching
...
+
+The second named reference ("`v1.0`") points to a manifest that is specific to the linux/ppc64le platform.
```
wking: You still have [both backticks and quotes][1] here.

[1]: https://github.com/opencontainers/image-spec/pull/716#discussion_r131225451

AkihiroSuda:
Fixed quotation

img/media-types.dot
```diff
@@ -6,6 +6,7 @@ digraph G {
     layer [shape=note, label="Layer tar archive\napplication/vnd.oci.image.layer.v1.tar\napplication/vnd.oci.image.layer.v1.tar+gzip\napplication/vnd.oci.image.layer.nondistributable.v1.tar\napplication/vnd.oci.image.layer.nondistributable.v1.tar+gzip"]
   }
 
+  imageIndex -> imageIndex [label="1..*"]
```
wking: If we're going to draw this loop in here, I think we also want to SHOULD support for indexes [here][1] (for both providers and consumers), at least for the index in `index.json`.

[1]: https://github.com/opencontainers/image-spec/blame/v1.0.0/image-index.md#L30

AkihiroSuda:
Updated, anyone PTAL?

AkihiroSuda:
lint failure on travis is unrelated: https://github.com/opencontainers/image-spec/issues/735

stevvooe:
LGTM

The main use case I see is an archive that has multiple multi-platform images.

AkihiroSuda:
ping @opencontainers/image-spec-maintainers

cyphar:
LGTM.

/cc @opencontainers/image-spec-maintainers

vbatts:
LGTM

PR #719, opened 2017-08-02
image-layout: make index.json OPTIONAL

AkihiroSuda:
The index.json is OPTIONAL, according to the description in
image-index.md and media-types.dot.

discussed in https://github.com/opencontainers/image-spec/pull/716

### DIFF
--- a/image-layout.md
+++ b/image-layout.md
@@ -139,7 +139,7 @@ This section defines the `application/vnd.oci.layout.header.v1+json` [media type
 
 ## index.json file
 
-This REQUIRED file is the entry point for references and descriptors of the image-layout.
+This OPTIONAL file is the entry point for references and descriptors of the image-layout.
 The [image index](image-index.md) is a multi-descriptor entry point.
 
 This index provides an established path (`/index.json`) to have an entry point for an image-layout and to discover auxiliary descriptors.


jonboulle:
index support _within_ a layout is required, otherwise a layout doesn't make much sense.
I agree it's a little awkward. This is consistent with the image index being optional in media-types.dot

Maybe you could add a line [here](https://github.com/opencontainers/image-spec/blame/master/image-index.md#L4) saying that index support is necessary for implementations working with layouts. Or we just strike that line. the provider/consumer distinction is unclear.

wking:
More previous discussion [here][1].  Note that even though `index.json` is currently required, you don't have to list any references inside it.

[1]: https://github.com/opencontainers/image-spec/pull/533#discussion_r99419341

stevvooe:
What does this achieve? While refs are one function of the index, without an index, it is fairly unsafe to be inferring types. Is there something that cannot be with a required index?

stevvooe:
I think I misread this initially.

The current wording is correct. `index.json` is required.

Remember that layouts are not the primary transport of OCI by any means, so make sure to interpret the statements under the assumption that there is a blob access model. To be clear, https://github.com/opencontainers/image-spec/blame/master/image-index.md#L4 says that images don't have to be packed into indexes, but they should be able to support consumption of indexes.

Moving for close here.

philips:
I think I agree this is something to close.

stevvooe:
@AkihiroSuda I'm going to go ahead and close this.

If there is something I've missed here, please reach out!

PR #742, opened 2018-02-27
Fix the schema validation API

### LINKED ISSUE #686: Fix the schema validation API

q384566678:
Fix #686 
In image-tools, warn is handled as follows:

```
	for _, e := range err.Warns {
			logrus.Warn(e)
	}
...
}
```

Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>


q384566678:
This is a preliminary idea, and if it works, I will continue.

I also have the idea of using `logrus.warnf` to handle warnings directly.
@stevvooe @vbatts @wking @xiekeyang PTAL

vbatts:
_house keeping:_ i *think* this is okay, but i'm not sure who all this will break with such a change. This makes me inclined to do nothing.

vbatts:
cleaning up as there is no further conversation

PR #745, opened 2018-04-17
Allow for multiple authors

taqtiqa-mark:
Arises where a container is a composition of several containers. See Buildah issue #591 specifically the current use case given in [comment](https://github.com/projectatomic/buildah/issues/591#issuecomment-381722873)

### DIFF
--- a/config.md
+++ b/config.md
@@ -98,7 +98,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
 
 - **author** *string*, OPTIONAL
 
-  Gives the name and/or email address of the person or entity which created and is responsible for maintaining the image.
+  Gives the names and/or email addresses of the persons or entities which created and are responsible for maintaining the image. Multiple names should be comma delimted.
 
 - **architecture** *string*, REQUIRED
 


taqtiqa-mark:
Will add "Signed-off-by"

wking:
You can already use the free-form [`org.opencontainers.image.authors`][1] anywhere that takes annotations (which is at least [indexes][2], [manifests][3], and [descriptors][4]).  I think that covers all image-spec Merkle trees.

[1]: https://github.com/opencontainers/image-spec/blob/v1.0.1/annotations.md#pre-defined-annotation-keys
[2]: https://github.com/opencontainers/image-spec/blob/v1.0.1/image-index.md#image-index-property-descriptions
[3]: https://github.com/opencontainers/image-spec/blob/v1.0.1/manifest.md#image-manifest-property-descriptions
[4]: https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#properties

wking:
> Does that mean the Config:Properties should be updated to be consistent with the pre-defined annotation keys. For example `vendor`, `revision`, `version` are not mentioned in Config:Properties?

My impression is that the config properties are deprecated properties which are kept around for backwards compat, and that new metadata should go into `annotations`.  But I'm not sure how widely supported those annotations are in tooling.  Are they're checked in the absence of the analogous config property?  How do tools handle both being set?  Is setting `org.opencontainers.image.authors` on the manifest sufficient, or should it be set on just the config, or both for maximum cross-tool portability?  Can a maintainer weigh in with some suggestions (ideally in the form of a implementation-note PR ;)?

PR #750, opened 2018-09-17
Fix duplicate "id" values in JSON schema

mtrmac:
The `"id"` values in JSON schema files must be unique, per RFC draft 8.3.1:
> A schema MAY (and likely will) have multiple URIs, but there is no way for a URI to identify more than one schema.

and recent `gojsonschema` fails when handling such inputs (fairly nontransparently, it silently fails to resolve `$ref` references to absolute URIs and reports something like
> Reference defs.json#/definitions/mapStringString must be canonical

.  This is visible in `make test`.)

In particular, the `https://opencontainers.org/schema/image/descriptor/annotations` `id` value had three definitions.  To resolve this:
- Leave the definition in `image-index-schema.json`; although using the `/descriptor` subnamespace for the `"manifests"` array is a bit surprising, the `/image/` part clearly belongs to `image-index-schema.json`.
- Rename the `id` definition in `content-descriptor.json`, to use the generic "blob descriptor" namespace.
- Remove the definition in `defs-descriptor.json`; that seems to be an "utility" schema file describing common structures, but it's better for users to reference schema fragments by purpose than by common structure (so that we can let the structure diverge in the future if necessary).

Finally, changing the content-descriptor.json `"id"` value changes the resolved absolute value of the reference to `defs-descriptor.json`, so add another namespace to be handled by `fsLoaderFactory`.

(As in #739, all I want to do is make the validation work, or, well, not fail. I am expressing no opinion on JSON schema or its use in this project.)

### DIFF
--- a/schema/content-descriptor.json
+++ b/schema/content-descriptor.json
@@ -21,7 +21,7 @@
       "$ref": "defs-descriptor.json#/definitions/urls"
     },
     "annotations": {
-      "id": "https://opencontainers.org/schema/image/descriptor/annotations",
+      "id": "https://opencontainers.org/schema/descriptor/annotations",
       "$ref": "defs-descriptor.json#/definitions/annotations"
     }
   },
--- a/schema/defs-descriptor.json
+++ b/schema/defs-descriptor.json
@@ -20,7 +20,6 @@
       }
     },
     "annotations": {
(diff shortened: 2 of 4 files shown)


cyphar:
LGTM.

PR #756, opened 2018-12-20
Linting and table formating

vbatts:
various tidying for makrdown linting.
Also arranging the MediaType compatibility matrix from a clunky list into a proper table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # OCI Image Format Specification
+
 <div>
 <a href="https://travis-ci.org/opencontainers/image-spec">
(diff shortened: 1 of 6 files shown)


vbatts:
omg. I think these lint changes are good, but the linter itself is bombing.
Looking at gometalinter, the upstream has been archived. Now is redirecting users to https://github.com/golangci/golangci-lint, which is not quite a drop-in replacement.
This issue is blocked until that is fixed.

vbatts:
two years is enough time to say we don't need this...

PR #761, opened 2019-02-19
Add platform.Feature

ChristianKniep:
As discussed on the [mailing list](https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/PCvWK6rEcqE) a PR to bring platform.features back into the image spec.
Motivation stated in https://github.com/moby/moby/issues/38715
Blog post providing context:
- [Optimized Container Images for AI/ML and HPC]()http://qnib.org/2019/02/12/optimized-images-for-aiml-hpc/
- [Match Node-Specific Needs Using Manifest Lists](http://qnib.org/2019/02/14/manifest-list-to-pick-optimized-images/)

Not sure about the version bump tho - please advise how to handle that one.

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -87,7 +87,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     - **`features`** *array of strings*
 
-        This OPTIONAL property specifies an array of strings, each specifying a mandatory hardware/host feature. Examples are:
+        This OPTIONAL property specifies an array of strings, each specifying an optional hardware/host feature. Examples are:
          - CPU type (*haswell*, *broadwell*, *skylake*, *ryzen*)
          - GPU [NVIDIA Compute Capability](https://developer.nvidia.com/cuda-gpus) (3.7, 6.1)
          - Host name, to separate different types in a data center without flushing out all the specifics (*compute-20190213*, *storage-2019*)
--- a/image-index.md
+++ b/image-index.md
@@ -90,7 +90,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
         This OPTIONAL property specifies an array of strings, each specifying a mandatory hardware/host feature. Examples are:
          - CPU type (*haswell*, *broadwell*, *skylake*, *ryzen*)
          - GPU [NVIDIA Compute Capability](https://developer.nvidia.com/cuda-gpus) (3.7, 6.1)
-         - Host name, to separate different types in a data center without flushing out all the specifics (*ComputeNode-20190213*, *storage-2019*)
+         - Host name, to separate different types in a data center without flushing out all the specifics (*compute-20190213*, *storage-2019*)
         By using this feature flag, an image index is able to provide specific images for certain hosts.
 
 - **`annotations`** *string-string map*
--- a/image-index.md
+++ b/image-index.md
@@ -87,7 +87,11 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     - **`features`** *array of strings*
 
-        This property is RESERVED for future versions of the specification.
+        This OPTIONAL property specifies an array of strings, each specifying a mandatory hardware/host feature. Examples are:
+         - CPU type (*haswell*, *broadwell*, *skylake*, *ryzen*)
+         - GPU [NVIDIA Compute Capability](https://developer.nvidia.com/cuda-gpus) (3.7, 6.1)
+         - Host name, to separate different types in a data center without flushing out all the specifics (*ComputeNode-20190213*, *storage-2019*)
+        By using this feature flag, a manifest list is able to provide specific images for certain hosts.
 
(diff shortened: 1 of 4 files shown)


stevvooe:
You need to update the spec and the schema. Also, I think we need to be explicit about how this field is matched.

ChristianKniep:
Ah - ok missed that. Will update. Thx

image-index.md
```diff
@@ -87,7 +87,11 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     - **`features`** *array of strings*
 
-        This property is RESERVED for future versions of the specification.
+        This OPTIONAL property specifies an array of strings, each specifying a mandatory hardware/host feature. Examples are:
+         - CPU type (*haswell*, *broadwell*, *skylake*, *ryzen*)
+         - GPU [NVIDIA Compute Capability](https://developer.nvidia.com/cuda-gpus) (3.7, 6.1)
+         - Host name, to separate different types in a data center without flushing out all the specifics (*ComputeNode-20190213*, *storage-2019*)
+        By using this feature flag, a manifest list is able to provide specific images for certain hosts.
```
jonjohnsonjr: nit: s/manifest list/image index/

ChristianKniep: @jonjohnsonjr changed it - thx for the correction

ChristianKniep:
@stevvooe sorry for the delay - a lot going on. re: your comment... 
Do you think the matching should be discussed within the spec? Might be something TBD within the runtime? Even though I am not sure - just a gut feeling.

image-index.md
```diff
@@ -87,7 +87,11 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     - **`features`** *array of strings*
 
-        This property is RESERVED for future versions of the specification.
+        This OPTIONAL property specifies an array of strings, each specifying a mandatory hardware/host feature. Examples are:
+         - CPU type (*haswell*, *broadwell*, *skylake*, *ryzen*)
+         - GPU [NVIDIA Compute Capability](https://developer.nvidia.com/cuda-gpus) (3.7, 6.1)
+         - Host name, to separate different types in a data center without flushing out all the specifics (*ComputeNode-20190213*, *storage-2019*)
+        By using this feature flag, an image index is able to provide specific images for certain hosts.
```
cyphar: In order for interoperability, we need to have either a standard set of well-understood names, or some way parsing well-structured feature names (`cpu-feature:avx512` and `cpu-type:haswell` for instance) in order to be able to handle unknown names.

I am **very** un-nerved by the hostname example, and would like to know how do we handle an image-spec implementation which wants to extract an image but doesn't have that hostname? Should it ignore the field? If so, how is the field mandatory if you can ignore it?

image-index.md
```diff
@@ -87,7 +87,11 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     - **`features`** *array of strings*
 
-        This property is RESERVED for future versions of the specification.
+        This OPTIONAL property specifies an array of strings, each specifying a mandatory hardware/host feature. Examples are:
```
cyphar: "mandatory" doesn't really have meaning in this spec, you need to explain what the behaviour needs to be and how a runtime should check the features.

ChristianKniep:
@cyphar Thank you for your feedback. The previous use of the feature flag tried (as I see it) to come up with an agreed on and objective description of hardware features; hence, the focus was on (discoverable) CPU flags.
My proposal is different in that it allows a site to define such feature flags as it fits their environment. That is what is meant be using flags like `compute-20190213` - it provides a reference to a certain node and freezes his host installation in time, so that one can create site-specific or even only node type specific containers.
As discussed with @AkihiroSuda in Japan last month and as brought up by your comment; for public image repositories a naming scheme should be agreed on; similar to label best-practises or what you provided as an example.
Where I see this feature flag missing is when you want to create containers that rely on something that the host provides but are not controlled by the kernel. The prime example is GPUs, where the runtime (nvidia-docker) breaks the immutability of the container FS to allow a container workflow to take place.
Allowing runtimes to use the OCI image spec to the fullest would inform a discussion about what feature flags are used (/useful) for public repositories.
This features provides a solution for communities with a need for either host-specific dependencies or the need for hardware optimisation by allowing them to define their own flags and hand over the distribution decision of optimised images to the runtime.

ChristianKniep:
@cyphar / @stevvooe Any additional input on that?
As said; IMHO predefining all possible features strings is hard to impossible. And in an environment where mixed workloads exists I reckon that using a hardware identifier like `host-type:HPE-Model2000` is a good start as it does not define which components of that host are used and optimised for. Workloads one might want to optimise for CPU+GPU and another one only for the CPU.

ChristianKniep:
I am addressing the issue in a different way - so the PR is obsolete.
http://www.qnib.org/2019/06/12/metahub/
https://github.com/qnib/metahub

PR #770, opened 2019-03-13
Artifact spec additions

SteveLasker:
Add spec enhancements to support new registry artifact types. 
Spec focuses on `manifest.config.mediaType` as the means to identify the new artifact types. 
Changes FAQ to future work on distribution to the OCI distribution spec


artifacts/mediaTypes.md
```diff
@@ -0,0 +1,211 @@
...
+Artifact authors MAY publicize their schema in [./mediaTypeMappings.json]() through a pull request.
+
+Artifact authors understand supporting additional artifact types is optional, and distribution parsing of new config schemas is optional. However, new config schemas may be useful to that artifact tooling experience. 
+
+## Layer mediaTypes
+
+While OCI container images have ordinal layers, supporting overlaying of a unified file system, new artifact authors MAY define how they persist their layers, which may be individual files, or collections of files. The tooling specific to that artifact type owns the processing of the files and how they are extracted on the client. 
+
+The [layer mediaType descriptor in the OCI Image spec](https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions) identifies the `mediaType` MUST support one of the existing layer formats. 
+
```
vsoch: Question! So I'm allowed to do whatever I like for the config mediaType, but then in the layers list, it MUST be one that has been pre-approved? Or is layer formats referring to the array (and not the objects themselves?)

artifacts/mediaTypes.md
```diff
@@ -0,0 +1,211 @@
...
+- Define a `null` schema, where the artifact only uses the `config.mediaType` to identify it's unique artifact type. 
+
+Artifact authors MAY publicize their schema in [./mediaTypeMappings.json]() through a pull request.
+
+Artifact authors understand supporting additional artifact types is optional, and distribution parsing of new config schemas is optional. However, new config schemas may be useful to that artifact tooling experience. 
+
+## Layer mediaTypes
+
+While OCI container images have ordinal layers, supporting overlaying of a unified file system, new artifact authors MAY define how they persist their layers, which may be individual files, or collections of files. The tooling specific to that artifact type owns the processing of the files and how they are extracted on the client. 
+
```
vsoch: Here is sounds like I have freedom to define custom types for the layers? And then it follows that with custom content types, the distribution spec doesn't have to server containers at all?

artifacts/mediaTypes.md
```diff
@@ -0,0 +1,211 @@
...
+    
+Artifact authors can extend these mediaTypes to identify their unique usage, using the following format: 
+
+`application/vnd.`[governingOrg/vendor]`.`[artifactName]`.layer.`[version][`.tar`|`+tar+gzip`]
+
+**Helm Example**
+
+`application/vnd.cncf.helm.chart.layer.v3.tar`
+`application/vnd.cncf.helm.chart.values.v3.tar`
+
```
vsoch: Just curious - who maintains these namespaces? What does vnd mean?

artifacts/mediaTypes.md
```diff
@@ -0,0 +1,211 @@
...
+  },
+  "layers": [
+    {
+      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
+      "size": 23133155,
+      "digest": "sha256:9a1a13172ed974323f7c35153e8b23b8fa1c85355b6b26cc3127e640e45ef0aa"
+    }
+  ]
+```
+
```
vsoch: weird that it's being rendered all funky here down.

artifacts/mediaTypes.md
SteveLasker: The custom layer types are still layers. What I'm trying to say here is they don't have to culminate in a unified file system where all the layers overlay each other. They can be completely separate collections the artifact tool can extract as they wish. A new artifact may extract layer0 into the current directory, while it might extract layer1 into a per-user config directory. 
My next round is to do a PR on the distribution spec. What this should be conveying is distribution would accept and server layers. As long as it's a blob, like using the ORAS (https://github.com/deislabs/oras/) client, the registry shouldn't care. 
Does that clarify it?

artifacts/mediaTypes.md
vsoch: If I'm deploying a registry, am I allowed to create my own mediaType for an artifact? (e.g., `application/vnd.dinosaur.dump.v1` or would I need to use only "one of the existing layer formats?"

vsoch: (that's clarification for what I was asking :point_up: )

artifacts/mediaTypes.md
vsoch: Yes I understand that bit - and maybe my question would be better asked on your next PR to the distribution spec.  What I am curious about is if the user is allowed custom mediaTypes (for example, docs) to be served in an image-spec (for download / otherwise handling by some client) then what is stopping someone from only serving (non container layer) artifacts in an image spec served by a registry? Technically, couldn't a registry exclusively serve one or more types, none of which are filesystem / layer related at all?

artifacts/mediaTypes.md
SteveLasker: We've discussed a "clearing house" where artifact authors can submit their artifacts for others to consume. This is in the ./artifacts/mediaTypeMappings.json file.  Registry authors could choose to keep these unique to them as well.

vsoch: okay that makes sense. And validation of any media type is out of scope for the spec, and is responsibility of a client or registry?

SteveLasker: Correct. I figured I'd cover that in the distribution spec.

artifacts/mediaTypes.md
SteveLasker: Yes. The artifact registry enhancements don't "judge" what's being saved, as long as they follow some basic semantics to be reasoned over. For instance, if a team was doing some work and wanted to store foo artifacts in a registry, leveraging all the investments made in that PaaS or product; and either don't save any images, or they choose to save them in a different registry instance, have at it.

vsoch: That's really neat :) It makes the spec useful beyond what it was originally intended for.

artifacts/mediaTypes.md
vsoch: Much better! My eyes say thank you :eyes:

artifacts/mediaTypes.md
atlaskerr: @vsoch  `vnd` stands for `vendor` and it represents a third-party mediatype registered with IANA.

vsoch: Oh interesting - first I was looking around IANA trying to find something container related:

 - [event names](https://www.iana.org/assignments/message-store-events/message-store-events.xhtml#message-store-events-2)
 - [event parameters](https://www.iana.org/assignments/message-store-events/message-store-events.xhtml#message-store-events-4)
 - [MQTP](https://www.iana.org/assignments/mtqp-options/mtqp-options.xhtml#mtqp-options-2)

but actually the better description is on [wikipedia](https://en.wikipedia.org/wiki/Media_type) that explains what you said above - vendor specific mime types. Then I found [this listing](https://www.iana.org/assignments/media-types/media-types.xhtml) but why don't we see anything from Docker there? Is it not registered?

atlaskerr: People use `vnd` and don't register. I know docker doesn't have their mediatypes registered but coreos has one for etcd I think.

vsoch: Interesting. I guess we can make up whatever we like then :wink:

artifacts/mediaTypes.md
sajayantony: Open for debate - media type constraints should be left to registry implementors. 
I don't know if the reference implementation should include a default validator with a configurable list of valid types. 
/cc @yuwaMSFT2

artifacts/mediaTypes.md
atlaskerr: Haha yeah basically.

SteveLasker: I'm hopeful most artifact owners will use the clearing house/PR approach to broadcast their types. I imagine most registries would limit artifact types to whitelisting known types. Otherwise, we could only present UNKOWN, or show the `mediaType` which would be sorta ugly. I'm hopeful we can make this easy enough for registries to import the script with minimal internal configuration. But, notice I keep saying hopeful. Only time will tell.

README.md
```diff
@@ -35,12 +35,19 @@ This entire workflow supports the UX that users have come to expect from contain
 
 To support this UX the OCI Image Format contains sufficient information to launch the application on the target platform (e.g. command, arguments, environment variables, etc).
 
-## FAQ
+## Additional Artifact Types
+
+The OCI image spec provides the core elements required for additional artifact types, including [Helm](https://helm.sh/), [CNAB](https://cnab.io) and existing artifacts like Azure Resource Manager (ARM) or AWS Cloud Formation Templates. By utilizing the image-spec core elements, additional tool authors and registry implementors can easily support new artifact types, benefiting from the common infrastructure to deploy container images.
```
clareliguori: Nit: CloudFormation is misspelled (no space), but also it seems odd to call this out as an example in the official spec when I'm not aware of any plans for us (AWS) to add this as a media type in the near term.  Will that be a community contribution? :)

artifacts/mediaTypes.md
```diff
@@ -0,0 +1,214 @@
...
+      "digest": "sha256:8a7d13172ed974323f7c35153e8b23b8fa1c85355b6b26cc3127e640712927dd"
+    }
+  ]
+}
+```
+
+
+## OCI image index
+
+An OCI index is a higher-level manifest which contains a collection of image manifests, typically used for one or more platforms. In this example, the image index is used to group an image an it's associated docs.
```
clareliguori: Nit: an it's -> and it's

README.md
SteveLasker: Hey Clare, 
I was just looking for examples to demonstrate the range of ideas. If you prefer I remove, I can do that. I've been thinking of doing the doc artifactType as a concrete example, in addition to helm. Just let me know what you prefer.

SteveLasker: Per Clare: removing as the work isn't currently planned, but very interested in the feature.

vbatts:
@SteveLasker is this PR still needed, as artifacts has become its own thing?

vbatts:
house cleaning

PR #772, opened 2019-03-21
Fixed a small typo

woernfl:
Fixed a small typo in the header of the table comparing `org.opencontainers.image` and `org.label-schema` prefix.


woernfl:
Hey @vbatts @jonboulle ,

It seems that Travis is failing with some linting jobs. However, I can not see that failling because of the change I made, any clue what's happening?

vbatts:
The megalinter and golang versions have fights and need to be fixed. It's unrelated.

PR #774, opened 2019-04-03
Fix pandoc parameter.

yyyeerbo:
It is --pdf-engine instead of --latex-engine.

Fixes: #773

Signed-off-by: Yang Bo <bo@hyper.sh>


vbatts:
oh cool. I had recently looked into that, and also to try newer versions.
https://quay.io/repository/oci/pandoc?tag=latest&tab=tags
Might you try with one of these newer images from the image `quay.io/oci/pandoc`?

yyyeerbo:
my pandoc is 2.2.1 . It  looks like the latest version.

vbatts:
I've rolled this PR into #846

Issue #747, opened 2018-07-09
Proposal: New Mediatype - Container Image Encryption


lumjjb:
# Overview

We would like to propose a new media type for encrypted layers of a container image. This addition would facilitate the ecosystem for encrypted container images. This allows users with stricter trust requirements to be able ensure end-to-end encryption from build to runtime. In addition, it allows users to use a centralized managed repository (i.e. Docker Hub) without any risk of their images being compromised. 

Proposal written and discussed by: 
Brandon Lum (@lumjjb), Dimitrios Pendarakis, Hani Jamjoom (@jamjoom), James Bottomley (@jejb), Phil Estes (@estesp), Stefan Berger (@stefanberger) within IBM.

This is a follow-up from a discussion with @stevvooe , and @dmcgowan at DockerCon.

# Goals
In coming up with a proposal, we considered the following:
- Application creators should not need to worry about details of encryption. Flows in the DevOps process and the runtime orchestration would be responsible for facilitating encryption.
- Re-usability of layer de-duplication among container images
- Encryption should be end-to-end, from the build step to when the container needs to be run on the worker machine. (i.e. registry owner should not be able to view the encrypted contents).
- Encrypted images created can be for multiple trusted entities, and key management is responsible for managing that trust.
- Specification should be easily integrated to generic with and across key management systems.
- Secondary experimental goal: Allow more granular control of security, to provide possibility of desegregation and more fine-grained  security controls. (i.e. base OS, middleware, application can be encrypted separately).

# Proposed Changes
We propose the following change:
- Adding a mediaType: application/vnd.oci.image.layer.v1.tar+gzip+enc to signify an encrypted layer (encrypted gzip).
- For each layer with the above mediaType, add the following fields (either via annotations or within the spec):
  - enc.algo - The encryption algorithm used to encrypt the layer (encryption algorithm naming and implementation will follow a registry such as IANA [https://www.iana.org/assignments/aead-parameters/aead-parameters.xhtml]) 
  - enc.keyid - The reference to the key to use to perform the decryption of the layer. (NOTE: This is not the key! Elaborated on in "Additional Details" of document). 
  - enc.* - Other optional parameters that may be needed by encryption algorithms.
  - Cipher data should use Explicit IV if cipher requires use of IV.

With the changes, we expect an example of creating an encrypted image to look like the following:

[...]

PR #775, opened 2019-04-27
Proposal to add Encrypted Layer Mediatype

lumjjb:
This is a first draft of modifications to the OCI spec on Encrypted Container Images. Work around this has been on-going for quite a while (since the start of [the issue](https://github.com/opencontainers/image-spec/issues/747)). The current design is based on the discussions we've had in [the issue](https://github.com/opencontainers/image-spec/issues/747) and with the [runtime implementation in containerd](https://github.com/containerd/containerd/pull/3134).

## Abstract

We would like to propose a new media type for encrypted layers of a container image. This addition would facilitate the ecosystem for encrypted container images. This allows users with stricter trust requirements to be able ensure end-to-end encryption from build to runtime. In addition, it allows users to use a centralized managed repository (i.e. Docker Hub) without any risk of their images being compromised. 

## Issue Details
Based on the discussion of https://github.com/opencontainers/image-spec/issues/747

## The specification and libraries to perform encryption/decryption

The implementation to perform encryption and decryption of this spec is in:
- https://github.com/containers/ocicrypt
  - `func EncryptLayer(ec *config.EncryptConfig, encOrPlainLayerReader io.Reader, desc ocispec.Descriptor) (io.Reader, EncryptLayerFinalizer, error)`
  - `func DecryptLayer(dc *config.DecryptConfig, encLayerReader io.Reader, desc ocispec.Descriptor, unwrapOnly bool) (io.Reader, digest.Digest, error)`

### Container Runtimes

Implemented and upstreamed for containerd stack
- [containerd (1.3+)](https://github.com/containerd/containerd/blob/master/docs/cri/decryption.md)
- https://github.com/containerd/imgcrypt

Implemented and upstreamed for RedHat stack
- [cri-o 1.17+](https://github.com/cri-o/cri-o/blob/master/tutorials/decryption.md)
- [buildah 1.5+](https://github.com/containers/buildah)
- [skopeo](https://github.com/containers/skopeo)

### Registry

Current docker distribution supports: https://github.com/docker/distribution. Tested with v2.7.1.

### Build toolchain

The plan for this is to have docker CLI, skopeo and buildah. In the meantime, we also support encryption of images in containerd via `imgcrypt` ctr client.

Current implementations for RedHat stack
- [buildah 1.5+](https://github.com/containers/buildah)
- [skopeo](https://github.com/containers/skopeo)

Current implementations for containerd stack
- [ctr-enc](https://github.com/containerd/imgcrypt)
- To integrate into docker CLI, we are currently waiting on https://github.com/moby/moby/issues/38043. Tracking with https://github.com/moby/buildkit/issues/714

### Kubernetes Integration

It is important to note that kubernetes integration is not required to use encrypted container images. i.e. in cri-o you are able to specify a directory containing keys.

A way to do this is via an operator [k8s-enc-image-operator](https://github.com/IBM/k8s-enc-image-operator)

Signed-off-by: Brandon Lum <lumjjb@gmail.com>

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -49,6 +49,10 @@ This specification defines the following annotation keys, intended for but not l
   * This SHOULD be the immediate image sharing zero-indexed layers with the image, such as from a Dockerfile `FROM` statement.
   * This SHOULD NOT reference any other images used to generate the contents of the image (e.g., multi-stage Dockerfile builds).
   * If the `image.base.name` annotation is specified, the `image.base.digest` annotation SHOULD be the digest of the manifest referenced by the `image.ref.name` annotation.
+* **org.opencontainers.image.org.opencontainers.image.enc.keys.pkcs7** Base64 comma separated [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
+* **org.opencontainers.image.org.opencontainers.image.enc.keys.jwe** Base64 comma separated [JWE(RFC7516)](https://tools.ietf.org/html/rfc7516) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
+* **org.opencontainers.image.org.opencontainers.image.enc.keys.openpgp** Base64 comma separated [OpenPGP(RFC4880)](https://tools.ietf.org/html/rfc4880) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
+* **org.opencontainers.image.org.opencontainers.image.enc.pubopts** Base64 encoded PublicLayerBlockCipherOptions JSON object for [encryption](layer.md#layer-encryption)
 
 ## Back-compatibility with Label Schema
 
--- a/annotations.md
+++ b/annotations.md
@@ -50,9 +50,9 @@ This specification defines the following annotation keys, intended for but not l
   * This SHOULD NOT reference any other images used to generate the contents of the image (e.g., multi-stage Dockerfile builds).
   * If the `image.base.name` annotation is specified, the `image.base.digest` annotation SHOULD be the digest of the manifest referenced by the `image.ref.name` annotation.
 * **org.opencontainers.image.org.opencontainers.image.enc.keys.pkcs7** Base64 comma separated [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
-* **org.opencontainers.image.org.opencontainers.image.enc.keys.jwe** Base64 comma separated [JWE(RFC7516)](https://tools.ietf.org/html/rfc7516) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
-* **org.opencontainers.image.org.opencontainers.image.enc.keys.openpgp** Base64 comma separated [OpenPGP(RFC4880)](https://tools.ietf.org/html/rfc4880) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
-* **org.opencontainers.image.org.opencontainers.image.enc.pubopts** Base64 encoded PublicLayerBlockCipherOptions JSON object for [encryption](layer.md#layer-encryption)
+* **org.opencontainers.image.enc.keys.jwe** Base64 comma separated [JWE(RFC7516)](https://tools.ietf.org/html/rfc7516) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
+* **org.opencontainers.image.enc.keys.openpgp** Base64 comma separated [OpenPGP(RFC4880)](https://tools.ietf.org/html/rfc4880) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
+* **org.opencontainers.image.enc.pubopts** Base64 encoded PublicLayerBlockCipherOptions JSON object for [encryption](layer.md#layer-encryption)
 
 ## Back-compatibility with Label Schema
 
--- a/layer.md
+++ b/layer.md
@@ -4,7 +4,7 @@ This document describes how to serialize a filesystem and filesystem changes lik
 One or more layers are applied on top of each other to create a complete filesystem.
 This document will use a concrete example to illustrate how to create and consume these filesystem layers.
 
-This section defines the `application/vnd.oci.image.layer.v1.tar`, `application/vnd.oci.image.layer.v1.tar+gzip`, `application/vnd.oci.image.layer.v1.tar+zstd`, `application/vnd.oci.image.layer.nondistributable.v1.tar`, `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`, and `application/vnd.oci.image.layer.nondistributable.v1.tar+zstd` [media types](media-types.md).
+This section defines the `application/vnd.oci.image.layer.v1.tar`, `application/vnd.oci.image.layer.v1.tar+gzip`, `application/vnd.oci.image.layer.v1.tar+zstd`, `application/vnd.oci.image.layer.nondistributable.v1.tar`, `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`, `application/vnd.oci.image.layer.nondistributable.v1.tar+zstd`, `application/vnd.oci.image.layer.v1.tar+encrypted`, `application/vnd.oci.image.layer.v1.tar+gzip+encrypted`, `application/vnd.oci.image.layer.nondistributable.v1.tar+encrypted` and `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip+encrypted` [media types](media-types.md).
 
 ## `+gzip` Media Types
 
@@ -15,6 +15,13 @@ This section defines the `application/vnd.oci.image.layer.v1.tar`, `application/
 
 * The media type `application/vnd.oci.image.layer.v1.tar+zstd` represents an `application/vnd.oci.image.layer.v1.tar` payload which has been compressed with [zstd][rfc8478].
 * The media type `application/vnd.oci.image.layer.nondistributable.v1.tar+zstd` represents an `application/vnd.oci.image.layer.nondistributable.v1.tar` payload which has been compressed with [zstd][rfc8478].
+
+## `+encrypted` Media Types
+
+* The media type `application/vnd.oci.image.layer.v1.tar+encrypted` represents an `application/vnd.oci.image.layer.v1.tar` payload which has been [encrypted](#layer-encryption).
+* The media type `application/vnd.oci.image.layer.v1.tar+gzip+encrypted` represents an `application/vnd.oci.image.layer.v1.tar+gzip` payload which has been [encrypted](#layer-encryption).
+* The media type `application/vnd.oci.image.layer.nondistributable.v1.tar+encrypted` represents an `application/vnd.oci.image.layer.nondistributable.v1.tar` payload which has been [encrypted](#layer-encryption).
+* The media type `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip+encrypted` represents an `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip` payload which has been [encrypted](#layer-encryption).
 
 ## Distributable Format
 
@@ -331,6 +338,96 @@ Non-distributable layers SHOULD be tagged with an alternative mediatype of `appl
 Implementations SHOULD NOT upload layers tagged with this media type; however, such a media type SHOULD NOT affect whether an implementation downloads the layer.
 
 [Descriptors](descriptor.md) referencing non-distributable layers MAY include `urls` for downloading these layers directly; however, the presence of the `urls` field SHOULD NOT be used to determine whether or not a layer is non-distributable.
+
+
+# Layer Encryption
+
+To be able to protect the confidentiality of the data in layers, encryption of the layer data blobs can be done to prevent unauthorized access to layer data. Encryption is performed on the data blob of a layer by specifying a mediatype with the `+encrypted` suffix. For example, `application/vnd.oci.image.layer.v1.tar+encrypted` is an layer representation of an encrypted `application/vnd.oci.image.layer.v1.tar` layer. 
+
+When using the `+encrypted` mediatype, the layer data blobs are encrypted. In order to decrypt an image, the encryption metadata is required. 
+
+## Encryption Metadata
+
+The encryption metadata consists of 2 parts: PublicLayerBlockCipherOptions and PrivateLayerBlockCipherOptions. The PublicLayerBlockCipherOptions contain encryption metadata that is public, i.e. cipher type, HMAC, etc. and PrivateLayerBlockCipherOptions contains the encryption metadata which should be confidential, i.e. symmetric key, nonce (optional), etc. These are stored in the respective [**org.opencontainers.image.enc** prefixed annotations](annotations.md).
+
+Below are golang definitions of these JSON objects:
+
+```
+// LayerCipherType is the ciphertype as specified in the layer metadata
+type LayerCipherType string
+
+// PublicLayerBlockCipherOptions includes the information required to encrypt/decrypt
+// an image which are public and can be deduplicated in plaintext across multiple
+// recipients
+type PublicLayerBlockCipherOptions struct {
+    // CipherType denotes the cipher type according to the list of OCI suppported
+    // cipher types.
+    CipherType LayerCipherType `json:"cipher"`
+
+    // Hmac contains the hmac string to help verify encryption
+    Hmac []byte `json:"hmac"`
+
+    // CipherOptions contains the cipher metadata used for encryption/decryption
+    // This field should be populated by Encrypt/Decrypt calls
+    CipherOptions map[string][]byte `json:"cipheroptions"`
+}
+
+// PrivateLayerBlockCipherOptions includes the information required to encrypt/decrypt
+// an image which are sensitive and should not be in plaintext
+type PrivateLayerBlockCipherOptions struct {
+	// SymmetricKey represents the symmetric key used for encryption/decryption
+	// This field should be populated by Encrypt/Decrypt calls
+	SymmetricKey []byte `json:"symkey"`
+
+	// Digest is the digest of the original data for verification.
+	// This is NOT populated by Encrypt/Decrypt calls
+	Digest digest.Digest `json:"digest"`
+
+	// CipherOptions contains the cipher metadata used for encryption/decryption
+	// This field should be populated by Encrypt/Decrypt calls
+	CipherOptions map[string][]byte `json:"cipheroptions"`
+}
+```
+
+Details of the algorithms and protocols used in the encryption of the data blob are defined in these JSON objects. Here are some examples of the Public/Private LayerBlockCipherOptions.
+- The `cipher` field specifies the encryption algorithm to use according to the [list of cipher types supported](#cipher-types).
+- The `symkey` field specifies the base64 encoded bytes of the symmetric key used in decryption.
+- The `cipherOptions` field specifies additional parameters used in the decryption process of the specified algorithm. This should be in accordance with the RFC standard of the algorithm used.
+```
+PublicLayerBlockCipherOption
+{
+    "cipher": "AES_256_CTR_HMAC_SHA256",
+    "hmac": "M0M5OTA5QUZFQzI1MzU0RDU1MURBRTIxNTkwQkIyNkUzOEQ1M0YyMTczQjhEM0RDM0VFRTRDMDQ3RTdBQjFDMQ=="
+    "cipheroptions": {}
+}
+
+PrivateLayerBlockCipherOption
+{
+    "symkey": "54kiln1USEaKnlYhKdz+aA==",
+    "cipheroptions": {
+        "nonce": "AdcRPTAEhXx6uwuYcOquNA==",
+        ...
+    }
+}
+
+```
+
+The PublicLayerBlockCipherOptions JSON object is stored base64 encoded in the layer annotation **org.opencontainers.image.enc.pubopts**. 
+
+The PrivateLayerBlockCipherOptions JSON object is not stored in plaintext due to the sensitive nature of the contents. Instead, the object goes through a cryptographic wrapping process. This ensures that only authorized parties are able to decrypt the layers, the decryption metadata objects are wrapped as encrypted messages to the authorized recipients in accordance with encrypted message standards such as [OpenPGP(RFC4880)](https://tools.ietf.org/html/rfc4880), [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315), [JWE(RFC7516)](https://tools.ietf.org/html/rfc7516). 
+
+The following annotations are used to communicate these encrypted messages:
+- `org.opencontainers.image.enc.keys.pkcs7` - An array of base64 comma separated encrypted messages that contain PrivateLayerBlockCipherOptions to perform decryption of the layer data in accordance with [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315)
+- `org.opencontainers.image.enc.keys.jwe` - An array of base64 comma separated encrypted messages that contain PrivateLayerBlockCipherOptions to perform decryption of the layer data in accordance with [JWE(RFC7516)](https://tools.ietf.org/html/rfc7516)
+- `org.opencontainers.image.enc.keys.openpgp` - An array of base64 comma separated encrypted messages that contain PrivateLayerBlockCipherOptions to perform decryption of the layer data in accordance with [OpenPGP(RFC4880)](https://tools.ietf.org/html/rfc4880)
+- `org.opencontainers.image.enc.keys.*` - An array of base64 comma separated encrypted messages that contain PrivateLayerBlockCipherOptions to perform decryption of the layer data in accordance with an appropriate standard of specified protocol.
+
+The decryption of the image can be performed by unwrapping the PrivateLayerBlockCipherOptions using the `org.opencontainers.image.enc.keys.*` annotations and using the appropriate cipher with the unwrapped PrivateLayerBlockCipherOptions to decrypt the layer data blob.
+
+### Cipher Types
+
+The current list of cipher types supported are:
+- `AES_256_CTR_HMAC_SHA256` - Encryption with `AES_256_CTR` algorithm [FIPS-197](https://csrc.nist.gov/csrc/media/publications/fips/197/final/documents/fips-197.pdf)  with Encrypt-then-mac [RFC7366](https://tools.ietf.org/html/rfc7366). The protocols used in this cipher type is in accordance to FIPS 140-2 compliant standards. 
 
 [libarchive-tar]: https://github.com/libarchive/libarchive/wiki/ManPageTar5#POSIX_ustar_Archives
 [gnu-tar-standard]: http://www.gnu.org/software/tar/manual/html_node/Standard.html
--- a/layer.md
+++ b/layer.md
@@ -352,7 +352,7 @@ The encryption metadata consists of 2 parts: PublicLayerBlockCipherOptions and P
 
 Below are golang definitions of these JSON objects:
 
-```
+```go
 // LayerCipherType is the ciphertype as specified in the layer metadata
 type LayerCipherType string
 
--- a/media-types.md
+++ b/media-types.md
@@ -13,6 +13,10 @@ The following media types identify the formats described here and their referenc
 - `application/vnd.oci.image.layer.nondistributable.v1.tar`: ["Layer", as a tar archive with distribution restrictions](layer.md#non-distributable-layers)
 - `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`: ["Layer", as a tar archive with distribution restrictions](layer.md#gzip-media-types) compressed with [gzip][rfc1952]
 - `application/vnd.oci.image.layer.nondistributable.v1.tar+zstd`: ["Layer", as a tar archive with distribution restrictions](layer.md#zstd-media-types) compressed with [zstd][rfc8478]
+- `application/vnd.oci.image.layer.v1.tar+encrypted` : ["Layer", as an encrypted tar archive](layer.md#enc-media-types)
+- `application/vnd.oci.image.layer.v1.tar+gzip+encrypted` : ["Layer", as an encrypted gzipped tar archive](layer.md#enc-media-types)
+- `application/vnd.oci.image.layer.nondistributable.v1.tar+encrypted` : ["Layer", as an encrypted tar archive with distribution restrictions](layer.md#enc-media-types)
+- `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip+encrypted` : ["Layer", as an encrypted gzipped tar archive with distribution restrictions](layer.md#enc-media-types)
 
 ## Media Type Conflicts
 
--- a/specs-go/v1/annotations.go
+++ b/specs-go/v1/annotations.go
@@ -59,4 +59,16 @@ const (
 
 	// AnnotationBaseImageName is the annotation key for the image reference of the image's base image.
 	AnnotationBaseImageName = "org.opencontainers.image.base.name"
+
+	// AnnotationEncryptionKeysPkcs7 is the annotation key for the base64 encoded encrypted
+	// messages containing metadata for decrypting encrypted data blobs.
+	AnnotationEncryptionKeysPkcs7 = "org.opencontainers.image.enc.keys.pkcs7"
+
+	// AnnotationEncryptionKeysJwe is the annotation key for the base64 encoded encrypted
+	// messages containing metadata for decrypting encrypted data blobs.
+	AnnotationEncryptionKeysJwe = "org.opencontainers.image.enc.keys.jwe"
+
+	// AnnotationEncryptionKeysOpenPgp is the annotation key for the base64 encoded encrypted
+	// messages containing metadata for decrypting encrypted data blobs.
+	AnnotationEncryptionKeysOpenPgp = "org.opencontainers.image.enc.keys.openpgp"
 )
--- a/specs-go/v1/encryption.go
+++ b/specs-go/v1/encryption.go
@@ -1,0 +1,23 @@
+// Copyright 2016 The Linux Foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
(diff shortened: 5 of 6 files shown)


layer.md
```diff
@@ -327,6 +334,38 @@ Implementations SHOULD NOT upload layers tagged with this media type; however, s
...
 [Descriptors](descriptor.md) referencing non-distributable layers MAY include `urls` for downloading these layers directly; however, the presence of the `urls` field SHOULD NOT be used to determine whether or not a layer is non-distributable.
 
+
+# Layer Encryption
+
+To be able to protect the confidentiality of the data in layers, encryption of the layer data blobs can be done to prevent unauthorized access to layer data. Encryption is performed on the data blob of a layer by specifying a mediatype with the `+enc` suffix. For example, `application/vnd.oci.image.layer.v1.tar+enc` is an layer representation of an encrypted `application/vnd.oci.image.layer.v1.tar` layer. 
+
+When using the `+enc` mediatype, the layer data blobs are encrypted with a symmetric encryption algorithm (i.e. AES_128_GCM, AES_128_CBC, etc.) according to the [IANA Registry](https://www.iana.org/assignments/aead-parameters/aead-parameters.xhtml). 
+
+Details of the algorithms and protocols used in the encryption of the data blob are defined in a JSON object below. We note that:
```
stevvooe: If I'm understanding this correctly, this structure ends up in the annotations below. It might be helpful to describe the annotations first, then their contents.

I also think for annotations, they need to be added to the annotations.md.

lumjjb: Agreed on adding to `annotations.md`. 

 

For the order of explanation, I've found explaining it via the process of the actual encryption works well for users who are not as familiar with encryption.

Here's a stab at phrasing it the other way. 


# Layer Encryption

To be able to protect the confidentiality of the data in layers, encryption of the layer data blobs can be done to prevent unauthorized access to layer data. Encryption is performed on the data blob of a layer by specifying a mediatype with the `+enc` suffix. For example, `application/vnd.oci.image.layer.v1.tar+enc` is an layer representation of an encrypted `application/vnd.oci.image.layer.v1.tar` layer.

When using the `+enc` mediatype, the layer data blobs are encrypted with a symmetric encryption algorithm (i.e. AES_128_GCM, AES_128_CBC, etc.) according to the [IANA Registry](https://www.iana.org/assignments/aead-parameters/aead-parameters.xhtml). These symmetric keys and metadata   of the encryption algorithms used are [wrapped](https://en.wikipedia.org/wiki/Key_Wrap) via an asymmetric key so that only authorized users are able to obtain the symmetric key to decrypt the layer.

The symmetric key and the encryption metadata is defined as the [LayerBlockCipherOptions](specs-go/v1/encryption.go).  The LayerBlockCipherOptions are wrapped in accordance with encrypted message standards such as [OpenPGP(RFC4880)](https://tools.ietf.org/html/rfc4880),                [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315), [JWE(RFC7516)](https://tools.ietf.org/html/rfc7516).

The following annotations are used to communicate the encrypted symmetric keys:
- `org.opencontainers.image.enc.keys.pkcs7` - An array of base64 comma separated encrypted messages that contain LayerBlockCipherOptions to perform decryption of the layer data in accordance with [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315)
- `org.opencontainers.image.enc.keys.jwe` - An array of base64 comma separated encrypted messages that contain LayerBlockCipherOptions to perform decryption of the layer data in accordance with [JWE(RFC7516)](https://tools.ietf.org/html/rfc7516)
- `org.opencontainers.image.enc.keys.openpgp` - An array of base64 comma separated encrypted messages that contain LayerBlockCipherOptions to perform decryption of the layer data in accordance with [OpenPGP(RFC4880)](https://tools.ietf.org/html/rfc4880)
- `org.opencontainers.image.enc.keys.*` - An array of base64 comma separated encrypted messages that contain LayerBlockCipherOptions to perform decryption of the layer data in accordance with an appropriate standard of specified protocol.

These annotations contains an encrypted message containing the LayerBlockCipherOptions object.

The LackerBlockCipherOPtions object is defined by the JSON object below. We note that:
- The `cipher` field specifies the encryption algorithm in accordance with the [IANA Registry](https://www.iana.org/assignments/aead-parameters/aead-parameters.xhtml).
- The `symkey` field specifies the base64 encoded bytes of the symmetric key used in decryption.
- The `cipherOptions` field specifies additional parameters used in the decryption process of the specified algorithm. This should be in accordance with the RFC standard of the algorithm used.
```
{
    "cipher": "AES_128_GCM",
    "symkey": "54kiln1USEaKnlYhKdz+aA==",
    "cipheroptions": {
        "nonce": "AdcRPTAEhXx6uwuYcOquNA==",
        ...
    }
}
```

Therefore, the decryption of the image can be performed by unwrapping the [LayerBlockCipherOptions](specs-go/v1/encryption.go) structure using the `org.opencontainers.image.enc.keys.*` annotations and using the appropriate cipher with the unwrapped LayerBlockCipherOptions to decrypt   the layer data blob.

lumjjb:
@vbatts As per our conversation a couple days ago, Based on the [RFC6838](https://tools.ietf.org/html/rfc6838) and [RFC6839](https://tools.ietf.org/html/rfc6839).

A specification like nondistributable layers would work:
- `application/vnd.oci.image.layer.enc.tar.v1`
- `application/vnd.oci.image.layer.enc.tar.gzip.v1`

The other way to do this is to have **one** encrypted layer mediatype `application/vnd.oci.image.layer.enc.v1` and specify the original underlying mediatype in an annotation. But i'm unsure if this may make handling on the runtimes a bit more difficult.

cc: @estesp @dmcgowan @crosbymichael @mrunalp @mtrmac @vrothberg

dmcgowan:
I don't think reproducing what was done for nondistributable layers is a good idea. Runtimes end up having to special case these since they are hard to interpret. For example `vnd.oci.image.layer.` may indicate that an object should be treated as a layer, but then `.tar.v1` indicates it is a tar. What is sandwiched in between those is some important metadata. In the nondistributable case it is just a hint of how an artifact is distributed, in the encrypted case, it is a requirement for how the tar must be processed. I think we really need to format the content types in a way that indicates how they should be processed, ensuring that modifications on those types are wrapping. Personally I like using `+` to indicate encoding (whether that is encryption or formatting). I think we should discuss at an upcoming OCI weekly meeting about these types and their relation to [RFC6838](https://tools.ietf.org/html/rfc6838).

lumjjb:
Updated based on discussions from OCI weekly dev discussion to move from `+enc` to `+encrypted`. We agreed on using suffixes on mediatypes and having these defined on top of artifacts.

lumjjb:
@mikebrow Responding to https://github.com/opencontainers/image-spec/issues/791#issuecomment-533394562 in this thread instead since it is more specific to encryption details.

The main integration points for Encrypted Container Images are:
- The specification and libraries to perform encryption/decryption
- Container runtimes
- The build toolchain
- Registry
- Kubernetes Integration

Do let me know if there's something specific you are looking for and I can provide a direct link. 

## The specification and libraries to perform encryption/decryption

The definition of encryption indication and how encryption metadata is defined in: 
- https://github.com/opencontainers/image-spec/pull/775

The implementation to perform encryption and decryption of this spec is in:
- https://github.com/containers/ocicrypt
  - `func EncryptLayer(ec *config.EncryptConfig, encOrPlainLayerReader io.Reader, desc ocispec.Descriptor) (io.Reader, EncryptLayerFinalizer, error)`
  - `func DecryptLayer(dc *config.DecryptConfig, encLayerReader io.Reader, desc ocispec.Descriptor, unwrapOnly bool) (io.Reader, digest.Digest, error)`

## Container Runtimes

Current implementations for containerd stack
- https://github.com/stefanberger/imgcrypt (To be moved to containerd repository)
- Use with containerd 1.3 with these PRs:
  - https://github.com/containerd/containerd/pull/3649
  - https://github.com/containerd/containerd/pull/3482

Current implementations for RedHat stack
- https://github.com/harche/cri-o/tree/pr_branch
- https://github.com/containers/image/pull/673 (Library to support use in cri-o)

## Registry

Current docker distribution supports: https://github.com/docker/distribution. Tested with v2.7.1.

## Build toolchain

The plan for this is to have docker CLI, skopeo and buildah. In the meantime, we also support encryption of images in containerd via `imgcrypt` ctr client.

Current implementations for RedHat stack
- https://github.com/containers/image/pull/673 (Library to support use in containers sub-projects)
- https://github.com/harche/skopeo/tree/pr_branch
- https://github.com/harche/cri-o/tree/pr_branch

Current implementations for containerd stack
- ctr-enc: https://github.com/stefanberger/imgcrypt (To be migrated to containerd repo)
- To integrate into docker CLI, we are currently waiting on https://github.com/moby/moby/issues/38043. Tracking with https://github.com/moby/buildkit/issues/714

## Kubernetes Integration

KEP: Add ImageDecryptSecrets
- https://github.com/kubernetes/enhancements/pull/1066 (Tagged 1.17)

With KEP implementation in 1.17, containerd-cri and cri-o will need to add support the new API changes.

It is important to note that kubernetes integration is not required to use encrypted container images. i.e. in cri-o you are able to specify a directory containing keys.

cc: @stefanberger, @harche

vbatts:
@lumjjb ok, finally discussing this in the weekly call.
Since this encrypted layer discussion has begun, the https://github.com/opencontainers/artifacts has come along. To a large degree, this is another mimetype/mediaType that would be documented there, and allowed as an extension to the image-spec. 

There is a lot of good work here in the PR, that should transfer to something like a README for these mime-types in the artifacts repo. 
Does this make sense?

lumjjb:
>@lumjjb ok, finally discussing this in the weekly call.

Awesome, sounds good.. I will come by the next OCI call on Wednesday.

>There is a lot of good work here in the PR, that should transfer to something like a README for these mime-types in the artifacts repo.
Does this make sense?

Yup! I will create an `encryption.md` PR to the artifacts repo, and I suppose it would be an edit pointing to that file from https://github.com/opencontainers/artifacts/pull/13/ ?

AkihiroSuda:
What's current status of this?

vbatts:
@AkihiroSuda it seems to be stalled on https://github.com/opencontainers/artifacts/pull/15

lumjjb:
As a quick update on this, I've updated the list of links in the comments above.

 The encryption support is now part of upstreamed containerd, cri-o, buildah, docker registry and skopeo tools.

npmccallum:
@vbatts I'm a bit late to the party. But this has recently come into scope for me.

It isn't clear to me from the various pieces of documentation strewn about whether the integrity protection for a given layer also provides integrity protection for all the layers beneath it.

For example, you have an image with multiple layers:
* Layer 1
* Layer 2
* Layer 3 (encrypted)

Does the integrity protection of layer 3 also include measurements of layers 2 and 1?

If not, it would be possible to modify one of the intermediate layers in order to gain backdoor access to the plaintext.

lumjjb:
Hi @npmccallum The encryption is only targeting confidentiality, although it has some side effects of integrity on a per layer basis. This was designed to be used in conjunction with Docker Content Trust / Notary / RH Simple signing, they should be able to work hand in hand.

npmccallum:
@lumjjb If you don't have integrity of the whole, you don't have confidentiality of the whole.

lumjjb:
If i understand your statement correctly, I believe those signing technologies ensure the integrity of the image manifest, and the integrity of the layers are ensured as being part of the signed manifest hash. This is done as a side effect of layers being content hashes.

npmccallum:
@lumjjb Okay. So this is working off the presumption that you have integrity protection for the whole manifest which contains measurements of all the layers. Therefore, there is integrity protection for the whole and confidentiality protection for the parts.

During deployment, do you ensure that the manifest is validly signed and all the layers match their measurements before you decrypt? Or is it possible to decrypt without validating a signature of the manifest and validating the measurement of the layers?

lumjjb:
>During deployment, do you ensure that the manifest is validly signed and all the layers match their measurements before you decrypt? Or is it possible to decrypt without validating a signature of the manifest and validating the measurement of the layers?

Yes, this is usually done by the pull/admission phase of the image, so integrity is measured before decryption is done during the unpacking phase.

npmccallum:
@lumjjb Do you refuse to decrypt if the manifest is unsigned?

stefanberger:
> If not, it would be possible to modify one of the intermediate layers in order to gain backdoor access to the plaintext.

How would that be possible? We are encrypting the tar'ed and gzip'ed layers using `AES_256_CTR_HMAC_SHA256` and a random IV. The symmetric encryption parameters, including the key and IV, are contained in a JSON document which is encrypted for each recipient using an RSA key for encryption with JWK, PKCS7, PKCS11, and gpg.

lumjjb:
>@lumjjb Do you refuse to decrypt if the manifest is unsigned?

That is the choice of the runtime policy, the encryption technology is agnostic to this, depending on the setup

npmccallum:
> > If not, it would be possible to modify one of the intermediate layers in order to gain backdoor access to the plaintext.
> 
> How would that be possible? We are encrypting the tar'ed and gzip'ed layers using `AES_256_CTR_HMAC_SHA256` and a random IV. The symmetric encryption parameters, including the key and IV, are contained in a JSON document which is encrypted for each recipient using an RSA key for encryption with JWK, PKCS7, PKCS11, and gpg.

How you encrypt the individual layer is irrelevant for this attack since the attack works by modifying one of the unencrypted layers.

stefanberger:
> How you encrypt the individual layer is irrelevant for this attack since the attack works by modifying one of the unencrypted layers.

How do you get to decrypt the encrypted layers if you don't have the RSA private key?

npmccallum:
> > @lumjjb Do you refuse to decrypt if the manifest is unsigned?
> 
> That is the choice of the runtime policy, the encryption technology is agnostic to this, depending on the setup

That seems to me like the wrong choice to preserve the security of the system as a whole. If you enable an encrypted layer, you should prevent decryption whenever integrity of the whole cannot be verified.

stefanberger:
The integrity of the whole can be verified with (previous) Notary or whatever may be in use these days -- we didn't want to re-invent this part. We do NOT guarantee that the unencrypted layers are not containing garbage but I would like to see how one can decrypt the RSA-encrypted layers.

lumjjb:
>That seems to me like the wrong choice to preserve the security of the system as a whole. If you enable an encrypted layer, you should prevent decryption whenever integrity of the whole cannot be verified.

My personal recommendation is to sign/verify images always. But I would defer this discussion to the orchestration components (i.e. kubernetes, cri-o, etc.), they are the decision makers in these matters.

npmccallum:
> How do you get to decrypt the encrypted layers if you don't have the RSA private key?

Under this attack, the attacker has access to:
1. one of the unencrypted layers
2. the unsigned manifest

The goal of the attacker is to extract the plaintext from the encrypted layer. However, the attacker does not have the key to perform decryption. Therefore, the attacker wants to trick the runtime that deploys the image into disclosing the plaintext on his behalf.

The attacker modifies the unencrypted layer to add a backdoor to an executable within the execution path (for example: bash). The attacker additionally modifies the manifest to contain the measurement of the backdoored layer.

When the container is deployed, the attacker does not have access to the deployment system. However, the deployment system validates the layer measurements (including the fraudulent measurement from the modified manifest) and assembles a container that contains both the backdoored binary (i.e. bash) and the plaintext from the encrypted layer. When the deployment system executes the container, the backdoored binary exfiltrates the plaintext.

Therefore, the confidentiality of the encrypted layer fails by modifying an unencrypted layer when there is no signature on the manifest. To defend against this attack, the deployment host must refuse to decrypt data whenever the manifest has no integrity protection (i.e. signature).

npmccallum:
> My personal recommendation is to sign/verify images always. But I would defer this discussion to the orchestration components (i.e. kubernetes, cri-o, etc.), they are the decision makers in these matters.

That requirement is probably onerous. Many people are happy to deploy images with no integrity/encryption. But, as I outlined above, you should refuse to decrypt an encrypted layer if there is no integrity protection on the manifest (or if there is some way to assemble a container from a layer with no measurement). This is because if you don't have integrity of all layers, you don't have confidentiality of any particular layer. Since the party that produce the container obviously wants confidentiality of a layer, a failure to sign the manifest in the presence of an encrypted layer should be treated as a misconfiguration that produces a security vulnerability.

npmccallum:
An alternate implementation would include the measurements of each dependent layer inside the AEAD of the encrypted layer.

lumjjb:
@npmccallum I don't disagree with your point that confidentiality != integrity. But I think that that is handled on a higher level, the proposal here is just talking about the **mechanics** of encryption, not the big picture. Kind of like DH is part of TLS or IPSec.

The audience for your problem is really towards the orchestration. So i'm going to tag a couple folks that will be able to provide more concrete actionable statements. I would recommend also bringing this to SIG-Node or a relevant kubernetes sig, and would be happy to be involved with the discussions there.

@mtrmac @mrunalp @mikebrow

stefanberger:
> An alternate implementation would include the measurements of each dependent layer inside the AEAD of the encrypted layer.

Would it have to be each dependent layer or if it makes it easy an 'accumulation' of the hashes of each of the previous layers? So like TPM PCR measurements: next_hash=sha256(cur_hash || layer_hash). That should work as well, or not?

npmccallum:
@lumjjb My point is not that confidentiality != integrity. My point is that you are assembling an executable from parts, some of which are supposed to be secret. But if you allow an attacker to modify other parts of your executable and then combine the plaintext with it, you are unable to guarantee the privacy of the plaintext. This is not an orchestration problem. It is either a mechanics implementation problem or a protocol problem.

mtrmac:
I think it’s very valid to be concerned that encryption can be deployed in an insecure way. I’m not sure yet that that has to be reflected in the encryption implementation, let’s nail down the required security properties of the implementation first.

---

I don’t see that the attacker should modify a _previous_ layer: backdooring a child layer on top of the victim encrypted one would work just as well (or maybe even better, because the attacker doesn’t have to worry about the victim layer overwriting the backdoor).

Or the attacker could just modify the image config so that the container’s start command is the backdoor, without modifying the layers; many container images contain enough tools to make that possible.

So we would really need full integrity verification of the image _as a whole_, including the manifest; treating layers as individual units or as “parent images” wouldn’t help.

But then again, once we are talking about integrity verification of the image as a whole, while encrypting only the layers (and not the config), this is getting pretty close to an implementation detail of the client (which _can_ easily just refuse to decrypt images referenced by tag without a signature or a trusted digest value); it’s not obvious to me that we need to codify anything for the encryption format details (or maybe we just need a “security considerations: implementations SHOULD” reminder).

---

On the other hand, this conversation might be a good enough reason to revisit the idea of going one layer at a time, or of not encrypting the config. Did we even have a detailed conversation about these things in the past?

lumjjb:
>On the other hand, this conversation might be a good enough reason to revisit the idea of going one layer at a time, or of not encrypting the config. Did we even have a detailed conversation about these things in the past?

The discussion included the config, and having the `+encryption` be usable on configs and any OCI objects, but the decided scope because of how implementation was handled was to start with layers and extend it in the future.

npmccallum:
> > An alternate implementation would include the measurements of each dependent layer inside the AEAD of the encrypted layer.
> 
> Would it have to be each dependent layer or if it makes it easy an 'accumulation' of the hashes of each of the previous layers? So like TPM PCR measurements: next_hash=sha256(cur_hash || layer_hash). That should work as well, or not?

Yes. Though there are some concerns in the cryptographic community about such strategies because the final hash contains a stronger representation of bits from the more recent layers than the earliest layers. This weakness would have to be combined with a second preimage attack to be effective. So I'm not generally worried about it.

However the dependent layers also include the layers above (since the result is one flat executable unit).

I'm not sure precisely of the use cases you're intending to solve here. But once there is a single encrypted layer involved, integrity of the whole needs to be enforced. There's also a usability concern with having multiple layers encrypted by multiple keys. The interactions between layers get really hairy once you have multiple encrypted layers encrypted by multiple keys (parties). It becomes difficult to reason about.

stefanberger:
> However the dependent layers also include the layers above (since the result is one flat executable unit).

Yes, though now we are back to the Notary and the like since those protect the whole image manifest tree and layers. It could be a recommendation to sign encrypted images since these two technologies may be used independent of each other. Maybe it should be enforced on the client side when downloading and unpacking an encrypted image that it also has been signature-verified. Maybe it could be a recommendation to the integrity-conscious user to perform the signature verification.

> 
> I'm not sure precisely of the use cases you're intending to solve here. But once there is a single encrypted layer involved, 

The primary use case is protecting the image contents at rest.

> integrity of the whole needs to be enforced. There's also a usability concern with having multiple layers encrypted by multiple 

Yes, and we did think of existing technologies for doing this, just that it's not 'enforced' on the client sides.

npmccallum:
> I think it’s very valid to be concerned that encryption can be deployed in an insecure way. I’m not sure yet that that has to be reflected in the encryption implementation, let’s nail down the required security properties of the implementation first.
> 
> I don’t see that the attacker should modify a _previous_ layer: backdooring a child layer on top of the victim encrypted one would work just as well (or maybe even better, because the attacker doesn’t have to worry about the victim layer overwriting the backdoor).
> 
> Or the attacker could just modify the image config so that the container’s start command is the backdoor, without modifying the layers; many container images contain enough tools to make that possible.
> 
> So we would really need full integrity verification of the image _as a whole_, including the manifest; treating layers as individual units or as “parent images” wouldn’t help.

Correct in all cases.

> But then again, once we are talking about integrity verification of the image as a whole, while encrypting only the layers (and not the config), this is getting pretty close to an implementation detail of the client (which _can_ easily just refuse to decrypt images referenced by tag without a signature or a trusted digest value); it’s not obvious to me that we need to codify anything for the encryption format details (or maybe we just need a “security considerations: implementations SHOULD” reminder).

As someone who has authored and otherwise been involved with numerous cryptographic standards, this needs a `MUST` statement since the security of the system depends on the correct implementation. I've taken the `SHOULD` approach in the past and regretted it.

> On the other hand, this conversation might be a good enough reason to revisit the idea of going one layer at a time, or of not encrypting the config. Did we even have a detailed conversation about these things in the past?

I suspect there is a desire to allow the layering of encrypted layers on top of unencrypted layers. It might be enough to codify:

1. Unencrypted layers `MUST NOT` be built on top of encrypted layers.
2. Encrypted layers `MAY` be built on top of unencrypted layers.
3. All encrypted layers `MUST` use the same encryption key.
4. An encrypted layer `MUST` include the measurements of all previous layers in its AEAD.

The above is just a proposal. But without some similar constraints, I think it will be nearly impossible to reason about the confidentiality of the resulting container.

lumjjb:
I like the idea about codifying some implementation guidelines (SHOULD/MUST) for clients implementing the spec. I am still leaning towards re-using container signing capabilities in these recommendations. 

However, I think it will be helpful @npmccallum if you can note some of the threat model / usecases that you are targeting and we can dig out our discussions from before on this to compare notes. It does sound like you are looking to expand the original scope of this outside of just encryption at rest, which is not what we had in mind for this, but are open to discussions and see where this goes.

stefanberger:
> * Unencrypted layers `MUST NOT` be built on top of encrypted layers.

I don't think that this can be enforced. If someone was hell-bent to manipulate an image he would change the build tool or add a layer manually and re-calculate the hashes and updated them in the manifest and then push the stuff to a repo.

The only way to solve this is via signature over the whole image and that's where image integrity enforcement would have to be done via Notary etc.

lumjjb:
Also tagging @SteveLasker @justincormack for relevancy

mtrmac:
>  I think it will be helpful @npmccallum if you can note some of the threat model / usecases that you are targeting a

@lumjjb same for your approach — what use cases benefit from encrypting only some of the layers (as was implemented in c/image), and why are you not concerned about encrypting the config?

lumjjb:
>@lumjjb same for your approach — what use cases benefit from encrypting only some of the layers (as was implemented in c/image), and why are you not concerned about encrypting the config?

Yea will have to go back to these meeting notes and dig them up. I remember us being concerned about encrypting the config, but this was very non-trivial to do this in containerd, so the discussion was to put it on the side till the OCI artifacts spec was done and make it work on generic OCI images.

mtrmac:
> > * Unencrypted layers `MUST NOT` be built on top of encrypted layers.
> 
> I don't think that this can be enforced.

Huh? It seems conceptually pretty trivial to enforce in the c/image decryption implementation.

> If someone was hell-bent to manipulate an image he would change the build tool or add a layer manually and re-calculate the hashes and updated them in the manifest and then push the stuff to a repo.

Of course someone can create an image that violates these rules, and someone else can write decryption code that decrypts code even if it violates these rules.

I think of the `opencontainers` project as a place to establish standards for _interoperability_, and documenting “such images MUST be rejected by compliant implementations” is as close as we can get, by just publishing a document in this repo, if we don‘t have a formal product validation process as a precondition to being allowed to use a trademark or any similar enforcement mechanism.

stefanberger:
>> Unencrypted layers MUST NOT be built on top of encrypted layers. 

> Huh? It seems conceptually pretty trivial to enforce in the c/image decryption implementation.

Yes, my fault. This would be enforced on the side running the image not building it.

stefanberger:
An article we wrote about this is titled ["Encrypted container images for container image security *at rest*"](https://developer.ibm.com/solutions/security/articles/encrypted-container-images-for-container-image-security-at-rest/).

"Another security concern is making sure that the container you are going to run is the one you requested to run and that it hasn’t been modified. To address this issue, you can use digital signatures securely stored in a notary to protect containers from modification."

npmccallum:
@lumjjb @mtrmac These are the threat models I'm currently dealing with.

# Threat Models
## Physical Disk

Under this scenario, the attacker is able to gain access to the physical disk containing a container image layer which contains secret data. Examples include:
* a dumpster diver grabbing an EOL disk
* an unauthorized technician at a disk warranty repair center
* theft of the disk when transferred to a low security area

In most cases, the data, once accessed, will not be used. The existing proposal covers this scenario well.

However, in some cases - for example a repaired disk - the data may be put back into production if good data reuse policies are not in place or are not followed. The existing proposal does not handle this scenario.

## Container Repository

Under this scenario, the attacker gains access to a container repository. Examples include:
* a rogue repository admin at a repository maintainer
* a rogue admin at the cloud where the repository is hosted
* any attacker who gains access to the above systems through a vulnerability
* a nation state which engages in active surveillance or sabotage using the repo

Unlike the **Physical Disk** model, it is always presumed that the compromised data will be deployed in production. Therefore, any mitigation must take steps to prevent the execution of tampered code since it could be used to disclose the data during deployment. The existing proposal does not handle this scenario.

## Confidential Computing

Hardware vendors have been producing trusted execution environments (TEEs) now for several years, including:
1. Intel SGX/TDX
2. AMD SEV/SEV-ES/SEV-SNP
3. POWER PEF
4. Arm TrustZone

These systems encrypt or otherwise isolate memory in use. The intent is to remove the host platform from the trust model. Therefore, attackers may include:
* a rogue admin at the cloud host
* any attacker who gains access to the cloud host
* a nation state which engages in active surveillance or sabotage using the host

Deploying a container into these environments requires careful measurement to ensure that the container image is not viewable or modifiable by the host. Therefore, we need to protect the container image from the tenant, through the repo **and** through the host into the protected area. For an example of this, [see how this is being handled for Kata](https://github.com/kata-containers/kata-containers/issues/1332#issue-795892219). The existing proposal does not handle this scenario.

npmccallum:
> An article we wrote about this is titled ["Encrypted container images for container image security _at rest_"](https://developer.ibm.com/solutions/security/articles/encrypted-container-images-for-container-image-security-at-rest/).
> 
> "Another security concern is making sure that the container you are going to run is the one you requested to run and that it hasn’t been modified. To address this issue, you can use digital signatures securely stored in a notary to protect containers from modification."

To be clear, I'm absolutely okay with punting on image integrity to a notary so long as the runtime refuses to decrypt a layer at deployment time if a notary is not used.

jejb:
> > An article we wrote about this is titled ["Encrypted container images for container image security _at rest_"](https://developer.ibm.com/solutions/security/articles/encrypted-container-images-for-container-image-security-at-rest/).
> > "Another security concern is making sure that the container you are going to run is the one you requested to run and that it hasn’t been modified. To address this issue, you can use digital signatures securely stored in a notary to protect containers from modification."
> 
> To be clear, I'm absolutely okay with punting on image integrity to a notary so long as the runtime refuses to decrypt a layer at deployment time if a notary is not used.

Before this gets over designed, the use case we're thinking of in the cloud is to deploy a container by name (i.e. sha256sum).  We're planning on having this presented to the confidential environment over the control channel either by the Tenant, in the case of direct connection, or by the tenant owned agent in the cloud.  This means the deployer inside the enclave knows the expected hash of the image which means no-one can tamper with the layers without that hash mismatching (and, on behalf of the tenant, it will refuse to deploy the container if that happens).

I think all clouds will adopt a deployment method like this meaning there's absolutely no need for signatures as well ... although they can be used if the tenant chooses, so there's no requirement for a notary.

stefanberger:
> I suspect there is a desire to allow the layering of encrypted layers on top of unencrypted layers. It might be enough to codify:
> 
>     1. Unencrypted layers `MUST NOT` be built on top of encrypted layers.
> 
>     2. Encrypted layers `MAY` be built on top of unencrypted layers.
> 
>     3. All encrypted layers `MUST` use the same encryption key.
> 
>     4. An encrypted layer `MUST` include the measurements of all previous layers in its AEAD.

@npmccallum Would it not be sufficient to require that the last layer must be encrypted for each key that was used? At that layer the accumulation of the hashes of all previous layers, encrypted or plain, can be enforced.

lumjjb:
Have not forgotten about this - have allocated some time next week to go through the previous discussions and synthesize them.

lumjjb:
Below are the goals and usecases/scenarios from the initial proposal

## Content encryption of OCI artifacts (Protection at rest from build -> runtime)

All artifacts have their content kept confidential from the build phase to the runtime or whenever the artifact needs to be consumed. This attempts to address the attack vectors between build and runtime. The threat model is that the producer and consumers of the image are trusted. The proposed PR is https://github.com/opencontainers/artifacts/pull/15.

The discussions of this was surrounding the work going around Docker Content Trust / Notary around that time. It was the intention that DCT or other image manifest signing technologies together with the artifact design of being content addressible would be used to attest the integrity and provenance of the payload.


### Initial Scope

The initial focus of the ocicrypt project was on the container images. The idea was to focus on the encryption of the content aspects. This consists of the config file and the layers. 

The work was later scoped down to only layer artifacts. This was due to the implementation scope as a POC in containerd. The mechanics of introducing this into configs were much more involved than the layers, resulting in the initial scope.


## Geofencing execution

Another scenario was towards providing geofencing capabilities. Encrypting content to tie it to key material which can then be securely distributed to an authorized platform. The specific motivator of this usecase was geofencing and compliance, and export control/DRM were other examples.

Keys would be used or distributed with platform attestation-backed authorization. Thus with key management, The images would only be usable if they were on nodes that are authorized. This is talked about a bit more in https://kccnceu20.sched.com/event/Zepc.

## Container layer flexibility

One of the main themes for the design on layers was to try to preserve characteristics of container images that people like. This includes ability to build on top of other images, layer reuse, external image scanning. 

The decision for encrypting each layer with a generated DEK, and wrapping the key was to be able to provide layer re-use. The wrapped metadata is therefore stored within the annotations and not the layer itself, as authorizing a new consumer would have changed the hash of the layer.

Having the wrapped keys as annotations (these wrapped keys could be artifacts on their own pointing to the layers), would also allow approved image scanners or trusted CI/CD tooling to perform analysis on the layers. Thus one could encrypt an image for both its runtime and a trusted party.

Another notable discussion point was that of multiple owners of different parts of the stack. The idea is one that is difficult to do with current tooling due to the way the builders construct images, but not impossible. For example, if the middleware team maintains the middleware layer and the application team is responsible for the application. An encrypted image could be created by the layering of the middleware encrypted layer, and the application could be
added on top without accessing the middleware, based on a file based contract provided between teams. The runtime could then run the container image only if it is authorized to use components of both teams.


@stefanberger

annotations.md
```diff
@@ -40,6 +40,10 @@ This specification defines the following annotation keys, intended for but not l
     ```
 * **org.opencontainers.image.title** Human-readable title of the image (string)
 * **org.opencontainers.image.description** Human-readable description of the software packaged in the image (string)
+* **org.opencontainers.image.org.opencontainers.image.enc.keys.pkcs7** Base64 comma separated [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
```
vbatts, line 43:
```suggestion
* **org.opencontainers.image.enc.keys.pkcs7** Base64 comma separated [PKCS7(RFC2315)](https://tools.ietf.org/html/rfc2315) encrypted messages that contain wrapped PrivateLayerBlockCipherOptions JSON objects for [encryption](layer.md#layer-encryption)
```

vbatts:
@lumjjb would like to bubble this work up. What is the current situation and your approach? has it evolved since you first opened this PR? I'm thinking this discussion could start up again.

lumjjb:
Hey @vbatts ! Certainly , for some reason this was buried in my inbox and surfaced when i was searching for something! Would like to get back to this discussion, where shall be a good place to do this? Should we chat on the sidebar first or bring this up at one of the weekly oci calls?

vbatts:
On 21/07/21 07:59 -0700, Brandon Lum wrote:
>Hey @vbatts ! Certainly , for some reason this was buried in my inbox and surfaced when i was searching for something! Would like to get back to this discussion, where shall be a good place to do this? Should we chat on the sidebar first or bring this up at one of the weekly oci calls?

Here, mailing-list ***@***.***), or the weekly.

layer.md
```diff
@@ -332,6 +339,96 @@ Implementations SHOULD NOT upload layers tagged with this media type; however, s
...
+
+When using the `+encrypted` mediatype, the layer data blobs are encrypted. In order to decrypt an image, the encryption metadata is required. 
+
+## Encryption Metadata
+
+The encryption metadata consists of 2 parts: PublicLayerBlockCipherOptions and PrivateLayerBlockCipherOptions. The PublicLayerBlockCipherOptions contain encryption metadata that is public, i.e. cipher type, HMAC, etc. and PrivateLayerBlockCipherOptions contains the encryption metadata which should be confidential, i.e. symmetric key, nonce (optional), etc. These are stored in the respective [**org.opencontainers.image.enc** prefixed annotations](annotations.md).
+
+Below are golang definitions of these JSON objects:
+
+```
```
AkihiroSuda, line 355: nit: ` ``` ` ->  ` ```go ` for readability

lumjjb: Done!

layer.md
```diff
@@ -4,7 +4,7 @@ This document describes how to serialize a filesystem and filesystem changes lik
...
 This document will use a concrete example to illustrate how to create and consume these filesystem layers.
 
-This section defines the `application/vnd.oci.image.layer.v1.tar`, `application/vnd.oci.image.layer.v1.tar+gzip`, `application/vnd.oci.image.layer.v1.tar+zstd`, `application/vnd.oci.image.layer.nondistributable.v1.tar`, `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`, and `application/vnd.oci.image.layer.nondistributable.v1.tar+zstd` [media types](media-types.md).
+This section defines the `application/vnd.oci.image.layer.v1.tar`, `application/vnd.oci.image.layer.v1.tar+gzip`, `application/vnd.oci.image.layer.v1.tar+zstd`, `application/vnd.oci.image.layer.nondistributable.v1.tar`, `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip`, `application/vnd.oci.image.layer.nondistributable.v1.tar+zstd`, `application/vnd.oci.image.layer.v1.tar+encrypted`, `application/vnd.oci.image.layer.v1.tar+gzip+encrypted`, `application/vnd.oci.image.layer.nondistributable.v1.tar+encrypted` and `application/vnd.oci.image.layer.nondistributable.v1.tar+gzip+encrypted` [media types](media-types.md).
```
optimistyzy, line 7: May I ask a question why tar+ encrypted + zstd or tar + zstd + encrypted is not supported? Thanks.

lumjjb: at the time, there wasn't `+zstd` yet, but yes it should support. `+encrypted` should be treated as a suffix and thus should be handled on all mediatypes that end with `+encrypted`

lumjjb:
In that case, you may want to just skip the compression since encrypted
bytes have inherent randomness and high entropy , which results in
compression not being effective.

PR #777, opened 2019-05-06
Add CPU variant to image config

pricec:
This commit adds the CPU variant to the image config type. It also
refactors the image index specification to isolate the platform
variant specifications, allowing a reference from the config spec.
The go specs are updated to include the new field in the v1.Image
struct, and tests are updated to include the new field.

Signed-off-by: Chris Price <chris.price@docker.com>

### DIFF
--- a/config.md
+++ b/config.md
@@ -105,6 +105,11 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
   The CPU architecture which the binaries in this image are built to run on.
   Configurations SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOARCH`][go-environment].
 
+- **variant** *string*, OPTIONAL
+
+  The variant of the specified CPU architecture.
+  Configurations SHOULD use, and implementations SHOULD understand, `variant` values listed in the [Platform Variants](image-index.md#platform-variants) table.
+
 - **os** *string*, REQUIRED
 
   The name of the operating system which the image is built to run on.
--- a/conversion.md
+++ b/conversion.md
@@ -46,17 +46,19 @@ These fields all affect the `annotations` of the runtime configuration, and are 
 | ------------------- | --------------- | ----- |
 | `os`                | `annotations`   | 1,2   |
 | `architecture`      | `annotations`   | 1,3   |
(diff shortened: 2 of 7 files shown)


dmcgowan:
This is also upstreaming a change in buildkit and soon to be in moby. https://github.com/moby/buildkit/blob/9a18f8040332acebbb8a9a21772560ab99d7fdaa/frontend/dockerfile/dockerfile2llb/image.go#L55

Note this change is useful for builders which create indexes from a list of manifests. The manifest links to the config, however today the config does not contain all the information needed to correctly construct an index containing 32-bit ARM images. A larger change such as adding the entire platform would complicate client compatibility, further hindering the ARM use cases. With this change clients need only to update the `Image` structure and use the variant if copying to a platform structure or no change at all if parsing into a platform structure, however, adding a new platform field would require clients to update their `Image` structure and add logic to clean up the structure EVERY time the structure is parsed just to support this ARM use case.

vbatts:
LGTM

tianon:
This came up in today's OCI call (cc @ibuildthecloud), and I think this only solves part of the issue, right?  We're still missing `os.version` and friends needed for Windows, so wouldn't it make more sense to add the full `platform` object to the image config?  (at the very least, all the fields it supports would be useful)

Toasterson:
Adding the full `platform` object to the Image config has the added bonus of keeping things in sync later down the line.

jonjohnsonjr:
> Adding the full platform object to the Image config has the added bonus of keeping things in sync later down the line.

I'd be in favor of [embedding](https://golang.org/doc/effective_go.html#embedding) the whole [`v1.Platform`](https://github.com/opencontainers/image-spec/blob/79b036d80240ae530a8de15e1d21c7ab9292c693/specs-go/v1/descriptor.go#L45-L64) into [`v1.Image`](https://github.com/opencontainers/image-spec/blob/79b036d80240ae530a8de15e1d21c7ab9292c693/specs-go/v1/config.go#L82) to achieve this.

cpuguy83:
Agree, let's bring in the whole platform object.
Does this need to be carried?

cpuguy83:
Carried these changes in https://github.com/opencontainers/image-spec/pull/809

PR #779, opened 2019-05-29
update link to weekly meeting

bacongobbler:


bacongobbler:
hold up, need to fix the DCO signoff

jonboulle:
thanks for the fix, but I think #786 is slightly preferable, so going to land that instead

PR #795, opened 2019-11-26
index: add media type

vrothberg:
Add a media type to the image index.  The spec mentions that an index
media type is reserved for use, to maintain compatibility.  Although
not being strictly required, some tools might want to set the media type
to be more explicit.

Signed-off-by: Valentin Rothberg <rothberg@redhat.com>

### DIFF
--- a/specs-go/v1/index.go
+++ b/specs-go/v1/index.go
@@ -21,6 +21,9 @@ import "github.com/opencontainers/image-spec/specs-go"
 type Index struct {
 	specs.Versioned
 
+	// MediaType is the media type of the index.
+	MediaType string `json:"mediaType,omitempty"`
+
 	// Manifests references platform specific manifests.
 	Manifests []Descriptor `json:"manifests"`
 


vrothberg:
@vbatts @cyphar PTAL

jonjohnsonjr:
> Although not being strictly required, some tools might want to set the media type to be more explicit.

I'm curious which tools you have in mind?

vrothberg:
> > Although not being strictly required, some tools might want to set the media type to be more explicit.
> 
> I'm curious which tools you have in mind?

We had a request for Buildah asking to set the media type. It's being set for Docker v2s2 and the users assumed there's a bug in Buildah for OCI indexes which motivated me to check what the spec states.

cyphar:
Deciding to not include (or rather, deciding to *remove*) embedded media-types was the conclusion of a fairly long-winded argument from a few years ago (see #411). The shortest possible version of the reasoning is "what Docker originally did was wrong, typed descriptors invalidate the need for embedded media-types".

I was originally in favour of having them embedded in the blob (because it would allow you to have tools that operate on blobs without needing to have a descriptor), but it turns out that embedding them has several flaws:

  1. Making use of the embedded media-type requires parsing JSON without knowing what type it is. This already causes problems with most statically-typed languages' JSON libraries. Not a deal-breaker necessarily, but it's definitely not ideal. It should be just as easy to write an image tool in Rust or Go as it is to write one in Python.

  2. The number of usecases where you need to parse an individual blob without knowing its media-type beforehand is fairly limited. In almost all cases you will be dealing with a full image or at the very least you will know (contextually) that you're operating on an index or other such object.

  3. If there is a conflict between the media-type and the type in the descriptor, which should win? Should that be invalid (probably)? In which case, should a program which wants to use an alternative media-type modify the blob (probably)? What if the original blob is signed as part of an image (then the signature is invalid)?

  4. If we upgrade a media-type in a backwards-compatible way (but wish to bump the version number) then you will get needless duplication of JSON blob because the media-type needs to be updated (even though the blob is unchanged).

  5. In other projects and protocols, media-types are rarely embedded into the object itself. Their primary purpose is to describe a blob (which may not be specific to a particular protocol)'s type.

So, unfortunately (given that the reasoning behind this is well-documented and well-discussed in #411) I have to NACK this -- unless there is a strong counterargument to the consensus reached two years ago.

vrothberg:
Thanks for the very detailed reply, @cyphar. The [spec states](https://github.com/opencontainers/image-spec/blob/master/image-index.md#image-index-property-descriptions):

> _mediaType string_
>
>_This property is reserved for use, to maintain compatibility. When used, this field contains the media type of this document, which differs from the descriptor use of mediaType._

I guess my main issue is that the upper description (which presumably is meant to summarize #411)  is hard to understand for me. Does it mean:
* optional as in "can be set at wish but is not required"?
* reserved as in "not to be used"?

I interpret it as not "not to be used" hence optional, which drove me to open this PR.

vbatts:
(as history would have it, I am in favor of having the self-describing `mediaType` field present)

cyphar:
@vrothberg "Reserved for use" means "not to be used [until we allow it]". The idea is that the specification reserves the right to use that field for something in the future -- I will admit it's a bit strangely worded (the spec explicitly says don't use it, so I don't understand why we're describing what its behaviour might be if we add it in the future) but we have a different way of indicating that something is optional (including the word `OPTIONAL` in all caps).

PR #800, opened 2020-03-23
Remove go4.org dependency

zhsj:
The offset to line/col translation is easy to implement. Not worthy
to bring a third-party dependency.

PS. I think this code is never used, but removing it is too aggressive
since it exposes public api.

Signed-off-by: Shengjing Zhu <zhsj@debian.org>

### DIFF
--- a/schema/error.go
+++ b/schema/error.go
@@ -15,10 +15,9 @@
 package schema
 
 import (
+	"bufio"
 	"encoding/json"
 	"io"
-
-	"go4.org/errorutil"
 )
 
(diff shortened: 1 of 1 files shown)


vbatts:
that's fair.
LGTM

jonjohnsonjr:
LGTM

jonjohnsonjr (approved):

jonjohnsonjr:
I think we need to bounce this somehow to reset the checks.

vbatts (approved):

PR #809, opened 2020-09-30
Embed platform fields in Image spec

cpuguy83:
Carries #777 + adds other platform fields
Closes #777

### DIFF
--- a/config.md
+++ b/config.md
@@ -105,6 +105,11 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
   The CPU architecture which the binaries in this image are built to run on.
   Configurations SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOARCH`][go-environment].
 
+- **variant** *string*, OPTIONAL
+
+  The variant of the specified CPU architecture.
+  Configurations SHOULD use, and implementations SHOULD understand, `variant` values listed in the [Platform Variants](image-index.md#platform-variants) table.
+
 - **os** *string*, REQUIRED
 
   The name of the operating system which the image is built to run on.
--- a/config.md
+++ b/config.md
@@ -105,15 +105,28 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
   The CPU architecture which the binaries in this image are built to run on.
   Configurations SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOARCH`][go-environment].
 
-- **variant** *string*, OPTIONAL
-
-  The variant of the specified CPU architecture.
-  Configurations SHOULD use, and implementations SHOULD understand, `variant` values listed in the [Platform Variants](image-index.md#platform-variants) table.
-
 - **os** *string*, REQUIRED
 
   The name of the operating system which the image is built to run on.
   Configurations SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOOS`][go-environment].
+
+- **os.version** *string*, OPTIONAL
+
+  This OPTIONAL property specifies the version of the operating system targeted by the referenced blob.
+  Implementations MAY refuse to use manifests where `os.version` is not known to work with the host OS version.
+  Valid values are implementation-defined. e.g. `10.0.14393.1066` on `windows`.
+
+- **os.features** *array of strings*, OPTIONAL
+
+  This OPTIONAL property specifies an array of strings, each specifying a mandatory OS feature.
+  When `os` is `windows`, image indexes SHOULD use, and implementations SHOULD understand the following values:
+
+  - `win32k`: image requires `win32k.sys` on the host (Note: `win32k.sys` is missing on Nano Server)
+
+- **variant** *string*, OPTIONAL
+
+  The variant of the specified CPU architecture.
+  Configurations SHOULD use, and implementations SHOULD understand, `variant` values listed in the [Platform Variants](image-index.md#platform-variants) table.
 
 - **config** *object*, OPTIONAL
 
--- a/conversion.md
+++ b/conversion.md
@@ -47,18 +47,22 @@ These fields all affect the `annotations` of the runtime configuration, and are 
 | `os`                | `annotations`   | 1,2   |
 | `architecture`      | `annotations`   | 1,3   |
 | `variant`           | `annotations`   | 1,4   |
-| `author`            | `annotations`   | 1,5   |
-| `created`           | `annotations`   | 1,6   |
+| `os.version`         | `annotations`   | 1,5   |
+| `os.features`        | `annotations`   | 1,6   |
+| `author`            | `annotations`   | 1,7   |
+| `created`           | `annotations`   | 1,8   |
 | `Config.Labels`     | `annotations`   |       |
-| `Config.StopSignal` | `annotations`   | 1,7   |
+| `Config.StopSignal` | `annotations`   | 1,9   |
 
 1. If a user has explicitly specified this annotation with `Config.Labels`, then the value specified in this field takes lower [precedence](#annotations) and the converter MUST instead use the value from `Config.Labels`.
 2. The value of this field MUST be set as the value of `org.opencontainers.image.os` in `annotations`.
 3. The value of this field MUST be set as the value of `org.opencontainers.image.architecture` in `annotations`.
 4. The value of this field MUST be set as the value of `org.opencontainers.image.variant` in `annotations`.
-5. The value of this field MUST be set as the value of `org.opencontainers.image.author` in `annotations`.
-6. The value of this field MUST be set as the value of `org.opencontainers.image.created` in `annotations`.
-7. The value of this field MUST be set as the value of `org.opencontainers.image.stopSignal` in `annotations`.
+5. The value of this field MUST be set as the value of `org.opencontainers.image.os.version` in `annotations`.
+6. The value of this field MUST be set as the value of `org.opencontainers.image.os.features` in `annotations`.
+7. The value of this field MUST be set as the value of `org.opencontainers.image.author` in `annotations`.
+8. The value of this field MUST be set as the value of `org.opencontainers.image.created` in `annotations`.
+9. The value of this field MUST be set as the value of `org.opencontainers.image.stopSignal` in `annotations`.
 
 ## Parsed Fields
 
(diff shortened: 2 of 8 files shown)


tianon:
(not a maintainer) I'm +1 on this, but I think it's probably worth at least mentioning pros/cons of including all the `platform` fields vs deprecating existing fields and adding a full proper `platform` object instead. :innocent:

cpuguy83:
Major con I can see with that is builders will need to set both the old fields and the new `platform` field to maintain backwards compatibility.

Also requires allocating another object, but 🤷

caniszczyk:
@cyphar any comments here?

ping @opencontainers/image-spec-maintainers 

It seems this would warrant a bump in a minor or micro release, nothing breaks here (optional field)

config.md
```diff
@@ -110,6 +110,26 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
...
+  Valid values are implementation-defined. e.g. `10.0.14393.1066` on `windows`.
+
+- **os.features** *array of strings*, OPTIONAL
+
+  This OPTIONAL property specifies an array of strings, each specifying a mandatory OS feature.
+  When `os` is `windows`, image indexes SHOULD use, and implementations SHOULD understand the following values:
+
+  - `win32k`: image requires `win32k.sys` on the host (Note: `win32k.sys` is missing on Nano Server)
+
+  When `os` is not `windows`, values are implementation-defined and SHOULD be submitted to this specification for standardization
```
cyphar, line 126: My only comment is that we don't have any other JSON fields that have `.` in their name, and this probably(?) will play silly buggers with `jq` commands, especially since there is a separate `os` field in the same struct. Can we call them `osVersion` and `osFeatures` or something similar (kinda sucks we can't make the `os` a struct without breaking compatibility).

cyphar: Also "SHOULD be submitted to this specification" is kind of a weird thing to put in a standard, since SHOULD is entirely about rules for implementations. I get this was copied from the original `variants` table text, but it would be a good idea to remove it.

conversion.md
```diff
@@ -46,17 +46,21 @@ These fields all affect the `annotations` of the runtime configuration, and are
 | ------------------- | --------------- | ----- |
 | `os`                | `annotations`   | 1,2   |
 | `architecture`      | `annotations`   | 1,3   |
-| `author`            | `annotations`   | 1,4   |
-| `created`           | `annotations`   | 1,5   |
+| `variant`           | `annotations`   | 1,4   |
+| `os.version`        | `annotations`   | 1,5   |
+| `os.features`       | `annotations`   | 1,6   |
```
cyphar, line 51: There isn't a corresponding footnote for either annotation mapping.

conversion.md
```diff
@@ -46,17 +46,21 @@ These fields all affect the `annotations` of the runtime configuration, and are
...
+| `Config.StopSignal` | `annotations`   | 1,9   |
 
 1. If a user has explicitly specified this annotation with `Config.Labels`, then the value specified in this field takes lower [precedence](#annotations) and the converter MUST instead use the value from `Config.Labels`.
 2. The value of this field MUST be set as the value of `org.opencontainers.image.os` in `annotations`.
 3. The value of this field MUST be set as the value of `org.opencontainers.image.architecture` in `annotations`.
-4. The value of this field MUST be set as the value of `org.opencontainers.image.author` in `annotations`.
-5. The value of this field MUST be set as the value of `org.opencontainers.image.created` in `annotations`.
-6. The value of this field MUST be set as the value of `org.opencontainers.image.stopSignal` in `annotations`.
+4. The value of this field MUST be set as the value of `org.opencontainers.image.variant` in `annotations`.
+5. The value of this field MUST be set as the value of `org.opencontainers.image.author` in `annotations`.
```
cyphar, line 61: I just noticed that this is actually meant to be `authors`, whoops. Oh well...

cyphar (changes requested):

cyphar:
I think a `platform` object would be a nicer solution, but :shrug: from my side too.

cpuguy83:
Updated.

config.md
cpuguy83: This was all copied from the `platform` definition in index.
Note that the fields in `platform` are `os.*` which is why I did this here.

I went ahead and updated this with your suggestion to camel case it.

conversion.md
cpuguy83: 🤦 Fixed.

config.md
cyphar: Ah I missed that. I just checked and you can get `jq` to deal with `.`s in field names so while I think `os.variant` might've been a mistake initially we should probably stick with it for consistency's sake -- sorry for the churn.

cpuguy83: No worries! Updated.

cpuguy83:
Ping

caniszczyk:
ping @opencontainers/image-spec-maintainers

vrothberg (commented):
Thanks for working on this, @cpuguy83!

I would love to help things push forward if needed.  @opencontainers/image-spec-maintainers, are there any open issues that prevent the changes from getting in?

I have an immediate need to know the variant of an image.  Once an image was pulled, we lose that information which is currently causing me headaches.

jonjohnsonjr:
You could poke some folks in the [weekly meeting](https://github.com/opencontainers/.github#meetings) (today) if you want to get some activity.

vrothberg:
> You could poke some folks in the [weekly meeting](https://github.com/opencontainers/.github#meetings) (today) if you want to get some activity.

Thanks.  I wanted to join for a long time but the time is tough (11PM in my time zone).  Maybe, I'll make it once to push this issue forward :)

conversion.md
```diff
@@ -46,17 +46,23 @@ These fields all affect the `annotations` of the runtime configuration, and are
...
 3. The value of this field MUST be set as the value of `org.opencontainers.image.architecture` in `annotations`.
-4. The value of this field MUST be set as the value of `org.opencontainers.image.author` in `annotations`.
-5. The value of this field MUST be set as the value of `org.opencontainers.image.created` in `annotations`.
-6. The value of this field MUST be set as the value of `org.opencontainers.image.stopSignal` in `annotations`.
+4. The value of this field MUST be set as the value of `org.opencontainers.image.variant` in `annotations`.
+5. The value of this field MUST be set as the value of `org.opencontainers.image.os.version` in `annotations`.
+6. The value of this field MUST be set as the value of `org.opencontainers.image.os.features` in `annotations`.
+7. The value of this field MUST be set as the value of `org.opencontainers.image.author` in `annotations`.
+8. The value of this field MUST be set as the value of `org.opencontainers.image.created` in `annotations`.
+9. The value of this field MUST be set as the value of `org.opencontainers.image.stopSignal` in `annotations`.
```
vbatts, line 65: For the record: unrelated to this PR, after a discussion in today's image-spec grooming, I think this entire stanza is bogus and not actually done in the wild. 

I move that we could remove this list entirely.

@cyphar you have the blame history on you. Do you recall a use-case of converting an image config into a runtime config that _actually_ does this?

image-index.md
```diff
@@ -97,6 +89,16 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
     See [Pre-Defined Annotation Keys](annotations.md#pre-defined-annotation-keys).
 
+## Platform Variants
+When the variant of the CPU is not listed in the table, values are implementation-defined and SHOULD be submitted to this specification for standardization.
```
jonjohnsonjr, line 93: Nit: newline after header

vbatts (approved):
This lines up with what's already being done (apart from my one comment), and it matches with the os and variant fields in the image-index descriptors.

schema/validator.go
```diff
@@ -219,7 +246,7 @@ func checkPlatform(OS string, Architecture string) {
 					return
 				}
 			}
-			fmt.Printf("warning: combination of %q and %q is invalid.\n", OS, Architecture)
+			fmt.Printf("warning: combination of OS %q and architecture %q is invalid.\n", OS, Architecture)
```
jonjohnsonjr, line 249: nit: maybe uncapitalize `OS` to match the field name?

conversion.md
tianon: (This discussion has/should move to https://github.com/opencontainers/image-spec/issues/864 now -- @cpuguy83 is just keeping these new bits consistent with the old bits :innocent:)

schema/validator.go
```diff
@@ -200,6 +202,31 @@ func validateConfig(r io.Reader) error {
 	return nil
 }
 
+func checkArchitecture(Architecture string, Variant string) {
+	validCombins := map[string][]string{
+		"arm":      {"v6", "v7", "v8"},
+		"arm64":    {"v8"},
```
tianon, line 208: Shouldn't these both include the empty string?

image-index.md
cpuguy83: Fixed

schema/validator.go
cpuguy83: Fixed

schema/validator.go
cpuguy83: Fixed.

cpuguy83:
```
# cd .; git clone https://github.com/opencontainerd/image-spec /home/runner/work/image-spec/image-spec/go/src/github.com/opencontainerd/image-spec
Cloning into '/home/runner/work/image-spec/image-spec/go/src/github.com/opencontainerd/image-spec'...
fatal: could not read Username for 'https://github.com': terminal prompts disabled
package github.com/opencontainerd/image-spec/specs-go/v1: exit status 128
```

vbatts:
@cpuguy83 hrm. Is that a go.mod issue?

tianon:
Looks like a network fluke in CI that probably just needs a restart to me 😅

schema/validator.go
```diff
@@ -22,8 +22,8 @@ import (
 	"io/ioutil"
 	"regexp"
 
+	v1 "github.com/opencontainerd/image-spec/specs-go/v1"
```
jonjohnsonjr, line 25:
```suggestion
	v1 "github.com/opencontainers/image-spec/specs-go/v1"
```

😉

jonjohnsonjr: I think if you actually accept this suggestion through the UI it might break DCO, so caveat emptor.

cpuguy83: Love that muscle memory 😆 
It was also in my $GOPATH.

cpuguy83:
All green now.

jonjohnsonjr (approved):

stevvooe (approved):
LGTM

cpuguy83:
@cyphar Does this LGTY? I made the adjustments you'd requested before. Merging is currently blocked due to prior review.

Thanks!

vbatts (approved):
😸

Jjmo12345 (commented):
Good

Jjmo12345 (commented):
👍

PR #810, opened 2020-10-01
Add go.mod and pin dependencies

ashokponkumar:
Signed-off-by: Ashok Pon Kumar <ashokponkumar@gmail.com>

### DIFF
--- a/specs-go/go.mod
+++ b/specs-go/go.mod
@@ -1,0 +1,5 @@
+module github.com/opencontainers/image-spec/specs-go
+
+go 1.15
+
+require github.com/opencontainers/go-digest v1.0.0
--- a/specs-go/go.sum
+++ b/specs-go/go.sum
@@ -1,0 +1,2 @@
+github.com/opencontainers/go-digest v1.0.0 h1:apOUWs51W5PlhuyGyz9FCeeBIOUDA/6nW8Oi/yOhh5U=
+github.com/opencontainers/go-digest v1.0.0/go.mod h1:0JzlMkj0TRzQZfJkVvzbP0HBR3IKzErnv2BNG4W4MAM=


vbatts:
not a bad idea, but it fails travis CI

vbatts:
@ashokponkumar could you rebase?

go.sum
```diff
@@ -0,0 +1,241 @@
+cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
```
vbatts, line 1: this go.sum file does not even need to be checked in

go.mod
```diff
@@ -0,0 +1,12 @@
+module github.com/opencontainers/image-spec
```
vbatts, line 1: after some discussion on the call, I'm not sure that we want to have a single `go.mod` at the top level of the project, but instead to have a `go.mod` in any immediate subdirectory that contains golang source.

The reasoning behind this is: if say `./schema/` brings in a dependency that is wholly not needed if someone only needs to import `./specs-go/`, importers need only fetch `specs-go/go.mod` related deps (which is likely always none).

We just did similar on the distribution-spec

vbatts (changes requested):

go.sum
ashokponkumar: Its generally the practise to checkin go.sum too. But if we don't want for this repo, we can remove it. (Ex: https://github.com/kubernetes/kubernetes/blob/master/go.sum)

go.mod
ashokponkumar: Sure. I have added for specs-go. Adding to schema to creates some ambiguous import issue.

```ashoks-mbp:schema ashokponkumar$ go mod tidy
go: finding module for package github.com/opencontainers/image-spec/specs-go/v1
go: found github.com/opencontainers/image-spec/specs-go/v1 in github.com/opencontainers/image-spec v1.0.1
github.com/opencontainers/image-spec/schema: ambiguous import: found package github.com/opencontainers/image-spec/schema in multiple modules:
	github.com/opencontainers/image-spec/schema (/Users/ashokponkumar/go/src/github.com/ashokponkumar/image-spec/schema)
	github.com/opencontainers/image-spec v1.0.1 (/Users/ashokponkumar/go/pkg/mod/github.com/opencontainers/image-spec@v1.0.1/schema)```

vbatts: oh dang. Just did some research on this, and it's a go modules mess ... and sounds like what @tianon mentioned during the image-spec grooming call with having to do special git tags also.

I found that the popular import `github.com/ugorji/go/codec` went through this when they brought in go modules.
https://github.com/ugorji/go/issues/299
They ended up having to add circular referencing go.mod files, based on tag versions, and also to make multiple tags for each of the sub-directories as their own modules https://github.com/ugorji/go/tags

I'm now wondering if this is all worth it, or is it just simpler to keep the go.mod file at the topic level directory like you began with :-\

@opencontainers/image-spec-maintainers thoughts?

(p.s. i'm curious why the we didn't hit this in the distribution-spec?)

ashokponkumar: The reason the issue is not there in distribution-spec is that, it has only one go module in the repo.

vbatts:
thank you @ashokponkumar for your PR! and congrats on your first commit!

ashokponkumar:
Thanks @vbatts for taking it through the process and getting it merged.

PR #817, opened 2020-12-23
media-types.md: clarify differences from Docker media types

AkihiroSuda:
OCI media types are slightly different from Docker ones,

e.g., Docker manifests must have `.mediaType` field while OCI may not have.
Also, OCI descriptors may have `.annotations` while Docker may not.

### DIFF
--- a/media-types.md
+++ b/media-types.md
@@ -38,25 +38,40 @@ This section shows where the OCI Image Specification is compatible with formats 
 
 **Similar/related schema**
 
-- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list) - mediaType is different
+- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/distribution/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list)
+  - `.mediaType`: only present in Docker, and reserved in OCI
+  - `.annotations`: only present in OCI
+  - `.[]manifests.annotations`: only present in OCI
+  - `.[]manifests.urls`: only present in OCI
 
 ### application/vnd.oci.image.manifest.v1+json
 
 **Similar/related schema**
 
-- [application/vnd.docker.distribution.manifest.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#image-manifest-field-descriptions)
+- [application/vnd.docker.distribution.manifest.v2+json](https://github.com/distribution/distribution/blob/master/docs/spec/manifest-v2-2.md#image-manifest-field-descriptions)
+  - `.mediaType`: only present in Docker, and reserved in OCI
+  - `.annotations`: only present in OCI
+  - `.config.annotations`: only present in OCI
+  - `.config.urls`: only present in OCI
+  - `.[]layers.annotations`: only present in OCI
 
 ### application/vnd.oci.image.layer.v1.tar+gzip
 
 **Interchangeable and fully compatible mime-types**
 
-- [application/vnd.docker.image.rootfs.diff.tar.gzip](https://github.com/docker/docker/blob/master/image/spec/v1.md#creating-an-image-filesystem-changeset)
+- [application/vnd.docker.image.rootfs.diff.tar.gzip](https://github.com/moby/moby/blob/v20.10.8/image/spec/v1.2.md#creating-an-image-filesystem-changeset)
 
 ### application/vnd.oci.image.config.v1+json
 
 **Similar/related schema**
 
-- [application/vnd.docker.container.image.v1+json](https://github.com/docker/docker/blob/master/image/spec/v1.md#image-json-description)
+- [application/vnd.docker.container.image.v1+json](https://github.com/moby/moby/blob/v20.10.8/image/spec/v1.2.md#image-json-description) (Docker Image Spec v1.2)
+  - `.config.Memory`: only present in Docker, and reserved in OCI
+  - `.config.MemorySwap`: only present in Docker, and reserved in OCI
+  - `.config.CpuShares`: only present in Docker, and reserved in OCI
+  - `.config.Healthcheck`: only present in Docker, and reserved in OCI
+
+`.config.StopSignal` and `.config.Labels` are accidentally undocumented in Docker Image Spec v1.2, but these fields are not OCI-specific concepts.
 
 ## Relations
 
--- a/config.md
+++ b/config.md
@@ -183,6 +183,22 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
 
     The field contains the system call signal that will be sent to the container to exit. The signal can be a signal name in the format `SIGNAME`, for instance `SIGKILL` or `SIGRTMIN+3`.
 
+  - **Memory** *integer*, OPTIONAL
+
+    This property is *reserved* for use, to [maintain compatibility](media-types.md#compatibility-matrix).
+
+  - **MemorySwap** *integer*, OPTIONAL
+
+    This property is *reserved* for use, to [maintain compatibility](media-types.md#compatibility-matrix).
+
+  - **CpuShares** *integer*, OPTIONAL
+
+    This property is *reserved* for use, to [maintain compatibility](media-types.md#compatibility-matrix).
+
+  - **Healthcheck** *object*, OPTIONAL
+
+    This property is *reserved* for use, to [maintain compatibility](media-types.md#compatibility-matrix).
+
 - **rootfs** *object*, REQUIRED
 
    The rootfs key references the layer content addresses used by the image.


AkihiroSuda:
@dmcgowan @tonistiigi @ktock PTAL

media-types.md
```diff
@@ -38,25 +38,38 @@ This section shows where the OCI Image Specification is compatible with formats
 
 **Similar/related schema**
 
-- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list) - mediaType is different
+- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list)
+  - `.mediaType`: only present on Docker
+  - `.annotations`: only present on OCI
+  - `.[]manifests.annotations`: only present on OCI
+  - `.[]manifests.platform`: ARM v6 is represented as `{"architecture":"arm","variant":"v6"}` on OCI, while represented as `{"variant:armv6l"}` on Docker
```
ktock, line 45: `.[]manifests.urls` is not defined in `application/vnd.docker.distribution.manifest.list.v2+json`.

media-types.md
```diff
@@ -38,25 +38,38 @@ This section shows where the OCI Image Specification is compatible with formats
...
 
 ### application/vnd.oci.image.manifest.v1+json
 
 **Similar/related schema**
 
 - [application/vnd.docker.distribution.manifest.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#image-manifest-field-descriptions)
+  - `.mediaType`: only present on Docker
+  - `.annotations`: only present on OCI
+  - `.config.annotations`: only present on OCI
```
ktock, line 54: `.config.urls` is not defined in `application/vnd.docker.distribution.manifest.v2+json`.

media-types.md
```diff
@@ -38,25 +38,38 @@ This section shows where the OCI Image Specification is compatible with formats
...
 
 **Similar/related schema**
 
-- [application/vnd.docker.container.image.v1+json](https://github.com/docker/docker/blob/master/image/spec/v1.md#image-json-description)
+- [application/vnd.docker.container.image.v1+json](https://github.com/moby/moby/blob/v20.10.1/image/spec/v1.2.md#image-json-description) (Docker Image Spec v1.2)
+  - `.config.Memory`: only present on Docker
+  - `.config.MemorySwap`: only present on Docker
+  - `.config.CpuShares`: only present on Docker
+  - `.config.Healthcheck`: only present on Docker
+  - `.config.Labels`: only present on OCI
```
ktock, line 72: `.config.StopSignal` is not defined in `application/vnd.docker.container.image.v1+json`.

media-types.md
AkihiroSuda: updated, thanks

media-types.md
AkihiroSuda: updated, thanks

media-types.md
AkihiroSuda: updated, thanks

media-types.md
```diff
@@ -38,25 +38,41 @@ This section shows where the OCI Image Specification is compatible with formats
 
 **Similar/related schema**
 
-- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list) - mediaType is different
+- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list)
+  - `.mediaType`: only present on Docker
```
cyphar, line 42: `s/on/in/g` -- this applies to all of the list entries in your change.

media-types.md
```diff
@@ -38,25 +38,41 @@ This section shows where the OCI Image Specification is compatible with formats
...
+  - `.annotations`: only present on OCI
+  - `.[]manifests.annotations`: only present on OCI
+  - `.[]manifests.platform`: ARM v6 is represented as `{"architecture":"arm","variant":"v6"}` on OCI, while represented as `{"variant:armv6l"}` on Docker
```
tonistiigi, line 45: This needs to be a mistake in https://github.com/distribution/distribution/pull/1592/files#diff-6b5275ccafa2776e05d590a9e729b0ec6762300317c928d34adc856d9f11714fR108, right? I'm quite sure no implementation has ever supported anything like `armv6l`. Quotes are wrong as well.

media-types.md
tonistiigi: Neither labels nor stopsignal are new OCI concepts so it is wrong to give an impression like they are OCI only. They were copied over from Docker but missing descriptions were not contributed back. The out-of-sync issue seems to be caused by the linked spec file reusing the `RunConfig` type from Docker API that is defined separately from image. We can just make sure the spec file lists all relevant fields so there is no confusion. `OnBuild`, `ArgsEscaped`, `Shell`, `StopTimeout` are also not defined in OCI. https://github.com/moby/buildkit/blob/master/frontend/dockerfile/dockerfile2llb/image.go#L32

media-types.md
AkihiroSuda: thanks, fixed

media-types.md
AkihiroSuda: Improved text. PTAL.

tonistiigi: Just fix that erroneous example in upstream. No need to copy it around more. Afaics platform was moved to OCI exactly how it was implemented in Docker (with all the os.* weirdnesses). Theoretically, you can use any value in there. OCI spec even makes it clear that it is implementation-defined. The values that OCI recommends for arm are the same values that Docker uses (and used before that docs example was changed to an invalid value).

media-types.md
tonistiigi: ^ @AkihiroSuda

media-types.md
AkihiroSuda: Removed the note about `.[]manifests.platform` and opened PR https://github.com/distribution/distribution/pull/3371 for fixing Docker spec.

AkihiroSuda:
rebased

media-types.md
```diff
@@ -38,25 +38,40 @@ This section shows where the OCI Image Specification is compatible with formats
 
 **Similar/related schema**
 
-- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list) - mediaType is different
+- [application/vnd.docker.distribution.manifest.list.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#manifest-list)
```
vbatts, line 41: Do you want to point to a tag here too (instead of master)?

media-types.md
```diff
@@ -38,25 +38,40 @@ This section shows where the OCI Image Specification is compatible with formats
...
+  - `.mediaType`: only present in Docker
+  - `.annotations`: only present in OCI
+  - `.[]manifests.annotations`: only present in OCI
+  - `.[]manifests.urls`: only present in OCI
 
 ### application/vnd.oci.image.manifest.v1+json
 
 **Similar/related schema**
 
 - [application/vnd.docker.distribution.manifest.v2+json](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#image-manifest-field-descriptions)
```
vbatts, line 51: Ditto

media-types.md
vbatts: "and is _reserved_ in OCI"?

vbatts (changes requested):
A couple of nits, but is looking good

media-types.md
AkihiroSuda: > Do you want to point to a tag here too (instead of master)?

Not in this PR. There is no tag for the latest revision of the file.
https://github.com/distribution/distribution/commit/c432849e44f960c95937e43cff2d010187f7cc0b

media-types.md
AkihiroSuda: Updated to remove `StopSignal` and `Labels`

> The fix is to update docs where needed.

Do we need to update all `v1.md`, `v1.1.md`, and `v1.2.md` in https://github.com/moby/moby/tree/v20.10.8/image/spec ? Or maybe we should have `v1.2.1.md`?

media-types.md
AkihiroSuda: Added "and is reserved in OCI"?

vbatts (approved):

jonjohnsonjr (approved):

vrothberg (approved):
LGTM

media-types.md
tonistiigi: I don't think we need to increase version. If you want to can look at the publish date for the old ones as I do think at least most of them are very old but just updating `v1.2` is good enough for me.

media-types.md
tonistiigi: Looks like the comment about "OnBuild, ArgsEscaped, Shell, StopTimeout" went unnoticed.

PR #829, opened 2021-03-24
Add ArgsEscaped field to image config

kevpar:
This PR is an attempt to fix containerd/containerd#5067, which currently prevents certain Docker-built Windows images from working in containerd. See the containerd issue for full details. Brief overview below.

Docker has been relying on the unofficial `ArgsEscaped` field in the image config for several years now. It is added when a Windows image is built using a "shell form" `ENTRYPOINT`/`CMD` in the Dockerfile, and indicates that the first element of the `Entrypoint`/`Cmd` field contains a full command line, rather than just a path to a binary.

This change was made in Docker to improve support for Windows containers, given differences in how Windows and Linux handle command line args and escaping. Given that there are existing images in the world that rely on this field being set, the easiest way to address the problem appears to be simply adding `ArgsEscaped` officially to the image spec.

For context, this is how `ArgsEscaped` is represented in Docker's image config type: https://github.com/moby/moby/blob/46cdcd206c56172b95ba5c77b827a722dab426c5/api/types/container/config.go#L57

### DIFF
--- a/config.md
+++ b/config.md
@@ -165,6 +165,10 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
 
     The field contains the system call signal that will be sent to the container to exit. The signal can be a signal name in the format `SIGNAME`, for instance `SIGKILL` or `SIGRTMIN+3`.
 
+  - **ArgsEscaped** *boolean*, OPTIONAL
+
+    This field is used for Windows images to indicate that the `Entrypoint` and `Cmd` fields will have only a single element, which has already been escaped. In this case their values should be used as-is without further escaping.
+
 - **rootfs** *object*, REQUIRED
 
    The rootfs key references the layer content addresses used by the image.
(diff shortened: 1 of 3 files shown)


kevpar:
After some internal discussion, going to look into another option for fixing this. Might re-open this in the future if that doesn't pan out.

jterry75:
@kevpar - Can we revisit this? Were there any objections from OCI to adding ArgsEscaped to the config spec here? It doesn't seem harmful to add given that Docker (/Buildkit) support `ArgsEscaped` and are the largest producer of images. It's easy to document that `ArgsEscaped==true` means that the `EntryPoint` is a pre-escaped single entry array and should not be additionally escaped. Additionally, we already have a provision in the Runtime spec for `CommandLine` rather than `Args` specifically to address this issue on Windows. It seems reasonable to have this IMO and then I can make the change to containerd to support it natively. Thoughts?

estesp:
Seems with the suggestion of using this in containerd via containerd/containerd#6479 we should re-open this for consideration?

PR #851, opened 2021-05-08
Create EMERITUS.md to recognize old maintainers

caniszczyk:
Previous maintainers should be thanked for their work.

Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

### DIFF
--- a/EMERITUS.md
+++ b/EMERITUS.md
@@ -1,0 +1,7 @@
+We would like to acknowledge previous OCI image spec maintainers and their huge contributions to our collective success:
+
+* Brandon Philips (@philips)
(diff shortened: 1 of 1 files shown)


vbatts:
Oh nice. I wondered if we would do this in the opencontainers/org repo or in each respective repo

cyphar:
`gitvalidation` doesn't like the trailing whitespace in `EMERITUS.md` (first line).

caniszczyk:
@vbatts for now my thought is to do it in each respective repo

cyphar:
LGTM. I will submit a similar patch to runc as well.

PR #856, opened 2021-06-18
Removing Link Introduction

vsoch:
There doesn't seem to be an introduction in this file (and the link goes to itself?) so I think we probably want to remove it (or link to an introduction elsewhere?)

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -7,7 +7,6 @@ The goal of this specification is to enable the creation of interoperable tools 
 
 ### Table of Contents
 
-- [Introduction](spec.md)
 - [Notational Conventions](#notational-conventions)
 - [Overview](#overview)
     - [Understanding the Specification](#understanding-the-specification)


spec.md
```diff
@@ -42,7 +41,7 @@ Typically, these manifests may provide different implementations of the image, p
 
 ![](img/build-diagram.png)
 
-Once built the OCI Image can then be discovered by name, downloaded, verified by hash, trusted through a signature, and unpacked into an [OCI Runtime Bundle](https://github.com/opencontainers/runtime-spec/blob/master/bundle.md).
+Once built the OCI image can then be discovered by name, downloaded, verified by hash, trusted through a signature, and unpacked into an [OCI Runtime Bundle](https://github.com/opencontainers/runtime-spec/blob/master/bundle.md).
```
jonjohnsonjr, line 44: I see `OCI Image` used throughout the spec, so I don't know if I agree with this line.

vsoch: There are also a lot of lowercase - I guess I don't see it as a proper noun so I made it lowercase. But if you are requesting to make it capital again, happy to do that.

vsoch: All set.

vrothberg (approved):
LGTM

vbatts:
It looks like the DCO/signoff may have an issue with a difference in the commit email and the signoff email? not sure

vsoch:
I did change both emails to be my GitHub no reply email when I left Stanford.

vsoch:
Where do you see the issue? It looks green to me!

jonboulle (approved):

vbatts:
https://pullapprove.com/opencontainers/image-spec/pull-request/856/requirements/
looks like a prefix of `814322+` in the commit email?

vsoch:
@vbatts GitHub holds an internal representation of that address with and without the number. It's kind of complicated but without signing with the number it doesn't link to me, so the gpg key actually has all of those emails (and my original stanford) so I'm not sure why one is explicitly chosen. I wrote more about it here: https://vsoch.github.io/2021/signing-with-myself/.

vsoch:
I did create the original commit in the GitHub interface and then pulled and rebased, so that might be why the two are different? I can close here and re-do again if you want... this seems like an extraordinary amount of work to just remove one line! I'm not sure why it's always so hard to contribute to these repos :/

vbatts:
Fair. And we have 2 DCO checks, and only one fail? 🤔

vbatts (approved):
🌃

vbatts:
@caniszczyk do we need the pullapprove DCO check also, since it's being a little strict?

Issue #895, opened 2022-02-08
Support for other languages


cardil:
Is the support for other languages planned? Or, if someone provides it, it may change to be merged?

I'm interested in particular on Rust implementation. As a minimum for cross-language support, some external standard could be provided - like for ex.: [OpenAPI](https://swagger.io/specification/). I think such an artifact could easily be computed (see: [openapi-go](https://github.com/swaggest/openapi-go)) and attached to the releases page. That would allow building upon this spec, in other languages.

cardil:
Scratch this issue, in favor of https://github.com/containers/oci-spec-rs

vbatts:
good call out. I've opened #896 to reference it

PR #896, opened 2022-02-16
implementations||adopters: point to C and Rust libraries

vbatts:
having seen #895, it's worth ensuring these other languages are listed
implementations

### DIFF
--- a/implementations.md
+++ b/implementations.md
@@ -15,6 +15,8 @@ Projects or Companies currently adopting the OCI Image Specification
 * [Containers](https://github.com/containers/)
   - [containers/build](https://github.com/containers/build)
   - [containers/image](https://github.com/containers/image)
+  - [containers/oci-spec-rs](https://github.com/containers/oci-spec-rs)
+  - [containers/libocispec](https://github.com/containers/libocispec)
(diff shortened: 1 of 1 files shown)


cardil:
I've opened the #895, having Rust in mind. I end up using [krustlet/oci-distribution](https://github.com/krustlet/oci-distribution) instead, of [containers/oci-spec-rs](https://github.com/containers/oci-spec-rs). The `containers/oci-spec-rs` is really basic - you could only parse manifest file with it, which isn't much.

The `krustlet/oci-distribution` has a far greater scope. It's still under heavy development, though. Here's a code I just wrote that fetches a remote image and saves it in the user's home:

https://github.com/cardil/ocilot/blob/d5ac007e5804bd9c2f5b353ccb29300d337446a0/oci/src/registry.rs#L21-L38

There are is also: 
 * https://github.com/camallo/dkregistry-rs
 * https://github.com/containers/containers-image-proxy-rs

stevvooe (approved):

jonjohnsonjr (approved):

vbatts:
this merged fast during the call, but I've opened https://github.com/opencontainers/image-spec/pull/897 to track the krustlet library you mentioned, @cardil

PR #908, opened 2022-04-21
CODEOWNERS: point to team rather than individuals

vbatts:

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,1 @@
-* @jonjohnsonjr @jonboulle @stevvooe @vbatts @cyphar
+* @opencontainers/image-spec-maintainers


jonboulle (dismissed):

jdolitsky (approved):
LGTM. This github team is now up-to-date

tianon:
I'm not an image-spec maintainer, but figured maybe my thoughts on this are useful (being I'm a runtime-spec maintainer and this PR might end up there too).  While I agree with this in theory (DRY maintainer listing), the thing I like about `CODEOWNERS` listing individuals is that we get this:

> requested review from [brendandburns](https://github.com/brendandburns), [jonjohnsonjr](https://github.com/jonjohnsonjr), [jstarks](https://github.com/jstarks), [jonboulle](https://github.com/jonboulle), [stevvooe](https://github.com/stevvooe) and [cyphar](https://github.com/cyphar) as [code owners](https://github.com/opencontainers/image-spec/blob/02efb9a75ee11e05937b535cc5f228f9343ab2f5/CODEOWNERS#L1)

It's even easier to ignore an async ping when you can hide behind a group alias than when it's pinging one single person, so IMO this is acceptable as long as folks continue to stay active. :sweat_smile:

If this lists the team alias, do we still get individuals listed/pinged such that there's an individual call-out vs a group call-out?

vbatts:
@tianon that is fair.

vbatts:
I like DRYing things up, but not sure the best solution here. Closing this for cleanup. We can revisit if someone feels strongly.

PR #921, opened 2022-05-17
Delete annotations.md

Gitnitdun:


sudo-bmitch (changes requested):
This file should not be deleted.

PR #826, opened 2021-03-12
Define the data field

jonjohnsonjr:
This should contain an embedded representation of the referenced
content, which is useful for avoiding extra hops to access small pieces
of content.

Fixes https://github.com/opencontainers/image-spec/issues/825
Closes #675

PR #934, opened 2022-08-09
Working Group Proposal for Reference Types

lachie83:
This is a culmination of many people's work over the last few months to introduce reference types to OCI as part of the [OCI Reference Types WG](https://github.com/opencontainers/tob/blob/main/proposals/wg-reference-types.md).

This PR contains the changes from our chosen proposal, [Proposal E](https://github.com/opencontainers/wg-reference-types/blob/main/docs/proposals/PROPOSAL_E.md), relevant to the image-spec. We're looking forward to hearing feedback from image-spec maintainers on these changes!

Associated distribution-spec PR - https://github.com/opencontainers/distribution-spec/pull/335

### DIFF
--- a/artifact.md
+++ b/artifact.md
@@ -1,0 +1,74 @@
+# OCI Artifact Manifest Specification
+
+The goal of the Artifact Manifest Specification is to define content addressable artifacts in order to store them along side container images in a registry.
+Like [OCI Images](manifest.md), OCI Artifacts may be referenced by the hash of their manifest.
+Unlike OCI Images, OCI Artifacts are not meant to be used by any container runtime.
+
+Examples of artifacts that may be stored along with container images are Software Bill of Materials (SBOM), Digital Signatures, Provenance data, Supply Chain Attestations, scan results, and Helm charts.
+
+This section defines the `application/vnd.oci.artifact.manifest.v1+json` [media type](media-types.md).
+For the media type(s) that this is compatible with see the [matrix](media-types.md#compatibility-matrix).
+
+# Artifact Manifest
+
+## *Artifact Manifest* Property Descriptions
+
+- **`mediaType`** *string*
+
+  This property MUST be used and contain the media type `application/vnd.oci.artifact.manifest.v1+json`.
+
+- **`artifactType`** *string*
+
+  This property SHOULD be used and contain the mediaType of the referenced artifact.
+  If defined, the value MUST comply with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2], and MAY be registered with [IANA][iana].
+
+- **`blobs`** *array of objects*
+
+  This OPTIONAL property is an array of objects and each item in the array MUST be a [descriptor](descriptor.md).
+  Each descriptor represents an artifact of any IANA mediaType.
+  The list MAY be ordered for certain artifact types like scan results.
+
+- **`refers`** *[descriptor](descriptor.md)*
+
+  This OPTIONAL property specifies a [descriptor](descriptor.md) of another manifest.
+  This value, used by the [`referrers` API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers), indicates a relationship to the specified manifest.
+
+- **`annotations`** *string-string map*
+
+  This OPTIONAL property contains additional metadata for the artifact manifest.
+  This OPTIONAL property MUST use the [annotation rules](annotations.md#rules).
+
+  See [Pre-Defined Annotation Keys](annotations.md#pre-defined-annotation-keys).
+
+  Annotations MAY be used to filter the response from the [`referrers` API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#listing-referrers).
+
+## Examples
+
+*Example showing an artifact manifest for an example SBOM referencing an image:*
+
+```jsonc,title=Manifest&mediatype=application/vnd.oci.artifact.manifest.v1%2Bjson
+{
+  "mediaType": "application/vnd.oci.artifact.manifest.v1+json",
+  "artifactType": "application/vnd.example.sbom.v1"
+  "blobs": [
+    {
+      "mediaType": "application/gzip",
+      "size": 123,
+      "digest": "sha256:87923725d74f4bfb94c9e86d64170f7521aad8221a5de834851470ca142da630"
+    }
+  ],
+  "refers": {
+    "mediaType": "application/vnd.oci.image.manifest.v1+json",
+    "size": 1234,
+    "digest": "sha256:cc06a2839488b8bd2a2b99dcdc03d5cfd818eed72ad08ef3cc197aac64c0d0a0"
+  },
+  "annotations": {
+    "org.opencontainers.artifact.created": "2022-01-01T14:42:55Z",
+    "org.example.sbom.format": "json"
+  }
+}
+```
+
+[iana]:         https://www.iana.org/assignments/media-types/media-types.xhtml
+[rfc6838]:      https://tools.ietf.org/html/rfc6838
+[rfc6838-s4.2]: https://tools.ietf.org/html/rfc6838#section-4.2
--- a/descriptor.md
+++ b/descriptor.md
@@ -1,11 +1,12 @@
 # OCI Content Descriptors
 
-* An OCI image consists of several different components, arranged in a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
-* References between components in the graph are expressed through _Content Descriptors_.
-* A Content Descriptor (or simply _Descriptor_) describes the disposition of the targeted content.
-* A Content Descriptor includes the type of the content, a content identifier (_digest_), and the byte-size of the raw content.
-* Descriptors SHOULD be embedded in other formats to securely reference external content.
-* Other formats SHOULD use descriptors to securely reference external content.
+- An OCI image consists of several different components, arranged in a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
+- References between components in the graph are expressed through _Content Descriptors_.
+- A Content Descriptor (or simply _Descriptor_) describes the disposition of the targeted content.
+- A Content Descriptor includes the type of the content, a content identifier (_digest_), and the byte-size of the raw content.
+  Optionally, it includes the type of artifact it is describing.
+- Descriptors SHOULD be embedded in other formats to securely reference external content.
+- Other formats SHOULD use descriptors to securely reference external content.
 
 This section defines the `application/vnd.oci.descriptor.v1+json` [media type](media-types.md).
 
@@ -51,7 +52,15 @@ The following fields contain the primary properties that constitute a Descriptor
   The decoded data MUST be identical to the referenced content and SHOULD be verified against the [`digest`](#digests) and `size` fields by content consumers.
   See [Embedded Content](#embedded-content) for when this is appropriate.
 
+- **`artifactType`** *string*
+
+  This OPTIONAL property contains the type of an artifact when the descriptor points to an artifact.
+  This is the value of `artifactType` when the descriptor references an [artifact manifest](artifact.md).
+  This is the value of the config descriptor `mediaType` when the descriptor references an [image manifest](manifest.md).
+
 Descriptors pointing to [`application/vnd.oci.image.manifest.v1+json`](manifest.md) SHOULD include the extended field `platform`, see [Image Index Property Descriptions](image-index.md#image-index-property-descriptions) for details.
+
+Descriptors pointing to [`application/vnd.oci.artifact.manifest.v1+json`](artifact.md) SHOULD include the extended field `artifactType`.
 
 ### Reserved
 
@@ -68,7 +77,7 @@ The _algorithm_ specifies the cryptographic hash function and encoding used for 
 
 A digest string MUST match the following [grammar](considerations.md#ebnf):
 
-```
+```ebnf
 digest                ::= algorithm ":" encoded
 algorithm             ::= algorithm-component (algorithm-separator algorithm-component)*
 algorithm-component   ::= [a-z0-9]+
@@ -104,12 +113,14 @@ Implementations MAY employ [canonicalization](considerations.md#canonicalization
 ### Digest calculations
 
 A _digest_ is calculated by the following pseudo-code, where `H` is the selected hash algorithm, identified by string `<alg>`:
-```
+
+```text
 let ID(C) = Descriptor.digest
 let C = <bytes>
 let D = '<alg>:' + Encode(H(C))
 let verified = ID(C) == D
 ```
+
 Above, we define the content identifier as `ID(C)`, extracted from the `Descriptor.digest` field.
 Content `C` is a string of bytes.
 Function `H` returns the hash of `C` in bytes and is passed to function `Encode` and prefixed with the algorithm to obtain the digest.
@@ -190,6 +201,17 @@ In the following example, the descriptor indicates that the referenced manifest 
 }
 ```
 
+In the following example, the descriptor indicates the type of artifact it is referencing:
+
+```json,title=Content%20Descriptor&mediatype=application/vnd.oci.descriptor.v1%2Bjson
+{
+  "mediaType": "application/vnd.oci.artifact.manifest.v1+json",
+  "size": 123,
+  "digest": "sha256:87923725d74f4bfb94c9e86d64170f7521aad8221a5de834851470ca142da630",
+  "artifactType": "application/vnd.example.sbom.v1"
+}
+```
+
 [rfc3986]: https://tools.ietf.org/html/rfc3986
 [rfc4634-s4.1]: https://tools.ietf.org/html/rfc4634#section-4.1
 [rfc4634-s4.2]: https://tools.ietf.org/html/rfc4634#section-4.2
@@ -198,3 +220,4 @@ In the following example, the descriptor indicates that the referenced manifest 
 [rfc6838-s4.2]: https://tools.ietf.org/html/rfc6838#section-4.2
 [rfc7230-s2.7]: https://tools.ietf.org/html/rfc7230#section-2.7
 [sha256-vs-sha512]: https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/hsMw7cAwrZE
+[iana]: https://www.iana.org/assignments/media-types/media-types.xhtml
--- a/manifest.md
+++ b/manifest.md
@@ -65,6 +65,11 @@ Unlike the [image index](image-index.md), which contains information about a set
 
         Entries in this field will frequently use the `+gzip` types.
 
+- **`refers`** *[descriptor](descriptor.md)*
+
+    This OPTIONAL property specifies a [descriptor](descriptor.md) of another manifest.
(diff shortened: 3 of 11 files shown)


vbatts:
OHMAN

artifact.md
```diff
@@ -0,0 +1,84 @@
...
+  Additionally, the following annotations SHOULD be used when deploying multi-arch container images:
+
+  - `org.opencontainers.platform.architecture`: CPU architecture for binaries
+  - `org.opencontainers.platform.os`: operating system for binaries
+  - `org.opencontainers.platform.os.version`: operating system version for binaries
+  - `org.opencontainers.platform.variant`: variant of the CPU architecture for binaries
+
+    Also, see [Pre-Defined Annotation Keys](annotations.md#pre-defined-annotation-keys).
+
+  User defined annotations MAY be used to filter various artifact types, e.g. signature public key hash, attestation type, and SBOM schema.
```
mounchin, line 53: As the spec allows user defined annotations, do we want to cap the number of annotations stored or returned as part of `referrers` API ?

sajayantony: The current proposal is that all annotations would be returned. If we were to trim down the list, then we need define the criteria.  For example, only the first 10 annotations would be returned.  That still doesn't cap the size of the payload since a single key could actually be large enough. 

I think it would be good to raise this issue in the distribution spec since the referrers response payload and registry implementations needs to now ensure that data from all manifests are returned.  https://github.com/opencontainers/distribution-spec/pull/335 /cc @michaelb990

sudo-bmitch: The index we return is rather minimal, so having enough annotations to cause this to exceed the recommended 4MB limit is likely to also exceed that limit in the referrer's manifest too. I think any limits we impose should be set on the original manifest, and not this API, to minimize complexity.

mounchin: Lets say worst case, there are like 20 artifacts in the referrers API reponse and each had say about ~3MB of annotations, then the API reponse payload would be very large. In these cases what should the API return? May be capping the reponse size per artifact?

sudo-bmitch:
@opencontainers/image-spec-maintainers given the size of this, we are giving this 2-4 weeks for any objections before merging. It was presented on the OCI call today and hopefully a recording will be posted soon. After merging it still won't be a tagged release, so if there are any issues raised, we can always open another PR.

lachie83:
We would kindly like to ask for reviews and/or approval from maintainers by August 25th, 2022. Please let us know if you don't think this is attainable.

lachie83:
Adding link to similar comment on distribution-spec PR. https://github.com/opencontainers/distribution-spec/pull/335#issuecomment-1212296424

artifact.md
sudo-bmitch: At that point, the response would trigger pagination, with one descriptor per page. It also wouldn't be supported to include more than one of those descriptors in the tag schema fallback. My hope is that manifests like that result in a bad enough UX, that bug reports get opened on the source and alternatives are used.

The downside of capping what is returned in the API response, is that any partial descriptor would then need to be pulled as a full manifest to determine if the client side filters match (the filtered annotation could be one that was trimmed). We would also need to provide an indicate that the descriptor is a partial response so clients know when they need to fall back to pulling the full manifest directly. When thinking of client implementations, that would probably be one pull for the API, one of the manifest by the artifact listing method, and then another pull of the manifest by the artifact get method, unless the manifest is cached between those methods.

descriptor.md
```diff
@@ -51,8 +56,14 @@ The following fields contain the primary properties that constitute a Descriptor
   The decoded data MUST be identical to the referenced content and SHOULD be verified against the [`digest`](#digests) and `size` fields by content consumers.
   See [Embedded Content](#embedded-content) for when this is appropriate.
 
+- **`artifactType`** *string*
```
shizhMSFT, line 59: The `artifactType` is defined twice. Should we merge both or drop one?

lachie83: Opened https://github.com/oci-playground/image-spec/pull/14

manifest.md
```diff
@@ -65,6 +65,11 @@ Unlike the [image index](image-index.md), which contains information about a set
 
         Entries in this field will frequently use the `+gzip` types.
 
+- **`refers`** *[descriptor](descriptor.md)*
+
+    This OPTIONAL property specifies a [descriptor](descriptor.md) of another image or [artifact](artifact.md).
```
mikebrow, line 70: can we clarify if there is a restriction or not for media type of the refers descriptor.. image manifest and image index manifest for example..

mikebrow (commented):
can we, should we add something over in image-index.md, such as another example, for using the image.index manifest to hold artifact manifests in the list of manifests? See example provided in distribution api..

artifact.md
```diff
@@ -0,0 +1,84 @@
...
+
+## *Artifact Manifest* Property Descriptions
+
+- **`mediaType`** *string*
+
+  This property MUST be used and contain the media type `application/vnd.oci.artifact.manifest.v1+json`.
+
+- **`artifactType`** *string*
+
+  This property SHOULD specify the mediaType of the referenced content and MAY be registered with [IANA][iana].
```
mikebrow, line 20: the MAY language, here, doesn't sync with the "is" and MUST language in other spots..

might be useful to make media-types.md media-and-artifact-types and put a paragraph there talking about any MUST/MAY language wrt artifact types?

artifact.md
```diff
@@ -0,0 +1,84 @@
...
+
+- **`blobs`** *string*
+
+  This OPTIONAL property contains a list of [descriptors](descriptor.md).
+  Each descriptor represents an artifact of any IANA mediaType.
```
wmesard, line 25: I have a use case where each element in the `blobs` array is associated with the corresponding element in the `layers` array of the referent OCI image. In other words, the length of the two arrays are always equal. However, a blob in the referring artifact may be omitted for any particular image layer.

This can be addressed by adding to this line: "Any element in the list may also be `null`."

Posting here for comments. I can open a new PR if that's easier.

sudo-bmitch: I suspect a null descriptor entry may cause issues for tools. What if an annotation in the descriptor was used to indicate which layer it is referencing. That could be a layer digest, or just a list index number.

manifest.md
sudo-bmitch: We shouldn't have a restriction here. An artifact shipped with an Artifact manifest or this Image manifest can refer to any other manifest, including an Index. (I.e. You can attach a signature on a multi-platform manifest.)

mikebrow: before suggesting text changes here to include that clarification..  Do we have to worry about/restrict wider dependency loops here outside the obvious don't refer to yourself loop? E.g. an image manifest with a refers to it's own image.index.. an image manifest with a refers to an artifact that refers to this image manifest...

sudo-bmitch: For loops, I'm looking at whether the refers descriptor can point to a normal child of the manifest. If all children are blobs (as with Image and Artifact manifests), and not manifests (like we have with an Index), you can't point to a manifest that's a child.

It's a reverse pointer, so we aren't worried about parents (both pointers would be going the same direction). And you can't know the digest of the parent in advance without a hash collision/prediction.

artifact.md
wmesard: This PR is defining the spec that the tools will have to conform to, no? But okay; I can go with the annotation approach. Thanks.

manifest.md
mikebrow: there may be an opportunity in platform image index.. where an artifact manifest included in the platform image index could be to an artifact that does not have a refers.. and in that case the artifact could be implied to be for that index (refer) by inclusion... thus allowing an index to have an artifact that refers to itself without incurring the loop restriction.. thinking out loud.

artifact.md
```diff
@@ -0,0 +1,72 @@
...
+- **`mediaType`** *string*
+
+  This property MUST be used and contain the media type `application/vnd.oci.artifact.manifest.v1+json`.
+
+- **`artifactType`** *string*
+
+  This property SHOULD be used and contain the mediaType of the referenced artifact.
+  If defined, the value MUST comply with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2], and MAY be registered with [IANA][iana].
+
+- **`blobs`** *string*
```
sudo-bmitch, line 23: We need to change "string" to an array of descriptors.

artifact.md
```diff
@@ -0,0 +1,72 @@
...
+For the media type(s) that this is compatible with see the [matrix](media-types.md#compatibility-matrix).
+
+# Artifact Manifest
+
+## *Artifact Manifest* Property Descriptions
+
+- **`mediaType`** *string*
+
+  This property MUST be used and contain the media type `application/vnd.oci.artifact.manifest.v1+json`.
+
```
fawadasaurus, line 17: Would it be possible to allow `config` here as `OPTIONAL`?  In my use case, I want to use an OCI artifact for packaging an application and want to use the `config` to provide execution details for the application.

sudo-bmitch: You can push an additional blob with a distinct media type. Or you can use the image manifest with a custom config. I don't believe anything we are doing requires an artifact to use the artifact manifest. The existing OCI Artifact guidance using the image manifest is still valid, and I'd even recommend it since it will take time for registries to support the new artifact manifest type.

fawadasaurus: > you can use the image manifest with a custom config

Yes.  Thats exactly what I am doing now, but I was hoping that when the artifact spec is released, that I would not be having to have a technically non-standard image.  My application does not use layers so the image spec with `layers` is not technically valid.

I understand that the spec will take some time to be supported but I would like that once the time comes, I am able to use the OCI distribution without breaking the `rules`:

Here is the spec description for `layers`
> The array MUST have the base layer at index 0. Subsequent layers MUST then follow in stack order (i.e. from layers[0] to layers[len(layers)-1]). The final filesystem layout MUST match the result of [applying](https://github.com/opencontainers/image-spec/blob/main/layer.md#applying-changesets) the layers to an empty directory. The [ownership, mode, and other attributes](https://github.com/opencontainers/image-spec/blob/main/layer.md#file-attributes) of the initial empty directory are unspecified.

By adding another blob, it is making it more difficult for existing registry distributions to be used because (for example) harbor already reads the `config` for an image automatically do display in the UI.  So now anyone that wants to use an artifact will need to make their own custom flows that are specific to their own use case to download an individual arbitrary blob from the image as opposed to just downloading a standard manifest config.

Your point is valid that these can be worked around, but I am hoping that the solution is built flexible enough that novel use cases are not `hacks` but can be widely used across all OCI-compliant registry implementations.

sajayantony: The current goal of the working group was to support attaching references and that was the primary motivation of creating the artifact manifest for the reference type working group without having to hack the image manifest to attach blobs of signatures or sboms. 

The artifact manifest does have potential, and it would be nice to consider if we want to expand its scope to help define other types of images/applications. If we do want to enhance the scope of the artifact manifest I would recommend we can do this once reference type work is merged and discuss adding more fields (like the data field in the descriptor https://github.com/opencontainers/image-spec/pull/826). 

Does that sound reasonable?

sudo-bmitch: > By adding another blob, it is making it more difficult for existing registry distributions to be used because (for example) harbor already reads the `config` for an image automatically do display in the UI. So now anyone that wants to use an artifact will need to make their own custom flows that are specific to their own use case to download an individual arbitrary blob from the image as opposed to just downloading a standard manifest config.

I haven't pulled up the Harbor UI in a while. What metadata are they extracting from the `config`? One of the goals we had was to move the config mediaType to artifactType, and then any data could be an annotation. I'm wondering if we have all the data needed without introducing another blob.

There's also nothing preventing us from defining a new artifact metadata json schema in the future, with a well known media type, that could be added as a blob to any artifact.

fawadasaurus: Thanks for the feedback.  I am certainly not wanting to derail any efforts.  I will wait for the spec to get merged and then will post any feedback as an issue and we can address in the future.

> What metadata are they extracting from the config?

Harbor reads the JSON in the config file and displays it directly in the UI:

config.json
```
{
    "flows": [
        "flow.wafl"
    ],
    "permissions": [
        "permissions.wafl"
    ],
    "interfaces": [
        "add.apex",
        "delete.apex",
        "edit.apex",
        "list.apex"
    ]
}
```
image

specs-go/v1/descriptor.go
```diff
@@ -44,6 +44,9 @@ type Descriptor struct {
 	//
 	// This should only be used when referring to a manifest.
 	Platform *Platform `json:"platform,omitempty"`
+
+	// ArtifactType is the IANA media type of this artifact.
+	ArtifactType string `json:"artifactType,omitempty"`
```
mikebrow, line 49: ?

sudo-bmitch: This is used in the Referrers listing.

sudo-bmitch (approved):

specs-go/v1/descriptor.go
mikebrow: if this refers descriptor example is correct, the definition could be improved by saying
```suggestion
        // a refers descriptor use of artifactType SHOULD be to specify the media type of the artifact being referred to
	ArtifactType string `json:"artifactType,omitempty"`
```

vbatts:
looking at the git history, it seems there was a merge done instead of rebase, and now there is a bubble with #919 in it.

artifact.md
```diff
@@ -0,0 +1,72 @@
...
+  This property SHOULD be used and contain the mediaType of the referenced artifact.
+  If defined, the value MUST comply with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2], and MAY be registered with [IANA][iana].
+
+- **`blobs`** *array of objects*
+
+  This OPTIONAL property is an array of objects and each item in the array MUST be a [descriptor](descriptor.md).
+  Each descriptor represents an artifact of any IANA mediaType.
+  The list MAY be ordered for certain artifact types like scan results.
+
+- **`refers`** *[descriptor](descriptor.md)*
```
vbatts, line 29: for clarity, this is _the_ field that "refers to" the originating object?

artifact.md
```diff
@@ -0,0 +1,72 @@
+# OCI Artifact Manifest Specification
+
+The goal of the Artifact Manifest Specification is to define content addressable artifacts in order to store them along side container images in a registry. Like [OCI Images](manifest.md), OCI Artifacts may be referenced by the hash of their manifest. Unlike OCI Images, OCI Artifacts are not meant to be used by any container runtime.
```
vbatts, line 3: nit, each sentence to it's own line

descriptor.md
```diff
@@ -1,11 +1,12 @@
 # OCI Content Descriptors
 
-* An OCI image consists of several different components, arranged in a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
-* References between components in the graph are expressed through _Content Descriptors_.
-* A Content Descriptor (or simply _Descriptor_) describes the disposition of the targeted content.
-* A Content Descriptor includes the type of the content, a content identifier (_digest_), and the byte-size of the raw content.
-* Descriptors SHOULD be embedded in other formats to securely reference external content.
-* Other formats SHOULD use descriptors to securely reference external content.
+- An OCI image consists of several different components, arranged in a [Merkle Directed Acyclic Graph (DAG)](https://en.wikipedia.org/wiki/Merkle_tree).
```
vbatts, line 3: Is this a space to clarify that descriptors are not exclusive to "An OCI  image"?

descriptor.md
```diff
@@ -68,7 +77,7 @@ The _algorithm_ specifies the cryptographic hash function and encoding used for
 
 A digest string MUST match the following [grammar](considerations.md#ebnf):
 
-```
+```text
```
vbatts, line 80: I wonder if github markdown handles ebnf instead of just text here (and below)?

specs-go/v1/manifest.go
```diff
@@ -30,6 +30,9 @@ type Manifest struct {
 	// Layers is an indexed list of layers referenced by the manifest.
 	Layers []Descriptor `json:"layers"`
 
+	// Refers is an optional link to any existing manifest within the repository.
+	Refers *Descriptor `json:"refers,omitempty"`
```
vbatts, line 34: particular reason for using a pointer here? Looks like we're using a mix, and usually there was reasoning for choosing one way or the other

vbatts (dismissed):
This looks good. Nothing breaking and will provide clarity on a tremendous pattern for clients and platforms to provide.

There are a couple nits that need to be cleaned up, but they can be follow-up PRs as they don't affect behavior/functionality.

One unfortunate aspect of merging this from the oci-playground branch is that a number of the commits are noisy as they get shadowed or cleaned up later... That isn't ideal and the _only_ benefit I see in leaving it as is would be so that all the authors preserve their involvement in the commit history.

specs-go/v1/descriptor.go
mikebrow: needs some explanation regarding the parts of the spec that are descriptors and what the artifact type field currently means in those contexts.. will try to draft something probably needs more examples

artifact.md
sudo-bmitch: Yes. The phrasing there can use work. I think it has bounced around several editors until getting punted to "read the distribution-spec".

descriptor.md
sudo-bmitch: That would make a good follow up PR. This line is showing up in the change log because I switched the list bullets from `*` to `-` to get my markdown linter to stop screaming at the rest of the file. :)

specs-go/v1/manifest.go
sudo-bmitch: To support `omitempty`. A struct is never "empty" to Go because it can't be compared to a zero value.

descriptor.md
sudo-bmitch: Thanks, I added this.

artifact.md
sudo-bmitch: Thanks, fixed.

artifact.md
```diff
@@ -0,0 +1,72 @@
...
+
+- **`blobs`** *array of objects*
+
+  This OPTIONAL property is an array of objects and each item in the array MUST be a [descriptor](descriptor.md).
+  Each descriptor represents an artifact of any IANA mediaType.
+  The list MAY be ordered for certain artifact types like scan results.
+
+- **`refers`** *[descriptor](descriptor.md)*
+
+  This OPTIONAL property specifies a [descriptor](descriptor.md) of another manifest.
```
lachie83, line 33: cc @vbatts @sudo-bmitch - Does this provide more clarity in response to https://github.com/opencontainers/image-spec/pull/934/files/5240cdaf5cfb2f62f1b6b8860e5561332437270e#r955050333

```suggestion
  This OPTIONAL property specifies a [descriptor](descriptor.md) of another manifest. If defined, the artifact `refers` to, or has a reference to the specified manifest
```

vbatts: I think so. (please put that sentence on its own newline)

descriptor.md
vbatts: yeah fair

vbatts:
during discussion on the call, this has enough features and fields that it merits a point release, but not a major version change.

sudo-bmitch (dismissed):

sajayantony (approved):

estesp:
Thanks to all for the work on this in the WG and getting all this together (and dealing with all the feedback) and ready! IANAM but LGTM! 🎉

sudo-bmitch (approved):

sajayantony (approved):

PR #945, opened 2022-09-06
Fix whitespace consistency in config.md

tianon:
*Some* of the sub-items of `config` were inconsistently indented (some using three spaces, some using two) -- adjusted to two to match the rest of the file.

### DIFF
--- a/config.md
+++ b/config.md
@@ -133,51 +133,51 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
   The execution parameters which SHOULD be used as a base when running a container using the image.
   This field can be `null`, in which case any execution parameters should be specified at creation of the container.
 
-   - **User** *string*, OPTIONAL
+  - **User** *string*, OPTIONAL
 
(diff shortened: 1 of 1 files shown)


tianon:
I was looking at why https://specs.opencontainers.org/image-spec/config/?v=v1.0.1#properties is rendering the sub-items of `config` as being even with it instead, and noticed this inconsistency.  I'm not 100% sure which engine that's using to render (or how to most effectively test it), but IMO this seems like a sane change even if it's not the cause of that rendering glitch. :sweat_smile:

tianon:
The CI-generated PDF renders correctly, for what that's worth!

image

(It was probably fine with the janky whitespace too, though)

sudo-bmitch (approved):
LGTM. Children of rootfs and history should also be adjusted to 2 spaces for consistency. And the bold and italic markup also needs work. But those can all be a separate PR, especially since it needs to happen in other files, too. I've got on my list to add a markdown linter here after doing a cleanup.

vbatts (approved):
:cupcake:

PR #950, opened 2022-09-09
Rename refers field to subject

sajayantony:
Addresses the renaming of `refers` --> `subject`  for #940 

### DIFF
--- a/specs-go/v1/manifest.go
+++ b/specs-go/v1/manifest.go
@@ -30,8 +30,8 @@ type Manifest struct {
 	// Layers is an indexed list of layers referenced by the manifest.
 	Layers []Descriptor `json:"layers"`
 
-	// Refers is an optional link to any existing manifest within the repository.
-	Refers *Descriptor `json:"refers,omitempty"`
+	// Subject is an optional link from the image manifest to another manifest forming an association between the image manifest and the other manifest.
+	Subject *Descriptor `json:"subject,omitempty"`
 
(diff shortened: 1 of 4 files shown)


specs-go/v1/manifest.go
```diff
@@ -30,8 +30,8 @@ type Manifest struct {
 	// Layers is an indexed list of layers referenced by the manifest.
 	Layers []Descriptor `json:"layers"`
 
-	// Refers is an optional link to any existing manifest within the repository.
-	Refers *Descriptor `json:"refers,omitempty"`
+	// Subject is an optional link to a manifest.
```
mikebrow, line 33:
```suggestion
	// Subject is an optional link from the image manifest to another manifest forming an association between the image manifest and the other manifest.
```

mikebrow (commented):
see suggested edit

michaelb990 (commented):
IANAM but LGTM 👍

sajayantony:
Thanks @sudo-bmitch for tooling fixes and @mikebrow for the working updates.

sudo-bmitch (approved):
LGTM. We need a similar update to distribution-spec.

mikebrow (approved):
LGTM

stevvooe (approved):
LGTM

Assuming this is safe. `Target` might be better, since `subject` usually refers to a set of users in auth systems, but this looks fine.

PR #969, opened 2022-11-01
feat: dependabot workflow automation for updating dependency

Rajpratik71:
Signed-off-by: Pratik Raj <Rajpratik71@gmail.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  # Dependencies listed in go.mod
+  - package-ecosystem: "gomod"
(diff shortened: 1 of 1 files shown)


sparr (approved):
I was surprised to see that `"/"` is the appropriate directory for the github-actions pacakge-ecosystem. https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directory

sudo-bmitch (changes requested):
I agree with Tianon on this one. I thought this was for the security updates, but we already have those enabled on the project. For other dependencies, that can be done on an as-needed basis.

PR #1005, opened 2023-02-02
Remove filtersApplied from image-spec

imjasonh:
See https://github.com/opencontainers/distribution-spec/pull/380 which proposes changing how the registry communicates that it applied filters to be "out-of-band" of the image manifest response, so this doesn't need to be documented in image-spec at all.

Signed-off-by: Jason Hall <jason@chainguard.dev>


sajayantony:
I feel this is a good change and as pointed out keeps the request semantics out of content/index. I feel we should merge this after  distribution spec changes have merged.

jonjohnsonjr (approved):

jonboulle (approved):
+1, I don't think this should have landed in the image spec. Let's keep it to distribution.

sajayantony (approved):

PR #1010, opened 2023-02-04
Require IANA mediaType for image config.mediaType and layers.mediaType

imjasonh:
This guidance is carried over from the artifacts repo, and matches exactly the [guidance for artifact manifests](https://github.com/opencontainers/image-spec/blob/main/artifact.md#artifact-manifest-property-descriptions).

This is intended as progress toward https://github.com/opencontainers/artifacts/issues/63

Signed-off-by: Jason Hall <jason@chainguard.dev>

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -40,6 +40,8 @@ Unlike the [image index](image-index.md), which contains information about a set
 
         Manifests concerned with portability SHOULD use one of the above media types.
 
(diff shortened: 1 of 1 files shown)


vbatts (approved):
yeah this seems sane. Compliance, but not requiring IANA registration

sudo-bmitch (approved):
LGTM

PR #1026, opened 2023-02-13
Image Manifest config must allow any mediaType?

brackendawson:
Following investigations in #1025 I noticed that a registry behaving like Quay.io can comply with the 1.1 spec but only allow an Image Manifest artifact with an `artifactType` of `application/vnd.oci.image.config.v1+json`.

Should this be a MUST or a SHOULD, or is the existing text already appropriate?

Signed-off-by: Bracken Dawson <abdawson@gmail.com>

### DIFF
--- a/manifest.md
+++ b/manifest.md
@@ -34,13 +34,9 @@ Unlike the [image index](image-index.md), which contains information about a set
     - **`mediaType`** *string*
 
         This [descriptor property](descriptor.md#properties) has additional restrictions for `config`.
-        Implementations MUST support at least the following media types:
+        Implementations MUST support any media type that complies with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2], and MAY be registered with [IANA][iana].
 
-        - [`application/vnd.oci.image.config.v1+json`](config.md)
-
-        Manifests concerned with portability SHOULD use one of the above media types.
-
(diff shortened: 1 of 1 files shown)


manifest.md
```diff
@@ -34,13 +34,9 @@ Unlike the [image index](image-index.md), which contains information about a set
     - **`mediaType`** *string*
 
         This [descriptor property](descriptor.md#properties) has additional restrictions for `config`.
-        Implementations MUST support at least the following media types:
-
-        - [`application/vnd.oci.image.config.v1+json`](config.md)
+        Implementations MUST support any media type that complies with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2], and MAY be registered with [IANA][iana].
```
jonboulle, line 37: I am a bit confused by this change - it now mandates implementations to support _any_ media types, which isn't feasible.

Would it be clearer for you if we added a single line on line 40 like this:
```
 Implementations MUST support at least the following media types:
        - [`application/vnd.oci.image.config.v1+json`](config.md)
 
        **Manifests MAY use, and implementations MAY support, other media types not defined here.**

```

brackendawson: I suppose it depends what kind of implementation you are, a registry can trivially support any media type, it just has to store the blob. A container runtime needs the information in there, so it can only support known media types, or maybe also their ducks.

sudo-bmitch (changes requested):
I agree with the comments that this isn't feasible for container runtimes (like containerd and runc) to support any config mediaType. We'd need to split the wording for different scenarios (runtimes, registries, and other content producers/consumers).

brackendawson:
I think if any change is needed to the config.mediaType description it probably belongs in #1043 now.

PR #1028, opened 2023-02-16
Clarify ignore language, ensure future compat

AaronFriel:
The usage of "ignore[d]" appears may have resulted in ambiguous parsing of the spec. 

One interpretation is that OCI 1.0 _already_ required registries to accept Artifact and future media types, and implementations - clients and registries - that fail to handle arbitrary media types for blobs, arbitrary media types for image configs.

### DIFF
--- a/config.md
+++ b/config.md
@@ -247,7 +247,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
        This field is used to mark if the history item created a filesystem diff.
        It is set to true if this history item doesn't correspond to an actual layer in the rootfs section (for example, Dockerfile's [ENV](https://docs.docker.com/engine/reference/builder/#/env) command results in no change to the filesystem).
 
-Any extra fields in the Image JSON struct are considered implementation specific and MUST be ignored by any implementations which are unable to interpret them.
+Any extra fields in the Image JSON struct are considered implementation specific and MUST NOT generate an error by any implementations which are unable to interpret them.
 
 Whitespace is OPTIONAL and implementations MAY have compact JSON with no whitespace.
 
--- a/considerations.md
+++ b/considerations.md
@@ -1,6 +1,8 @@
 # Extensibility
 
-Implementations that are reading/processing [manifests](manifest.md) or [image indexes](image-index.md) MUST NOT generate an error if they encounter an unknown property.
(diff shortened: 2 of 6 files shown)


config.md
```diff
@@ -247,7 +247,7 @@ Note: Any OPTIONAL field MAY also be set to null, which is equivalent to being a
        This field is used to mark if the history item created a filesystem diff.
        It is set to true if this history item doesn't correspond to an actual layer in the rootfs section (for example, Dockerfile's [ENV](https://docs.docker.com/engine/reference/builder/#/env) command results in no change to the filesystem).
 
-Any extra fields in the Image JSON struct are considered implementation specific and MUST be ignored by any implementations which are unable to interpret them.
+Any extra fields in the Image JSON struct are considered implementation specific and MUST NOT generate an error by any implementations which are unable to interpret them.
```
tianon, line 250: This is specific to _registry_ implementations, right? (as opposed to "client implementations" which feels included in that "any")

In other words, the intent is that we expect Docker Hub and friends to accept Helm charts, but it's still 100% reasonable and expected for the Docker CLI/Engine to bail with a helpful error if you try to `docker run` a Helm chart. :sweat_smile:

AaronFriel: Yeah, hm.

An example of a client that's in-the-middle is, e.g., `oras copy`. I expect ORAS to faithfully copy manifests and blobs from one registry to another. But I suppose that's a "MAY".

AaronFriel:
@sudo-bmitch @tianon PTAL, I've amended this in two ways:

## Grandfathering OCI 1.1-rc behavior

Documenting deviation from the `descriptor` spec, but limiting the blast radius. That is, rather than destroy the value of the media type, codify it as opaque and inaccurate in a limited number of cases.

This preserves the intent of the `descriptor` spec, which I fear with #999 will be irreparably damaged.

This change should not conflict with #1029, except #1029 should be changed to indicate that this "SCRATCH" specification exists only for artifacts using image manifests and going forward, implementations SHOULD use artifact manifest.

## Distinguishing behavior of clients vs registries/proxies

Implementations fall into two categories:
1. Storing and copying content, which MUST accept arbitrary media types and content for manifests and blobs.
2. Processing content, which MUST NOT generate an error when encountering unknown properties. (Implied: they MAY generate an error when encountering unknown manifests.)

AaronFriel:
PTAL

AaronFriel:
Closing in favor of #1030

PR #1051, opened 2023-04-24
Remove artifact media type reference

djdongjin:
The artifact manifest media type (and the `artifact.md`) was removed in https://github.com/opencontainers/image-spec/pull/999. This PR removes expired reference to them.


tianon (approved):
Good catch :+1:

sudo-bmitch (approved):
LGTM, thanks!

PR #1054, opened 2023-04-27
Fix minor typo in doc comment of specs-go

imjasonh:
`ppc64` is not a valid architecture, but `ppc64le` is.


tianon (approved):
I agree 100% with this change, but do want to clarify that `ppc64` _is_ valid for big endian 64bit POWER and this isn't necessarily a typo or incorrect, but `ppc64le` is much more common and thus a better example 😅

(https://go.dev/doc/install/source#environment)

sudo-bmitch (approved):

Issue #1053, opened 2023-04-27
disambiguate wasi by renaming it to wasip1


codefromthecrypt:
Comments happened on https://github.com/opencontainers/image-spec/pull/964#issuecomment-1412108382 post merge, about how "wasi" is ambiguous, and [Go disambiguated it with "wasip1"](https://github.com/golang/go/issues/58141). As the issue was already merged, comments didn't really lead to action.

Since then, [rust](https://github.com/rust-lang/rust/pull/110596) and [JS](https://nodejs.org/en/blog/announcements/v20-release-announce#progress-on-web-assembly-system-interface-wasi) are beginning to enforce a wasi version similar, if not exactly the same as the Go discussion. I think we shouldn't really keep "wasi" knowing this, especially as the whole thing is new might as well do it right.

My suggestion is to keep the short name go uses "wasip1" as we know it is incompatible with subsequent versions. Without this, runtimes can't use the OS to figure out if the code might run or not, invalidating the concept entirely.

PR #1055, opened 2023-04-27
Remove special guidance around wasm

imjasonh:
Go is going to support a platform like `wasip1/wasm`, `wasip2/wasm`, and eventually `wasi/wasm`. We should remove the guidance that Wasm is a special case, and fallback to following Go's guidance when describing a platform.

Fixes https://github.com/opencontainers/image-spec/issues/1053

This exception was added in https://github.com/opencontainers/image-spec/pull/964 -- see there for original (outdated!) discussion.

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -58,7 +58,6 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
         This REQUIRED property specifies the operating system.
         Image indexes SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOOS`][go-environment2].
-        An exception to this guidance is provided for Wasm applications that rely on the WASI system interface, where the `os` SHOULD be `wasi`.
 
     - **`os.version`** *string*
 


imjasonh:
A noteworthy excerpt of the PR that added it:

> Diverging from Go here provides some small opportunity for trouble for us down the road:
> 
> - Go could add official support for WASI, but choose to do so using a `GOOS` value that isn't `wasi` -- this would confuse our guidance more, since there'd be a valid `GOOS` for WASI that isn't our recommended `.platform.os` value.

Well, that's what happened! Let's back it out before we cut v1.1 so we're not stuck with this decision any longer than necessary.

tianon:
> Go is going to support a platform like `wasm/wasip1`, `wasm/wasip2`, and eventually `wasm/wasi`. We should remove the guidance that Wasm is a special case, and fallback to following Go's guidance when describing a platform.

Scratch that, reverse it (`wasi` is the `os` value, not the `arch` value) :sweat_smile:

> Go is going to support a platform like `wasip1/wasm`, `wasip2/wasm`, and eventually `wasi/wasm`. We should remove the guidance that Wasm is a special case, and fallback to following Go's guidance when describing a platform.

I'd like to solicit comments from @rumpl, who has done a lot of work on container runtimes (moby/moby engine maintainer) _and_ WASM/WASI, since I think he's got a somewhat unique perspective with a good amount of experience in both halves of this. :bow: :heart:

I also think https://github.com/opencontainers/image-spec/pull/964#issuecomment-1462323781 is worth revisiting, specifically whether we should consider using `os.version` for this (which Go doesn't have an analogue for that I'm aware of, so they couldn't have considered using it) -- it's hard to say what the future of `wasi` or `wasm` holds, but I'm guessing this isn't going to be the last time we have incompatibilities between builds and which versions runtimes can reasonably support. :disappointed:

imjasonh:
> Scratch that, reverse it (wasi is the os value, not the arch value) 😅

Thank you for pointing this out! I'll update the original comment to avoid any (more) confusion.

> I also think https://github.com/opencontainers/image-spec/pull/964#issuecomment-1462323781 is worth revisiting, specifically whether we should consider using os.version for this (which Go doesn't have an analogue for that I'm aware of, so they couldn't have considered using it) -- it's hard to say what the future of wasi or wasm holds, but I'm guessing this isn't going to be the last time we have incompatibilities between builds and which versions runtimes can reasonably support. 😞

I think it's simplest to say "just follow Go's precedent" for as long as we can, until there's some case where Go's semantics don't support what the OCI community needs. This very nearly happened with Wasm, and I think our response to it was good. Little did I realize that Go would leap back into action and catch up and surpass us. 😆 

So I think `wasip1/wasm` is better than `wasi/wasm`-with-`os.version=p1`, because we can still just follow Go's lead, and the rules are relatively simpler.

If/when in the future some incompatibility comes up where Go precedent doesn't satisfy OCI's needs, we can carve out another exemption, or say "for `os=newthing`, follow `$otherlang`'s platform semantics," which may involve leveraging `os.version` (or `variant`, or a new field we specify, who knows)

sudo-bmitch (approved):

codefromthecrypt:
FWIW `wasip1/wasm` was the result of a sensible deliberation amongst a lot of different people. It wasn't random or not thought through, and it was made knowing about the existing "wasi" conventions and the problems with it. There are a growing number on 👍 this PR because it makes sense and fits into existing tool work. We really don't want a world of preventable "if statements" in Makefiles etc. Thanks for raising this!

tianon (approved):

PR #1062, opened 2023-05-12
Add `org.opencontainers.image.source.subpath`

### LINKED ISSUE #1046: Proposal: additional Pre-Defined Annotation Keys
While attempting to use the [Pre-Defined Annotation Keys] to track container images back to the corresponding source code, we (re)discovered that `org.opencontainers.image.source` (URL to get source code for building the image) is insufficient to determine the source code associated with the image when building from a monorepo.  (See #886 for an earlier, less action-oriented version of this issue.)

The particular flavor of this issue is that `org.opencontainers.image.source` is likely to be something like `https://github.com/opencontainers/image-spec.git`, but those URLs don't provide a way to specify a sub-path within a git repository, which may be useful when building multiple images from a single repo.  This is a common practice for many projects, including Kubernetes, Tekton, and Knative (three projects I've interacted with the most).

I'd like to propose documenting the following additional Pre-Defined Annotation Key:

`org.opencontainers.image.source.subpath`: A relative path within the source repository used as the base directory to build the container (string)

A few questions on definition which we can either ignore, document, or finesse to avoid defining:
* Windows: do windows image builds use windows path conventions?
* Default / empty value: Does this mean `.`, or unknown?  If it means "unknown", does that mean tools should generally set `.`?  If it means `.`, how much do all the existing images mess with this?

I'm happy to propose a PR if this would be useful.  @imjasonh @sudo-bmitch as the last two to shepherd changes to this file.

evankanderson:
Fixes #1046

Defines an additional subpath within the source repository.  In combination with the `source` attribute, this functions as a correlation ID for multiple images produces from the same source repository (e.g. the different containers produced by https://github.com/knative/serving, or any other mono-repo type arrangement which builds related artifacts from a common source tree).

`source.subpath` is explicitly tool-specific; I'd expect that most source repositories use a single tool for building images, or at the very least don't use two tools with the same `subpath` value to produce different images.  There's probably at least one awesome counter-example out there... I don't know if I want to see it.

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -23,6 +23,7 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.url** URL to find more information on the image (string)
 * **org.opencontainers.image.documentation** URL to get documentation on the image (string)
 * **org.opencontainers.image.source** URL to get source code for building the image (string)
+* **org.opencontainers.image.source.subpath** A tool-specific path within the `source` repository which be used to distinguish different build targets in the same source repository
 * **org.opencontainers.image.version** version of the packaged software
   * The version MAY match a label or tag in the source code repository
   * version MAY be [Semantic versioning-compatible](https://semver.org/)


tianon:
(soft) NACK; https://github.com/opencontainers/image-spec/issues/1046#issuecomment-1546687225

(Soft because I'm still willing to be convinced, but I think we need more spec maintainer opinions to be expressed for this to really move one way or the other)

annotations.md
```diff
@@ -23,6 +23,7 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.url** URL to find more information on the image (string)
 * **org.opencontainers.image.documentation** URL to get documentation on the image (string)
 * **org.opencontainers.image.source** URL to get source code for building the image (string)
+* **org.opencontainers.image.source.subpath** A tool-specific path within the `source` repository which be used to distinguish different build targets in the same source repository
```
AkihiroSuda, line 26: A tool-specific path should be just a third party annotation with their reversed-domain prefix

AkihiroSuda: And probably it doesn’t need to be a tool-specific:

```suggestion
* **org.opencontainers.image.source.subpath** A relative path within the `source` repository which be used to distinguish different build targets in the same source repository. The path separator MUST be `slash`. The path MUST NOT contain “..” (upper dir) component.
```

sudo-bmitch (changes requested):
I think we have three options here depending on what we want to do:

1. Decline this and use a tool specific annotation (e.g. something like `io.buildpacks.source.subpath`) if the goal is to have that tool define unique builds.
2. Change this to `org.opencontainers.image.instance` or some other unique key to dedup multiple builds in the same source repo. The value would be free form, based on the build tooling. E.g. makefiles may specify a target name. And docker builds may include the context, dockerfile, target, platform, build args, and any other inputs.
3. Leave this proposal as is, but make it generic and remove any expected results since different tooling will use it in ways that do not guarantee it is a unique build within the git repo. With other popular build tools, two different images can be built from the same subpath by altering other inputs.

annotations.md
```diff
@@ -23,6 +23,7 @@ This specification defines the following annotation keys, intended for but not l
 * **org.opencontainers.image.url** URL to find more information on the image (string)
```
CHEFSUSHIWILL1, line 23: thank you

CHEFSUSHIWILL1 (changes requested):

tianon:
I'm still partial to Brandon's option 1 (have each tool define their own additional fields for uniqueness if that's a property that they want/need instead of relying on us to somehow be the arbiter of all the possible uniqueness vectors for every build tool). :see_no_evil: :heart:

annotations.md
evankanderson: Does the path separator need to be `/` even on Windows systems (genuinely don't know enough to answer here).

CHEFSUSHIWILL1 (approved):

PR #1066, opened 2023-05-25
Add artifactType to image index

AaronFriel:
It would be useful for implementations which generate multiple artifacts, each with their own platform or other metadata and perhaps, associated signatures, attestations, to be able to index them all in a single image index.

As an implementer, I'm considering having a top level "package" stored as an image index, which then references all the artifacts in the package as artifacts (image manifests). The artifactType of the image index would be "application/...package", and the artifactType of the image manifests could be:

* "application/...plugin" with annotations and/or platform fields set to indicate that it is the plugin for, e.g., macOS with Apple Silicon (darwin-arm64)
* "application/...schema" with annotations indicating this is a separate artifact
* ... other artifact types as needed

Using an image index in this use case is helpful as we would want to associate SBOMs, signatures to the individual artifacts references, so that a client only needs to download and verify the artifacts it consumes selects from the index.

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -22,6 +22,11 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
   When used, this field MUST contain the media type `application/vnd.oci.image.index.v1+json`.
   This field usage differs from the [descriptor](descriptor.md#properties) use of `mediaType`.
 
+- **`artifactType`** *string*
+
+  This OPTIONAL property contains the type of an artifact when the manifest is used for an artifact.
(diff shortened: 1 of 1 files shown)


sajayantony (approved):
This aligns with #1020 and defines these fields that can be used by artifact authors. These are optional fields and having them enables distribution spec to return indexes that may be referrers to other artifacts.

tianon (approved):
IMO this is reasonable (even separately from #1020, now that I've looked at it again with fresh eyes) :bow:

sajayantony:
Need to update the json schema and the go-specs. I can do that up as follow up PR.

PR #1105, opened 2023-08-12
Add `riscv64` arch to `check{Architecture|Platform}`

michalbiesek:


sudo-bmitch (approved):

schema/validator.go
```diff
@@ -232,7 +233,7 @@ func checkPlatform(OS string, Architecture string) {
 		"darwin":    {"386", "amd64", "arm", "arm64"},
 		"dragonfly": {"amd64"},
 		"freebsd":   {"386", "amd64", "arm"},
-		"linux":     {"386", "amd64", "arm", "arm64", "ppc64", "ppc64le", "mips64", "mips64le", "s390x"},
+		"linux":     {"386", "amd64", "arm", "arm64", "ppc64", "ppc64le", "mips64", "mips64le", "s390x", "riscv64"},
```
tianon, line 236: IMO we should sort these, but I'm also not sure what this file is for (why are we hand-maintaining a list of things we've delegated to Go to define specifically partly so we don't have to hand-maintain such a list?).

```suggestion
		"linux":     {"386", "amd64", "arm", "arm64", "mips64", "mips64le", "ppc64", "ppc64le", "riscv64", "s390x"},
```

I don't feel super strongly though given that the list is already not in proper sorted order (and am thus approving).

tianon (approved):

PR #1122, opened 2023-09-15
add ORAS to implementations.md

FeynmanZhou:
[ORAS](https://oras.land/) is an OCI registry client and a set of libraries that provide a way to distribute OCI Artifacts to and from OCI Registries. It is a CNCF Sandbox project. ORAS is compliant with OCI [image-spec v1.1.0-rc4](https://github.com/opencontainers/image-spec/blob/v1.1.0-rc4/manifest.md#oci-image-manifest-specification) and [distribution-spec v1.1.0-rc3](https://github.com/opencontainers/distribution-spec/tree/v1.1.0-rc3).


sudo-bmitch (approved):

shizhMSFT (approved):
LGTM but IANAM

sajayantony (approved):

Issue #1131, opened 2023-09-26
Descriptors, the platform object, and the artifactType


neersighted:
In the aftermath of #999/the removal of the artifact mediaType, I believe that the guidance for descriptors has become misleading.

Artifacts are now represented using the standard image manifest and mediaType:

https://github.com/opencontainers/image-spec/blob/93f6e65855a1e046b42afbad0ad129a3d44dc971/manifest.md?plain=1#L23-L27

When we had a separate mediaType/artifact manifest, the following line in the descriptor specification was informative, as it implies that "images" (read: runnable content) should have a platform, but artifacts (arbitrary content) may not:

https://github.com/opencontainers/image-spec/blob/93f6e65855a1e046b42afbad0ad129a3d44dc971/descriptor.md?plain=1#L61

However, now that there is one manifest mediaType, this advice appears to be rather unhelpful; **all manifests** are now the same type, so this implies that all manifests SHOULD have a platform in their descriptor.

In practice, this has lead to implementations including/believing they should include `{"os": "unknown", "architecture": unknown"}` when describing non-image content (artifacts), which I believe to be misguided/incorrect, as values for these fields SHOULD be understood by the Go toolchain:

https://github.com/opencontainers/image-spec/blob/93f6e65855a1e046b42afbad0ad129a3d44dc971/image-index.md?plain=1#L57-L65

I'd like to drop this line from the descriptor spec, or amend it to acknowledge `artifactType`, and the `config.mediaType` fallback (collectively "artifacts"); but between these options I have no strong preference.

tianon:
At the risk of oversimplifying (please correct me if I've misunderstood!), you're proposing we loosen/tighten the language around `platform` to make sure it's clear it doesn't (have to) apply to non-image artifacts (even though they technically share the image `mediaType`)?

neersighted:
Yes, that's essentially my thought. When artifacts had a separate mediaType, they did not have a `SHOULD` for the platform; now that they share a mediaType, this applies.

I think the state before the separate mediaType/manifest definition was dropped made more sense, and that this line only ended up applying to artifacts by happenstance. I think usage of `platform` in the descriptor should be supported/encouraged where it makes sense, but requiring all artifacts to specify a "nonsense" platform (read: meaningless to the Go toolchain) doesn't make sense to me.

neersighted:
[...]

PR #1141, opened 2023-10-17
Artifacts clarification

neersighted:
- Closes https://github.com/opencontainers/image-spec/issues/1131
- Relates to https://github.com/opencontainers/image-spec/pull/1137
- Follow-up to #999

Attempt to better clarify the conceptual differences between artifacts and images (which I tie to the config as described in this spec), and provide guidance on how to detect artifacts from code.

Also attempt to clarify how descriptors interact with artifacts in the aftermath of #999.


artifacts-guidance.md
```diff
@@ -1,6 +1,29 @@
 # Guidance for Artifacts Authors
 
-Content other than OCI container images MAY be packaged using the image manifest.
-When this is done, the `config.mediaType` value should not be a known OCI image config [media type](media-types.md).
-Historically, due to registry limitations, some tools have created non-OCI conformant artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType` and values specific to the artifact in `layer[*].mediaType`.
+## Artifacts and Images
+
+This specification is primarily concerned with packaging two kinds of content: Artifacts, and Images. Both are representing using a [manifest](manifest.md).
```
tianon, line 5:
```suggestion
This specification is primarily concerned with packaging two kinds of content: Artifacts and Images.
Both are represented using a [manifest](manifest.md).
```

(minor grammatical fix + one sentence per line to keep Brandon happy)

artifacts-guidance.md
```diff
@@ -1,6 +1,29 @@
...
+Images are defined in this specification as conformant content with a [config](config.md), designed to be interpreted by a runtime that implements the [runtime-spec][]. 
+Conversely, an Artifact is any other conformant content that **does not contain a config to be interpreted by a runtime-spec implementation.**
```
tianon, line 7: As a runtime-spec maintainer also, I'm very confused by this block -- the config object holds the data used to construct a runtime-spec "bundle", but it is not parsed/specified as part of the runtime-spec.  The two are very loosely related. :sweat_smile:

artifacts-guidance.md
```diff
@@ -1,6 +1,29 @@
...
+
+## Creating an Artifact
+
+Content other than Images MAY be packaged using the [manifest]; this is otherwise known as an Artifact.
+When this is done, the `artifactType` should be set to a custom media type, or the `config.mediaType` should not be a known Image config [media type](media-types.md).
 Implementation details and examples are provided in the [image manifest specification](manifest.md#guidelines-for-artifact-usage).
+
+Note: Historically, due to registry limitations, some tools have created non-conformant Artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType`.
```
tianon, line 15: IMO we really can't get away from specifying that "images" have strict layer media type requirements too -- if a given runtime can't parse one of the layers, it can't reasonably assume it can (safely) do anything with the rest of that registry object (manifest, artifact, image, whatever you want to call it).

tianon (commented):
I haven't done a full review, but just noting some of my thoughts so far to hopefully spark more discussion.

artifacts-guidance.md
neersighted: Yeah, I'm not quite 100% happy with this either, but there are other parts of the spec that allude to the runtime spec, and "config that is meant to construct a runtime spec config base" still is a useful distinction, I believe.

artifacts-guidance.md
neersighted: I don't want to address that in this PR yet. As a follow-up to this, I want us to discuss rootfs vs non-rootfs layer types and how we can consistently decide which layers must be interpreted and which are "optional."

artifacts-guidance.md
jsturtevant: This also seems at contradiction to the statement below: `It is possible that software implementing the runtime-spec may also be able to interpret Artifacts; however that is outside the scope of this spec.`  Couldn't a runtime generate a runtime-spec based on something other than a classic image-config?  The loose coupling seems like this would make this possible.

I am thinking eventually we will want a WASM Artifact that doesn't use the image config format but still gets converted to a runtime-spec for runtimes that support the artifact type.

neersighted: It definitely could; does anyone have suggestions on how to describe interpretation of the classical config.md config without muddying the waters with runtime-spec references?

descriptor.md
```diff
@@ -54,11 +54,11 @@ The following fields contain the primary properties that constitute a Descriptor
...
+  This OPTIONAL property contains the type of an artifact when the descriptor points to an [artifact](artifacts-guidance.md).
+  This property MUST be the same as the `artifactType` of the referenced [manifest](manifest.md), or the `mediaType` of the `config` descriptor if an `artifactType` is not set.
   If defined, the value MUST comply with [RFC 6838][rfc6838], including the [naming requirements in its section 4.2][rfc6838-s4.2], and MAY be registered with [IANA][iana].
 
-Descriptors pointing to [`application/vnd.oci.image.manifest.v1+json`](manifest.md) SHOULD include the extended field `platform`, see [Image Index Property Descriptions](image-index.md#image-index-property-descriptions) for details.
+Descriptors referencing an image (as defined in the [artifacts guidance](artifacts-guidance.md)) SHOULD include the extended field `platform`, see [Image Index Property Descriptions](image-index.md#image-index-property-descriptions) for details. Artifacts as defined in the artifacts guidance MAY include the extended field `platform` when it improves discoverability or interoperability of the artifact.
```
sudo-bmitch, line 61:
```suggestion
Descriptors referencing an Image (as defined in the [artifacts guidance](artifacts-guidance.md)) SHOULD include the extended field `platform`.
See [Image Index Property Descriptions](image-index.md#image-index-property-descriptions) for details.
Artifacts as defined in the artifacts guidance MAY include the extended field `platform` when it improves discoverability or interoperability of the artifact. 
```

sudo-bmitch (changes requested):
I have a high level concern that we're trying to document the mixture of runnable Images and Artifacts in the same Index. Is that something we want to standardize in OCI, or would runtimes prefer that they are only asked to run an entry from an Index of only Images?

The concern I have is there's no limit to unique media types an Artifact may use, and there's likely a desire by new use cases to define new types of runnable content that may also have new media types that a current runtime would not have seen before. I'm wondering if we want to differentiate between a runtime unsure how to run the requested content vs a runtime not finding any Image content in an Index.

Feedback from @opencontainers/runtime-spec-maintainers is welcome.

artifacts-guidance.md
```diff
@@ -1,6 +1,29 @@
...
+
+## Interacting with Artifacts
+
+Runtimes implementing the [runtime-spec] SHOULD ignore unknown Artifacts when selecting content from an [index](image-index.md).
```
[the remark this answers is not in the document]
neersighted: This should be dependent on the `artifactType`, yes.

neersighted:
> I have a high level concern that we're trying to document the mixture of runnable Images and Artifacts in the same Index. Is that something we want to standardize in OCI, or would runtimes prefer that they are only asked to run an entry from an Index of only Images?

Yes, this is a realistic use-case with inline metadata today, e.g. what BuildKit does. 

> I'm wondering if we want to differentiate between a runtime unsure how to run the requested content vs a runtime not finding any Image content in an Index.

I think this is behavior defined by the runtime; what I want to make clear is that a runtime can interpret new artifact types if it has explicit support for them. But the default should be "ignore any content I do not know how to interpret" to obviate the need for `platform: unknown/unknown` and similar hacks with regard to platform selection.

descriptor.md
neersighted: I've avoided using the capital/definite Image here (restricting it to artifacts-guidance); do we think it's a good idea to leak the definite article into this file?

sudo-bmitch:
> > I have a high level concern that we're trying to document the mixture of runnable Images and Artifacts in the same Index. Is that something we want to standardize in OCI, or would runtimes prefer that they are only asked to run an entry from an Index of only Images?
> 
> Yes, this is a realistic use-case with inline metadata today, e.g. what BuildKit does.

I want to be careful that we don't say "buildkit implemented it, therefore OCI needs to recommend it". It's not unlike how cosign pushed manifests that looked like an OCI image with a different layer type to get around registries that filtered unknown config mediaType values. Even though the behavior works and doesn't violate the spec, I'm not incline to recommend that as a method to push artifacts.

Before I weigh in on the mixed content in an Index scenario, I'd like to hear from the impacted users that would be parsing that content (runtime maintainers).

artifacts-guidance.md
neersighted: I think that referring to `conversion.md` and the mechanism therein is the clarification I want.

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
 Implementation details and examples are provided in the [image manifest specification](manifest.md#guidelines-for-artifact-usage).
+
+Note: Historically, due to registry limitations, some tools have created non-conformant Artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType`.
+
+## Interacting with Artifacts
+
+Runtimes implementing the [runtime-spec][] and following the process described in [conversion.md](conversion.md) SHOULD ignore unknown Artifacts (as determined by the presence of a descriptor `artifactType`) when selecting content from an [index](image-index.md).
+It is possible that software implementing the runtime-spec may also be able to interpret Artifacts; however that is outside the scope of this spec.
+
+Artifacts can be detected at runtime using by checking two keys:
```
jonboulle, line 23:
```suggestion
Artifacts can be detected at runtime by checking two keys:
```

neersighted:
cc @dmcgowan re: runtime behavior with mixed indexes.

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
+Content other than Images MAY be packaged using the [manifest]; this is otherwise known as an Artifact.
+When this is done, the `artifactType` should be set to a custom media type, or the `config.mediaType` should not be a known Image config [media type](media-types.md).
 Implementation details and examples are provided in the [image manifest specification](manifest.md#guidelines-for-artifact-usage).
+
+Note: Historically, due to registry limitations, some tools have created non-conformant Artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType`.
+
+## Interacting with Artifacts
+
+Runtimes implementing the [runtime-spec][] and following the process described in [conversion.md](conversion.md) SHOULD ignore unknown Artifacts (as determined by the presence of a descriptor `artifactType`) when selecting content from an [index](image-index.md).
+It is possible that software implementing the runtime-spec may also be able to interpret Artifacts; however that is outside the scope of this spec.
```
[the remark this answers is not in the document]
neersighted: I'd say implementation-defined; I think that if we tried to standardize a way to interpret certain artifacts in the OCI spec itself it would be in conversion.md and the runtime-spec.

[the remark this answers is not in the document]
jsturtevant: This makes sense. I could see some type of guidance on what must/should get converted to the runtime spec.

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
 
-Content other than OCI container images MAY be packaged using the image manifest.
-When this is done, the `config.mediaType` value should not be a known OCI image config [media type](media-types.md).
-Historically, due to registry limitations, some tools have created non-OCI conformant artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType` and values specific to the artifact in `layer[*].mediaType`.
+## Artifacts and Images
+
+This specification is primarily concerned with packaging two kinds of content: Artifacts and Images.
+Both are representing using a [manifest](manifest.md).
+Images are defined in this specification as conformant content with a conformant [config](config.md), processed according to [conversion.md](conversion.md) to derive a [runtime-spec][] configuration blob.
+Conversely, an Artifact is any other conformant content that **does not contain a config to be interpreted by a runtime-spec implementation using the conversion mechanism.
```
jsturtevant, line 8:
```suggestion
Conversely, an Artifact is any other conformant content that **does** not contain a [config](config.md) to be interpreted by a runtime-spec implementation using the conversion mechanism.
```

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
+## Interacting with Artifacts
+
+Software following the process described in [conversion.md](conversion.md) to create a [runtime-spec][] configuration blob SHOULD ignore unknown Artifacts (as determined by the presence of a descriptor `artifactType`) when selecting content from an [index](image-index.md).
+It is possible that implementations may also be able to interpret known Artifact types; however that is outside the scope of this spec.
+
+Artifacts can be detected at runtime using by checking two keys:
+1. Is an `artifactType` present in the descriptor, or in the [manifest](manifest.md)?
+2. Is the `config.mediaType` of the manifest something other than a [known media type](media-types.md) for [config](config.md)?
+
+If either of these tests is true, then the content is an Artifact.
```
sajayantony, line 27: My understanding is that the wasm folks want to use the artifact type. It's runnable by the wasm runtime and not something that is aligned with the runtime spec. 
How can we support them by enabling the usage of `artifactType` ?  @cpuguy83 @tianon what do you folks think?

neersighted: That is covered by the case of "runtime has special knowledge of how to handle an artifact."

For the WASM-with-existing-containerd case, they should not set the artifact type as they will just be a "regular" image with special layers and a WASM platform object.

sajayantony: I’m probably at the risk of swimming against the current here but I honestly was hoping that the artifact type could be set by anyone and not limited to. 
if the runtime spec doesn’t want the artifact type set then it can mandate that but WASM runtimes don’t need to honor that.

neersighted: The issue is that I need a way to tell "not runnable" from "runnable" in containerd; I'd rather not have "an image" that is also "an artifact" at the same time as we get back to the `platform: none/none` pattern.

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
+
+Note: Historically, due to registry limitations, some tools have created non-conformant Artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType`.
+
+## Interacting with Artifacts
+
+Software following the process described in [conversion.md](conversion.md) to create a [runtime-spec][] configuration blob SHOULD ignore unknown Artifacts (as determined by the presence of a descriptor `artifactType`) when selecting content from an [index](image-index.md).
+It is possible that implementations may also be able to interpret known Artifact types; however that is outside the scope of this spec.
+
+Artifacts can be detected at runtime using by checking two keys:
+1. Is an `artifactType` present in the descriptor, or in the [manifest](manifest.md)?
```
mikebrow, line 24:
```suggestion
1. Is an {`artifactType`, _string_} present, e.g. as a `subject` reference in an image/artifact manifest or as a self reference definition in the [image artifact manifest `artifactType` field](https://github.com/opencontainers/image-spec/blob/main/manifest.md#image-manifest-property-descriptions)?
```

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
+2. Is the `config.mediaType` of the manifest something other than a [known media type](media-types.md) for [config](config.md)?
```
mikebrow, line 25: seems like that's 3 keys.?  

probably a good idea to explain here, not following the if the `config.mediaType` is unknown it's an artifact thought..

mikebrow (commented):
see comments...

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
+
+Content other than Images MAY be packaged using the [manifest]; this is otherwise known as an Artifact.
+When this is done, the `artifactType` should be set to a custom media type, or the `config.mediaType` should not be a known Image config [media type](media-types.md).
 Implementation details and examples are provided in the [image manifest specification](manifest.md#guidelines-for-artifact-usage).
+
+Note: Historically, due to registry limitations, some tools have created non-conformant Artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType`.
+
+## Interacting with Artifacts
+
+Software following the process described in [conversion.md](conversion.md) to create a [runtime-spec][] configuration blob SHOULD ignore unknown Artifacts (as determined by the presence of a descriptor `artifactType`) when selecting content from an [index](image-index.md).
```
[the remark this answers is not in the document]
cpuguy83: This kind of makes sense, but also today no runtime ignores unknown artifact types because they aren't aware of the field since it is completely new.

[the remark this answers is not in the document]
neersighted: Yeah, this is somewhat of a new behavior, but at the same time is just an attempt to suggest what seems the only sane default. The idea here is that tools like BuildKit could stop producing `platform: unknown/unknown` to prevent their not-images from being considered for platform selection by simply switching to artifacts instead.

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
 # Guidance for Artifacts Authors
 
-Content other than OCI container images MAY be packaged using the image manifest.
-When this is done, the `config.mediaType` value should not be a known OCI image config [media type](media-types.md).
-Historically, due to registry limitations, some tools have created non-OCI conformant artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType` and values specific to the artifact in `layer[*].mediaType`.
+## Artifacts and Images
+
+This specification is primarily concerned with packaging two kinds of content: Artifacts and Images.
+Both are representing using a [manifest](manifest.md).
+Images are defined in this specification as conformant content with a conformant [config](config.md), processed according to [conversion.md](conversion.md) to derive a [runtime-spec][] configuration blob.
```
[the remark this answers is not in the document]
neersighted: It's being defined in contrast to Artifacts for now... I think it's okay; this isn't the best place to have this long-term, but for now this seems like a relevant place to try and explain what the difference between an image and an artifact is.

artifacts-guidance.md
neersighted: cc @jsturtevant; this is meant to enable (through clarity of what runtimes likely SHOULD support) efforts to make "native WASM images" work with "existing runtimes;" I see no reason that something that is otherwise a conformant image (and just uses a `wasm` platform) needs to set `artifactType`.

This is also somewhat related to https://github.com/opencontainers/image-spec/pull/1141#discussion_r1371576029, which I still want to keep as a follow-up as the scope of this PR has grown quite a bit.

artifacts-guidance.md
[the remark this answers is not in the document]
cpuguy83: The issue here is not specifying a garbage platform will cause older implementations to blow up.
I'm not sure what that means for the wording here.
Putting anything in an index that is not an image manifest is not backwards compatible, I guess.

artifacts-guidance.md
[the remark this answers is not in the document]
mikebrow: fair.. 

```suggestion
Image with containing conformant content and with conformant [config](config.md) are processed according to [conversion.md](conversion.md) to derive a [runtime-spec][] configuration blob.
```

artifacts-guidance.md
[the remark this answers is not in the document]
neersighted: Most tools should ignore artifacts already (but not as eagerly/not as codified as I'd like) by virtue of e.g. the config mediaType or the layer types, thankfully. See https://github.com/opencontainers/image-spec/issues/1131#issuecomment-1738134251 for some research into this (though I could definitely do more testing with existing implementations and mixed indexes).

sudo-bmitch:
Given the various scenarios we're seeing, and knowing there will be more scenarios in the future, perhaps instead of being prescriptive in exactly what fields are defined and we should keep it at a higher level. Something like "container images are any content pushed to the registry designed to be executed by a container runtime, artifacts are everything else."

There will be artifacts that want to have a valid platform defined, and other artifacts that want to use the image config media type. We may be able to avoid the latter with guidance in the spec, but that value isn't visible in the Index. And expecting runtimes to avoid trying to run an artifact image based on a new field in the descriptor (i.e. `artifactType`) is going to be error prone for existing runtimes that aren't aware of new fields.

The two options I can think of are either to ensure artifacts are listed after any container images in an index, so that the selector for container runtimes selects the first match, or OCI recommends against mixing container images and artifacts in the same index for the best support of older runtimes.

artifacts-guidance.md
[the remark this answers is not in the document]
cpuguy83: I guess in this case the artifact type should be listed below valid runtime image manifests?

mikebrow:
> Something like "container images are any content pushed to the registry designed to be executed by a container runtime, artifacts are everything else."

IMO it is not a good idea to redefine what a container image is (restrict it's definition to only that which is executed by a container runtime) inside the artifacts guidance documentation.. and we support data in images, so what does execute mean, for processing?  Why would we prohibit container runtimes, and their fat clients, from executing/processing artifacts?

sudo-bmitch:
>  Why would we prohibit container runtimes, and their fat clients, from executing/processing artifacts?

My phrasing is that container runtimes trying to execute/process it makes it a container image. It's not restricting what runtimes can do, it's defining whether it's a container image or artifact based on whether runtimes could execute/process it. It leaves flexibility for runtimes to extend to process more content in the future without us needing to adjust the image-spec for every runtime change.

mikebrow:
> Given the various scenarios we're seeing, and knowing there will be more scenarios in the future, perhaps instead of being prescriptive in exactly what fields are defined and we should keep it at a higher level. Something like "container images are any content pushed to the registry designed to be executed by a container runtime, artifacts are everything else."
> 
> There will be artifacts that want to have a valid platform defined, and other artifacts that want to use the image config media type. We may be able to avoid the latter with guidance in the spec, but that value isn't visible in the Index. And expecting runtimes to avoid trying to run an artifact image based on a new field in the descriptor (i.e. `artifactType`) is going to be error prone for existing runtimes that aren't aware of new fields.
> 
> The two options I can think of are either to ensure artifacts are listed after any container images in an index, so that the selector for container runtimes selects the first match, or OCI recommends against mixing container images and artifacts in the same index for the best support of older runtimes.

IMO it is not a good idea to redefine (is/are language) what a container image is inside the artifacts guidance documentation.

Let's not leave out the index may or may not include platform filters, and "implementations SHOULD support" "application/vnd.oci.image.index.v1+json (nested index)" and "an encountered mediaType that is unknown to the implementation MUST NOT generate an error."

mikebrow:
> My phrasing is that container runtimes trying to execute/process it makes it a container image. It's not restricting what runtimes can do, it's defining whether it's a container image or artifact based on whether runtimes could execute/process it. It leaves flexibility for runtimes to extend to process more content in the future without us needing to adjust the image-spec for every runtime change.

I think it's an oci container image if it's in a format conforming to the image spec for storage or transport, and then at runtime, by conversion and augmentation via a container runtime, if it is determined by a runc compatible runtime to then adhere to the runtime spec config and rootfs requirements of a runtime bundle for a platform.

mikebrow:
artifacts are a grey space wrt. whether they can be considered a part of or augmentation to layers/config/index, guidance, filtering mechanisms, additional image config, even possibly a key store, alg, or content for a new digest type... runnable or not is probably up to the iana type or we possibly break with existing spec restrictions. 

100% agree we need to avoid a client reading an artifact, via matching, as a default platform image if/when that is not the intended result.

mikebrow:
I like the high level definition here: https://github.com/opencontainers/image-spec/blob/main/spec.md#overview

> At a high level the image manifest contains metadata about the contents and dependencies of the image including the content-addressable identity of one or more [filesystem layer changeset](https://github.com/opencontainers/image-spec/blob/main/layer.md) archives that will be unpacked to make up the final runnable filesystem. The image configuration includes information such as application arguments, environments, etc. The image index is a higher-level manifest which points to a list of manifests and descriptors. Typically, these manifests may provide different implementations of the image, possibly varying by platform or other attributes.

sudo-bmitch:
> Let's not leave out the index may or may not include platform filters, and "implementations SHOULD support" "application/vnd.oci.image.index.v1+json (nested index)" and "an encountered mediaType that is unknown to the implementation MUST NOT generate an error."

An issue attempting to be solved in this PR is runtimes picking container images from a mixed index with artifacts included in the manifest list. The artifact entries could have matching platform values, and known media type values. My comments above are looking at what options, if any, we have to resolve that. I don't believe "MUST NOT generate an error" helps in this scenario since the values would be known.

mikebrow:
> An issue attempting to be solved in this PR is runtimes picking container images from a mixed index with artifacts included in the manifest list. The artifact entries could have matching platform values, and known media type values. My comments above are looking at what options, if any, we have to resolve that. I don't believe "MUST NOT generate an error" helps in this scenario since the values would be known.

absolutely... 100% agree we need to avoid a client reading an artifact, via matching, as a default platform image if/when that is not the intended result.

my point in including the additional cites was as a reminder:
1) that said artifacts might be in a nested index vs.. the platform index..
2) if the manifest type is unknown it mush not generate an error...

and those current rules may be useful in finding a resolution..

For example, a) push said artifacts into an index under the platform index... (and check if clients actually recurse today for indexes that do not have a platform filter, thus suggesting platform index is for container image manifests not artifact manifest irregardless of the desire to filter artifacts by platform
b) add a new manifest type for an artifact index

artifacts-guidance.md
neersighted: It's certainly possible; as discussed on the last PR having the standard config mediaType with a custom artifactType is rather outside what most would consider idiomatic, however.

The `platform` field on the descriptor should be the same as the platform in the `config` blob, so the platform of the image should be accessible via the same mechanism as the top-level `artifactType`.

Additionally, I do think that an annotation with e.g. "WASM image standard version" or similar would be perfectly idiomatic if you need some additional metadata.

WDYT?

artifacts-guidance.md
rchincha: Now that image-spec and dist-spec officially have "artifact" support, perhaps revisit this? converging to config.mediaType=empty and basically only do 1. above?

mikebrow: Sure I see know what @neersighted meant.

Proposed text:

An oci image manifest can be identified as an artifact of some type other than an specified oci container image by 1) first checking if the artifact type value is set and if so whether it is valid by RFC6838, 2) otherwise checking if the config mediaType value is not one of the known OCI Image Types and is valid by RFC6838.


Proposed pseudo explanation:

func checkForArtifactType(manifest) artifactType, error 
{
manifestArtifactTypeMustBeSet = UNKNOWN
manifestArtifactTypeIsSet = UNKNOWN
manifestArtifactTypeMustBeValid = UNKNOWN
manifestArtifactTypeIsValid = UNKNOWN
configMediaTypeIsOCIImageType = UNKNOWN
configMediaTypeIsValidArtifact = UNKNOWN

if manifest.config.mediaType is "application/vnd.oci.empty.v1+json" { 
   manifestArtifactTypeMustBeSet = TRUE;
}

manifestArtifactTypeIsSet = is manifest.artifactType unset or nil? 

if manifestArtifactTypeIsSet {
   manifestArtifactTypeMustBeValid = TRUE;
}

if manifestArtifactTypeMustBeValid {
   manifestArtifactTypeIsValid = checkRFC6838(manifest.artifactType);
   if !manifestArtifactTypeIsValid {
      return with error(manifest.artifactType + "is not valid")
   }
   return manifest.artifactType // we have the artifact type we were looking for
}

configMediaTypeIsOCIImageType = isKnownOCIImageConfigMediaType(manifest.config.mediaType)

if configMediaTypeIsOCIImageType { 
   return image .. no artifact found
}

// not "oci.empty" and not an oci image media type
configMediaTypeIsValidArtifact = checkRFC6838(manifest.config.mediaType);

if !configMediaTypeIsValidArtifact {
   return with error(manifest.config.mediaType + "is not valid") 
}

return manifest.config.mediaType // we have the artifact config media type we were looking for
} // end checkForArtifactType()

func checkRFC6838 (type string) BOOL 
{
  // returns true if not nil and complies with [RFC 6838](https://tools.ietf.org/html/rfc6838), including the [naming requirements in its section 4.2](https://tools.ietf.org/html/rfc6838#section-4.2)
}

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
+Content other than Images MAY be packaged using the [manifest]; this is otherwise known as an Artifact.
+When this is done, the `artifactType` should be set to a custom media type, or the `config.mediaType` should not be a known Image config [media type](media-types.md).
 Implementation details and examples are provided in the [image manifest specification](manifest.md#guidelines-for-artifact-usage).
+
+Note: Historically, due to registry limitations, some tools have created non-conformant Artifacts using the `application/vnd.oci.image.config.v1+json` value for `config.mediaType`.
+
+## Interacting with Artifacts
+
+Software following the process described in [conversion.md](conversion.md) to create a [runtime-spec][] configuration blob SHOULD ignore unknown Artifacts (as determined by the presence of a descriptor `artifactType`) when selecting content from an [index](image-index.md).
+It is possible that implementations may also be able to interpret known Artifact types; however that is outside the scope of this spec.
```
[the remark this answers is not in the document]
rchincha: Can the phrase "known Artifact" be clarified here? 
Does it mean artifactType field exists or not?
Does it mean artifactType field exists but uses a well-known media-type?
Does it mean well-known media-type to outside world but not to this spec?
etc ... etc ...

artifacts-guidance.md
rchincha: This would be a nice subroutine to add to this repo - likely to be heavily used going forward. 

1. Runnable container images
2. Non-runnable not-container images (referrer or otherwise)
3. Runnable non-container images

^ if I were to summarize.

rchincha: Just a thought ... OCI = packaging + distribution + "bring your own runtime (BYOR)"?

artifacts-guidance.md
```diff
@@ -1,6 +1,30 @@
...
+## Artifacts and Images
+
+This specification is primarily concerned with packaging two kinds of content: Artifacts and Images.
+Both are representing using a [manifest](manifest.md).
+Images are defined in this specification as conformant content with a conformant [config](config.md), processed according to [conversion.md](conversion.md) to derive a [runtime-spec][] configuration blob.
+Conversely, an Artifact is any other conformant content that **does not contain a config to be interpreted by a runtime-spec implementation using the conversion mechanism.
+
+## Creating an Artifact
+
+Content other than Images MAY be packaged using the [manifest]; this is otherwise known as an Artifact.
```
sudo-bmitch, line 12:
```suggestion
Content other than Images MAY be packaged using the [manifest](manifest.md); this is otherwise known as an Artifact.
```

PR #1172, opened 2024-02-27
Reformat "Platform Variants", especially to add amd64, ppc64le, riscv64

tianon:
This changes the table to be a line per ISA instead of per variant; adds x86-64, POWER8+, and RISC-V; and notes the appropriate "Go analog" in each case (because these values mirror Go's choices, like our use of `GOOS` and `GOARCH`).

Of particular note, `GOARM64` and `GORISCV64` are not fully implemented or released in Go yet, but they are accepted proposals with in-progress implementations or already merged to main:

- `GOARM64`: https://github.com/golang/go/issues/60905
- `GORISCV64`: https://github.com/golang/go/issues/61476#issuecomment-1791156741


image-index.md
```diff
@@ -105,13 +105,15 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
...
 
 When the variant of the CPU is not listed in the table, values are implementation-defined and SHOULD be submitted to this specification for standardization.
-
-| ISA/ABI         | `architecture` | `variant`   |
-|-----------------|----------------|-------------|
-| ARM 32-bit, v6  | `arm`          | `v6`        |
-| ARM 32-bit, v7  | `arm`          | `v7`        |
-| ARM 32-bit, v8  | `arm`          | `v8`        |
-| ARM 64-bit, v8  | `arm64`        | `v8`        |
+These values SHOULD match (or be similar to) their analog listed in [the Go Language document][go-environment2].
```
tianon, line 108: > (or be similar to)

This part makes me nuts, but I felt like I had to somehow address that `GOARM` does not use a `v` prefix, `GOARM64` will not (at least initially) support `v8`, we don't do extra things like `softfloat` or `sse2` in _this_ field, etc SOMEHOW in this wording.

jonjohnsonjr (approved):

image-index.md
xnox: Hm, i would be more explicit in tags that are supported. One can reference go constants as valid ones, but specify them explicitly.

There will always be variants that are not yet specified or relevant for go, but do make sense to be explicit for container images that contain code that was written in other languages. For example, whilst GOARM64 is already defined, it doesn't actually do anything different in go - but it does for gcc for example.

tianon: The table below does include explicit examples, though :sweat_smile:

tianon:
Putting these links here because I keep wanting them and having to find them again (will maybe update this over time with more interesting/useful links):
- https://ubuntu.com/blog/optimising-ubuntu-performance-on-amd64-architecture
- https://developers.redhat.com/blog/2021/01/05/building-red-hat-enterprise-linux-9-for-the-x86-64-v2-microarchitecture-level
- https://gitlab.archlinux.org/archlinux/rfcs/-/blob/master/rfcs/0002-march.rst
- https://fedoraproject.org/wiki/Changes/x86-64_micro-architecture_update

tianon:
In case it's helpful to someone else, I'll note here that I've copied the current `debian:trixie-slim` to `tianon/test:variants`, filtered the list to those with an appropriate `variant` value that _could_ be set, and added the appropriate and explicit `variant` values in both the config and index so folks can test runtime behavior:

- https://oci.dag.dev/?image=tianon/test:variants
- https://oci.dag.dev/?image=tianon/test@sha256:fa25b6cfb7923a7c4fdd439b814afc0f7cc97186aa8d1573bd02ffa8b18f713d (`tianon/test@sha256:fa25b6cfb7923a7c4fdd439b814afc0f7cc97186aa8d1573bd02ffa8b18f713d`)

Each architecture _also_ exists as a single-architecture tagged image manifest so runtime edge case behavior can be more easily fully tested:

- `tianon/test:variant-amd64`
- `tianon/test:variant-arm64v8`
- `tianon/test:variant-arm32v7`
- `tianon/test:variant-arm32v5`
- `tianon/test:variant-ppc64le`
- `tianon/test:variant-riscv64`

PR #695, opened 2017-06-09
spec: describe components of EBNF grammar

stevvooe:
To clarify the grammar definitions, we define the subset of EBNF used by
this specification to specify various field formats.

Signed-off-by: Stephen J Day <stephen.day@docker.com>

Closes #692

PR #1196, opened 2024-08-08
Remove misleading "must" in `ref.name` requirements

tianon:
~~See b692deece6f3e93d43de1615fa592b86f0f27c34 (https://github.com/opencontainers/image-spec/pull/695) for where this was originally added (making clear this was the original intent).~~

~~IMO it _slightly_ conflicts with the bullet point above it being simply SHOULD, but it has been part of the spec since v1.0 (and in a disagreement between MUST and SHOULD, the stronger constraint wins and that's MUST).~~

Updated: https://github.com/opencontainers/image-spec/pull/1196#pullrequestreview-2243187911

### DIFF
--- a/annotations.md
+++ b/annotations.md
@@ -33,7 +33,7 @@ This specification defines the following annotation keys, intended for but not l
 - **org.opencontainers.image.ref.name** Name of the reference for a target (string).
   - SHOULD only be considered valid when on descriptors on `index.json` within [image layout](image-layout.md).
   - Character set of the value SHOULD conform to alphanum of `A-Za-z0-9` and separator set of `-._:@/+`
-  - The reference must match the following [grammar](considerations.md#ebnf):
+  - A valid reference matches the following [grammar](considerations.md#ebnf):
 
     ```ebnf
     ref       ::= component ("/" component)*


tianon:
Perhaps this _should_ be SHOULD instead to match the earlier requirement?  I'm happy to update.

tianon:
(however, to be explicit, I do think using "must" here is a bug and we should fix it to either MUST or SHOULD so the intent is 100% clear and no longer ambiguous -- MUST is more consistent with the language used, but SHOULD is more consistent with the rest of the annotation's stated limitations)

sudo-bmitch (commented):
This makes me really want to define a reference somewhere in OCI, and then have this as a pointer to that definition. Given that the previous line says "SHOULD", I think converting this to a hard MUST is incompatible with the previous line and could break anyone that is treating themselves as an exception to the SHOULD.

I'd lean towards removing "must" completely, and just say "A valid reference matches the following grammar". That would straddle the "here's what you should do" with the hint of "don't be surprised if there are implementations doing invalid stuff". I'd also accept changing this from "must" to "SHOULD".

tianon:
Sure, no strong argument here -- updated!

tianon:
(CI failure is unrelated :upside_down_face:)

hassanrawan887:
ما هاذا
rawanhassan

في الاثنين، ١٩ أغسطس ٢٠٢٤، ١١:٥٧ م Tianon Gravi ***@***.***>
كتب:

sudo-bmitch (approved):
LGTM

sajayantony (approved):

Gabomerino:
> ~~See b692deece6f3e93d43de1615fa592b86f0f27c34 (😕😕🙁

PR #1238, opened 2025-01-23
Fix grammar nit

mkenigs:
Total nit but thought I'd tweak cause I noticed it

`This field` reads more naturally than `The field`, and it's consistent with the use of `This field` and `This property` throughout the rest of the document


tianon (approved):
Thanks!

sudo-bmitch (approved):
LGTM

PR #1290, opened 2025-10-16
Update OCI Image Implementations list

BobDu:
Updated the list of projects adopting the OCI Image Specification by adding 'containers/skopeo' and 'oras-project/rust-oci-client', and removed 'projectatomic/skopeo' and 'krustlet/oci-distribution'. Also noted the end of 'box-builder/box'.

### DIFF
--- a/implementations.md
+++ b/implementations.md
@@ -2,7 +2,6 @@
 
 Projects or Companies currently adopting the OCI Image Specification
 
-- [projectatomic/skopeo](https://github.com/projectatomic/skopeo)
 - [Amazon Elastic Container Registry (ECR)](https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-manifest-formats.html) ([announcement](https://aws.amazon.com/about-aws/whats-new/2017/01/amazon-ecr-supports-docker-image-manifest-v2-schema-2/))
 - [Azure Container Registry (ACR)](https://docs.microsoft.com/azure/container-registry/container-registry-image-formats#oci-images)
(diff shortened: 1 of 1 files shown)


tianon (approved):

implementations.md
```diff
@@ -17,18 +16,19 @@ Projects or Companies currently adopting the OCI Image Specification
...
   - [containers/libocispec](https://github.com/containers/libocispec)
-- [krustlet/oci-distribution](https://github.com/krustlet/oci-distribution)
-- [box-builder/box](https://github.com/box-builder/box)
+  - [containers/skopeo](https://github.com/containers/skopeo)
 - [coolljt0725/docker2oci](https://github.com/coolljt0725/docker2oci)
 - [regclient/regclient](https://github.com/regclient/regclient)
 - [ORAS](https://oras.land/)
   - [oras-project/oras](https://github.com/oras-project/oras/)
   - [oras-project/oras-go](https://github.com/oras-project/oras-go)
+  - [oras-project/rust-oci-client](https://github.com/oras-project/rust-oci-client)
```
tianon: I was worried about this one a little bit because the description claims distribution spec, but the image spec representation is in there, particularly in `manifest.rs`:

https://github.com/oras-project/rust-oci-client/blob/e53edd48afcd90f7ecdb25420faa065ba6004f98/src/manifest.rs

13helloyello13-stack: [whip-output.cpp](https://github.com/user-attachments/files/24610061/whip-output.cpp)

PR #1305, opened 2026-02-03
Document the supported platforms

ProbstDJakob:
This PR adds a table of supported platforms in order to be independent of the go library and directly list officially, by the spec, supported platforms. The platforms are extracted out of the go documentation and merged with the current table (`arm64/v8.1` is not listed in go).

### DIFF
--- a/image-index.md
+++ b/image-index.md
@@ -57,12 +57,12 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
     - **`architecture`** *string*
 
       This REQUIRED property specifies the CPU architecture.
-      Image indexes SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOARCH`][go-environment2].
+      Image indexes SHOULD use, and implementations SHOULD understand, `architecture` values listed in the [Platforms section](#platforms) table.
 
     - **`os`** *string*
 
       This REQUIRED property specifies the operating system.
-      Image indexes SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOOS`][go-environment2].
+      Image indexes SHOULD use, and implementations SHOULD understand, `os` values listed in the [Platforms sections](#platforms) table.
 
     - **`os.version`** *string*
 
@@ -82,7 +82,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
     - **`variant`** *string*
 
       This OPTIONAL property specifies the variant of the CPU.
-      Image indexes SHOULD use, and implementations SHOULD understand, `variant` values listed in the [Platform Variants](#platform-variants) table.
+      Image indexes SHOULD use, and implementations SHOULD understand, `variant` values listed in the [Platforms section](#platforms) table.
 
     - **`features`** *array of strings*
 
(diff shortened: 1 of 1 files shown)


image-index.md
```diff
@@ -102,18 +102,28 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
...
 These values SHOULD match (or be similar to) their analog listed in [the Go Language document][go-environment2].
-
-| ISA/ABI    | `architecture` | `variant`             | Go analog   |
-|------------|----------------|-----------------------|-------------|
-| ARM 32-bit | `arm`          | `v6`, `v7`, `v8`      | `GOARM`     |
-| ARM 64-bit | `arm64`        | `v8`, `v8.1`, …       | `GOARM64`   |
-| POWER8+    | `ppc64le`      | `power8`, `power9`, … | `GOPPC64`   |
-| RISC-V     | `riscv64`      | `rva20u64`, …         | `GORISCV64` |
-| x86-64     | `amd64`        | `v1`, `v2`, `v3`, …   | `GOAMD64`   |
+Defaults are subject to change at any time.
```
ProbstDJakob: Not sure if they should really be subject to change at any time, but that would be the case when relying on go and they would change the default.

sudo-bmitch (changes requested):
I'm a nack on this with several high level concerns:

1. "Supported" is a strong word that different people could interpret in different ways. Would every implementation need to support everything listed in the spec? Would implementations that generate other values no longer be valid according to the spec?
2. When possible, several of us prefer to delegate to external specifications. Usually that's an RFC, but in this case we chose to delegate to the Go language. That means every time a new architecture or variant comes along, we don't need to update our spec, and there isn't a risk of the our spec getting out of sync with an upstream source.
3. I'm not sure what "default" means in this context.
4. People use the spec for more than just container images (see the docs on packaging artifacts). So I don't think it's safe to limit the list to what popular container runtimes support today.

ProbstDJakob:
1. I personally would prefer a spec which states what is officially supported and the rest would be implementation specific so that there not one implementation which only accepts `i386` and another which only accepts `386`, but that is definitely not the place for such an argument so I've adapted the section to follow better the rest of the document.
2. I think an RFC is completely fine to link, but documentation is subject to change at any time. For any implementation not written in go they must closely watch the docs to get not out of sync as especially the default variants could change at any time. Speaking of default variants, if go ever decides to change the default variant for example for `amd64` implementations which follow the go documentation closely will use the new variant, e.g. `v3`, and other implementations which still depend on the old variant `v1` would then try to pull images not supported by their platform. Yes, you currently do not explicitly state what variant should be assumed if non is supplied (but also do not say that it is implementation specific), but I guess most implementation now use whatever is documented in go.
3. I've updated the text to be more clear.
4. Ok fair, applications like `oras` could benefit from every platform combination (though I am not sure if it currently does).

ProbstDJakob:
> A runtime that only supports a single platform would still be OCI conformant. In fact, without things like binfmt-misc a lot of tooling only supports a single platform.

I think you've misunderstood me, I do not mean that every implementations must/should implement all listed platforms, but when they support one platform listed in the table, then I would prefer that they must use the platform as specified in the table. So that not `docker` pushes images with `linux/386` and `podman` tries to pull images with `linux/i386`. This would fail as the image index does not have this platform though it actually contains it just with another name. But again, this is not what I want to "solve" with this PR.

> Go has a 1.0 guarantee that would prevent them from changing something like that. They can add new architectures, and deprecate architectures they no longer support. [...]

The [Go 1 guarantees do not include the tooling](https://go.dev/doc/go1compat#tools) and explicitly states that they may change behaviour and may break scripts depending on properties of the tools. Regarding addition and removal of supported architectures by Go, if one implements an application before the addition or after the removal by Go the application maintainers are not bound to a specific name thus may run into the same problem as described above. So for example Go removes support for `386` and a new OCI conformant application is created with support for `386` they may name it as `x86` instead.
Furthermore Go currently supports only a minimal subset of what Linux (and possible other OSs) supports. Linux has support for [13 more groups of architecture](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch?id=b20624608f350c5dadd74577629e90715d351e2c) than Go. Docker (or more precisely `containerd` in version 1.4.3 to which [the docker docs link to](https://docs.docker.com/reference/cli/docker/buildx/build/#platform)) has [support for more](https://github.com/containerd/containerd/blob/v1.4.3/platforms/database.go) than what Go supports, but still not everything Linux supports (and that is fine as stated above).
Moreover the linked Go documentation is neither a URN nor a URL to a versioned spec and thus is prone to change. Event though Go has some guarantees about version 1, the history has shown that guarantees or promises may still change (though I hope this will not happen with Go's guarantees and Go still has versioned specs like the memory model or the language and std lib itself). Versioned specs (e.g. this one or the RFC Series (ISSN 2070-1721)) on the other hand are immutable once released. Of course, one could change an already released spec, but they usually are distributed and thus harder to change than a single website (e.g. this spec has 808 Forks with possible more unknown to GitHub).

> [...] tooling should not break if it ever encounters [new architectures and deprecated architectures]. So tooling that is still aware of ppc64 after Go deprecates it would still be valid. And if tooling adds a platform before Go adds it to their documentation, that shouldn't be a breaking change, applications would just skip it if they are unfamiliar. Most Docker images today ship with `unknown/unknown` platforms in their index, which isn't part of the spec, but get ignored as unknown platforms.

Definitely, same as with the media type unknown platforms should not break the application and just be skipped for example, but previously known platforms should not be implicitly removed from the spec and instead be kept at least until a breaking change. Otherwise something as written above with `linux/386` and `linux/i386` or as written a bit later with `386` and `x86` could happen.

> I've had luck with both oras and regclient/regctl.

Just curious, what do you mean with "you had luck"?

---

Regarding Linux and `containerd` supporting more architectures, this table still misses a lot of platforms (not only architectures) which probably should be added either in this PR or another one, if this PR gets accepted. Also in case it gets accepted, maybe a JSON (and possibly a `.go` file) should be added containing the known platforms.

Speaking of known, wouldn't _Known Platforms_ be a good name for the section instead of simply _Platforms_?

image-index.md
```diff
@@ -57,12 +57,12 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
     - **`architecture`** *string*
 
       This REQUIRED property specifies the CPU architecture.
-      Image indexes SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOARCH`][go-environment2].
+      Image indexes SHOULD use, and implementations SHOULD understand, `architecture` values listed in the [Platforms section](#platforms) table.
 
     - **`os`** *string*
 
       This REQUIRED property specifies the operating system.
-      Image indexes SHOULD use, and implementations SHOULD understand, values listed in the Go Language document for [`GOOS`][go-environment2].
```
chetchewbetcha, line 65: Crooks

sudo-bmitch:
I'm going to close this since there don't appear to be any maintainers interested in approving it.

PR #1314, opened 2026-04-16
Clarify that whiteout filenames require a non-empty basename

### LINKED ISSUE #1301: clarify what a `.wh.` file does
Please clarify the behavior of a whiteout file being applied to an empty dirname and basename.

For example, if I create a layer with the following content:

```
.wh.
etc/config.txt
```

Is this the expected behavior?
* wipes all content from previous layers
* adds file etc/config.txt

mahitha-ada:
## Description

Fixes #1301.

Clarifies in the whiteout specification that a `.wh.` prefix MUST be
followed by a non-empty basename. The current spec does not explicitly
address the edge case of a bare `.wh.` entry (with no basename), which
leads to inconsistent behavior across implementations — for example,
containerd rejects it with "invalid whiteout name: .wh." while other
runtimes may handle it unpredictably.

This change:
- Adds an explicit requirement that the basename must be non-empty
- Specifies that implementations MUST error on bare `.wh.` entries
- Points readers to opaque whiteouts for directory-level deletion

Signed-off-by: Mahitha Adapa <mahitha.ada@gmail.com>

### DIFF
--- a/layer.md
+++ b/layer.md
@@ -151,7 +151,7 @@ Create a new directory and initialize it with a copy or snapshot of the prior ro
 Example commands that can preserve [file attributes](#file-attributes) to make this copy are:
 
 - [cp(1)](https://linux.die.net/man/1/cp): `cp -a rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
-- [rsync(1)](https://linux.die.net/man/1/rsync):  `rsync -aHAX rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
+- [rsync(1)](https://linux.die.net/man/1/rsync): `rsync -aHAX rootfs-c9d-v1/ rootfs-c9d-v1.s1/`
 - [tar(1)](https://linux.die.net/man/1/tar): `mkdir rootfs-c9d-v1.s1 && tar --acls --xattrs -C rootfs-c9d-v1/ -c . | tar -C rootfs-c9d-v1.s1/ --acls --xattrs -x` (including `--selinux` where supported)
 
 Any [changes](#change-types) to the snapshot MUST NOT change or affect the directory it was copied from.
@@ -159,7 +159,7 @@ Any [changes](#change-types) to the snapshot MUST NOT change or affect the direc
 For example `rootfs-c9d-v1.s1` is an identical snapshot of `rootfs-c9d-v1`.
 In this way `rootfs-c9d-v1.s1` is prepared for updates and alterations.
 
-**Implementor's Note**: *a copy-on-write or union filesystem can efficiently make directory snapshots*
+**Implementor's Note**: _a copy-on-write or union filesystem can efficiently make directory snapshots_
 
 Initial layout of the snapshot:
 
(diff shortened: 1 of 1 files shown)


layer.md
```diff
@@ -100,7 +100,7 @@ With this approach, the link map and links names of a directory could be compare
 Implementations on Windows MUST support these additional attributes, encoded in [PAX vendor
 extensions](https://github.com/libarchive/libarchive/wiki/ManPageTar5#pax-interchange-format) as follows:
 
-- [Windows file attributes](https://msdn.microsoft.com/en-us/library/windows/desktop/gg258117(v=vs.85).aspx) (`MSWINDOWS.fileattr`)
+- [Windows file attributes](<https://msdn.microsoft.com/en-us/library/windows/desktop/gg258117(v=vs.85).aspx>) (`MSWINDOWS.fileattr`)
```
sudo-bmitch, line 103: Is this change needed?

layer.md
```diff
@@ -246,6 +246,7 @@ In all other cases, the implementation MUST do the semantic equivalent of the fo
 
 - A whiteout file is an empty file with a special filename that signifies a path should be deleted.
 - A whiteout filename consists of the prefix `.wh.` plus the basename of the path to be deleted.
+- A whiteout filename MUST include a non-empty basename after the `.wh.` prefix. A bare `.wh.` entry with no basename is invalid and implementations MUST return an error when encountering such an entry. To remove all children of a directory, use an opaque whiteout (`.wh..wh..opq`) as described below.
```
sudo-bmitch:
```suggestion
- A `.wh.` file, without a basename to delete, is invalid and implementations SHOULD return an error when encountering such an entry.
```

I think the language should be softened to a SHOULD since not every implementation is a container runtime, e.g. a scanner looking for leaked credentials in the layer should not be blocked by a maliciously crafted layer. This can also be simplified to the one sentence.

sudo-bmitch (changes requested):

layer.md
mahitha-ada: not needed

layer.md
mahitha-ada: Makes sense, accepted the suggestion.

layer.md
sudo-bmitch:
```suggestion
- [Windows file attributes](https://msdn.microsoft.com/en-us/library/windows/desktop/gg258117(v=vs.85).aspx) (`MSWINDOWS.fileattr`)
```

In that case, lets revert it for consistency.

sudo-bmitch:
Thanks for the updates @mahitha-ada . Please [squash the commits](https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together) on this.

sudo-bmitch (approved):
LGTM.

prince200212 (approved):

tianon (approved):

leoslay623-rgb (commented):
[porn spam]

lindaburton0987-source (approved):

Issue #350, opened 2016-09-29
Why is `size` defined as an int64 for OCI content descriptor?


asridharan:
I would think `size` should be described as uint64 ?

wking:
On Thu, Sep 29, 2016 at 09:06:11AM -0700, asridharan wrote:
> I would think `size` should be described as uint64 ?

More discussion on this in #153.

asridharan:
I still don't understand the reasoning behind this decision. The only thing I see from the conversation seems to be its okay to use int64 since we "don't" allow negative values, and somehow the motivation seems to be that "go" has done it this way. A spec should not be language specific, so seems very arbitrary at this point.

Also, the fact that size is expected to be positive and validation needs to be performed has not been stated in the spec. Just being "picky" here since this is the spec and should be treated as a gold standard. The reasoning has to be beyond question.

wking:
On Thu, Sep 29, 2016 at 12:04:34PM -0700, asridharan wrote:
> I still don't understand the reasoning behind this decision.

I would also prefer to see an unsigned integer, but was unable to make
a convincing-enough case for it.

> … and somehow the motivation seems to be that "go" has done it this
> way.

The compat argument wasn't about Go, it was that docker/distribution
uses int64 [1].  But docker/distribution uses different media types,
and all reasonable size values will fit into both an int64 and a
uint64, so I don't see a problem with switching the OCI type to use
uints that carry at least 64 bits (more discussion in #153 where I
argue for following C and JSON in not capping the maximum size).

> Also, the fact that size is expected to be positive and validation
> needs to be performed has not been stated in the spec.

The non-negative-ness is covered by the existing [2]:

  This REQUIRED property specifies the size, in bytes, of the raw
  content.

because you can't have negatively-sized content.

Validation is covered by the existing [3]:

  If the length of the retrieved content does not match the specified
  length, the content SHOULD NOT be trusted.

The only thing wrong about using a signed integer to store that
information is that you're wasting a bit and inviting these “why is
this signed?” discussions ;).

[1]: https://github.com/opencontainers/image-spec/pull/153#issuecomment-226622038
[2]: https://github.com/opencontainers/image-spec/blame/32a3cb9fe339aab5070ebae5bb022f733eb36c30/descriptor.md#L31
[3]: https://github.com/opencontainers/image-spec/blame/32a3cb9fe339aab5070ebae5bb022f733eb36c30/descriptor.md#L33

asridharan:
[...]

PR #1334, opened 2026-08-20
Describe integer fields with JSON types and bound the descriptor size

BobDu:
#1249 made the I-JSON limitations of RFC 7493 a MUST for all JSON content in this specification, and RFC 7493 section 2.2 states that a sender cannot expect a receiver to treat an integer outside [-(2^53)+1, (2^53)-1] as an exact value. For the specification to be self-consistent the range permitted for `size` therefore has to be a subset of that interval, and today it is not: `defs-descriptor.json` allows values up to 9223372036854776000.

RFC 7493 attaches no requirement keyword to that interval; it only says that values outside it cannot be expected to be exact. Specifications built on I-JSON generally tighten it into a requirement so that values stay exact across implementations. RFC 8620 section 1.3 defines UnsignedInt as an integer where "the value MUST be in the range 0 <= value <= 2^53-1", and section 6.1 applies it to a directly comparable field: the size in octets of an immutable blob addressed by an id. RFC 9535 section 2.1 requires integers relevant to processing to be "within the range of exact integer values defined in Internet JSON (I-JSON) ..., namely within the interval [-(2^53)+1, (2^53)-1]".

This narrows the upper bound in theory, but it puts no valid existing data at risk: with the hardware available today an image cannot approach petabyte scale, and if that ever changes it will be a v2 concern.

Separately, this specification should be language neutral, and `int64` is a Go type name. The specification only needs to state that the value is an integer and give its bounds; which data type an implementation uses is the implementation's choice and outside the scope of the specification. `int` is not a JSON type either, so `schemaVersion` in manifest.md and image-index.md is changed to `integer` for consistency.

Changing the prose does not imply changing the Go types, which would be a breaking change. It only clarifies the specification document so that it is self-consistent and leaves no room for ambiguity, and it describes the current situation more closely, since oci-spec-rs defines the field as `u64`.

Related: #153, #350, https://github.com/youki-dev/oci-spec-rs/blob/42a237132bcaf7c6a753be0e1d7e309b7298fb5a/src/image/descriptor.rs#L40

### DIFF
--- a/descriptor.md
+++ b/descriptor.md
@@ -29,12 +29,13 @@ The following fields contain the primary properties that constitute a Descriptor
   This REQUIRED property is the _digest_ of the targeted content, conforming to the requirements outlined in [Digests](#digests).
   Retrieved content SHOULD be verified against this digest when consumed via untrusted sources.
 
-- **`size`** *int64*
+- **`size`** *integer*
 
   This REQUIRED property specifies the size, in bytes, of the raw content.
   This property exists so that a client will have an expected size for the content before processing.
   If the length of the retrieved content does not match the specified length, the content SHOULD NOT be trusted.
   The size MUST NOT be negative.
+  The size MUST NOT be greater than 9007199254740991 (2^53-1).
 
 - **`urls`** *array of strings*
 
--- a/image-index.md
+++ b/image-index.md
@@ -9,7 +9,7 @@ For the media type(s) that this document is compatible with, see the [matrix][ma
 
 ## _Image Index_ Property Descriptions
 
-- **`schemaVersion`** *int*
+- **`schemaVersion`** *integer*
 
   This REQUIRED property specifies the image manifest schema version.
   For this version of the specification, this MUST be `2` to ensure backward compatibility with older versions of Docker.
--- a/manifest.md
+++ b/manifest.md
@@ -15,7 +15,7 @@ Unlike the [image index](image-index.md), which contains information about a set
 
 ## _Image Manifest_ Property Descriptions
 
-- **`schemaVersion`** *int*
+- **`schemaVersion`** *integer*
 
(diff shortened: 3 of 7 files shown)


tianon:
Can you please elaborate succinctly on the specific problem you're trying to solve?  I have read your description, but I don't really see a specific problem in it.

BobDu:
The initial issue I found was that both of the following schemas use 9223372036854776000 as the maximum value of an int64, even though the actual maximum is 9223372036854775807:

https://github.com/opencontainers/image-spec/blob/af26a05fba5ee648512f4ea3c9fda1fcc1b6d6dc/schema/defs-descriptor.json#L13

https://github.com/opencontainers/image-spec/blob/af26a05fba5ee648512f4ea3c9fda1fcc1b6d6dc/schema/defs.json#L22

I initially intended to submit a PR that simply corrected the value to 9223372036854775807.

However, I then realized that this would not address the underlying interoperability issue. RFC 7493 states that a sender cannot expect a receiver to preserve integers greater than 2^53 - 1 exactly (error in the maximum value here is likely due to this reason), because many JSON implementations represent numbers using IEEE 754 binary64 values.

Because size represents an exact byte count, a more fundamental solution is to restrict it to the I-JSON exact-integer range, with a maximum of 2^53 - 1.

sudo-bmitch:
Please rebase this to resolve the CI failure.

tianon:
I'm mostly neutral on the spec changes here (maybe *marginally* negative simply because we are a very Go heavy spec and I think that's mostly fine), and would neither support nor block them, officially. 

However, I think the json schema change perhaps deserves the same question from https://github.com/opencontainers/image-spec/pull/1340#issuecomment-5716208613: do you have a project which uses these schema files?

PR #1336, opened 2026-08-20
Deprecate ImageID

### LINKED ISSUE #1333: ImageID: allow hashing index instead of config
https://github.com/opencontainers/image-spec/blob/af26a05fba5ee648512f4ea3c9fda1fcc1b6d6dc/config.md?plain=1#L85-L89
(first appeared in e3d8984157fe736543867900eba8ce96568e8c3c , from https://github.com/moby/moby/pull/22264)

The spec demands "ImageID" to be computed by hashing the config, however, containerd has been hashing the index as the image ID.
Docker has begun to use containerd's image ID too:
- https://github.com/moby/moby/discussions/53307

So I suggest amending the spec to match the reality.
We can also consider saying that "image ID is opaque to implementations".

sudo-bmitch:
Fixes #1333


config.md
```diff
@@ -30,63 +30,12 @@ Layers SHOULD be packed and unpacked reproducibly to avoid changing the layer Di
 
 NOTE: Do not confuse DiffIDs with [layer digests](manifest.md#image-manifest-property-descriptions), often referenced in the manifest, which are digests over compressed or uncompressed content.
 
-### Layer ChainID
```
AkihiroSuda, line 33: https://github.com/opencontainers/image-spec/blob/main/identity/chainid.go should be deprecated in v1.2 and removed in v1.3-ish

sudo-bmitch: Good catch. I'll revert this to a deprecation, and add a similar deprecation to the Go code.

tianon (dismissed):

sudo-bmitch:
On a second viewing, ChainID is used by a lot of runtime tools, so I'm going to leave that.

https://pkg.go.dev/github.com/opencontainers/image-spec@v1.1.1/identity?tab=importedby

tianon (approved):

config.md
```diff
@@ -84,9 +84,10 @@ Most importantly, we can easily see that `ChainID(C) != ChainID(A|B|C)`, otherwi
 
 ### ImageID
 
-Each image's ID is given by the SHA256 hash of its [configuration JSON](#image-json).
-It is represented as a hexadecimal encoding of 256 bits, e.g., `sha256:a9561eb1b190625c9adb5a9513e72c4dedafc1cb2d4c5236c9a6957ec7dfd5a9`.
```
ahmad11333454: Htyh

config.md
ahmad11333454: What

PR #1337, opened 2026-08-21
Remove golangci-lint

sudo-bmitch:
golangci-lint was Go version dependent to build, and resulted in more maintenance overhead than code fixes. Instead, this switches to "go fmt" and "go vet" for some minimal compiler checks.


Makefile
```diff
@@ -85,10 +85,11 @@ check-license: ## check license headers in source files
...
 
 .PHONY: lint-go
-lint-go: .install.lint ## lint check of Go files using golangci-lint
+lint-go: ## run various checks on the Go files
 	@echo "checking Go lint"
-	@GO111MODULE=on $(GOPATH)/bin/golangci-lint run
-	@[ "$$BUILD_SPEC_MODULE_ONLY" = true ] || { cd schema && GO111MODULE=on $(GOPATH)/bin/golangci-lint run; }
+	@test -z "$(go fmt ./...)"
+	@test -z "$(go vet ./...)"
+	@[ "$$BUILD_SPEC_MODULE_ONLY" = true ] || ( cd schema && test -z "$(go fmt ./...)" && test -z "$(go vet ./...)" )
```
tianon: Why capturing?  Both command return non-zero with useful output IME when they're just run (and they both support `-C`):

```suggestion
	@go fmt ./...
	@go vet ./...
	@[ "$$BUILD_SPEC_MODULE_ONLY" = true ] || { go -C schema fmt ./... && go -C schema vet ./...; }
```

BobDu: `$(...)` inside a recipe is a make variable reference, not shell command substitution, so this target expands to `test -z ""`:

```
$ make --warn-undefined-variables -n lint-go
Makefile:90: warning: undefined variable 'go fmt ./...'
Makefile:91: warning: undefined variable 'go vet ./...'
Makefile:92: warning: undefined variable 'go fmt ./...'
Makefile:92: warning: undefined variable 'go vet ./...'
echo "checking Go lint"
test -z ""
test -z ""
[ "$BUILD_SPEC_MODULE_ONLY" = true ] || ( cd schema && test -z "" && test -z "" )
```

BobDu: `go -C` needs Go 1.21+ , `go.mod` says 1.18

BobDu (changes requested):

Makefile
sudo-bmitch: `go fmt` edits the file with a exits with a 0, which wouldn't be useful in the CI pipeline that discards the change. I was keeping the same pattern for `go vet` but removed it since that does exit with a non-zero. I was also using `()` instead of `{}` for the sub-command to avoid having the `cd schema` leak to the parent shell, which could get confusing if commands get chained either here or in a future copy/paste.

sudo-bmitch: Thanks for catching that, fixed.

tianon (approved):

sajayantony (approved):

9d7zwbvmpn-lgtm (commented):
image

9d7zwbvmpn-lgtm (commented):
image

